PR #4, opened 2016-01-14
Check "in eval?" variable instead of walking the call stack

karenetheridge:
The issue queue is closed in this repository, or I would have simply filed an issue.. apologies if you have a queue somewhere else. But see `perldoc perlvar` -- you can just check `$^S` to see if you're in an eval, rather than walking the call stack.

### DIFF
--- a/Bugzilla/Sentry.pm
+++ b/Bugzilla/Sentry.pm
@@ -301,13 +301,7 @@ sub _write_to_error_log {
 
 # lifted from Bugzilla::Error
 sub _in_eval {
-    my $in_eval = 0;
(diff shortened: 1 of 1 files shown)


globau:
Thanks for the contribution, but this is a read-only mirror of git.mozilla.org/webtools/bmo/bugzilla. Please open a bug and attach a review as per standard process. https://wiki.mozilla.org/Bugzilla:Submitting_Patches

Sorry for the inconvenience!

globau:
$^S won't work here because it's always true when running under mod_perl.

karenetheridge:
bummer, oh well!

PR #16, opened 2016-07-21
Bug 1257608 - Attaching a file can't be done in a new tab/window

SebastinSanty:
[Bug 125608](https://bugzilla.mozilla.org/show_bug.cgi?id=1257608)


SebastinSanty:
There is already work underway to make certain <a> tags to look like buttons.

PR #18, opened 2016-07-22
Bug 1262465 - ensure unprivileged users can mark a bug as security sensitive

SebastinSanty:
[Bug 1262465](https://bugzilla.mozilla.org/show_bug.cgi?id=1262465)


SebastinSanty:
New PR avoiding Hook.process

PR #19, opened 2016-07-29
Bug 1290580 - Rebase docker infra off of CentOS 6

dylanwh:
Alright, this should built pretty fast -- can you see if CI works?


docker/runtests.sh
```diff
@@ -48,15 +48,16 @@ echo -e "\n== Starting memcached"
...
 sed -e "s?%DB_NAME%?bugs_test?g" --in-place $BUGZILLA_ROOT/qa/config/checksetup_answers.txt
 sed -e "s?%USER%?$BUGZILLA_USER?g" --in-place $BUGZILLA_ROOT/qa/config/checksetup_answers.txt
 echo "\$answer{'memcached_servers'} = 'localhost:11211';" >> $BUGZILLA_ROOT/qa/config/checksetup_answers.txt
 
 if [ "$TEST_SUITE" == "checksetup" ]; then
     cd $BUGZILLA_ROOT/qa
-    /buildbot_step "Checksetup" ./test_checksetup.pl config/config-checksetup-$BUGS_DB_DRIVER
+    /buildbot_step "Checksetup" ./checksetup.pl config/config-checksetup-$BUGS_DB_DRIVER
     exit $?
 fi
```
dklawren: Remove this section completely. This was an old test we no longer do that tests upgrading older databases used by older versions of Bugzilla. so test_checksetup.pl was the correct script to run, not checksetup.pl.

docker/bugzilla.conf
```diff
@@ -11,6 +11,7 @@ ServerName localhost:80
         DirectoryIndex index.cgi
         Options Indexes FollowSymLinks ExecCGI
         AllowOverride All
-        Require all granted
+        Satisfy Any
+        Allow from all
     </Directory>
```
dklawren: We should just grab the real http conf from SCL3 webheads and use it. Minus the SSL certs, etc.

docker/runtests.sh
```diff
@@ -75,7 +76,7 @@ perl generate_test_data.pl
 echo -e "\n== Starting web server"
 sed -e "s?^#Perl?Perl?" --in-place /etc/httpd/conf.d/bugzilla.conf
 /usr/sbin/httpd &
-sleep 3
+sleep 30
```
dklawren: Do we really need 30?

dylanwh: after a lot of testing, I think 10 is a reasonably safe value

PR #22, opened 2016-08-31
Bug 1253535 - Detect if mysql can support utf8mb4, and if using utf8, convert to utf8mb4

dylanwh:
[Bug 1253535 - Detect if mysql can support utf8mb4, and if using utf8, convert to utf8mb4](https://bugzilla.mozilla.org/show_bug.cgi?id=1253535)

Just a rough sketch at how this might work. can test it later and iterate.


floatingatoll:
@dylanwh Do you still want this as a thing?

dylanwh:
Replaced with #282

PR #26, opened 2016-10-11
Bug 1309112 - Detect and linked GitHub issue and pull request in comment

SebastinSanty:
[1309112](https://bugzilla.mozilla.org/show_bug.cgi?id=1309112)

### DIFF
--- a/extensions/BMO/Extension.pm
+++ b/extensions/BMO/Extension.pm
@@ -713,6 +713,18 @@ sub bug_format_comment {
         }
     });
 
+    # link github pull requests and issues
+    push (@$regexes, {
+        match => qr#\b([\S]+\/[\S]+\#[0-9]+)\b#mx,
+        replace => sub {
+            my $args = shift;
+            my $owner = html_quote($args->{matches}->[0]);
+            my $repo = html_quote($args->{matches}->[1]);
+            my $number = html_quote($args->{matches}->[2]);
+            return qq#<a href="https://github.com/$owner/$repo/issues/$number">$owner/$repo\#$number</a>#;
+        }
+    });
+
     # Update certain links to git.mozilla.org to go to github.com instead
     # https://git.mozilla.org/?p=webtools/bmo/bugzilla.git;a=blob;f=Bugzilla/WebService/Bug.pm;h=d7a1d8f9bb5fdee524f2bb342a4573a63d890f2e;hb=HEAD#l657
     push(@$regexes, {


SebastinSanty:
@dylanwh : Still to test, pm'ed you on an issue

extensions/BMO/Extension.pm
```diff
@@ -713,6 +713,18 @@ sub bug_format_comment {
         }
     });
 
+    # link github pull requests and issues
+    push (@$regexes, {
+        match => qr#\b([\S]+\/[\S]+\#[0-9]+)\b#mx,
```
upsuper: As far as I can see, the repo name can only contains `[A-Za-z0-9_\.-]`. I guess the owner part is that way as well. Actually `[\S]` looks wrong anyway, since it could include `/` itself, right?

Also I think you may want to ensure that the character before is not a `/` either to avoid parsing part of a URL.

SebastinSanty: Yeah, true that. I missed it.

SebastinSanty: @upsuper I guess, it should also contain a space before and after, so that links like these "sample/to/repo#num" don't link?

extensions/BMO/Extension.pm
```diff
@@ -713,6 +713,18 @@ sub bug_format_comment {
         }
     });
 
+    # link github pull requests and issues
+    push (@$regexes, {
+        match => qr#\b([A-Za-z0-9_\.-]+\/[A-Za-z0-9_\.-]+\#[0-9]+)\b#mx,
```
dylanwh: Unfortunately this regular expression is so broad it will be a performance problem.
There are 11,764,265 comments on BMO.

For the short term, we can hard-code a list of projects globally,
and later on I will add a way of linking components to external resources.

I'm updating the bug with instructions on this.

SebastinSanty: Yeah, listing projects, seems to be a great idea. I'll have a look on BMO. Thanks for the review, Dylan!

extensions/BMO/Extension.pm
```diff
@@ -713,6 +713,18 @@ sub bug_format_comment {
...
     });
 
+    # link github pull requests and issues
+    push (@$regexes, {
+        match => qr#\s([A-Za-z0-9_\.-]+\/[A-Za-z0-9_\.-]+\#[0-9]+)\b#mx,
+        replace => sub {
+            my $args = shift;
+            my $owner = html_quote($args->{matches}->[0]);
+            my $repo = html_quote($args->{matches}->[1]);
+            my $number = html_quote($args->{matches}->[2]);
```
upsuper: I have a question. How does it know what is `[0]`, `[1]`, and `[2]`? It seems to me there are no such groups in the regexp. Should it be `([A-Za-z0-9_\.-]+)\/([A-Za-z0-9_\.-]+)\#([0-9]+)` instead? I'm not familiar with Perl, so I may be missing something.

SebastinSanty: My bad :/. Corrected it! Thanks for reviewing @upsuper

extensions/BMO/Extension.pm
```diff
@@ -715,13 +715,13 @@ sub bug_format_comment {
...
     push (@$regexes, {
-        match => qr#\s([A-Za-z0-9_\.-]+)\/([A-Za-z0-9_\.-]+)\#([0-9]+)\b#mx,
+        match => qr/\s([A-Za-z0-9_\.-]+)\/([A-Za-z0-9_\.-]+)\#([0-9]+)\b/,
         replace => sub {
             my $args = shift;
             my $owner = html_quote($args->{matches}->[0]);
             my $repo = html_quote($args->{matches}->[1]);
             my $number = html_quote($args->{matches}->[2]);
-            return qq#<a href="https://github.com/$owner/$repo/issues/$number">$owner/$repo\#$number</a>#;
+            return qq#<a target = "_blank" href="https://github.com/$owner/$repo/issues/$number"> $owner/$repo\#$number</a>#;
```
upsuper: Hmmm, probably you should do the same thing as the svn regexp that, capture `\s` as `[0]` and prepend to the result, instead of adding a whitespace, because `\s` can match other characters as well. Also it doesn't seem to me we usually do `target="_blank"` on links. Even if we do, there shouldn't be whitespaces around `=`.

SebastinSanty:
@upsuper Sorry for the delay. @dylanwh Can you have a look at this? Thanks.

PR #38, opened 2017-01-11
Bug 1330449 - Change Edit button text and add a tooltip

KWierso:
This changes the text for the Edit button and also adds a tooltip to explain what it does.

### DIFF
--- a/extensions/BugModal/template/en/default/bug_modal/edit.html.tmpl
+++ b/extensions/BugModal/template/en/default/bug_modal/edit.html.tmpl
@@ -278,7 +278,7 @@
       <div>
         <button type="button" id="cancel-btn" class="minor" style="display:none">Cancel</button>
         <button type="button" id="mode-btn" class="major">
-          <span id="mode-btn-readonly">Edit</span>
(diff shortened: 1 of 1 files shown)


KWierso:
Wordsmithing welcome :)

extensions/BugModal/template/en/default/bug_modal/edit.html.tmpl
```diff
@@ -278,7 +278,7 @@
       <div>
         <button type="button" id="cancel-btn" class="minor" style="display:none">Cancel</button>
         <button type="button" id="mode-btn" class="major">
-          <span id="mode-btn-readonly">Edit</span>
+          <span id="mode-btn-readonly" title="Toggle editing fields for bug metadata">Edit Fields</span>
```
hwine: Suggest "Enable editing...", as once you are in Edit mode, there are "Save" and "cancel" buttons, nothing to toggle.

dklawren:
I would prefer it to be called 'Edit Bug' because clicking the 'Edit' button enables more than just editing of bug fields. It also allows for toggling privacy on comments, restricting comments, ignoring email for the bug (user specific), etc. Dylan, ok with that too?

PR #40, opened 2017-01-18
Bug 1331305 - Auto resize the height of textarea of "add comment"

KuoE0:
[Bug 1331305 - Auto resize the height of textarea of "add comment"](https://bugzilla.mozilla.org/show_bug.cgi?id=1331305)

Please see https://bugzilla.mozilla.org/show_bug.cgi?id=1331305.

### DIFF
--- a/extensions/BugModal/web/bug_modal.css
+++ b/extensions/BugModal/web/bug_modal.css
@@ -689,11 +689,14 @@ h3.change-name {
 
 #comment {
     border: 1px solid #ccc;
+    resize: none;
+    overflow: auto;
 }
 
 #comment, #comment-preview {
     clear: both;
     width: 100%;
+    height: auto;
     box-sizing: border-box !important;
     margin: 0 0 1em;
 }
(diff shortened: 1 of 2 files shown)


wlach:
This is just a drive-by comment, but it seems as if with your patch the text area could grow arbitrarily large, which I'm not sure is what we want.

I would define a maximum height for the textarea that would not be exceeded. This is how github comments seem to work.

KuoE0:
I updated the patch with setting 80% height of the viewport as the max height.

KuoE0:
Hi @dylanwh, I cannot watch this video. It shows "This video is unavailable." Can you check that? Thx!

wlach:
Browsing on the web, it seems like this is not a simple problem.

http://stackoverflow.com/questions/454202/creating-a-textarea-with-auto-resize
http://stackoverflow.com/questions/19170083/automatically-resize-text-area-based-on-content

Could you perhaps just use http://www.jacklmoore.com/autosize/ ?

PR #41, opened 2017-01-18
Bug 1331302 - Restrict the max-width of the textarea of "add comment"

KuoE0:
Please see https://bugzilla.mozilla.org/show_bug.cgi?id=1331302.

### DIFF
--- a/extensions/BugModal/web/bug_modal.css
+++ b/extensions/BugModal/web/bug_modal.css
@@ -696,6 +696,7 @@ h3.change-name {
     width: 100%;
     box-sizing: border-box !important;
     margin: 0 0 1em;
+    max-width: 1024px;
 }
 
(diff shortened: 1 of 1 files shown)


dklawren:
The css you edited is actually for the add-comment div and we need to apply the max-width to the textarea element itself. Here is what I suggest in the same file:

 #comment, #comment-preview {
     clear: both;
-   width: 100%;
     margin: 0 0 1em;
+  max-width: 1024px;
  }

KuoE0:
I updated the patch. Did you mean that? Some codes you provided is converted to bullets :p.

dklawren:
To https://github.com/mozilla-bteam/bmo.git
   44f10d0..9382ce4  master -> master

This has been merged on command line. Thanks!

PR #43, opened 2017-01-31
Add missing shirt sizes in lowermost gear choices.

mhoye:
Minor fix, thanks.


dylanwh:
This never had a bug assigned, and is a code change so a bug is required

PR #44, opened 2017-02-23
Fix file permissions, remove Persona from generate_bmo_data, and add c9-build script (no bug)

dylanwh:


dklawren:
Looks good. Merge away.

PR #45, opened 2017-02-24
Bug 1342542 - Add special group partner-confidential, jbalaco@mozilla.com a…

dylanwh:
…nd mbest@mozilla.com to generate_bmo_data.pl

### DIFF
--- a/scripts/c9-build
+++ b/scripts/c9-build
@@ -1,6 +1,9 @@
 #!/bin/bash
 
 cd ~/workspace
+if ! ps aux | grep '[m]ysql' &>/dev/null; then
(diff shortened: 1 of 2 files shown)


dklawren:
This needs to be in mozilla-bteam/docker-bmo-ci/scripts/generate_bmo_data.pl, right?

We only have it in the bmo core core cause mozreview was using it at one time for their own containers.

dkl

dylanwh:
I need it for vagrant too. Or setting up an install on cloud9. and lots of places really.
Maybe https://github.com/mozilla-bteam/docker-bmo-ci/pull/3?

PR #46, opened 2017-03-03
Bug 1343663 - Comments are too wide

emceeaich:
Set #bugzilla-body to have em based widths so they scale with font size.

### DIFF
--- a/extensions/BugModal/web/bug_modal.css
+++ b/extensions/BugModal/web/bug_modal.css
@@ -812,9 +812,9 @@ h3.change-name {
 /* theme */
 
 #bugzilla-body {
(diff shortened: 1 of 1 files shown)


extensions/BugModal/web/bug_modal.css
```diff
@@ -812,9 +812,9 @@ h3.change-name {
 /* theme */
 
 #bugzilla-body {
-    max-width: 1200px;
-    min-width: 800px;
-    width: 100%;
+    max-width: 72em;
+    min-width: 60em;
+    width: 60em;
```
dylanwh: this should be width: 100%; or just delete it entirely.

PR #51, opened 2017-03-10
Bug 1344979 - When adding a patch to an unassigned bug, pre-check the reassignment checkbox

SebastinSanty:
[Bug 1344979 - When adding a patch to an unassigned bug, pre-check the reassignment checkbox](https://bugzilla.mozilla.org/show_bug.cgi?id=1344979)

### DIFF
--- a/template/en/default/attachment/create.html.tmpl
+++ b/template/en/default/attachment/create.html.tmpl
@@ -84,7 +84,7 @@ TUI_hide_default('attachment_text_field');
         <td>
           <em>If you want to assign this [% terms.bug %] to yourself,
               check the box below.</em><br>
-          <input type="checkbox" id="takebug" name="takebug" value="1">
+          <input type="checkbox" id="takebug" name="takebug" value="1" checked>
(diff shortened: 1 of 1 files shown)


template/en/default/attachment/create.html.tmpl
```diff
@@ -84,7 +84,7 @@ TUI_hide_default('attachment_text_field');
         <td>
           <em>If you want to assign this [% terms.bug %] to yourself,
               check the box below.</em><br>
```
dylanwh: I guess this should say "If you want to assign this bug to yourself, uncheck the box below."
otherwise LGTM.

floatingatoll: Er, how would unchecking the 'takebug' checkbox result in taking the bug?

dylanwh: correction: if we're going to pre-check the checkbox, the text should be updated to reflect that probably.

dylanwh:
Actually, it's fine as is. Commiting patch outside of github for consistency.

PR #52, opened 2017-03-10
Bug 1346162 - Needinfo is not checked for users that don't have canconfirm

SebastinSanty:
[Bug 1346162 - Needinfo is not checked for users that don't have editbugs](https://bugzilla.mozilla.org/show_bug.cgi?id=1346162)


dylanwh:
Will commit as patch to ensure correct commit message.

PR #54, opened 2017-03-14
Bug 1347237 - get_permissions.html is missing

dylanwh:


extensions/BMO/template/en/default/hook/global/user-error-error_message.html.tmpl
```diff
@@ -9,8 +9,8 @@
 [% IF error == 'illegal_change' || error == 'illegal_change_deps' %]
   <p>
     If you are attempting to confirm an unconfirmed [% terms.bug %] or edit the
-    fields of a [% terms.bug %], <a href="https://wiki.mozilla.org/BMO/UserGuide#Permissions_and_Groups">
-    find out how to get the necessary permissions</a>.
+    fields of a [% terms.bug %], <a href="page.cgi?id=get_permissions.html">find
```
emceeaich: Why did the link change? Is the wiki out of date or get_permissions.html more authoritative?

PR #55, opened 2017-03-17
Bug 1343805 - Use a darker font for comments, and also avoid Courier

SebastinSanty:


dylanwh:
Per recent comment on bug, s/Fira Code/Fira Mono/g

extensions/BugModal/web/bug_modal.css
```diff
@@ -598,7 +598,7 @@ h3.change-name {
 
 .comment-text {
     white-space: pre-wrap;
-    font: 13px/1.2 "Droid Sans Mono",Menlo,Monaco,"Courier New",Courier,monospace;
+    font: 13px/1.2 "Droid Sans Mono",Menlo,Monaco,"Fira Code",Consolas,monospace;
```
floatingatoll: There are ligature concerns with "Fira Code" in the bug; can we use a simple monospace variant of Fira instead?

skins/contrib/Mozilla/global.css
```diff
@@ -337,7 +337,7 @@ hr {
 #bugzilla-body {
     background: none;
     border: none;
-    color: #404040;
+    color: #202020;
```
floatingatoll: Technically, we could just use color: #000 here, but if the goal is to *intentionally* lower the contrast of the text vs. the white background, then it makes sense that we're up in the #222/444 ranges.

extensions/BugModal/web/bug_modal.css
floatingatoll: Fira Mono, apparently

SebastinSanty: Yes, that would be better. I'll update it.

skins/contrib/Mozilla/global.css
SebastinSanty: I am not sure about this, as it would affect the whole BMO. I would like to defer this to @dylanwh

dylanwh:
Closing this because the font of the bike shed is apparently very important.

PR #57, opened 2017-03-19
Bug 1347822 - References in comments to attachments on other bugs shouldn't be augmented with actions

SebastinSanty:


dylanwh:
Please see latest comments on the bug

PR #58, opened 2017-03-26
Bug 1350467 - Add method Bugzilla->preload_features() to be called in mod_perl.pl startup

dylanwh:


Bugzilla/CPAN.pm
```diff
@@ -104,6 +104,16 @@ sub feature {
     return $FEATURE_LOADED{$feature_name} = 1;
 }
 
+sub preload_features {
+    my ($class) = @_;
+    return 0 unless _CAN_HAS_FEATURE;
```
floatingatoll: Does it offer any material performance benefit to move this line above the my()@_ invocation?

Bugzilla/CPAN.pm
```diff
@@ -104,6 +104,16 @@ sub feature {
...
+    my $meta = $class->cpan_meta;
+
+    foreach my $feature ($meta->features) {
+        $class->feature($feature->identifier);
+    }
```
floatingatoll: This is more concise and with two less my() invocations, but the usual concerns about map() readability apply. ```map { $class->feature($_->identifier) } $class->cpan_meta->features;``` might work out.

dylanwh: map in void context is a thing I try to never do, it's bad style
(although perl is smart enough to throw the return values away)

Bugzilla/CPAN.pm
dylanwh: pretty much I've never been able to measure a difference from using variables or not, it seems like just method calls / function calls are slow.

PR #60, opened 2017-04-03
Bug 1352908 - Optimize certain filters in the template compiler

dylanwh:
[% ... FILTER none %] is optimized away. There is no cost for using it.
[% ... FILTER html %] is optimized to a direct call to
Bugzilla::Util::html_quote(), saving a method call to $stash->filter.
[% ... FILTER uri %] also saves a call to $stash->filter.
[% ... FILTER null %] avoids a filter call and always results in ''

without patch:
```
spent 2.53s (1.64+890ms) within Bugzilla::Template::Context::filter which was called 381645 times, avg 7µs/call
```

with patch:
```
spent 470ms (186+285) within Bugzilla::Template::Context::filter which was called 37008 times, avg 13µs/call
```

### DIFF
--- a/Bugzilla/Template/Directive.pm
+++ b/Bugzilla/Template/Directive.pm
@@ -1,0 +1,70 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+#
+# This Source Code Form is "Incompatible With Secondary Licenses", as
+# defined by the Mozilla Public License, v. 2.0.
+
+# This exists to implement the template-before_process hook.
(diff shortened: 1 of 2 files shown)


Bugzilla/Template/Directive.pm
```diff
@@ -0,0 +1,70 @@
...
+package Bugzilla::Template::Directive;
+
+use 5.10.1;
+use strict;
+use warnings;
+
+use base qw(Template::Directive);
+
+sub filter {
+    package Template::Directive;
```
floatingatoll: Is this a new dependency?

dylanwh: it's part of the TT package.

dylanwh: I also talked to some people at TPC, and while this is a bit crazy it's not that unusual.

PR #62, opened 2017-04-03
Bug 1352912 - Optimize certain variables and functions in the template compiler

dylanwh:
Removed Bugzilla.* and Hook.* plugins. The functionally
is now provided by Bugzilla::Template::Namespace.
Calls to those routines get generated directly into the equivalent perl code.

This results in drastically fewer calls to $stash->get().

For Hook.process, that code now lives in Bugzilla/Hook.pm.

There was a totally useless User plugin as well, which is now gone.

Bugzilla::Template::Namespace also handles constants.* and terms.*.
The former now catches typos at checksetup time -- we had one existing typo.
Also, extensions that add constants need to add them to Bugzilla::Constants now,
perhaps through monkey patching.

Support for static terms is a WIP, and should probably be controlled with an
option.

Meanwhile, the template functions Param() and display_value() are now optimized
as well -- this is done by hooking into the code
generate (Bugzilla::Template::Directive).

We also do this to ignore [% USE Bugzilla %] and [% USE Hook %].
These can be removed from the code at any time, but their presense is not harmful.


Bugzilla/Constants.pm
```diff
@@ -430,9 +426,6 @@ use constant FIELD_TYPE_BUG_URLS => 7;
 use constant FIELD_TYPE_KEYWORDS => 8;
 use constant FIELD_TYPE_DATE => 9;
 use constant FIELD_TYPE_INTEGER => 10;
-use constant FIELD_TYPE_USER     => 20;
-use constant FIELD_TYPE_USERS    => 21;
-use constant FIELD_TYPE_BUG_LIST => 22;
```
floatingatoll: Does a placeholder need to be left behind here reminding people that 20..22 were allocated at one point, or is it safe to reuse them immediately / accidentally?

Bugzilla/Hook.pm
```diff
@@ -55,6 +56,71 @@ sub in {
...
+    $template ||= $context->stash->{component}->{name};
+
+    # sanity check:
+    if (!$template =~ /[\w\.\/\-_\\]+/) {
+        ThrowCodeError('template_invalid', { name => $template });
+    }
+
+    my (undef, $path, $filename) = File::Spec->splitpath($template);
+    $path ||= '';
+    $filename =~ m/(.+)\.(.+)\.tmpl$/;
```
floatingatoll: It would be comforting if template filenames were more restrictive than regex ```.``` here.

Bugzilla/Template/Namespace.pm
```diff
@@ -0,0 +1,106 @@
...
+    "Bug"               => "Bug",
+    "abug"              => "a bug",
+    "Abug"              => "A bug",
+    "aBug"              => "a Bug",
+    "ABug"              => "A Bug",
+    "bugs"              => "bugs",
+    "Bugs"              => "Bugs",
+    "zeroSearchResults" => "Zarro Boogs found",
+    "Bugzilla"          => "Bugzilla",
+    "BugzillaTitle" => 'Bugzilla@Development',
```
floatingatoll: This one seems to be whitespace-variant from the rest.

Bugzilla/Template/Namespace.pm
```diff
@@ -0,0 +1,106 @@
...
+                    $expr .= "($obj_args)";
+                }
+            }
+        }
+    }
+
+    return tt_dot_op($expr, $ident);
+}
+
+
```
floatingatoll: Is this block of whitespace intentional? (As in, "missing code possibility" intentional, though "meets this community's whitespace specifications" is also a fair question.)

template/en/default/global/user-error.html.tmpl
```diff
@@ -675,7 +675,7 @@
   [% ELSIF error == "fieldvalue_name_too_long" %]
     [% title = "Field Value Is Too Long" %]
     The value of a field is limited to 
-    [%+ constants.FIELD_VALUE_MAX_SIZE FILTER none %] characters. 
+    [%+ constants.MAX_FIELD_VALUE_SIZE FILTER none %] characters.
```
floatingatoll: The commit message doesn't talk about this fix. Was it found using the rest of the commit?

extensions/BMO/Extension.pm
```diff
@@ -89,6 +89,50 @@ BEGIN {
     *Bugzilla::Attachment::fetch_github_pr_diff     = \&_attachment_fetch_github_pr_diff;
 }
 
+sub template_before_process_wants {
+    return {
+        'bug/create/create-user-engagement.html.tmpl'        => 1,
```
floatingatoll: In ```BugModal/Extension.pm``` this is done more cleanly with ```want{$_} = 1 for (```, which seems useful here as well. return(map { $_ => 1 } ...) or similar would let you format this list exactly as the one in BugModal, with only the entry/exit lines varying for context.

extensions/FlagTypeComment/Extension.pm
```diff
@@ -70,6 +70,12 @@ sub db_schema_abstract_schema {
 # Templates #
 #############
 
+sub template_before_process_wants {
+    my %wants;
+    $wants{$_}=1 for FLAGTYPE_COMMENT_TEMPLATES;
```
floatingatoll: I count six methods of doing this in the patch.

extensions/MozReview/Extension.pm
```diff
@@ -42,6 +42,16 @@ BEGIN {
     }
 }
 
+sub template_before_process_wants {
+    return {
+        'bug/edit.html.tmpl'          => 1,
+        'bug_modal/header.html.tmpl'  => 1,
+        'bug_modal/edit.html.tmpl'    => 1,
+        'attachment/create.html.tmpl' => 1,
+        'bug/show-header.html.tmpl'   => 1
```
floatingatoll: Missing trailing comma, because JSON. (Sorry, heh.)

template/en/default/global/user-error.html.tmpl
dylanwh: Yes, the new code finds typos in constants.

floatingatoll: I would technically prefer you submit this as a separate pull request with its own bug (or, at the very least, its own commit), and then reference it from this PR, so that we're not mixing "refactor a function" and "repair an error" in the same pass. But I'm not going to block it.

Bugzilla/Constants.pm
dylanwh: I don't think so.

template/en/default/global/user-error.html.tmpl
dylanwh: I just committed this typo fix as a bare commit.

PR #63, opened 2017-04-19
Bug 1357809 - Add __heartbeat__, __lbheartbeat__, and __version__ endpoints for future cloud-services integration

dylanwh:


extensions/BMO/Extension.pm
```diff
@@ -2522,7 +2524,36 @@ sub _check_default_product_security_group {
 sub install_filesystem {
     my ($self, $args) = @_;
     my $files = $args->{files};
+    my $create_files = $args->{create_files};
     my $extensions_dir = bz_locations()->{extensionsdir};
+    $create_files->{__lbheartbeat__} = {
+        perms    => Bugzilla::Install::Filesystem::WS_SERVE,
+        contents => '',
```
floatingatoll: I strongly encourage this to contain a valid string that can be used to confirm that the load balancer got a valid response back, and not just "an empty response" (which could be any number of causes).

extensions/BMO/Extension.pm
```diff
@@ -2522,7 +2524,36 @@ sub _check_default_product_security_group {
...
+    };
+
+
+    # version.json needs to have a source attribute pointing to
+    # our repository. We already have this information in the (static)
+    # contribute.json file, so parse that in
+    my $json            = JSON::XS->new->pretty->utf8->canonical();
+    my $contribute_json = read_file( bz_locations()->{cgi_path} . "/contribute.json" );
+    my $contribute      = $json->decode($contribute_json);
```
floatingatoll: Some error handling of the above and below lines might be valuable.

heartbeat.cgi
```diff
@@ -0,0 +1,31 @@
...
+use Bugzilla::Update;
+
+my $ok = eval {
+    my $memcached = Bugzilla->memcached;
+    my $dbh       = Bugzilla->dbh;
+    my $timestamp = $dbh->selectrow_array('SELECT LOCALTIMESTAMP(0)');
+    $memcached->set_config({key => "heartbeat", data => $timestamp});
+    my $mc_timestamp = $memcached->get_config({key => "heartbeat"});
+
+    $mc_timestamp eq $timestamp;
```
floatingatoll: If two load balancers both call heartbeat.cgi at the same time, spanning a one-second boundary per LOCALTIMESTAMP(0), then they will trigger a race condition and 'NOT OK' below, unless heartbeat.cgi is somehow under a global lock or whatever.

Perhaps ```le``` rather than ```eq``` here would help?

heartbeat.cgi
```diff
@@ -0,0 +1,31 @@
...
+use Bugzilla::Constants;
+use Bugzilla::Error;
+use Bugzilla::Update;
+
+my $ok = eval {
+    my $memcached = Bugzilla->memcached;
+    my $dbh       = Bugzilla->dbh;
+    my $timestamp = $dbh->selectrow_array('SELECT LOCALTIMESTAMP(0)');
+    $memcached->set_config({key => "heartbeat", data => $timestamp});
+    my $mc_timestamp = $memcached->get_config({key => "heartbeat"});
```
floatingatoll: We just wrote this value to memcache. Why is it being read back?

floatingatoll: (Specifically: Testing the return result of $memcached->set_config() should be sufficient to detect whether an error occurred during memcache write, unless we need this value for the ```eq``` comparison below?)

heartbeat.cgi
floatingatoll: (But even with le, it's still possible to race, you just need weird timing variations in the DB read and the Memcache write.)

dylanwh:
Think this good enough to merge? I know there's an r+ on the bug but we have changed a few things since then.

floatingatoll:
Yeah, it's fine.

PR #64, opened 2017-04-19
Bug 1348380 - Add param and user setting to control elasticsearch behavior

dylanwh:


Bugzilla/Config/Elastic.pm
```diff
@@ -16,18 +16,23 @@ use Bugzilla::Config::Common;
...
-    },
-    {
-        name    => 'elasticsearch_index',
-        type    => 't',
-        default => 'bugzilla',
-    },
-  );
+    return (
+        {
+            name    => 'elasticsearch',
```
floatingatoll: What are these three parameters? Why is this one called simply 'elasticsearch'? Are these documented elsewhere?

Bugzilla/Elastic.pm
```diff
@@ -16,6 +16,12 @@ with 'Bugzilla::Elastic::Role::HasIndexName';
 
 sub suggest_users {
     my ($self, $text) = @_;
+
+    unless (Bugzilla->params->{elasticsearch}) {
```
floatingatoll: If the type => b default => 0 above means 'boolean false', then this is fine. (If it means anything other than 'boolean', though)

Bugzilla/Elastic.pm
```diff
@@ -38,10 +44,15 @@ sub suggest_users {
     }
     else {
         warn "suggest_users error: $@";
-        my $users = Bugzilla::User::match($text, 25, 0);
-        return [ map { { real_name => $_->name, name => $_->login } } @$users];
+        # optimization: faster than a regular method call.
```
floatingatoll: This is a pretty hardcore optimization. Is it truly faster than ```return &_suggest_users_fallback(@_)``` ?

Bugzilla/Install.pm
```diff
@@ -175,6 +175,12 @@ sub SETTINGS {
         default => 'off',
         category => 'API'
     },
+    {
+        name     => 'use_elasticsearch',
```
floatingatoll: Is this name meant to match the ```'elasticsearch'``` seeminly-boolean thing seen previously above?

buglist.cgi
```diff
@@ -692,11 +692,18 @@ my $fallback_search = Bugzilla::Search->new(fields => [@selectcolumns],
...
 my $search;
-my $elastic = $cgi->param('elastic') // 1;
+my $elastic = $cgi->param('elastic') // $elastic_default;
 if (defined $cgi->param('elastic')) {
-    $vars->{was_elastic} = 1;
+    $vars->{was_elastic} = $elastic;
 }
+
+# If turned off in the admin section, it is always off.
+$elastic = 0 unless Bugzilla->params->{elasticsearch};
```
floatingatoll: if (Bugzilla->params->{elasticsearch}) { lines 695 .. 702 } else { $elastic = 0; } would be more efficient, so that you aren't wasting time with $cgi->param when it's globally disabled.

template/en/default/global/setting-descs.none.tmpl
```diff
@@ -56,6 +56,7 @@
    "possible_duplicates"              => "Display possible duplicates when reporting a new $terms.bug",
    "requestee_cc"                     => "Automatically add me to the CC list of $terms.bugs I am requested to review",
    "api_key_only"                     => "Require API key authentication for API requests",
+   "use_elasticsearch"                => "Use elasticsearch for $terms.bug searches when possible",
```
floatingatoll: "when possible" is, perhaps, a bit of a fine point. Why is it opt-in for logged-in users rather than opt-out? How could a user discover that in any way other than the above code patch?

Bugzilla/Install.pm
dylanwh: nope! those are two different things, one is user setting and the other is a admin parameter. They are two very different systems. :-/

template/en/default/global/setting-descs.none.tmpl
dylanwh: Excellent questions! it is opt-in for our users because we think is a good idea to not harm people's productivity by making search behave slightly different (if faster).
For logged-out users, we make it opt-out because less time will be spent on queries.

Discoverability of this user preference is not very good (but no worse than the other parameters), however we do advertize that elasticsearch is in use and provide a way to disable it on a per-query basis already -- this is visible on bugzilla-dev.allizom.org right now.

Bugzilla/Elastic.pm
floatingatoll: (I'm also fine accepting that goto is normal in this use, if that's fine.)

PR #66, opened 2017-04-24
Bug 1357775 - the needinfo dropdown for the triage owner should display the name and email address when selected

dylanwh:
Turns out I missed the logic for displaying the real name beside the needinfo
role selector. That logic is scary and weird and I would've replaced it had I
noticed it then.

This patch removes the template code that generated JS, and instead puts
populates a data-identity attribute with the real name (or login name).

I added a convenience method to Bugzilla::User for this, which can later be
upstreamed.

### DIFF
--- a/extensions/Needinfo/template/en/default/bug/needinfo.html.tmpl
+++ b/extensions/Needinfo/template/en/default/bug/needinfo.html.tmpl
@@ -95,20 +95,7 @@ $(function() {
     } else {
       $('#needinfo_from_container').addClass('bz_default_hidden');
       $('#needinfo_from').prop('disabled', true);
-      var identity = '';
-      if (role == 'reporter') {
-        identity = '[% bug.reporter.realname || bug.reporter.login FILTER js %]';
-      } else if (role == 'assigned_to') {
-        identity = '[% bug.assigned_to.realname || bug.assigned_to.login FILTER js %]';
-      } else if (role == 'qa_contact') {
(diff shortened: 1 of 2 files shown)


extensions/Needinfo/template/en/default/bug/needinfo.html.tmpl
```diff
@@ -202,24 +189,35 @@ $(function() {
...
+            <option
+                data-identity="[% user.name_or_login FILTER html %]"
+                value="user">myself</option>
           [% END %]
           [% FOREACH mentor = available_mentors %]
-            <option [% IF available_mentors.size > 1 %]title="mentor"[% END %] value="[% mentor.login FILTER html %]">
-              [% available_mentors.size == 1 ? "mentor" : mentor.login FILTER html %]
+            <option data-identity="[% mentor.name_or_login FILTER html %][% IF available_mentors.size > 1 %] (mentor)[% END %]"
+                [% IF available_mentors.size > 1 %]title="mentor"[% END %] value="[% mentor.login FILTER html %]">
+                [% available_mentors.size == 1 ? "mentor" : mentor.login FILTER html %]
```
dylanwh: I was wondering about title="" -- doesn't it get used for the displayed value in the select?

kyoshino: `title` is just a tooltip and it actually won't be displayed on `<option>`s. The label is the text within the `<option>` like `<option value="[VALUE]">[LABEL]</option>`.

PR #67, opened 2017-04-26
No Bug - Clean up here-docs

dylanwh:
here-docs should either be at indent level 0 (and stored in constants)
or replaced with other forms of quoting that can be idented more readibly.

### DIFF
--- a/Bugzilla/Install/Filesystem.pm
+++ b/Bugzilla/Install/Filesystem.pm
@@ -49,6 +49,56 @@ use constant HT_DEFAULT_DENY => <<EOT;
 deny from all
 EOT
 
+use constant HT_GRAPHS_DIR => <<EOT;
+# Allow access to .png and .gif files.
+<FilesMatch (\\.gif|\\.png)\$>
+  Allow from all
+</FilesMatch>
+
+# And no directory listings, either.
+Deny from all
+EOT
+
+use constant HT_WEBDOT_DIR => <<EOT;
+# Restrict access to .dot files to the public webdot server at research.att.com
+# if research.att.com ever changes their IP, or if you use a different
+# webdot server, you'll need to edit this
+<FilesMatch \\.dot\$>
+  Allow from 192.20.225.0/24
+  Deny from all
+</FilesMatch>
+
+# Allow access to .png files created by a local copy of 'dot'
+<FilesMatch \\.png\$>
+  Allow from all
+</FilesMatch>
+
+# And no directory listings, either.
+Deny from all
+EOT
+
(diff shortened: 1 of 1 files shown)


dylanwh:
I want to add the caching headers for the static files, and I think here is the most sensible place. Before doing that I wanted to cleanup this a little bit.

Bugzilla/Install/Filesystem.pm
```diff
@@ -49,6 +49,56 @@ use constant HT_DEFAULT_DENY => <<EOT;
 deny from all
 EOT
 
+use constant HT_GRAPHS_DIR => <<EOT;
```
floatingatoll: Can this be a single q, so we can dump all the backslashing?

Bugzilla/Install/Filesystem.pm
```diff
@@ -534,12 +534,8 @@ sub _remove_empty_css_files {
...
     my ($file) = @_;
     my $basename = basename($file);
-    my $empty_contents = <<EOT;
-/*
- * Custom rules for $basename.
- * The rules you put here override rules in that stylesheet.
- */
-EOT
+    my $empty_contents = qq{/* Custom rules for $basename.\n}
+        . qq{ * The rules you put here override rules in that stylesheet.  */};
```
floatingatoll: Why don't we just make this comment one line instead of three?

Bugzilla/Install/Filesystem.pm
```diff
@@ -481,14 +484,11 @@ sub update_filesystem {
...
-
-*** It appears that you still have an old index.html hanging around.
-    Either the contents of this file should be moved into a template and 
-    placed in the '$templatedir/en/custom' directory, or you should delete 
-    the file.
-
-EOT
+        print join("\n",
+            "*** It appears that you still have an old index.html hanging around.",
+            "Either the contents of this file should be moved into a template and ",
```
floatingatoll: Trailing spaces here will be appended by newline, which works out to be a whitespace error that git can't see.

Bugzilla/Install/Filesystem.pm
```diff
@@ -49,6 +49,56 @@ use constant HT_DEFAULT_DENY => <<EOT;
...
+<FilesMatch \\.(css|js)\$>
+  Allow from all
+</FilesMatch>
+
+# And no directory listings, either.
+Deny from all
+EOT
+
+use constant INDEX_HTML => <<EOT;
+<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.01 Transitional//EN">
```
floatingatoll: Must we?

PR #68, opened 2017-04-26
Bug 1359961 - Replace Expires with Cache-Control

floatingatoll:


floatingatoll:
Optionally, you may choose to reject 6ec705b "remove expires header", as it does no harm to keep it.

dylanwh:
I think we can/should delete this entire section of the root .htaccess file.

PR #69, opened 2017-04-26
Bug 1359971 - add GIF, ICO, JPG, PNG, SWF, WOFF to cacheable assets

floatingatoll:
This pull request is a single commit (c914cfa). The remaining commits are part of #68, upon which this commit depends.


dylanwh:
The problem here is the lifetime, the filenames for images/etc are not unique.

floatingatoll:
Hm. Okay, retracting. Which assets do we append `?timestamp` to?

PR #75, opened 2017-05-10
Bug 1361890 - Fix problems with current js and css concatenation

dylanwh:


Bugzilla/Template.pm
```diff
@@ -975,6 +868,14 @@ sub create {
             # Function for retrieving global parameters.
             'Param' => sub { return Bugzilla->params->{$_[0]}; },
 
+            asset_file => sub {
+                return Bugzilla->asset_manager->asset_file($_[0]);
+            },
+
+            asset_files => sub {
+                return [Bugzilla->asset_manager->asset_files(@{ $_[0] })];
```
floatingatoll: It weirds me out that the [] is placed here instead of upstream, but it clearly works, so shrug.

extensions/BugModal/template/en/default/bug_modal/edit.html.tmpl
```diff
@@ -298,7 +298,7 @@
         <button type="button" id="mode-btn" class="major">
           <span id="mode-btn-readonly" title="Enable editing fields for [% terms.bug %] metadata">Edit [% terms.Bug %]</span>
           <span id="mode-btn-loading">
-            <img id="edit-throbber" src="extensions/BugModal/web/throbber.gif" width="16" height="11">
+            <img id="edit-throbber" src="[% asset_file('extensions/BugModal/web/throbber.gif') FILTER html %]" width="16" height="11">
```
floatingatoll: I read this as '.gif FILTER html' and wodnered why we are emitting a .gif reinterpreted as HTML. What does this actually do in practical HTML output?

template/en/default/global/header.html.tmpl
```diff
@@ -137,31 +136,30 @@
...
-      [% ELSE %]
-        [% FOREACH asset_url = css_sets.custom %]
+        <style type="text/css">
+            [% style %]
+        </style>
+
+        [% FOREACH asset_url = asset_files(css_sets.custom) %]
+            [% PROCESS format_css_link %]
+        [% END %]
+    [% ELSE %]
```
floatingatoll: This ELSE seems to be indented very confusingly vs. the surrounding section.

Bugzilla/Constants.pm
```diff
@@ -702,7 +695,7 @@ sub _bz_locations {
         # The script should really generate these graphs directly...
         'webdotdir'      => "$datadir/webdot",
         'extensionsdir'  => "$libpath/extensions",
-        'assetsdir'      => "$datadir/assets",
+        'assetsdir'      => "$libpath/assets",
```
dylanwh: libpath is effectively the web root, and datadir is the NFS-mounted partition. I think this comment is about how confusing it is to use 'libpath' for this -- I think may I mean 'cgipath'... The whole nomenclature of bz_locations() is hairy though. 

Switching this means assets are per-web head, which is a huge improvement.

Bugzilla/CGI.pm
```diff
@@ -441,12 +441,16 @@ sub header {
         %headers = ('-type' => shift(@_));
     }
     else {
-        %headers = @_;
+        # we always want headers to be in lower case, to avoid
+        # for instance -Cache_Control vs. -cache_control.
+        my %tmp = @_;
+        %headers = map { lc($_) => $tmp{$_} } keys %tmp;
```
dylanwh: I think random is acceptable, but it would be good to call that out in a comment.
I'll do that.

Bugzilla/CGI.pm
```diff
@@ -441,12 +441,16 @@ sub header {
...
     }
 
     if ($self->{'_content_disp'}) {
         $headers{'-content_disposition'} = $self->{'_content_disp'};
     }
+    $headers{'-cache_control'} = 'no-cache, no-store, must-revalidate' unless $headers{'-cache_control'};
```
dylanwh: I think that is more readable. Or perhaps the //= operator would signal the intent even more.

Bugzilla/Install/AssetManager.pm
```diff
@@ -0,0 +1,224 @@
...
+    my ($self, $file) = @_;
+    return unless -f $file;
+    my $base_dir  = $self->base_dir;
+    my $asset_dir = $self->asset_dir;
+    my $asset_map = $self->state->{asset_map};
+
+    my $key = File::Spec->abs2rel( $file, $base_dir );
+    return if $asset_map->{$key};
+
+    if ($file =~ /\.(jpe?g|png|gif|ico|woff|js)$/i) {
```
dylanwh: I'm not sure I understand. Can you elaborate?

Bugzilla/Install/AssetManager.pm
```diff
@@ -0,0 +1,224 @@
...
+            die "failed to write $asset_file";
+        }
+    }
+}
+
+sub compile_all {
+    my ($self) = @_;
+    my $asset_map = $self->state->{asset_map};
+
+    %$asset_map = ();
```
dylanwh: That used to be the case when asset_map was itself an object attribute! Now that 'state' is the attribute, setting a fresh hashref is fine.

Bugzilla/Install/AssetManager.pm
```diff
@@ -0,0 +1,224 @@
...
+
+    my $wanted = sub {
+        $self->compile_file($File::Find::name);
+    };
+
+    find( { wanted => $wanted, no_chdir => 1 }, @{ $self->source_dirs });
```
dylanwh: Probably need to run the whole thing through perltidy -pbp

Bugzilla/Install/AssetManager.pm
floatingatoll: Could you run a test query for /favicon.ico through your live Dev instance and make sure it's served correctly with the expected (non-immutable?) headers?

Bugzilla/Install/AssetManager.pm
```diff
@@ -0,0 +1,224 @@
...
+    $self->_save_state();
+}
+
+sub _css_url_rewrite {
+    my ($self, $url, $file) = @_;
+    my $dir = dirname($file);
+    # rewrite relative urls as the unified stylesheet lives in a different
+    # directory from the source
+    $url =~ s/(^['"]|['"]$)//g;
+    if ($url =~ m!^(/|data:)!) {
```
dylanwh: I don't think so.

Bugzilla/Install/AssetManager.pm
floatingatoll: Could you run a test query for /favicon.ico through your live Dev instance and make sure it's served correctly with the expected (non-immutable?) headers?

floatingatoll: Could you run a test query for /favicon.ico through your live Dev instance and make sure it's served correctly with the expected (non-immutable?) headers?

Bugzilla/Install/Filesystem.pm
```diff
@@ -342,7 +353,7 @@ sub FILESYSTEM {
         $attachdir              => DIR_CGI_WRITE,
         $graphsdir              => DIR_CGI_WRITE | DIR_ALSO_WS_SERVE,
         $webdotdir              => DIR_CGI_WRITE | DIR_ALSO_WS_SERVE,
-        $assetsdir              => DIR_CGI_WRITE | DIR_ALSO_WS_SERVE,
+        $assetsdir              => DIR_WS_SERVE,
```
dylanwh: It means that the new assetdir is not writable by the user running the code ('apache' typically).

Bugzilla/Install/Filesystem.pm
```diff
@@ -551,7 +564,6 @@ sub update_filesystem {
 
     _remove_empty_css_files();
     _convert_single_file_skins();
```
dylanwh: I'm not sure.

Bugzilla/Template.pm
dylanwh: it's because TT2 stuff doesn't really handle lists. I suppose I could change it to the other way...

extensions/BugModal/template/en/default/bug_modal/edit.html.tmpl
dylanwh: FILTER html means "escape html characters"

Bugzilla/Install/AssetManager.pm
globau: it's used by jquery-datetimepicker, jquery-ui, and yui3

Bugzilla/Install/AssetManager.pm
```diff
@@ -0,0 +1,224 @@
...
+    my $asset_map = $self->state->{asset_map};
+
+    my $key = File::Spec->abs2rel( $file, $base_dir );
+    return if $asset_map->{$key};
+
+    if ($file =~ /\.(jpe?g|png|gif|ico|woff|js)$/i) {
+        my $ext            = $1;
+        my $digest         = $self->_digest_file($file);
+        my $asset_file     = File::Spec->catfile($asset_dir, "$digest.$ext");
+        cp($file, $asset_file);
```
globau: since you're not checking the result here, you don't care about copying the source file's permission bits.
you should be using copy() instead of cp().

Bugzilla/Install/AssetManager.pm
```diff
@@ -0,0 +1,224 @@
...
+        return sprintf( "url(%s)", $self->asset_file($ref_file, $self->asset_dir));
+    }
+}
+
+sub _new_digest { Digest::SHA->new($SHA_VERSION) }
+
+sub _digest_file {
+    my ($self, $file) = @_;
+    my $digest = $self->_new_digest;
+    $digest->addfile($file);
```
globau: use binary mode here.

Bugzilla/Install/AssetManager.pm
```diff
@@ -0,0 +1,224 @@
...
+    if ($file =~ /\.(jpe?g|png|gif|ico|woff|js)$/i) {
+        my $ext            = $1;
+        my $digest         = $self->_digest_file($file);
+        my $asset_file     = File::Spec->catfile($asset_dir, "$digest.$ext");
+        cp($file, $asset_file);
+        if ($digest eq $self->_digest_file($asset_file)) {
+            $asset_map->{$key} = File::Spec->abs2rel($asset_file, $asset_dir);
+        }
+        else {
+            die "failed to write $asset_file";
```
globau: addfile() in _digest_file() will croak if it can't read the file, so the error here can only mean the file contents were corrupted.

for write failures it would be better to check the result of the file copy and die there (mostly because you can include $! in the error message).

Bugzilla/Install/AssetManager.pm
```diff
@@ -0,0 +1,224 @@
...
+    $digest->add($string);
+    return $digest->hexdigest;
+}
+
+sub _build_base_dir  { Cwd::realpath(File::Spec->rel2abs(bz_locations->{cgi_path})) }
+sub _build_asset_dir {
+    my ($self) = @_;
+    my $dir = Cwd::realpath(File::Spec->rel2abs(bz_locations->{assetsdir}));
+
+    if ($dir && -d $dir) {
```
globau: when is $dir undefined, and why would it be correct to return that as the asset_dir?

Bugzilla/Install/AssetManager.pm
```diff
@@ -0,0 +1,224 @@
...
+
+sub _build_source_dirs {
+    my ($self) = @_;
+    my $base = $self->base_dir;
+
+    return [
+        "$base/skins",
+        "$base/js",
+        grep { -d $_ }
+        glob("$base/extensions/*/web")
```
globau: if $base can contain spaces, you want `glob("'$base/extensions/*/web'")`

Bugzilla/Install/AssetManager.pm
```diff
@@ -0,0 +1,224 @@
...
+    else {
+        $state = {};
+    }
+
+    $state->{asset_map} //= {};
+
+    return $state;
+}
+
+sub _build_json { JSON::XS->new->canonical->utf8 }
```
globau: i doubt you need the overhead of key sorting here

Bugzilla/Template.pm
```diff
@@ -398,7 +398,13 @@ sub css_files {
 
     unshift @requested_css, "skins/yui.css" unless $no_yui;
 
-    my @css_sets = map { _css_link_set($_) } @requested_css;
+    my @css_sets = map { _css_link_set($_) } sort {
+        my $first_a = $a =~ m!js/jquery!;
+        my $first_b = $b =~ m!js/jquery!;
```
globau: as this could be called a lot (because sort) it would be better to use index() for a substring search

Bugzilla/Template.pm
```diff
@@ -429,127 +430,19 @@ sub _css_link_set {
...
-
-    # build minified files
-    my @minified;
-    foreach my $source (@$sources) {
-        next unless -e "$cgi_path/$files{$source}";
-        my $file = $skins_path . '/' . md5_hex($source) . '.js';
-        if (!-e $file) {
-            my $content = read_file("$cgi_path/$files{$source}");
-
-            # minimal minification
```
globau: farewell comment that amuses me

extensions/BugModal/template/en/default/bug_modal/edit.html.tmpl
globau: shouldn't this be the `url` filter?  (repeat for all other occurrences)

template/en/default/global/header.html.tmpl
```diff
@@ -188,14 +182,28 @@
...
+      [% END %]
+    [% ELSE %]
+      [% 
+         SET js_all = [];
+         CALL js_all.import(starting_js_urls);
+         CALL js_all.import(javascript_urls);
+      %]
+      [% FOREACH asset_url = asset_files(js_all) %]
+        [% PROCESS format_js_link %]
+      [% END %]
```
globau: weird mix of styles: [% %] foreach line vs [% line; line; %]

dylanwh: No, href elements must be HTML escaped but don't need to be URI-escaped.

Bugzilla/Template.pm
dylanwh: In the case when the regex doesn't match, the regex is faster than index().
Though with the number of elements (and size of the strings) it doesn't matter either way.

Bugzilla/Install/AssetManager.pm
dylanwh: it's run at checksetup time, so I don't mind. I prefer the results to be consistent. :-)
The order would be random (> 5.10 perl anyway) otherwise.

Bugzilla/Install/AssetManager.pm
dylanwh: It is.

Bugzilla/Install/AssetManager.pm
dylanwh: I'll explicitly return undef, which will cause the type constraint to report an appropriate error.

Bugzilla/Install/AssetManager.pm
dylanwh: replaced with readdir()

extensions/BugModal/template/en/default/bug_modal/edit.html.tmpl
floatingatoll: I wish there was a ```FILTER href```, so that we could always use it in HTML href/src scenarios and be confident we're doing the right thing.

dylanwh:
This is now rebased on mary's css patch as there was one conflict.

Bugzilla/CGI.pm
```diff
@@ -441,12 +441,20 @@ sub header {
         %headers = ('-type' => shift(@_));
     }
     else {
-        %headers = @_;
+        # we always want headers to be in lower case, to avoid
+        # for instance -Cache_Control vs. -cache_control.
+        my %tmp = @_;
+        foreach my $key (keys %tmp) {
+            my $lc_key = lc $key;
+            warn "duplicate header: $key" if exists $headers{$lc_key};
```
floatingatoll: Advise warn "overwriting duplicate header: $key" to clearly state the actual warning.

Bugzilla/CGI.pm
```diff
@@ -441,12 +441,20 @@ sub header {
...
+            $headers{$lc_key} = $tmp{$key};
+        }
     }
 
     if ($self->{'_content_disp'}) {
         $headers{'-content_disposition'} = $self->{'_content_disp'};
     }
+    $headers{'-cache_control'} //= 'no-cache, no-store, must-revalidate';
```
floatingatoll: Please separate this from the content_disposition section with a newline and add a comment indicating why it's the default.

Consider adding 'private'.

Bugzilla/Install/AssetManager.pm
```diff
@@ -0,0 +1,218 @@
...
+    if ( $file =~ /\.(jpe?g|png|gif|ico|woff|js)$/i ) {
+        my $ext        = $1;
+        my $digest     = $self->_digest_file($file) or die "invalid digest for $file";
+        my $asset_file = File::Spec->catfile( $asset_dir, "$digest.$ext" );
+        cp( $file, $asset_file ) or die "failed to copy $file to $asset_file: $!";
+        if ( $digest eq $self->_digest_file($asset_file) ) {
+            $asset_map->{$key} = File::Spec->abs2rel( $asset_file, $asset_dir );
+        }
+        else {
+            die "failed to write $asset_file";
```
floatingatoll: "failed to write" might be more usefully phrased as "checksum mismatch after copying $asset_file".

Bugzilla/Install/AssetManager.pm
```diff
@@ -0,0 +1,218 @@
...
+
+        $content =~ s{(?<!=)url\(([^\)]+)\)}{$self->_css_url_rewrite($1, $file)}eig;
+        my $digest = $self->_digest_string($content);
+        my $asset_file = File::Spec->catfile( $asset_dir, "$digest.css" );
+        write_file( $asset_file, $content ) or die "failed to write $asset_file: $!";
+        if ( $digest eq $self->_digest_file($asset_file) ) {
+            $asset_map->{$key} = File::Spec->abs2rel( $asset_file, $asset_dir );
+        }
+        else {
+            die "failed to write $asset_file";
```
floatingatoll: As above.

Bugzilla/Install/AssetManager.pm
```diff
@@ -0,0 +1,218 @@
...
+        }
+    }
+    elsif ( $file =~ /\.css$/ ) {
+        my $content = read_file($file);
+
+        $content =~ s{(?<!=)url\(([^\)]+)\)}{$self->_css_url_rewrite($1, $file)}eig;
```
floatingatoll: What is this doing, precisely? A before/after doc block in a comment would make a tremendous difference here to the unfamiliar.

Bugzilla/Install/AssetManager.pm
```diff
@@ -0,0 +1,218 @@
...
+        my $digest = $self->_digest_string($content);
```
floatingatoll: Is it safe to trust that _digest_string will never return '/', or at least that the code will fail cleanly in that broken scenario?

Bugzilla/Install/AssetManager.pm
```diff
@@ -0,0 +1,218 @@
...
+        }
+        else {
+            die "failed to write $asset_file";
+        }
+    }
+}
+
+sub compile_all {
+    my ($self) = @_;
+    my $asset_map = $self->state->{asset_map} = {};
```
floatingatoll: You don't seem to use $asset_map at all here, though?

Bugzilla/Install/AssetManager.pm
```diff
@@ -0,0 +1,218 @@
...
+    my $wanted = sub {
+        $self->compile_file($File::Find::name);
+    };
+
+    find( { wanted => $wanted, no_chdir => 1 }, @{ $self->source_dirs } );
+
+    $self->_save_state();
+}
+
+sub _css_url_rewrite {
```
floatingatoll: It would be very encouraging to find that there are tests for this particular rewrite, but I wasn't sure if that's implied by existing tests already or not.

Bugzilla/Install/AssetManager.pm
```diff
@@ -0,0 +1,218 @@
...
+    my ( $self, @files ) = @_;
+
+    return \@files;
+}
+
+sub asset_sri {
+    my ( $self, $asset_file ) = @_;
+    my ($hex) = $asset_file =~ m!([[:xdigit:]]+)\.\w+$!;
+    my $data = pack "H*", $hex;
+    return "sha$SHA_VERSION-" . encode_base64( $data, "" );
```
floatingatoll: Please validate that $SHA_VERSION is SRI-compatible before proceeding here. (256, 384, 512.)

Bugzilla/Install/AssetManager.pm
```diff
@@ -0,0 +1,218 @@
...
+    }
+}
+
+sub _new_digest { Digest::SHA->new($SHA_VERSION) }
+
+sub _digest_file {
+    my ( $self, $file ) = @_;
+    my $digest = $self->_new_digest;
+    $digest->addfile( $file, "b" );
+    return $digest->hexdigest;
```
floatingatoll: Is it safe to assume that ->hexdigest croaks on errors here?

Bugzilla/Install/AssetManager.pm
```diff
@@ -0,0 +1,218 @@
...
+    }
+    else {
+        return $dir;
+    }
+}
+
+sub _build_source_dirs {
+    my ($self)  = @_;
+    my $base    = $self->base_dir;
+    my $ext_dir = "$base/extensions";
```
floatingatoll: I would encourage a blank line here, followed by ```my @dirs;```, so that it's apparent what the next three lines of (relatively dense) code are for, distinct from $base and returnvalue stuff.

Bugzilla/Install/AssetManager.pm
```diff
@@ -0,0 +1,218 @@
...
+    opendir my $ext_dir_handle, $ext_dir or die "unable to open $ext_dir: $!";
+    my @dirs = grep { -d $_ } map {"$ext_dir/$_/web"} grep { !/^\.\.?$/ } readdir $ext_dir_handle;
+    closedir $ext_dir_handle;
+
+    return [ "$base/skins", "$base/js", grep { -d $_ } @dirs ];
+}
+
+sub _build_state_file {
+    my ($self) = @_;
+    return $self->asset_dir . "/state.json";
```
floatingatoll: Either use File::Spec->catdir or don't, but not both.

Bugzilla/Install/AssetManager.pm
```diff
@@ -0,0 +1,218 @@
...
+}
+
+sub _build_state {
+    my ($self) = @_;
+    my $state;
+    if ( open my $fh, '<:bytes', $self->state_file ) {
+        local $/ = undef;
+        my $json = <$fh>;
```
floatingatoll: It's odd to see you writing file-reading code here when (1) most json parsers can read directly from a file, and (2) you sometimes use a perl module to read in files instead of bespoke code.

Bugzilla/Install/AssetManager.pm
```diff
@@ -0,0 +1,218 @@
...
+        close $fh;
+        $state = $self->json->decode($json);
+    }
+    else {
+        $state = {};
+    }
+
+    $state->{asset_map} //= {};
```
floatingatoll: A comment might be valuable here to indicate what could result in this being necessary.

Bugzilla/Install/AssetManager.pm
```diff
@@ -0,0 +1,218 @@
...
+    return $state;
+}
+
+sub _build_json { JSON::XS->new->canonical->utf8 }
+
+sub _save_state {
+    my ($self) = @_;
+    open my $fh, '>:bytes', $self->state_file
+        or die "unable to write state file: $!";
+    print $fh $self->json->encode( $self->state );
```
floatingatoll: 'or die'

Bugzilla/Install/AssetManager.pm
```diff
@@ -0,0 +1,218 @@
...
+    close $fh;
```
floatingatoll: also 'or die'

Bugzilla/Install/Filesystem.pm
```diff
@@ -31,6 +31,7 @@ use File::Path;
 use File::Basename;
 use File::Copy qw(move);
 use File::Spec;
+use File::stat;
```
floatingatoll: Seriously, they made this *lowercase*? Augh. Unless it's a typo. Probably isn't!

Bugzilla/Install/Filesystem.pm
```diff
@@ -81,12 +82,16 @@ EOT
 
 use constant HT_ASSETS_DIR => <<'EOT';
 # Allow access to .css and js files
-<FilesMatch \.(css|js)$>
-  Allow from all
+<FilesMatch state\.json$>
```
floatingatoll: <Files "state.json"> would probably be simpler. Why is this content considered unsafe?

Bugzilla/Install/Filesystem.pm
```diff
@@ -81,12 +82,16 @@ EOT
...
+    Deny from all
 </FilesMatch>
 
+FileETag None
```
floatingatoll: I assume an Etag is set elsewhere in the code (even just the string "1" would be fine), as otherwise this may cause undesirable caching behavior from clients.

Bugzilla/Install/Filesystem.pm
```diff
@@ -446,8 +451,13 @@ sub FILESYSTEM {
         "$webdotdir/.htaccess"       => { perms => WS_SERVE,
                                           contents => HT_WEBDOT_DIR },
         "$assetsdir/.htaccess"       => { perms => WS_SERVE,
-                                          contents => HT_ASSETS_DIR },
+                                          contents  => HT_ASSETS_DIR },
```
floatingatoll: This whitespace change is slightly confusing, and may be accidental.

Bugzilla/Install/Filesystem.pm
```diff
@@ -446,8 +451,13 @@ sub FILESYSTEM {
...
     );
+    my $mtime = stat(__FILE__)->mtime;
+    foreach my $file (keys %htaccess) {
+        my $file_stat = stat($file);
+        $htaccess{$file}{overwrite} = $file_stat && $mtime > $file_stat->mtime;
```
floatingatoll: I don't like this optimization, because mtime can't be trusted in a lot of web content scenarios. Is this saving time *only* at buildsetup time? Wouldn't a digest comparison be more appropriate?

Bugzilla/Template.pm
```diff
@@ -401,7 +401,13 @@ sub css_files {
 
     unshift @requested_css, "skins/yui.css" unless $no_yui;
 
-    my @css_sets = map { _css_link_set($_) } @requested_css;
+    my @css_sets = map { _css_link_set($_) } sort {
```
floatingatoll: I don't mind that the sort may be inefficient, but I'd love to see a comment explaining why this sort is required.

PR #76, opened 2017-05-11
Bug 1361376 - Bugzilla::WebService::Bug: render_comment() should not pass a literal template string to call quoteUrls()

dylanwh:


floatingatoll:
Why shouldn't it?

dylanwh:
Because it's wasteful? Also, supporting that from the template preload patch is hard.

PR #77, opened 2017-05-11
Bug 1363736 - Ensure http configs (.htaccess) are forward-compatible …

dylanwh:
…with newer versions of apache


indygreg:
LGTM!

PR #79, opened 2017-05-16
give each cron job a name, so that they don't accumulate

dylanwh:
No bug -- since this is just a dev environment fix.


vagrant_support/cron.yml
```diff
@@ -1,44 +1,51 @@
 - name: set cron mailto
   cron:
+    name: set cron mailto
```
floatingatoll: It bugs me that they can't inherit the name from the parent, but I guess that's how ansible is.

PR #82, opened 2017-05-18
Bug 1365340 - Add /new-bug (new_bug.cgi)

SebastinSanty:


template/en/default/bug/new_bug.html.tmpl
```diff
@@ -0,0 +1,49 @@
...
+  #%]
+
+[% PROCESS "global/field-descs.none.tmpl" %]
+
+[% title = BLOCK %]Enter [% terms.Bug %] [% END %]
+
+[% PROCESS global/header.html.tmpl
+  title = title
+  generate_api_token = 1
+  yui = [ 'calendar', 'datatable', 'button' ]
```
dylanwh: don't need this, instead set no_yui = 1 here

template/en/default/bug/new_bug.html.tmpl
```diff
@@ -0,0 +1,49 @@
...
+<table>
+<tbody>
+  <tr>
+    <td colspan="4">
+      Coming Soon!
+    </td>
+  </tr>
+</tbody>
+</table>
+
```
dylanwh: It doesn't matter much, but probably this doesn't need to be a table.

template/en/default/bug/new_bug.html.tmpl
```diff
@@ -0,0 +1,49 @@
...
+  title = title
+  generate_api_token = 1
+  yui = [ 'calendar', 'datatable', 'button' ]
+  style_urls = [ 'skins/standard/attachment.css',
+                 'skins/standard/enter_bug.css',
+                 'skins/custom/create_bug.css' ]
+  javascript_urls = [ "js/attachment.js", "js/util.js",
+                      "js/field.js", "js/TUI.js", "js/bug.js",
+                      "js/create_bug.js" ]
+  onload = "init();"
```
dylanwh: remove style_urls, javascript_urls and onload for now. None of them are needed.
You'll eventually want the ones that bug_modal uses.

new_bug.cgi
```diff
@@ -0,0 +1,57 @@
...
+
+my $cgi = Bugzilla->cgi;
+my $dbh = Bugzilla->dbh;
+my $template = Bugzilla->template;
+my $vars = {};
+
+my $format = $template->get_format("bug/new_bug",
+                                   scalar $cgi->param('format'), 
+                                   scalar $cgi->param('ctype'));
+
```
dylanwh: I think we can avoid the whole "get_format" thing.

$template->process("bug/new_bug.html.tmpl", ...);

template/en/default/bug/new_bug.html.tmpl
```diff
@@ -0,0 +1,49 @@
...
+  # The Original Code is the Bugzilla Bug Tracking System.
+  #
+  # The Initial Developer of the Original Code is Netscape Communications
+  # Corporation. Portions created by Netscape are
+  # Copyright (C) 1998 Netscape Communications Corporation. All
+  # Rights Reserved.
+  #
+  # Contributor(s): Sebastin Santy <sebastinssanty@gmail.com>
+  #
+  #%]
```
dylanwh: This is the old (MPL v1 header) -- it should use the new (MPL v2) header -- I think this is the case in Bugzilla/CGI/ContentSecurityPolicy.pm for instance.

new_bug.cgi
```diff
@@ -0,0 +1,57 @@
...
+print $cgi->header($format->{'ctype'});
```
dylanwh: print $cgi->header();

.htaccess
```diff
@@ -57,6 +57,9 @@ RewriteRule ^__version__$ version.json [L]
 # heartbeat.cgi returns 200 if the DB and memcached are both working, and 500 otherwise.
 RewriteRule ^__heartbeat__$ heartbeat.cgi [L]
 
+# New single page interface for filing bugs
+RewriteRule ^new-bug$ new_bug.cgi [L]
```
dylanwh: I think this needs to be [L,QSA]:
https://httpd.apache.org/docs/current/rewrite/flags.html#flag_qsa

floatingatoll:
I encourage ^new[-_]bug$ since most of us think Bugzilla uses _, not -, in
all URLs.

template/en/default/bug/new_bug.html.tmpl
```diff
@@ -0,0 +1,21 @@
...
+
+[% title = BLOCK %]Enter [% terms.Bug %] [% END %]
+
+[% PROCESS global/header.html.tmpl
+  title = title
+  generate_api_token = 1
+  no_yui = 1
+%]
+
+
```
dylanwh: remove extra white space and use a <p> tag here, I think.

new_bug.cgi
```diff
@@ -0,0 +1,50 @@
...
+my $user = Bugzilla->login(LOGIN_REQUIRED);
+
+my $cgi = Bugzilla->cgi;
+my $template = Bugzilla->template;
+my $vars = {};
+
+print $cgi->header();
+$template->process("bug/new_bug.html.tmpl",
+                    $vars)
+  || ThrowTemplateError($template->error());
```
dylanwh: Using the named 'or' operator is a better style, write "or ThrowTemplateError(...)" here,
and run the whole file through perltidy -pbp. Otherwise it lgtm.

PR #85, opened 2017-05-26
Bug 1361439 - Create dockerflow-compliant container that runs a BMO web head

dylanwh:


docker_files/httpd.conf
```diff
@@ -0,0 +1,98 @@
+ServerTokens OS
```
dylanwh: I think there is a lot of stuff from production that needs to be added to this.
@klibby or @gozer can you look over this httpd.conf?

floatingatoll: Webops uses 'Prod' rather than 'OS' here, in case that simplifies things.

.dockerignore
```diff
@@ -0,0 +1,18 @@
...
+pm_to_blib
+template_cache
+vagrant_support
+\#*\#
+*/\#*\#
+*/*/\#*\#
+*/*/*/\#*\#
+*/*/*/*/\#*\#
+.DS_Store
+*/.DS_Store
```
floatingatoll: Docker technically is trying to support ```**/.DS_Store```, but they have 3 implementation bugs right now around it, and docker-compose wasn't handling it right in January, so this is fine, if aesthetically grumpy.

Bugzilla/Install/Localconfig.pm
```diff
@@ -40,10 +40,7 @@ our @EXPORT_OK = qw(
 
 sub _sensible_group {
     return '' if ON_WINDOWS;
-    my @groups     = qw( apache www-data _www );
```
floatingatoll: This is a permanent change to how localconfig works. Is it called out sufficiently in docs or whatever, that it will stop defaulting to group apache *over* EGID?

Bugzilla/Install/Localconfig.pm
```diff
@@ -297,7 +294,7 @@ sub update_localconfig {
         print colored(install_string('lc_new_vars', { localconfig => $filename,
                                                       new_vars => wrap_hard($newstuff, 70) }),
                       COLOR_ERROR), "\n";
-        exit;
+        exit unless $params->{use_defaults};
```
floatingatoll: Why do we newly permit fallthrough here?

README.rst
```diff
@@ -75,3 +86,83 @@ Most of the cron jobs and the jobqueue daemon are running.  The push connector
...
+This repository is also a runnable docker container.
+
+Container Arguments
+-------------------
+
+Currently, the entry point takes a single command argument.
+This can be **httpd** or **shell**.
+
+httpd
+    This will start apache listening for connections on ``$PORT``
```
floatingatoll: Trailing period needed here.

README.rst
```diff
@@ -75,3 +86,83 @@ Most of the cron jobs and the jobqueue daemon are running.  The push connector
...
+  The DNS name or IP address of the host that the database server runs on.
+
+BMO_db_name
+  The name of the database
+
+BMO_db_user
+  The database user to connect as.
+
+BMO_db_pass
+  The password for the user above
```
floatingatoll: Trailing period, ditto db_name and probably others.

README.rst
```diff
@@ -75,3 +86,83 @@ Most of the cron jobs and the jobqueue daemon are running.  The push connector
...
+
+BMO_site_wide_secret
+  This secret key is used by your installation for the creation and
+  validation of encrypted tokens. These tokens are used to implement
+  security features in Bugzilla, to protect against certain types of attacks.
+  It's very important that this key
```
floatingatoll: Rewrap needed here?

README.rst
```diff
@@ -75,3 +86,83 @@ Most of the cron jobs and the jobqueue daemon are running.  The push connector
...
+  is kept secret. It also must be very long.
```
floatingatoll: Is 'very long' validated and enforced? How?

README.rst
```diff
@@ -75,3 +86,83 @@ Most of the cron jobs and the jobqueue daemon are running.  The push connector
...
+
+BMO_inbound_proxies
+  This is a list of IP addresses that we expect proxies to come from.
```
floatingatoll: Why do we care if there are proxies?

README.rst
```diff
@@ -75,3 +86,83 @@ Most of the cron jobs and the jobqueue daemon are running.  The push connector
...
+  This can be '*' if only the load balancer can connect to this container.
```
floatingatoll: I can deduce that this means you'll trust the XFF header the list of BMO_inbound_proxies, but it might be useful to be more clear about that link.

README.rst
```diff
@@ -75,3 +86,83 @@ Most of the cron jobs and the jobqueue daemon are running.  The push connector
...
+
+BMO_memcached_namespace
+  The global namespace for the memcached server
+
+BMO_memcached_servers
+  A list of memcached servers (ip addresses or host names)
```
floatingatoll: Zero or more, or one or more?

README.rst
```diff
@@ -75,3 +86,83 @@ Most of the cron jobs and the jobqueue daemon are running.  The push connector
...
+  security features in Bugzilla, to protect against certain types of attacks.
+  It's very important that this key
+  is kept secret. It also must be very long.
+
+BMO_inbound_proxies
+  This is a list of IP addresses that we expect proxies to come from.
+  This can be '*' if only the load balancer can connect to this container.
+
+BMO_memcached_namespace
+  The global namespace for the memcached server
```
floatingatoll: "the memcached server" implies that there can only be one, when the following option says 1..N.

checksetup.pl
```diff
@@ -53,7 +53,8 @@ BEGIN
                      'cpanm:s', 'check-modules',
                      'make-admin=s', 'reset-password=s', 'version|V',
                      'no-assets',
-                     'no-permissions|p');
+                     'default-localconfig',
+                     'no-database', 'no-permissions|p');
```
floatingatoll: If 'no-database' was worth combining with 'no-permissions', should 'no-assets' be rolled into this single line as well?

docker_files/httpd.conf
```diff
@@ -0,0 +1,98 @@
...
+LoadModule mime_magic_module modules/mod_mime_magic.so
+LoadModule expires_module modules/mod_expires.so
+LoadModule deflate_module modules/mod_deflate.so
+LoadModule headers_module modules/mod_headers.so
+LoadModule usertrack_module modules/mod_usertrack.so
+LoadModule setenvif_module modules/mod_setenvif.so
+LoadModule mime_module modules/mod_mime.so
+LoadModule dav_module modules/mod_dav.so
+LoadModule status_module modules/mod_status.so
+LoadModule autoindex_module modules/mod_autoindex.so
```
floatingatoll: Is this a default config of some kind? This list of modules is somewhat excessive.

docker_files/init.pl
```diff
@@ -0,0 +1,81 @@
...
+my %localconfig = (webservergroup => 'app');
+
+my %override = (
+    'inbound_proxies'     => 1,
+    'memcached_namespace' => 1,
+    'memcached_servers'   => 1,
+    'shadowdb'            => 1,
+    'shadowdbhost'        => 1,
+    'shadowdbport'        => 1,
+    'shadowdbsock'        => 1
```
floatingatoll: apache_size_limit? I didn't fully audit the entire documented list of BMO_ env options, but it seemed like one that should be parsed.

dylanwh: So some settings are... backwards, and stored in data/params. Only those backwards setting get the 'override' treatment. apache_size_limit is a localconfig option and thus needs no special voo-doo. Those settings should have never been in data/params to begin with.

Bugzilla/Install/Localconfig.pm
dylanwh: The default value is never used in current production, as localconfig is always present. A comment indicating that this is a divergence from upstream is probably needed though.

docker_files/httpd.conf
dylanwh: I'll need to look up what ServerTokens even does...

Bugzilla/Install/Localconfig.pm
dylanwh: So that we can get some default options when the localconfig file doesn't exist.

README.rst
dylanwh: bmo used to run in a way that meant you could hit it directly from the outside world.

globau: "When inbound traffic to Bugzilla goes through a proxy, Bugzilla thinks that the IP address of the proxy is the IP address of every single user. If you enter a comma-separated list of IPs in this parameter, then Bugzilla will trust any `X-Forwarded-For` header sent from those IPs, and use the value of that header as the end user's IP address."

PR #86, opened 2017-05-30
Bug 1368739 - Make OrangeFactor extension work with SSO

edmorley:
OrangeFactor is now behind SSO, so cross-origin requests to the API must use `withCredentials` to ensure the session cookie is sent if it exists. OrangeFactor's API permits this via appropriately set `Access-Control-Allow-{Origin,Credentials}` headers.

Users will also now be reminded about the need to sign in via OrangeFactor first.

Unfortunately there's no way to catch the SSO HTTP 302 separately from other failure modes, since:
(a) the `XMLHttpRequest` spec doesn't allow for telling if a request redirected
(b) the workaround (sniffing content-type of the response and seeing if it's `text/html` rather than `application/json`) isn't possible since BMO's CSP blocks the request to auth0.com.

...however this extension is going to be replaced once OrangeFactor is decommissioned, so meh.

The redundant hiding of `#orange-graph` has also been removed.

### DIFF
--- a/extensions/OrangeFactor/web/js/orange_factor.js
+++ b/extensions/OrangeFactor/web/js/orange_factor.js
@@ -74,6 +74,9 @@ $(function() {
         var bugId = document.forms['changeform'].id.value;
         var request = {
             dataType: "json",
+            xhrFields: {
+                withCredentials: true
+            },
             url: "https://brasstacks.mozilla.com/orangefactor/api/count?" +
                  "bugid=" + encodeURIComponent(bugId) + "&tree=trunk"
         };
@@ -82,8 +85,7 @@ $(function() {
                 getOrangeCount(data);
             })
             .fail(function() {
-                $('#graph-count').hide();
-                $('#orange-graph').hide()
+                $('#orange-count').text('Please sign into OrangeFactor first');
             });
     }
 


dylanwh:
if it wasn't legacy, we could add a CSP for auth0.com -- the policies are per-resource, so show_bug.cgi could allow that specifically. But I agree it's not worth the minor risk.

floatingatoll:
It might be possible to block auth-redirects on the OrangeFactor API URL endpoints, so that they return 401 instead of 302. Would that help?

extensions/OrangeFactor/web/js/orange_factor.js
```diff
@@ -82,8 +85,7 @@ $(function() {
                 getOrangeCount(data);
             })
             .fail(function() {
-                $('#graph-count').hide();
-                $('#orange-graph').hide()
+                $('#orange-count').text('Please sign into OrangeFactor first');
```
floatingatoll: It would be really stellar if this included a sign-in link, or even just a link to the OF homepage.

dylanwh:
Just a note, the CSP for the show bug page would changed around here: https://github.com/mozilla-bteam/bmo/blob/bc705ae4c2fb5b4a22c1c1594ea2791545d12cc4/Bugzilla/CGI.pm#L70

edmorley:
> It would be really stellar if this included a sign-in link, or even just a link to the OF homepage.

There's a link to the right of the text due to the existing layout:
bmo-orangefactor

> It might be possible to block auth-redirects on the OrangeFactor API URL endpoints, so that they return 401 instead of 302. Would that help?

Yeah I agree this would be the preferred long-term solution, but I don't think it's worth it for an EOL site.

I'm pretty sure there are only single-digit users that actually have enabled this addon and use it regularly.

edmorley:
And to add - the replacement for OrangeFactor will use Treeherder's API, so if it was still desired to create an equivalent extension for BMO, it wouldn't need to use auth at all, since Treeherder's API is public.

PR #88, opened 2017-06-02
Bug 1348439 - History item "reporter accessible: true" is confusing

umohm1:


extensions/BugModal/template/en/default/bug_modal/activity_stream.html.tmpl
```diff
@@ -311,6 +311,8 @@
           <a href="#c[% change.comment.count FILTER none %]">
             Comment [% change.comment.count FILTER none %]</a> is private:
           [%+
+        ELSIF change.fieldname == 'reporter_accessible';
+          "";
```
dylanwh: nit: Probably should leave a comment here as to why this is empty.

PR #90, opened 2017-06-06
Bug 1365342 - Extract the preview comment functionality and make it a reusable template

SebastinSanty:


template/en/default/bug/new_bug.html.tmpl
```diff
@@ -5,15 +5,27 @@
...
 [% title = BLOCK %]Enter [% terms.Bug %] [% END %]
 
-[% PROCESS global/header.html.tmpl
-  title = title
-  generate_api_token = 1
-  no_yui = 1
-%]
+[% PROCESS global/variables.none.tmpl %]
+[% IF !header_done %]
+  [% PROCESS bug_modal/common-header.html.tmpl %]
```
dylanwh: as discussed on irc: this needs quoting because of the -, so
```
[% PROCESS "bug_modal/common-header.html.tmpl" %]
```

dylanwh: Alternatively, we could use _'s. I think _ is slightly more common for us than -, and I guess focusing on a single style would be nice. :-)

SebastinSanty: Yes, I'll heed to that.

SebastinSanty:
@dylanwh Can you check, if this was the functionality desired by the bug. I'll reduce the common-* files accordingly after your review.

extensions/BugModal/web/bug_modal.js
```diff
@@ -1335,7 +1333,7 @@ function confirmUnsafeURL(url) {
 // fix url after bug creation/update
 if (history && history.replaceState) {
     var href = document.location.href;
-    if (!href.match(/show_bug\.cgi/)) {
+    if (!href.match(/new-bug/) && !href.match(/show_bug\.cgi/)) {
```
dylanwh: instead of checking the url, let's add an element with a data attribute and check that.
Not sure what element this should be, but let's chat about it tomorrow.

dylanwh: Perfect! check if bug-modal is one of the classes of the body element!

extensions/BugModal/template/en/default/bug_modal/common_header.html.tmpl
```diff
@@ -0,0 +1,84 @@
...
+    "skins/custom/bug_groups.css",
+    "js/jquery/plugins/datetimepicker/datetimepicker.css",
+    "js/jquery/plugins/contextMenu/contextMenu.css"
+  );
+%]
+
+[% javascript = BLOCK %]
+
+  [%# expose useful data to js %]
+
```
dylanwh: I know this how it currently works, but it would be interesting to see if we can add this to the (template toolkit) varaible js_BUGZILLA, which gets serialized automatically.
you'll need to grep the code to see if that's possible, and this is by no means a requirement.

extensions/BugModal/template/en/default/bug_modal/new_comment.html.tmpl
```diff
@@ -13,14 +13,14 @@
...
       You are not allowed to make an additional comment on this [% terms.bug %].
     </div>
     [% RETURN %]
   [% END %]
 
-  [% IF user.is_insider %]
+  [% IF bug && user.is_insider %]
     <div id="add-comment-private"
       title="Make comment visible only to members of the '[% Param('insidergroup') FILTER html %]' group"
     >
```
dylanwh: There are a few more items around the comment that should be hidden:
remove these options

extensions/BugModal/web/bug_modal.js
```diff
@@ -1011,8 +1011,6 @@ $(function() {
     $('#product, #component')
         .change(function(event) {
             $('.set-default-container').show();
-            $('#set-default-assignee').prop('checked', $('#assigned_to').val() == BUGZILLA.default_assignee).change();
-            $('#set-default-qa-contact').prop('checked', $('#qa_contact').val() == BUGZILLA.default_qa_contact).change();
             slide_module($('#module-people'), 'show');
```
dylanwh: so where did these get moved to? we can't just delete them from bug_modal.js can we?

extensions/BugModal/template/en/default/bug_modal/common_new_comment.html.tmpl
```diff
@@ -0,0 +1,55 @@
...
+  # bug: bug object
+  #%]
+
+<div id="add-comment">
+
+  [% IF bug && !bug.check_can_change_field('longdesc', 0, 1) %]
+    <div id="new-comment-notice">
+      You are not allowed to make an additional comment on this [% terms.bug %].
+    </div>
+    [% RETURN %]
```
dylanwh: it's more efficient to use [% ELSE %] than an early return.
RETURN is implemented in terms of raising an exception.

template/en/default/bug/new_bug.html.tmpl
```diff
@@ -5,16 +5,29 @@
...
-  no_yui = 1
-%]
-
+[% IF !header_done %]
+  [% PROCESS bug_modal/common_header.html.tmpl 
+    title = title
+  %]
+  [% PROCESS global/header.html.tmpl %]
+  [% header_done = 1 %]
+[% END %]
```
dylanwh: the header_done stuff is from post_bug/show_bug because they do... weird things with the markup. Remove the header_done conditions -- this template is only meant to be called on its own

dylanwh: also, this isn't a problem here -- but be cautious of title -- it is presumed to be already html escaped.

SebastinSanty: Yes, I'll correct the header_done conditions

PR #92, opened 2017-06-06
Skipping some steps in ansible for rapid debugging

SebastinSanty:


SebastinSanty:
Use the following command for rapid debugging:
```
vagrant rsync && BMO_ANSIBLE_LAZY=1 vagrant provision web
```

Waiting fo #93 to be merged before I can put the above command in README.rst.

SebastinSanty:
Another approach to this would be to have an update provisioner script, as this approach has multiple `LAZY` conditionals, which makes it ugly.

Using named provisions:
```
config.vm.provision "update", type: "ansible" do |ansible|
    ansible.playbook = "update.yml"
  end
```
which can be called by: 
```
vagrant provision --provision-with update
````

**Comment by dylanwh** [comment]
ooh, I like this option!
@SebastinSanty  -- do it that way -- update.yml for fast updates. 
much easier to maintain (altough some tasks will need to be extracted from playbook.yml)

File: Vagrantfile [hunk 1]
```diff
@@ -36,7 +36,7 @@ Vagrant.configure('2') do |config|
   # For a complete reference, please see the online documentation at
   # https://docs.vagrantup.com.
 
-  config.vm.provision 'ansible_local', run: 'always' do |ansible|
+  config.vm.provision 'main', type: 'ansible', run: 'always' do |ansible|
```
dylanwh: what's the reason for this change?

vagrant_support/checksetup.yml
```diff
@@ -18,7 +20,8 @@
     owner: vagrant
     group: vagrant
     mode: 0644
-
+  when: LAZY == 0
+  
 - name: run checksetup
   become: false
   shell: sg apache -c '/usr/local/bin/bmo-checksetup'
```
dylanwh: To save even more time, we should pass --no-templates to checksetup.
It's probably safe to always do this.

Vagrantfile
SebastinSanty: It is to make it known that this is the main script, as there is another update script below. Not deemed necessary, but just to avoid confusion.

dylanwh: let's change this to ansible_local for both -- otherwise people have to also install ansible on their machines.

vagrant_support/checksetup.yml
```diff
@@ -18,10 +20,11 @@
     owner: vagrant
     group: vagrant
     mode: 0644
+  when: LAZY == 0
 
 - name: run checksetup
   become: false
-  shell: sg apache -c '/usr/local/bin/bmo-checksetup'
+  shell: sg apache -c '/usr/local/bin/bmo-checksetup --no-templates'
```
dylanwh: This won't work, although maybe it should.  You need to look at vagrant_support/bmo-checksetup.j2

dylanwh: to make this work, you could pass "$@" to checksetup.pl line

dylanwh: (including the quotes)

PR #95, opened 2017-06-07
Bug 1370973 - Add support for CONNECT proxy (add LWP::Protocol::connect)

dylanwh:
This patch:

1. adds LWP::Protocol::connect
2. moves LWP::UserAgent (also known as 'LWP') and Mozilla::CA from Bitly (which required the specific version) to the main, always-required part of Makefile.PL. 
3. Removes LWP::UserAgent from the 'updates' feature
4. As auth_delegation now doesn't require anything that isn't always present, there is no reason to define it as a 'feature' in Makefile.PL so that is also removed.

I've also pushed this to the 'carton' branch, and started a new build of the vendored dependencies. I'll update this bug with a link to build results shortly. (probably it will take an hour to finish building).

### DIFF
--- a/Makefile.PL
+++ b/Makefile.PL
@@ -49,11 +49,14 @@ my %requires = (
     'Email::Send'              => '1.911',
     'File::Slurp'              => '9999.13',
     'JSON::XS'                 => '2.01',
+    'LWP::Protocol::connect'   => 0,
+    'LWP::UserAgent'           => '5.835',
     'List::MoreUtils'          => $^V > v5.10.1 ? '0.418' : '0.22',
     'Math::Random::ISAAC'      => '1.0.1',
     'Module::Metadata'         => '1.000033',
     'Module::Runtime'          => 0,
     'Moo'                      => 2,
+    'Mozilla::CA'              => 0,
     'Parse::CPAN::Meta'        => '1.44',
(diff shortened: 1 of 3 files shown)


Makefile.PL
```diff
@@ -49,11 +49,14 @@ my %requires = (
     'Email::Send'              => '1.911',
     'File::Slurp'              => '9999.13',
     'JSON::XS'                 => '2.01',
+    'LWP::Protocol::connect'   => 0,
```
floatingatoll: This is the only line of this patch I'm truly comfortable saying "Yes, this looks correct, and I understand what will happen here."

dylanwh:
Here is the resulting change in dependencies from this patch:
https://github.com/mozilla-bteam/carton-bundles/commit/c20ac8379e16cc58452175b10adbebe229dfa20f

dylanwh:
You have to click "approve" for me to be able to merge this now. :-)

floatingatoll:
You're welcome to make me a reviewer with write access. I can sign a CLA or
chain MoCo or whatever :)

dylanwh:
I fixed it!
Reviewers now have the ability to push to master -- which is required for this to work.
I renamed 'devs' to 'peers' and made that group protect 'production'

PR #99, opened 2017-06-13
Bug 1372363 - Fix BMO defaults (bug modal pref, needinfo exclusions)

dylanwh:
By default, include needinfo everywhere.
Also set the default for bug modal to on.

### DIFF
--- a/extensions/Needinfo/Extension.pm
+++ b/extensions/Needinfo/Extension.pm
@@ -39,6 +39,8 @@ sub install_update_db {
     print "Creating needinfo flag ... " . 
           "enable the Needinfo feature by editing the flag's properties.\n";
 
+    # inclusions 0:0 maps to __ANY__ : __ANY__ in the UI,
+    # meaning needinfo is enabled for all products and components by default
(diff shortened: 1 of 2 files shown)


floatingatoll:
As a less-familiar participant, I wasn't able to determine what actual impact "include needinfo everywhere" will have on BMO itself. Is it more clear to you all?

extensions/Needinfo/Extension.pm
```diff
@@ -39,8 +39,6 @@ sub install_update_db {
     print "Creating needinfo flag ... " . 
           "enable the Needinfo feature by editing the flag's properties.\n";
 
-    # Initially populate the list of exclusions as __Any__:__Any__ to
```
floatingatoll: This comment is removed but not replaced, leaving the weird ```['0:0']``` inclusion less-documented than the exclusion it replaces.

dylanwh:
This does not change the behavior of production, as it's about the default settings that happen at install time (which is only relevant for contributors)

PR #100, opened 2017-06-13
Bug 1355169 - Add rate-limiting to show_bug.cgi and rest.cgi

umohm1:


Bugzilla.pm
```diff
@@ -155,7 +156,7 @@ sub init_page {
     }
 
     # If Bugzilla is shut down, do not allow anything to run, just display a
-    # message to the user about the downtime and log out.  Scripts listed in 
+    # message to the user about the downtime and log out.  Scripts listed in
```
floatingatoll: If at all possible, please extract the whitespace cleanup into its own commit. It's okay if it's in the pull request, but it shouldn't be mixed with a functionality-altering commit unless that specific code is being modified.

Bugzilla.pm
```diff
@@ -54,6 +54,7 @@ use File::Basename;
 use File::Spec::Functions;
 use Safe;
 use Sys::Syslog qw(:DEFAULT);
+use JSON::XS qw(decode_json);
```
floatingatoll: Is BMO generally fine with modules that require compilation? It seems reasonable to me, but in case that's a topic y'all have past decisions on.

Bugzilla.pm
```diff
@@ -767,6 +768,23 @@ sub elastic {
     $class->process_cache->{elastic} //= Bugzilla::Elastic->new();
 }
 
+sub check_rate_limit {
+    my ($class, $name, $id) = @_;
+    my $params = Bugzilla->params;
+    if ($params->{rate_limit_active}) {
```
floatingatoll: Moving this specific test up one level to the function calling ```check_rate_limit``` would improve the performance of rate limiting checks, very very slightly.

Normally I wouldn't suggest micro-optimizations, but since this function is used in the anonymous user request paths, if there's heavy load, any improvement may be worth considering.

If you choose to pursue this, you'd need a comment indicating that the caller is responsible for that test.

Bugzilla.pm
```diff
@@ -767,6 +768,23 @@ sub elastic {
...
+        my $rules = decode_json($params->{rate_limit_rules});
```
floatingatoll: Does an acceptable user experience occur if ```{rate_limit_rules}``` is invalid JSON?

Bugzilla.pm
```diff
@@ -767,6 +768,23 @@ sub elastic {
...
+        my $limit = $rules->{$name};
+        unless ($limit) {
+             warn "no rules for $name!";
```
floatingatoll: Can the rules be tested for this property at startup time, rather than logging once per request?

Bugzilla.pm
```diff
@@ -767,6 +768,23 @@ sub elastic {
...
+             return 0;
+        }
+        if (Bugzilla->memcached->should_rate_limit("$name:$id", @$limit)) {
```
floatingatoll: Does an acceptable user experience occur if $limit is not an array reference?

extensions/Needinfo/template/en/default/bug/needinfo.html.tmpl
```diff
@@ -92,11 +92,13 @@ $(function() {
       $('#needinfo_from_container').removeClass('bz_default_hidden');
       $('#needinfo_from').prop('disabled', false);
       $('#needinfo_role_identity').text('');
+      $('needinfo_from_textfield').prop('required', true);
```
floatingatoll: Are these changes meant to be included in this pull request? (They might go away if another one is merged, perhaps?)

t/css.t
```diff
@@ -0,0 +1,37 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
```
floatingatoll: Is this file part of rate limiting, or another open / in-flight pull request?

template/en/default/admin/params/admin.html.tmpl
```diff
@@ -23,6 +23,14 @@
    desc = "Set up account policies"
 %]
 
+[% rate_limit_rules_desc = BLOCK %]
+This parameter is a json object that allows for overriding the default rate limit.
+Each key takes an array in the form of [MAX_RATE, SECONDS] where MAX_RATE is the maximum
+number of requests that can occur over SECONDS. The default is [75, 60] or 75 requests
+over 60 seconds. Valid keys are <code>get_b[%''%]ug</code> which covers JSONRPC, XMLRPC, REST and BZAPI single
```
floatingatoll: As an outside, I'd love to understand more about what these ```[%''%]``` blocks are defending against.

template/en/default/admin/params/admin.html.tmpl
```diff
@@ -23,6 +23,14 @@
    desc = "Set up account policies"
 %]
 
+[% rate_limit_rules_desc = BLOCK %]
+This parameter is a json object that allows for overriding the default rate limit.
+Each key takes an array in the form of [MAX_RATE, SECONDS] where MAX_RATE is the maximum
```
floatingatoll: Consider some rewording here for flow:

"This parameter is a json object. It must have one or more valid keys, whose values are each an array [M,S]. There are two valid keys: <code>get; <code>show."

template/en/default/admin/params/admin.html.tmpl
```diff
@@ -23,6 +23,14 @@
...
+number of requests that can occur over SECONDS. The default is [75, 60] or 75 requests
```
floatingatoll: The syntax description above does not offer any way for the user to select the default.

template/en/default/global/user-error.html.tmpl
```diff
@@ -1812,6 +1812,10 @@
     The token you submitted does not exist, has expired, or has
     been canceled.
 
+  [% ELSIF error == "rate_limit" %]
+    [% title = "Rate Limit Exceeded" %]
+    You have exceeded the rate limit.
```
floatingatoll: We may wish to consider further wording here. This is a one-time opportunity to communicate to a distracted coder that we need them to behave more politely towards our instance. We might offer a reasonable default rate limit, request they set the user-agent to their email address, and encourage them to try again after fifteen minutes. (It's okay to open this as a separate issue, as bonus wording shouldn't block shipping a rate limit.)

Bugzilla.pm
dylanwh: that's already a dependency.

floatingatoll: Oh, sigh, you're right. I'm sorry.

template/en/default/admin/params/admin.html.tmpl
dylanwh: A stupid sanity check that prevents us from talking about "bugs"

Bugzilla.pm
dylanwh: It isn't allowed to be invalid (no way of saving an invalid value) although maybe covering that is a good idea.

Bugzilla.pm
dylanwh: Also checked by the config layer.

Bugzilla.pm
dylanwh: It can change at any time -- we could cache it, but that would require clearing when the params are changed -- possibly memcached get_config() could work here. I'll talk to @umohm1  about that.

dylanwh: At the current time, there is no support for caching param values. So we'll deal with the performance for now.

dylanwh: See: https://bugzilla.mozilla.org/show_bug.cgi?id=1355463

template/en/default/admin/params/admin.html.tmpl
umohm1: I'll make the change.

template/en/default/admin/params/admin.html.tmpl
umohm1: Could you clarify a bit on this ?

Bugzilla/Config/Admin.pm
```diff
@@ -44,11 +47,37 @@ sub get_param_list {
...
+sub check_rate_limit_rules {
+    my $rules = shift;
+
+    my $val = eval { decode_json($rules) };
+    return "failed to parse json" unless defined $val;
+    return "value is not HASH"    unless ref $val && ref($val) eq 'HASH';
+    return "rules are invalid"    unless all {
+        ref($_) eq 'ARRAY' && looks_like_number( $_->[0] ) && looks_like_number( $_->[1] )
+    } values %$val;
+
```
dylanwh: This needs to return "missing get_bug" when get_bug is missing, and "missing show_bug" when show_bug is missing.

Bugzilla.pm
```diff
@@ -767,6 +768,23 @@ sub elastic {
...
+            Bugzilla->audit("[rate_limit] $id exceeds rate limit $name: " . join("/", @$limit));
```
dylanwh: Just to cover here -- in the previous review glob wanted to add the remote_ip here, to protect against the future when $id might not be the remote ip address. We're specifically *not* doing this now because it will just make the log noisier. I'll either add it when we need it, or when we have structured logging. So this part is fine.

Bugzilla/Config/Admin.pm
umohm1: While I'm not sure what this code is doing. Would I be leaving this as is and adding in a conditional?

dylanwh:
```perl
foreach my $required (qw( show_bug get_bug )) {
    return "missing $required" unless exists $val->{$required};
}
```

Bugzilla.pm
dylanwh: Just a note that this performance thing was noted and at the moment ignored.

PR #101, opened 2017-06-13
Bug 1365344 - Extract the "status" and "null" modules

SebastinSanty:


extensions/BugModal/lib/WebService.pm
```diff
@@ -96,9 +96,10 @@ sub components {
...
             { function => 'BugModal.components', params => ['product'] } );
     }
-    my $product = Bugzilla::Product->check( { name => $params->{product_name}, cache => 1 } );
-    $product = Bugzilla->user->can_enter_product( $product, 1 );
-    return { components => _name( $product->components ) };
+    my $product = Bugzilla::Product->check({ name => $params->{product_name}, cache => 1 });
+    $product = Bugzilla->user->can_enter_product($product, 1);
+    my @components = map { { name => $_->name, description => Bugzilla::Component->check({ product => $product, name => $_->name })->description} } @{ $product->components };
+    return { components => \@components }
 }
```
dylanwh: You don't want this change, I think.

template/en/default/bug/new_bug.html.tmpl
```diff
@@ -9,10 +9,136 @@
 [% title = BLOCK %]Enter [% terms.Bug %] [% END %]
 [% PROCESS bug_modal/common_header.html.tmpl 
 title = title
+javascript_urls = ['extensions/BugModal/web/new_bug.js']
```
dylanwh: nit: some indentation here would be nice

extensions/BugModal/web/new_bug.js
```diff
@@ -0,0 +1,48 @@
+$(document).ready(function() {
```
dylanwh: looks like this is using 8-space indent -- needs to be 4-space indent.

extensions/BugModal/lib/WebService.pm
SebastinSanty: This is to facilitate retrieval of component description without making another ajax call.

extensions/BugModal/web/new_bug.js
```diff
@@ -0,0 +1,48 @@
...
+            },
+            function(data) {
+                $('#product').empty()
+                $('#product').append($('<option>', { value: 'Select Product', text: 'Select Product' }));
+                // populate select menus
+                $.each(data.products, function(key, value) {
+                    $('#product').append($('<option>', { value: value.name, text: value.name }));
+                });
+            },
+            function() {}
```
dylanwh: Should there be error handling here?

SebastinSanty: As of now, no, as this is just facilitating data loading in the drop-down. Maybe we can keep a small warning near to show that the data wasn't loaded. The previous error handling in bug_modal was to revert back to the non-edit mode, which we don't have here.

PR #102, opened 2017-06-14
Bug 1372653 - Upgrade LWP to 6.26 & add dependency for LWP::Protocol::https 6.07

dylanwh:
Sanity tests pass, bmo-base/bmo-ci have up-to-date deps.


Makefile.PL
```diff
@@ -49,8 +49,8 @@ my %requires = (
     'Email::Send'              => '1.911',
     'File::Slurp'              => '9999.13',
     'JSON::XS'                 => '2.01',
-    'LWP::Protocol::connect'   => 0,
-    'LWP::UserAgent'           => '5.835',
+    'LWP::UserAgent'           => '6.26',
+    'LWP::Protocol::https'     => 6,
```
floatingatoll: I noted that these two lines switched order - but this list is sorted - so please swap the two.

extensions/Persona/Config.pm
```diff
@@ -19,11 +19,6 @@ use constant REQUIRED_MODULES => [
         module  => 'JSON',
         version => 0,
     },
-    {
-        package => 'libwww-perl',
-        module  => 'LWP::UserAgent',
-        version => 0,
-    },
```
floatingatoll: Can we just remove extensions/Persona from BMO, or is this not BrowserID Persona?

Makefile.PL
floatingatoll: Also, the dependency on '6' doesn't match the described dependency on '6.07'.

dylanwh:
Since LWP 6.06: 
https://metacpan.org/changes/distribution/libwww-perl#L109

>  - correct behavior for https_proxy,
>       - e.g. don't send plain https:// requests to proxy, but instead
>            establish CONNECT tunnel and then send requests inside tunnel. This
>           change does together with a change in LWP::Protocol::https. The
>          change supports LWP::Protocol::https with the default IO::Socket::SSL
>            backend, but also with Net::SSL. Also:

extensions/Persona/Config.pm
dylanwh: So you're not going to like the response to this but BMO's CI breaks if I remove that code at the moment... even though it isn't used.

floatingatoll: Just as long as there's a bug on file documenting that, I'm happy to ignore it.

dylanwh: I have every intention of fixing that, but...

dylanwh: https://bugzilla.mozilla.org/show_bug.cgi?id=1283138 is the bug where we talked about removing Persona. It was just turned off -- I'll have a bug filed to remove it (and fix tests) later but that's outside of the scope of this task.

Makefile.PL
dylanwh: I'll fix the order. 6.07 is the latest --so I'll add that too.

dylanwh: 6.09 apparently.

floatingatoll:
I encourage 6.06 if you're setting a minversion, since you don't care if
it's newer than 6.06 or not. If we need to pin versions directly, then
latest is fine I guess :)

dylanwh:
I fixed the order and version -- note that this file specifies the minimum version, and other parts of the infra pin the specific versions. So as-deployed it will be the latest version for those two.
They won't be upgraded unless:
1) a new dependency requires a newer version
2) we manually force them to be upgraded.

dylanwh:
Related change: https://github.com/mozilla-bteam/carton-bundles/commit/071c43e3665b842135d3866142d824045bcd20c0

PR #103, opened 2017-06-14
Bug 1373000 - New BugModal API for product data and component+product-specific data

SebastinSanty:


extensions/BugModal/lib/WebService.pm
```diff
@@ -62,6 +81,25 @@ sub rest_resources {
...
+    };
+}
+
+sub components {
+    my ($self, $params) = @_;
+    my $product = Bugzilla::Product->check({ name => $params->{product_name}, cache => 1 });
+    my $components = _name($product->components);
+    my %options;
+    $options{component} = $components;
+    return \%options;
```
floatingatoll: Does ```components``` return the same @array of { name => $ } hashrefs that ```products``` does, in the end? If not, is there some legacy BMO reason why these data structures differ so significantly in construction for this new code?

SebastinSanty: Yes, it does return in the format you specified and is same as that of `products`.

SebastinSanty: @floatingatoll Should I change it to `@array` (hash) form like that of products to avoid confusion?

floatingatoll: Well, the components one doesn't obviously document the resulting output object. It requires you to implicitly understand what _name() returns, which I don't know, and so I couldn't predict what this will return from the API without reading _name().

You do use two different methods to construct the same output, which misleads me into thinking that you're constructing two different types of output. I do think it would be useful to make the two code paths *look* more similar, since the code is the only documentation available for the resulting data structure.

extensions/BugModal/lib/WebService.pm
```diff
@@ -27,6 +27,25 @@ use List::MoreUtils qw(any first_value);
...
+            },
+        },
+
+        # return all the components pertaining to the product.
+        # required by new-bug
+        qr{^/bug_modal/components}, {
+            GET => {
+                method => 'components',
+                params => sub {
+                    return { product_name => Bugzilla->input_params->{product} }
```
dylanwh: wow, this is a surprising use of this! I don't think we've ever used it this way, but it is interesting.
Probably this can be empty and just use Bugzilla->input_params from the product sub.

extensions/BugModal/lib/WebService.pm
```diff
@@ -62,6 +81,26 @@ sub rest_resources {
...
+    my @products = @{ $user->get_enterable_products };
+
+    @products = map { { name => $_->name } } @products;
+    return {
+        products     => \@products,
+    };
+}
+
+sub components {
+    my ($self, $params) = @_;
```
dylanwh: $product will need to be detainted here.
IMHO it's probably safe enough to untaint() if it's not a ref, e.g.
```perl
if (!ref $params->{product}) {
   untaint($params->{product});
}
else {
   ThrowUserError(TODO); # we have an error name for bad parameters, but I forget what it is
}
```

extensions/BugModal/lib/WebService.pm
```diff
@@ -62,6 +81,26 @@ sub rest_resources {
...
+    my $product = Bugzilla::Product->check({ name => $params->{product_name}, cache => 1 });
+
```
dylanwh: you want to check this with Bugzilla->user->can_enter_product() first

extensions/BugModal/lib/WebService.pm
```diff
@@ -62,6 +81,26 @@ sub rest_resources {
...
+    return {
+        products     => \@products,
+    };
+}
+
+sub components {
+    my ($self, $params) = @_;
+    my $product = Bugzilla::Product->check({ name => $params->{product_name}, cache => 1 });
+    
+    my @components = map { { name => $_->name } } @{ $product->components };
```
dylanwh: You can use _name() for this, which will also filter out inactive components. The second arg of _name() is optional, so should be safe.

SebastinSanty: I had done that, but to keep uniformity with the product code and to avoid confusion, I wrote it this way. For your reference atoll had already reviewed with this similar query: https://github.com/mozilla-bteam/bmo/pull/103#pullrequestreview-44132271

floatingatoll: I admit I don't quite follow this specific comment, sorry. In the other place I was noting that it was weird how the two data structures were constructed in completely different ways, when the code flow itself was the sole possible way to guess at the structure.

floatingatoll: But I think I do know a useful question to offer. What is the equivalent function for ```_name()``` that returns Products, rather than Components, in the same structure as ```_name()``` does?

SebastinSanty: Sorry, you couldn't follow the comment. `_name()` is a general function which checks the validity, and returns with only the name(property) of the object. I guess, we can approach both ways here, but as dylan suggests "`_name()` for this, which will also filter out inactive components". For products, it can be this way:
```perl
sub products {
    my $user = Bugzilla->user;
    my $products = _name($user->get_enterable_products);
    my %options;
    $options{product} = $products;
    return \%options;
}
```

floatingatoll: Oh, okay. That makes sense, thank you. ++ to using _name() in both places in the same sort of way.

extensions/BugModal/lib/WebService.pm
```diff
@@ -62,6 +82,24 @@ sub rest_resources {
...
+    return { products => _name($user->get_enterable_products) }
+}
+
+sub components {
+    my ($self, $params) = @_;
+    if (!ref $params->{product_name}) {
+        untaint($params->{product_name});
+    }
+    else {
+        ThrowUserError('params_required',{ function => 'BugModal.components', params => ['product'] });
```
dylanwh: it's a CodeError, not a usererror, sorry.
s/ThrowUserError/ThrowCodeError/

PR #105, opened 2017-06-15
Bug 1361439 - Create dockerflow-compliant container that runs a BMO webhead

dylanwh:
Already, this is against *master* now, and hopefully not broken.


dylanwh:
@dklawren -- need one more approval. Your proposed changes will be in bug 1376031 https://bugzilla.mozilla.org/show_bug.cgi?id=1376031

floatingatoll:
"Also a hack I added to init.pl was" - what?

dklawren:
Sorry. The init.pl change was that I added /app/docker_files/checksetup_answers.txt to the checksetup.pl line in the script.

+system("perl", "checksetup.pl", "--no-templates", "--no-permissions", '--no-assets', '/app/docker_files/checksetup_answers.txt');

dylanwh:
yeah, but we're not doing that in this patch. So can I get an an approval or is there some other problem with this patch?

PR #107, opened 2017-06-16
Replace the update provision command in README

SebastinSanty:


README.rst
```diff
@@ -56,9 +56,14 @@ After editing files in the bmo directory, you will need to run
 
 .. code-block:: bash
 
-    vagrant rsync && vagrant provision web
+    vagrant rsync && vagrant provision --provision-with update
 
-to see the changes applied to your vagrant VM.
+to see the changes applied to your vagrant VM. If db is changed, do a full
+provision:
+
```
dylanwh: also should mention that if the above command fails, a full provision might fix things too.

PR #108, opened 2017-06-16
Bug 1374396 - Add initial push connector for Phabricator

darkwing:
*Work in progress*

Bugs that contain an Phabricator attachment whose security level is changed should, for now, send an AUDIT message.


dklawren:
Ah forgot to mention. Please file a BMO bug for this and update the PR summary to include the bug number. Make sure to assign the bug to yourself as well.

extensions/Push/lib/Connector/Phabricator.pm
```diff
@@ -0,0 +1,81 @@
...
+use base 'Bugzilla::Extension::Push::Connector::Base';
+
+use Bugzilla::Bug;
+use Bugzilla::Constants;
+use Bugzilla::Extension::Push::Constants;
+use Bugzilla::Extension::Push::Util;
+use Bugzilla::User;
+
+use constant PHAB_CONTENT_TYPE => 'text/x-phabricator-request';
+use constant SECURITY_GROUP_KEY => '?'; # TODO:  Need to figure out what key this is
```
darkwing: @dklawren Is there a specific security group key we need to check for or `is_private` all we need?

extensions/Push/lib/Connector/Phabricator.pm
```diff
@@ -0,0 +1,81 @@
...
+
+# TODO:  Should we check if PhabBugz is enabled and bail if not?
```
darkwing: Is this a check we need or is enable/disable the push connector enough?

extensions/Push/lib/Connector/Phabricator.pm
```diff
@@ -0,0 +1,81 @@
...
+
+sub options {
+    return (
```
darkwing: Can I create push connector without any options?  I saw a bunch of `""` text and no "Enable/Disable" dropdown.

extensions/Push/lib/Connector/Phabricator.pm
dklawren: this will be part of phase 2 or 3 of the connector development. But the end goal would be we would look at the current list of groups of a bug (after change) and the list of groups we are currently syncing from BMO to Phab, and make the new custom policy of the revision to be the intersection of those two group lists.

The synced groups list will come from Bugzilla->params->{phabricator_sync_groups}

So we do not actuallt need the SECURITY_GROUP_KEY constant.

extensions/Push/lib/Connector/Phabricator.pm
dklawren: In should_send() near the top, we can just check for phabricator_enabled and return if not.

return 0 unless Bugzilla->params->{phabricator_enabled};

extensions/Push/lib/Connector/Phabricator.pm
```diff
@@ -0,0 +1,78 @@
...
+}
+
+sub should_send {
+    my ($self, $message) = @_;
+
+    my $logger = Bugzilla->push_ext->logger;
+    $logger->info('should_send');
+
+    # TODO:  Ensure the specific security group is what was changed
+    my $did_change_security = ($message->routing_key =~ /^(?:attachment|bug)\.modify:.*\bis_private\b/);
```
dklawren: Here we cannot look for is_private specifically as that just signifies if the bug is has transitioned from public to private or private to public. We could have a bug that has an additional security group added or one or more removed but not yet public.

So we need to key off 'bug_group' I think which is the internal field name. We can then load the bug object in send() later to get the current list of set bug groups to compare against the ones we are syncing.

extensions/Push/lib/Connector/Phabricator.pm
```diff
@@ -0,0 +1,78 @@
...
+
+sub _get_bug_data {
+    my ($self, $data) = @_;
+    my $target = $data->{event}->{target};
+    if ($target eq 'bug') {
+        return $data->{bug};
+    } elsif (exists $data->{$target}->{bug}) {
+        return $data->{$target}->{bug};
+    } else {
+        return;
```
floatingatoll: Is it normal that ```$message->payload_decoded()``` would have two separate ways to reference a bug here?

The line 71 exists will auto-instantiate $target inside $data if it's not already present, which may or may not matter.

darkwing:
OK, this is ready for a proper review.  Please let me know if I missed something per the scope of the trello bug (https://trello.com/c/bVAK4gwQ/393-13-when-i-add-or-remove-a-security-group-on-a-bug-with-a-phab-revision-stub-attachment-the-phabricator-push-connector-will-log-a).  

Also, the only outstanding question is with regard to `options(){}`, which without the stub option, cannot be enabled. (there's now Enable/Disable dropdown).

darkwing:
I based my checks on the payloads I could conjure up (https://gist.github.com/darkwing/f4a347f30b9cb2c0fc0c1058e2854416)

extensions/Push/lib/Connector/Phabricator.pm
```diff
@@ -0,0 +1,83 @@
...
+use Bugzilla::Constants;
+use Bugzilla::Extension::Push::Constants;
+use Bugzilla::Extension::Push::Util;
+use Bugzilla::User;
+
+use constant PHAB_CONTENT_TYPE => 'text/x-phabricator-request';
+
+sub options {
+    return (
+        {   name     => 'phabricator_url',
```
dylanwh: Not sure why thus got cuddled, but I'll fix after merge.

extensions/Push/lib/Connector/Phabricator.pm
```diff
@@ -0,0 +1,83 @@
...
+        }
+    );
+}
+
+sub should_send {
+    my ( $self, $message ) = @_;
+
+    return 0 unless Bugzilla->params->{phabricator_enabled};
+
+    if (!(  $message->routing_key
```
dylanwh: can this be... 
```perl
return 0 unless $message->routing_key =~ /^(?:attachment|bug)\.modify:.*\bbug_group\b/;
```

extensions/Push/lib/Connector/Phabricator.pm
```diff
@@ -0,0 +1,83 @@
...
+        return 0;
+    }
+
+    my $data = $message->payload_decoded;
+    my $bug_data = $self->_get_bug_data($data) || return 0;
+    my $bug = Bugzilla::Bug->new( { id => $bug_data->{id}, cache => 1 } );
+    my $has_phab_stub_attachment
+        = $bug->has_attachment_with_mimetype(PHAB_CONTENT_TYPE);
+
+    if ($has_phab_stub_attachment) {
```
dylanwh: I think 
```perl
return $bug->has_attachment_with_mimetype(PHAB_CONTENT_TYPE);
```
is probably much clearer.

darkwing:
Updated @dylanwh !

PR #110, opened 2017-06-19
Bug 1371446 - Don't allow needinfo requests from the void

umohm1:


extensions/Needinfo/template/en/default/bug/needinfo.html.tmpl
```diff
@@ -91,12 +91,14 @@ $(function() {
     if (role == 'other') {
       $('#needinfo_from_container').removeClass('bz_default_hidden');
       $('#needinfo_from').prop('disabled', false);
-      $('#needinfo_role_identity').text('');
+      $('#needinfo_role_identity').text
```
floatingatoll: How does this single specific change interact with the two 'required' property changes below, on either side of the if/else?

dylanwh: I'm pretty sure this is a bug.

extensions/Needinfo/template/en/default/bug/needinfo.html.tmpl
```diff
@@ -91,12 +91,14 @@ $(function() {
...
+      $('needinfo_from_textfield').prop('required', true);
```
dylanwh: also $('#needinfo_role_identity').text('')

extensions/Needinfo/template/en/default/bug/needinfo.html.tmpl
```diff
@@ -92,11 +92,13 @@ $(function() {
       $('#needinfo_from_container').removeClass('bz_default_hidden');
       $('#needinfo_from').prop('disabled', false);
       $('#needinfo_role_identity').text('');
+      $('needinfo_from').prop('required', true);
```
dylanwh: this selector doesn't match anything -- we'll have a chat to figure this out.

dylanwh:
I have some changes in https://github.com/umohm1/bmo/pull/2 that fix the red border thing we talked about on irc -- merge that in and I think this is done.

PR #111, opened 2017-06-21
Bug 1372693 Update to Recruiting Bugzilla Custom Form 

umohm1:
https://bugzila.mozilla.org/enter_bug.cgi?product=Recruiting


extensions/BMO/template/en/default/bug/create/create-recruiting.html.tmpl
```diff
@@ -183,6 +169,7 @@ function jobDescToggle(what) {
             <option value="Tyler Haugen">Tyler Haugen</option>
             <option value="Josh Kruse">Josh Kruse</option>
             <option value="Matt Frassica">Matt Frassica</option>
+            <option value="Matt Frassica">Huma Shah</option>
```
dylanwh: Need to update the =value too

emceeaich: <option value="Matt Frassica">Huma Shah</option>
 should be
<option value="Huma Shah">Huma Shah</option>

PR #113, opened 2017-06-23
Bug 1365345 - Modal Bug Entry Prototype

SebastinSanty:


dylanwh:
I mean, that's one option. Another option would be to figure out how to handle the back/forward cache.

extensions/BugModal/lib/WebService.pm
```diff
@@ -87,19 +87,24 @@ sub products {
...
-    return { components => \@components }
+    my $product = Bugzilla::Product->check( { name => $params->{product_name}, cache => 1 } );
+    $product = Bugzilla->user->can_enter_product( $product, 1 );
+    my @components = map {
+        {
+            name        => $_->name,
+            description => $_->description,
+        }
+    } @{ $product->components };
+    my @versions = map { { name => $_->name } } grep( $_->is_active, @{ $product->versions } );
```
dylanwh: so both grep and map have two forms: `grep code, list` and `grep { code } list`
It's better to just always use the block style.
so make this:
```perl
my @versions = map { { name => $_->name } } grep { $_->is_active } @{ $product->versions };

File: extensions/BugModal/web/new_bug.js [hunk 2]
```diff
@@ -1,48 +1,89 @@
...
-                $('#product').append($('<option>', { value: 'Select Product', text: 'Select Product' }));
-                // populate select menus
-                $.each(data.products, function(key, value) {
-                    $('#product').append($('<option>', { value: value.name, text: value.name }));
-                });
+                callback(data.products);
             },
-            function() {}
+            function() {
+                callback();
```
dylanwh: hmm, what does calling callback() with no arguments actually do?

SebastinSanty: No need of that here, as the data is already loaded (or else it goes out through the bugzilla_ajax' error function).

extensions/BugModal/template/en/default/bug_modal/field.html.tmpl
```diff
@@ -28,6 +28,7 @@
   #                 edit html instead of replacing it.  forces edit_only (default: false);
   # default: (string) default value (eg. used as a placeholder in user fields)
   # help: (string) optional url that describes the field (requires a label to be defined and visible)
+  # required: (boolean) if the field is required.
```
dylanwh: comment should note this is only implemented for FIELD_TYPE_FREETEXT

PR #117, opened 2017-07-01
Bug 1377596 - Add changed column to elasticsearch results

dylanwh:


Bugzilla/Elastic/Search.pm
```diff
@@ -90,8 +90,11 @@ sub data {
         die $@;
     }
     $self->_set_query_time($result->{took} / 1000);
+    my %stupid_name = (
```
floatingatoll: I'm not sure 'stupid_name' is necessarily the right variable name choice to try and merge into upstream someday. Why is it stupid? A code comment would be useful here.

Bugzilla/Elastic/Search.pm
```diff
@@ -90,8 +90,11 @@ sub data {
...
+        changeddate => 'delta_ts',
+    );
+    my @fields = map { $stupid_name{$_} // $_ } @{ $self->fields };
     my (@ids, %hits);
```
floatingatoll: If you move this line to just after 92 ->_set_query_time, the diff reads more cleanly. But I like this arrangement better anyways because these are the loop variables and they should be declared closer to the loop than the one-time @fields stuff anyways.

dylanwh:
No unit tests were harmed in the making of this, and this code is... difficult to test until I have a usable testing data set.

Bugzilla/Elastic/Search.pm
```diff
@@ -86,14 +97,10 @@ sub data {
             body => $body,
         );
     };
-    if (!$result) {
-        die $@;
-    }
+    die $@ unless $result;
```
floatingatoll: Cosmetic only, presumably.

floatingatoll:
I meant more like 'does it blend', not 'does it pass all variant units under subsection 2.1.3.4', for whatever it's worth.

PR #121, opened 2017-07-03
Bug 1377620 - Elasticsearch bulk indexer ignores changes that happened before newest additions

dylanwh:


Bugzilla/Elastic/Indexer.pm
```diff
@@ -109,7 +104,7 @@ sub _bulk_helper {
     );
 }
 
-sub find_largest_mtime {
```
floatingatoll: Was this a cosmetic change uncovered while fixing the core problem, but not necessarily related to the problem itself?

Bugzilla/Elastic/Indexer.pm
```diff
@@ -170,54 +165,44 @@ sub _debug_sql {
...
-
-    my $updated_ids = $self->_select_updated_ids($class);
-    if ($updated_ids) {
-        $self->_bulk_load_ids($bulk, $class, $updated_ids) if @$updated_ids;
-    }
+    my $bulk        = $self->_bulk_helper($class);
+    my $last_mtime  = $self->_find_largest_mtime($class);
+    my $last_id     = $self->_find_largest_id($class);
+    my $new_ids     = $self->_select_all_ids($class, $last_id);
+    my $updated_ids = $self->_select_updated_ids($class, $last_mtime);
```
floatingatoll: This is so much easier to read. Thank you. Now I understand what's going on.

Bugzilla/Elastic/Indexer.pm
```diff
@@ -170,54 +165,44 @@ sub _debug_sql {
...
-    my $dbh   = Bugzilla->dbh;
-    my $mtime = $self->find_largest_mtime($class);
-    if ($mtime && $mtime != $self->mtime) {
-        my ($updated_sql, $updated_params) = $self->_debug_sql($class->ES_SELECT_UPDATED_SQL($mtime));
-        return $dbh->selectcol_arrayref($updated_sql, undef, @$updated_params);
-    } else {
-        return undef;
-    }
+    my $dbh = Bugzilla->dbh;
+    my ($updated_sql, $updated_params) = $self->_debug_sql($class->ES_SELECT_UPDATED_SQL($last_mtime));
```
floatingatoll: What validates `$last_mtime` as being a valid mtime at all?

Bugzilla/Elastic/Indexer.pm
```diff
@@ -170,54 +165,44 @@ sub _debug_sql {
...
     $self->put_mapping($class);
-    $self->clear_mtime;
     $self->_bulk_load_ids($self->_bulk_helper($class), $class, $ids);
 }
 
 sub _bulk_load_ids {
     my ($self, $bulk, $class, $all_ids) = @_;
 
     my $iter  = natatime $class->ES_OBJECTS_AT_ONCE, @$all_ids;
-    my $mtime = $self->mtime;
```
floatingatoll: Why is the lazy-cache here removed? Is that related to the bug somehow?

dylanwh: it was sort of wrong.
You could forget to clear it and then have the mtime for users get inserted into the mixed into the mtimes for comments... the new code is much easier to follow.

Bugzilla/Elastic/Indexer.pm
dylanwh: It was because a previous review asked for sanity checking a value, and a desire to not repeat myself. So I merged two code paths into one implementation

Bugzilla/Elastic/Indexer.pm
dylanwh: it's checked to be a number, which is good enough for mysql, elasticsearch, and perl

Bugzilla/Elastic/Indexer.pm
```diff
@@ -104,34 +104,41 @@ sub _bulk_helper {
...
 
-    my $result = $self->client->search(
-        index => $self->index_name,
-        type  => $class->ES_TYPE,
-        body  => {
-            aggs => { $class->ID_FIELD => { extended_stats => { field => $class->ID_FIELD } } },
-            size => 0
-        }
-    );
+    return $self->_find_largest('es_mtime');
```
floatingatoll: Is a `$class->ES_MTIME_FIELD` available here?

dylanwh: no, because mtime doesn't come from the object, but from the indexer itself.

PR #122, opened 2017-07-03
Bug 1377807 - Move memcached options from data/params to localconfig

umohm1:


template/en/default/setup/strings.txt.pl
```diff
@@ -387,6 +387,15 @@ END
 WARNING: Dependency graph images are not accessible.
 Delete ##dir##/.htaccess and re-run checksetup.pl.
 END
+    memcached_servers => <<'END',
+If this option is set, Bugzilla will integrate with `Memcached
+<http://www.memcached.org/>`_. Specify one or more servers, separated by
```
floatingatoll: Please remove the `_` character here.

template/en/default/setup/strings.txt.pl
```diff
@@ -54,7 +54,7 @@ END
 to continue.
 END
     cpanfile_created => "##file## created",
-    cpan_bugzilla_home =>
```
floatingatoll: I had a hard time finding the actual code changes due to the sheer volume of unrelated whitespace fixes in this pull request. Please consider separating them into their own whitespace-only commit and/or pull request in the future.

template/en/default/setup/strings.txt.pl
dylanwh: yeah, the _ is string concatenation in the templates, but it isn't needed here.

template/en/default/setup/strings.txt.pl
```diff
@@ -387,6 +387,15 @@ END
 WARNING: Dependency graph images are not accessible.
 Delete ##dir##/.htaccess and re-run checksetup.pl.
 END
+    memcached_servers => <<'END',
```
dylanwh: these should be named localconfig_memcached_servers and localconfig_memcached_namespace and be moved up near the other options -- possibly in alphabetical order

PR #123, opened 2017-07-03
Bug 1377933 - Remove trailing whitespace from all perl files

dylanwh:


Bugzilla/DB/Pg.pm
```diff
@@ -118,7 +118,7 @@ sub sql_not_regexp {
 
     $self->bz_check_regexp($real_pattern) if !$nocheck;
 
-    return "${expr}::text !~* $pattern" 
+    return "${expr}::text !~* $pattern"
```
floatingatoll: The semicolon-free return pattern here might be worth a second pass in a different PR.

umohm1:
I don't see any conflicting changes.

PR #129, opened 2017-07-07
Bug 1361184 - Write script to remap milestones and versions

dylanwh:


dylanwh:
glob's irc feedback:
```
10:39 <glob> as there's no documentation or comments it's hard to be sure exactly what's going on here from a feedback perspective
10:39 <glob> it looks like the script creates a json file which contains a series of actions which are later executed.  by generating an explicit set of actions you make it possible that it will fail during execution should the state of the database be different, and it also makes it difficult to edit if required.
10:39 <glob> ideally the json should be declarative - defining the mappings of the names, with the actions required to perform the changes done at execution time.
10:40 <glob> but i'm concerned now that may be more work for you that you need at this time..
10:40 <glob> i love how clear the dry-run/summary is with your code
10:41 <glob> maybe drop the json file completely; always do a dry-run with confirmation
10:42 <glob> i would also recommend changing the classname as Bugzilla::ChangeSet could easily be misread as dealing with changesets within bugs (iirc that's the exact same clase that was proposed in a massive "refactor how we track changes" bug)
10:42 <glob> something like Bugzilla::BulkEdit::ChangeSet would be clearer
```

floatingatoll:
(The complete lack of explanation of what it's doing or what input it desires is obviously not good.)

PR #130, opened 2017-07-07
Bug 1378999 - Add update/new stats to bulk_indexer verbose mode

dylanwh:


scripts/bulk_index.pl
```diff
@@ -49,32 +49,40 @@ BEGIN
...
 $indexer->create_index;
 
+my $run_time = time;
 my $loop = IO::Async::Loop->new;
 my $timer = IO::Async::Timer::Periodic->new(
     first_interval => 0,
-    interval       => 15,
-    reschedule     => 'skip',
+    interval       => 20,
+    reschedule     => 'drift',
```
dylanwh: crap, this is an unrelated change -- but it results in a much more reliable runtime.

Bugzilla/Elastic/Indexer.pm
```diff
@@ -181,6 +181,11 @@ sub bulk_load {
 
     $self->_bulk_load_ids($bulk, $class, $new_ids) if @$new_ids;
     $self->_bulk_load_ids($bulk, $class, $updated_ids) if @$updated_ids;
+
+    return {
+        new     => scalar @$new_ids,
```
floatingatoll: I asked on IRC and confirmed that scalar is unary and doesn't eat past comma here.

floatingatoll: I asked on IRC and confirmed that scalar is unary and doesn't eat past comma here.

scripts/bulk_index.pl
```diff
@@ -49,32 +49,40 @@ BEGIN
...
-        say "indexing comments" if $verbose;
         my $start_comments = time;
-        $indexer->bulk_load('Bugzilla::Comment');
-        print "    ", time - $start_comments, " seconds\n" if $verbose > 1;
+        say "indexing comments" if $verbose;
+        my $comments = $indexer->bulk_load('Bugzilla::Comment');
+        bulk_load_stats($start_comments, $comments) if $verbose > 1;
 
         $loop->stop if $once;
+        $run_time = time;
```
floatingatoll: You should set this prior to $loop->stop, since otherwise the printf at +60 will never be accurate beyond a single run due to invalid start time. It's technically a minor point, but code outside this loop shouldn't have to worry about *whether* $run_time is valid, ever, and it costs little to make it right for all scenarios rather than just some.

scripts/bulk_index.pl
```diff
@@ -49,32 +49,40 @@ BEGIN
...
 
+my $run_time = time;
 my $loop = IO::Async::Loop->new;
 my $timer = IO::Async::Timer::Periodic->new(
     first_interval => 0,
     interval       => 15,
     reschedule     => 'skip',
 
     on_tick => sub {
+        printf "Running after %d seconds\n", time - $run_time;
```
floatingatoll: This isn't `bulk_load_stats($run_time, time);`?

scripts/bulk_index.pl
```diff
@@ -49,32 +49,40 @@ BEGIN
...
         my $start_users = time;
         say "indexing users" if $verbose;
-        $indexer->bulk_load('Bugzilla::User');
-        print "    ", time - $start_users, " seconds\n" if $verbose > 1;
+        my $users = $indexer->bulk_load('Bugzilla::User');
+        bulk_load_stats($start_users, $users) if $verbose > 1;
```
floatingatoll: I'm slightly surprised to see you calling bulk_load_stats() directly here. Shouldn't that be covered by $indexer->bulk_load() for you? Or, is the goal here to only collect these stats at *this* call to bulk_load, but not *all* calls?

scripts/bulk_index.pl
```diff
@@ -49,32 +49,40 @@ BEGIN
...
 my $loop = IO::Async::Loop->new;
 my $timer = IO::Async::Timer::Periodic->new(
     first_interval => 0,
     interval       => 15,
     reschedule     => 'skip',
 
     on_tick => sub {
+        printf "Running after %d seconds\n", time - $run_time;
         my $start_users = time;
         say "indexing users" if $verbose;
```
floatingatoll: Per the immediately-following review comment, rather than printing "indexing users" here, you could just do it as part of the bulk_load('Bugzilla::User') call. You already have the string 'Bugzilla::User' handy to emit in a log message, and that routine should be responsible for logging its own stats anyways IMO.

Bugzilla/Elastic/Indexer.pm
```diff
@@ -181,6 +181,11 @@ sub bulk_load {
 
     $self->_bulk_load_ids($bulk, $class, $new_ids) if @$new_ids;
     $self->_bulk_load_ids($bulk, $class, $updated_ids) if @$updated_ids;
+
+    return {
```
floatingatoll: Consider adding `class => $class` here, since it's free information for you to provide but impossible to distinguish from the results alone otherwise. (See also various review thoughts below about closer stats integration here.)

dylanwh:
Noted, I might make those changes next but for now it would be useful for debugging to get this out.

PR #133, opened 2017-07-10
Bug 1331305 - Auto-resize comment boxes

wlach:
Inspired by earlier work by @KuoE0 in https://github.com/mozilla-bteam/bmo/pull/40


wlach:
Video of it in action: http://wrla.ch/autosizer_comment.webm

extensions/BugModal/web/bug_modal.css
```diff
@@ -710,6 +710,7 @@ body.platform-Win32 .comment-text, body.platform-Win64 .comment-text {
     box-sizing: border-box !important;
     margin: 0 0 1em;
     max-width: 1024px;
+    max-height: 280px;
```
wlach: This is semi-arbitrarily based on github's current maximum (266px).

emceeaich:
This is introducing a new dependency with autosize.js, should that file go in the bug modal's extension path?

wlach:
That probably makes more sense now that I think of it, happy to make that change.

wlach:
@dylanwh the zoom text area stuff stuff is still needed/used for the "new bug" form (which does not currently use autosizing). So we'd need to convert that over before taking it out altogether-- seems like a worthwhile thing to do from what I can tell, but I'd rather roll that into a seperate change and just get this in (with whatever fixups you think are necessary).

wlach:
@dylanwh updated the location of autosizer and used the minified version

wlach:
I made it auto-resize when the reply button is clicked and a large amount of content is pasted inside.

PR #138, opened 2017-07-11
Bug 1380727 - Implement security group features for Phabricator 

darkwing:
Posting as a WIP for discussion.

### DIFF
--- a/extensions/PhabBugz/lib/Util.pm
+++ b/extensions/PhabBugz/lib/Util.pm
@@ -49,7 +49,7 @@ sub get_revisions_by_ids {
     ThrowUserError('invalid_phabricator_revision_id')
         unless (exists $result->{result}{data} && @{ $result->{result}{data} });
 
-    return $result->{result}{data};
+    return @{$result->{result}{data}};
 }
 
 sub create_revision_attachment {
--- a/extensions/PhabBugz/lib/Util.pm
+++ b/extensions/PhabBugz/lib/Util.pm
@@ -27,20 +27,20 @@ our @EXPORT = qw(
     get_bug_role_phids
     get_members_by_bmo_id
     get_project_phid
-    get_revision_by_id
+    get_revisions_by_ids
     intersect
     make_revision_public
     request
     set_project_members
 );
 
-sub get_revision_by_id {
-    my $id = shift;
+sub get_revisions_by_ids {
+    my ($ids) = @_;
 
     my $data = {
         queryKey => 'all',
         constraints => {
-            ids => [ int($id) ]
+            ids => $ids
         }
     };
 
@@ -49,7 +49,7 @@ sub get_revision_by_id {
     ThrowUserError('invalid_phabricator_revision_id')
         unless (exists $result->{result}{data} && @{ $result->{result}{data} });
 
-    return $result->{result}{data}[0];
+    return $result->{result}{data};
 }
 
 sub create_revision_attachment {
--- a/extensions/Push/lib/Connector/Phabricator.pm
+++ b/extensions/Push/lib/Connector/Phabricator.pm
@@ -74,7 +74,7 @@ sub send {
         foreach my $attachment (@attachments) {
             my ($rev_id) = ($attachment->filename =~ PHAB_ATTACHMENT_PATTERN);
             next if !$rev_id;
-            push(@rev_ids, $rev_id);
+            push(@rev_ids, int($rev_id));
         }
 
         if(@rev_ids) {
@@ -85,6 +85,7 @@ sub send {
             foreach my $rev_detail (@rev_details) {
                 my $rev_phid = $rev_detail->{phid};
                 $logger->info('Making revision $rev_phid public:');
+                $logger->info($rev_phid);
                 make_revision_public($rev_phid);
             }
         }
--- a/extensions/Push/lib/Connector/Phabricator.pm
+++ b/extensions/Push/lib/Connector/Phabricator.pm
@@ -1,9 +1,9 @@
 # This Source Code Form is subject to the terms of the Mozilla Public
-# # License, v. 2.0. If a copy of the MPL was not distributed with this
-# # file, You can obtain one at http://mozilla.org/MPL/2.0/.
-# #
-# # This Source Code Form is "Incompatible With Secondary Licenses", as
-# # defined by the Mozilla Public License, v. 2.0.
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+#
+# This Source Code Form is "Incompatible With Secondary Licenses", as
+# defined by the Mozilla Public License, v. 2.0.
 
 package Bugzilla::Extension::Push::Connector::Phabricator;
 
@@ -15,11 +15,14 @@ use base 'Bugzilla::Extension::Push::Connector::Base';
 
 use Bugzilla::Bug;
 use Bugzilla::Constants;
+use Bugzilla::Extension::PhabBugz::Util qw(intersect make_revision_public get_revisions_by_ids);
 use Bugzilla::Extension::Push::Constants;
 use Bugzilla::Extension::Push::Util;
 use Bugzilla::User;
+use List::Util qw(any);
 
 use constant PHAB_CONTENT_TYPE => 'text/x-phabricator-request';
+use constant PHAB_ATTACHMENT_PATTERN => qr/^phabricator-D(\d+)/;
 
 sub options {
     return (
@@ -47,9 +50,45 @@ sub should_send {
 }
 
 sub send {
+    my ( $self, $message ) = @_;
+
     my $logger = Bugzilla->push_ext->logger;
 
-    $logger->info('AUDIT');
+    my $data = $message->payload_decoded;
+    my $bug_data = $self->_get_bug_data($data) || return 0;
+    my $bug = Bugzilla::Bug->new( { id => $bug_data->{id}, cache => 1 } );
+
+    if(!is_public($bug)) {
+        $logger->info('Bailing on send because the bug is not public');
+        return PUSH_RESULT_OK;
+    }
+
+    my @attachments = grep {
+        $_->isobsolete == 0 &&
+        $_->contenttype eq PHAB_CONTENT_TYPE &&
+        $_->attacher->login eq 'phab-bot@bmo.tld'
+    } @{ $bug->attachments() };
+
+    if(@attachments){
+        my @rev_ids;
+        foreach my $attachment (@attachments) {
+            my ($rev_id) = ($attachment->filename =~ PHAB_ATTACHMENT_PATTERN);
+            next if !$rev_id;
+            push(@rev_ids, $rev_id);
+        }
+
+        if(@rev_ids) {
+            $logger->info('Getting info for revisions: ');
+            $logger->info(@rev_ids);
+
+            my @rev_details = get_revisions_by_ids(\@rev_ids);
+            foreach my $rev_detail (@rev_details) {
+                my $rev_phid = $rev_detail->{phid};
+                $logger->info('Making revision $rev_phid public:');
+                make_revision_public($rev_phid);
+            }
+        }
+    }
 
     return PUSH_RESULT_OK;
 }
--- a/extensions/PhabBugz/lib/WebService.pm
+++ b/extensions/PhabBugz/lib/WebService.pm
@@ -27,7 +27,7 @@ use Bugzilla::Extension::PhabBugz::Util qw(
     edit_revision_policy
     get_bug_role_phids
     get_project_phid
-    get_revision_by_id
+    get_revisions_by_ids
     intersect
     make_revision_public
     request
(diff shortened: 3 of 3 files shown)


floatingatoll:
Hi! Could you begin the discussion? I'm not quite sure how to begin understanding your intentions through a code push only here.

extensions/Push/lib/Connector/Phabricator.pm
```diff
@@ -59,9 +54,32 @@ sub should_send {
...
+    my $bug_data = $self->_get_bug_data($data) || return 0;
+    my $bug = Bugzilla::Bug->new( { id => $bug_data->{id}, cache => 1 } );
+
+    if ( any { $_->{name} =~ /core-security/ } @{$bug->groups()} ) {
+        $logger->info('HAS SECURITY!');
+
+        my @attachments = grep {
+            $_->isobsolete == 0 &&
+            $_->contenttype eq PHAB_CONTENT_TYPE &&
+            $_->{attacher}->{realname} eq 'phabricatorbot'
```
darkwing: I believe this needs to be changed to "phab-bot".

extensions/Push/lib/Connector/Phabricator.pm
```diff
@@ -59,9 +54,32 @@ sub should_send {
...
+    my ( $self, $message ) = @_;
+
     my $logger = Bugzilla->push_ext->logger;
 
-    $logger->info('AUDIT');
+    my $data = $message->payload_decoded;
+    my $bug_data = $self->_get_bug_data($data) || return 0;
+    my $bug = Bugzilla::Bug->new( { id => $bug_data->{id}, cache => 1 } );
+
+    if ( any { $_->{name} =~ /core-security/ } @{$bug->groups()} ) {
```
darkwing: Is this a PhabBugz setting, @dklawren ?

extensions/Push/lib/Connector/Phabricator.pm
dklawren: Should be:

$_->attacher->login eq 'phab-bot@bmo.tld'

or whatever we set up the login to be.

extensions/Push/lib/Connector/Phabricator.pm
```diff
@@ -59,9 +54,32 @@ sub should_send {
...
+        my @attachments = grep {
+            $_->isobsolete == 0 &&
+            $_->contenttype eq PHAB_CONTENT_TYPE &&
+            $_->{attacher}->{realname} eq 'phabricatorbot'
+        } @{ $bug->attachments() };
+
+        if(@attachments){
+            $logger->info('HAS PHAB ATTACHMENTS!');
+
+            my @attachment_ids = map { grep(PHAB_ATTACHMENT_PATTERN, $_->filename) } @attachments;
```
dklawren: I don't think this will work the way we want it to in that I don't think grep will return the value inside () but instead the full string of $_.

So we should do this instead:

my @rev_ids;
foreach my $attachment (@attachments) {
      my ($rev_id) = ($attachment->filename =~ PHAB_ATTACHMENT_PATTERN);
      next if !$rev_id;
      push(@rev_ids, $rev_id);
}

extensions/Push/lib/Connector/Phabricator.pm
```diff
@@ -1,10 +1,3 @@
-# This Source Code Form is subject to the terms of the Mozilla Public
```
darkwing: I had to axe comments locally because a copy/paste via `VI` in `bmo-extensions` would then comment every single line from then on down.  I'll add back before merge.

darkwing:
@dklawren I thought we mentioned needing to run this in all cases, not just a security group change.  Is it *only* during group changes? i.e. look for this:  https://gist.github.com/darkwing/f4a347f30b9cb2c0fc0c1058e2854416#file-changed-security-group-payload-json-L4  ?

extensions/Push/lib/Connector/Phabricator.pm
```diff
@@ -59,9 +54,32 @@ sub should_send {
...
+    my $bug_data = $self->_get_bug_data($data) || return 0;
+    my $bug = Bugzilla::Bug->new( { id => $bug_data->{id}, cache => 1 } );
+
+    my @synced_groups = split(/,/, Bugzilla->params->{phabricator_sync_groups});
+
+    if ( any { /$_->{name}/ ~~ @synced_groups } @{$bug->groups()} ) {
+        my @attachments = grep {
+            $_->isobsolete == 0 &&
+            $_->contenttype eq PHAB_CONTENT_TYPE &&
+            $_->attacher->{login_name} eq 'phab-bot@bmo.tld'
```
dklawren: $_->attacher->login eq 'phab-bot@bmo.tld'

extensions/Push/lib/Connector/Phabricator.pm
```diff
@@ -59,9 +54,32 @@ sub should_send {
...
     my $logger = Bugzilla->push_ext->logger;
 
-    $logger->info('AUDIT');
+    my $data = $message->payload_decoded;
+    my $bug_data = $self->_get_bug_data($data) || return 0;
+    my $bug = Bugzilla::Bug->new( { id => $bug_data->{id}, cache => 1 } );
+
+    my @synced_groups = split(/,/, Bugzilla->params->{phabricator_sync_groups});
+
+    if ( any { /$_->{name}/ ~~ @synced_groups } @{$bug->groups()} ) {
```
dklawren: What is '~~'?

Also we need to look at the groups that were added/removed as part of the bug change and see if any one of those are in @synced_groups. It will be this part in the event data:

   "event" : {
      "target" : "bug",
      "changes" : [
         {
            "removed" : "",
            "added" : "core-security, core-security-release",
            "field" : "bug_group"
         }
      ],
      ...
   }

You will need to split the groups on whitespace from 'added' and 'removed' and then combine them into a unique list. Then do compare the @changed_groups and @synced_groups to see if there are any intersected values. You can use the intersect() code from extensions/PhabBugz/lib/Util.pm.

extensions/Push/lib/Connector/Phabricator.pm
```diff
@@ -59,9 +54,32 @@ sub should_send {
...
+    my ( $self, $message ) = @_;
+
     my $logger = Bugzilla->push_ext->logger;
 
-    $logger->info('AUDIT');
+    my $data = $message->payload_decoded;
+    my $bug_data = $self->_get_bug_data($data) || return 0;
+    my $bug = Bugzilla::Bug->new( { id => $bug_data->{id}, cache => 1 } );
+
+    my @synced_groups = split(/,/, Bugzilla->params->{phabricator_sync_groups});
```
dklawren: Better to split on /[,\s]+/ in case they used spaces instead of commas.

extensions/Push/lib/Connector/Phabricator.pm
```diff
@@ -59,9 +54,32 @@ sub should_send {
...
+            $_->attacher->{login_name} eq 'phab-bot@bmo.tld'
+        } @{ $bug->attachments() };
+
+        if(@attachments){
+            my @rev_ids;
+            foreach my $attachment (@attachments) {
+                my ($rev_id) = ($attachment->filename =~ PHAB_ATTACHMENT_PATTERN);
+                next if !$rev_id;
+                push(@rev_ids, $rev_id);
+            }
```
dklawren: The code to make the actual changes should exist now in extension/PhabBug/lib/Util.pm to make the changes to the revision ids found.

extensions/PhabBugz/lib/Util.pm
```diff
@@ -36,11 +37,18 @@ our @EXPORT = qw(
 
 sub get_revision_by_id {
     my $id = shift;
+    my @result_data = get_revisions_by_ids(qw($id));
+
+    return $result_data[0];
+}
+
+sub get_revisions_by_ids {
+    my ($ids) = @_;
```
darkwing: I'm getting the following error here:

```
Phabricator: Message #8: TRANSIENT-ERROR Code: ERR-CONDUIT-CORE Reason: Error while reading "ids": Expected a
list, but value is not a list.
 at /loader/0x4728448/Bugzilla/Extension/PhabBugz/Util.pm line 270.
	Bugzilla::Extension::PhabBugz::Util::request(...) called at /loader/0x4728448/Bugzilla/Extension/PhabBugz/Util.pm line 48
	Bugzilla::Extension::PhabBugz::Util::get_revisions_by_ids(...) called at ./extensions/Push/lib/Connector/Phabricator.pm line 91
	Bugzilla::Extension::Push::Connector::Phabricator::send(...) called at /loader/0x49a0dc8/Bugzilla/Extension/Push/Push.pm line 129
	eval {...} called at /loader/0x49a0dc8/Bugzilla/Extension/Push/Push.pm line 128
	Bugzilla::Extension::Push::Push::push(...) called at /loader/0x49a0dc8/Bugzilla/Extension/Push/Push.pm line 55
	Bugzilla::Extension::Push::Push::start(...) called at /loader/0x49a0dc8/Bugzilla/Extension/Push/Daemon.pm line 94
	Bugzilla::Extension::Push::Daemon::gd_run(...) called at /usr/local/share/perl5/Daemon/Generic.pm line 192
	Daemon::Generic::new(...) called at /usr/local/share/perl5/Daemon/Generic.pm line 37
	Daemon::Generic::newdaemon(...) called at /loader/0x49a0dc8/Bugzilla/Extension/Push/Daemon.pm line 23
	Bugzilla::Extension::Push::Daemon::start(...) called at extensions/Push/bin/bugzilla-pushd.pl line 22
```

When I $logger->info() what I'm passing to this method, I'm simply seeing `5`; is there a way to typecast to ensure this is always list a list?  Help?

extensions/Push/lib/Connector/Phabricator.pm
```diff
@@ -59,9 +55,46 @@ sub should_send {
...
+    my ( $self, $message ) = @_;
+
     my $logger = Bugzilla->push_ext->logger;
 
-    $logger->info('AUDIT');
+    my $data = $message->payload_decoded;
+    my $bug_data = $self->_get_bug_data($data) || return 0;
+    my $bug = Bugzilla::Bug->new( { id => $bug_data->{id}, cache => 1 } );
+
+    if(!is_public($bug)) {
```
darkwing: We'd spoken about this earlier but would it make more sense to move this check to the `should_send`?  There doesn't seem to be a point to getting this far if the bug isn't public.

extensions/Push/lib/Connector/Phabricator.pm
```diff
@@ -59,9 +55,46 @@ sub should_send {
...
+            my ($rev_id) = ($attachment->filename =~ PHAB_ATTACHMENT_PATTERN);
+            next if !$rev_id;
+            push(@rev_ids, $rev_id);
+        }
+
+        if(@rev_ids) {
+            $logger->info('Getting info for revisions: ');
+            $logger->info(@rev_ids);
+
+            my @rev_details = get_revisions_by_ids(@rev_ids);
```
darkwing: Is it possible that, despite @rev_ids having `@` and pushing to it, that one item makes `@rev_ids` a string?  May explain the error in the Utils.pm file I mentioned above.

extensions/PhabBugz/lib/Util.pm
```diff
@@ -36,11 +37,18 @@ our @EXPORT = qw(
 
 sub get_revision_by_id {
     my $id = shift;
+    my @result_data = get_revisions_by_ids(qw($id));
```
darkwing: Will this properly add send the ID as a list?

extensions/PhabBugz/lib/Util.pm
```diff
@@ -36,11 +37,18 @@ our @EXPORT = qw(
...
+}
+
+sub get_revisions_by_ids {
+    my ($ids) = @_;
 
     my $data = {
         queryKey => 'all',
         constraints => {
-            ids => [ int($id) ]
+            ids => join(',', $ids)
```
darkwing: This should probably not be a string, should probably keep as an array

dklawren: yes we should keep as array

dklawren: array reference i meant to say. join() returns a list where is a [ ] is a reference to a list which is what JSON wants.

dklawren:
Also after think more, I feel we can combine and go with a single get_revisions_by_ids() and not even have get_revision_by_id(). We just always pass a reference to a list into get_revisions_by_ids() even for a single value in the list. Then have get_revisions_by_ids always return a reference to a list of revisions found. The calling code will be responsible for working only with [0] in the list.

dklawren:
Also after think more, I feel we can combine and go with a single get_revisions_by_ids() and not even have get_revision_by_id(). We just always pass a reference to a list into get_revisions_by_ids() even for a single value in the list. Then have get_revisions_by_ids always return a reference to a list of revisions found. The calling code will be responsible for working only with [0] in the list.

extensions/PhabBugz/lib/Util.pm
globau: you probably want..
`my @ids = @_;`
then
`ids => \@ids`

floatingatoll:
join() returns a string.

darkwing:
This is ready for another look

floatingatoll:
I wanted to note that this line made me uncomfortable: are we using qw()
instead of array references here?

get_revisions_by_ids(qw($id))

darkwing:
Hello @dklawren, I've added a second commit which ensures `get_revisions_by_id` returns an array properly and avoids a type error by using `int()`; and now everything should working as desired, presuming `make_revision_public` is in working order!  Thank you for your patience!

result

PR #142, opened 2017-07-12
Bug 1366143 - Fix alignment of select boxes

dylanwh:


kyoshino:
I think it still can be better, but anyway, since form controls are depending on the user's platform, it's pretty difficult to manage.

* Best solution: Use custom elements (maybe once Firefox adds the [native support](https://bugzilla.mozilla.org/show_bug.cgi?id=889230))
* Better solution: Use [`-moz-appearance: none`](https://developer.mozilla.org/en-US/docs/Web/CSS/-moz-appearance) to reset the appearance of `<select>`, then make it look similar to `<input>`. I'll try this approach.

floatingatoll:
(This was closed without merging in favor of #144.)

PR #146, opened 2017-07-13
Bug 1380408 - elasticsearch user completion should consider people with multiple ircnicks

dylanwh:


Bugzilla/User.pm
```diff
@@ -176,21 +176,28 @@ sub ES_PROPERTIES {
...
         suggest_user => {
             input => [ $self->login, $self->name ],
             output => $self->identity,
             payload => { name => $self->login, real_name => $self->name },
         },
     };
-    if ($self->name && $self->name =~ /:(\w+)/) {
-        my $ircnick = $1;
+    my @nicks;
+    if ($self->name) {
```
floatingatoll: What is this if() testing for? When would 'if ($self->name)' NOT product a warning, but then $name =~ /:.../ would?

floatingatoll: `my(@nicks) = ($self->name =~ m//mg);` being the alternate form I envisioned for 188...

Bugzilla/Elastic/Indexer.pm
```diff
@@ -179,6 +179,7 @@ sub bulk_load {
     my $new_ids     = $self->_select_all_ids($class, $last_id);
     my $updated_ids = $self->_select_updated_ids($class, $last_mtime);
 
+    $self->put_mapping($class);
```
dylanwh: This is actually fixing a pretty bad mistake in the old code, that wouldn't update the mapping if we ever deleted the ES index.

floatingatoll: Has it been submitted upstream, and/or is it relevant to anyone other than BMO?

PR #148, opened 2017-07-13
Bug 1373420 - Font for User Story is Courier instead of Fira Sans

kyoshino:
https://bugzilla.mozilla.org/show_bug.cgi?id=1373420

### DIFF
--- a/extensions/BugModal/web/bug_modal.css
+++ b/extensions/BugModal/web/bug_modal.css
@@ -310,6 +310,7 @@ input[type="number"] {
 
 #user-story {
     margin: 0;
+    padding: 4px;
     white-space: pre-wrap;
     min-height: 2em;
 }
--- a/skins/standard/global.css
+++ b/skins/standard/global.css
@@ -35,7 +35,8 @@
     }
     /* monospace is much smaller than Verdana by default, so we make it a bit bigger. */
     pre, code, kbd {
-        font-size: medium;
+        font-family: "Droid Sans Mono", Menlo, Monaco, "Courier New", Courier, monospace;
+        font-size: small;
     }
 /* global (end) */
 
@@ -234,7 +235,8 @@
     }
 
     textarea {
-        font-family: monospace;
+        font-family: "Droid Sans Mono", Menlo, Monaco, "Courier New", Courier, monospace;
+        font-size: small;
     }
 /* generic (end) */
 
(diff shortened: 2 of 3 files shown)


skins/standard/global.css
```diff
@@ -35,7 +35,8 @@
     }
     /* monospace is much smaller than Verdana by default, so we make it a bit bigger. */
     pre, code, kbd {
-        font-size: medium;
+        font-family: "Droid Sans Mono", Menlo, Monaco, "Courier New", Courier, monospace;
+        font-size: small;
```
floatingatoll: Why is the font size of the User Story being reduced as part of changing the font, when one of the listed permitted fonts is the original font?

skins/standard/global.css
```diff
@@ -35,7 +35,8 @@
     }
     /* monospace is much smaller than Verdana by default, so we make it a bit bigger. */
     pre, code, kbd {
-        font-size: medium;
+        font-family: "Droid Sans Mono", Menlo, Monaco, "Courier New", Courier, monospace;
```
floatingatoll: Does this set of changes bring the User Story font-size and font-family into alignment with the other instances of our use of Droid?

extensions/BugModal/web/bug_modal.css
```diff
@@ -310,6 +310,7 @@ input[type="number"] {
 
 #user-story {
     margin: 0;
+    padding: 4px;
```
floatingatoll: There are undocumented padding changes included in this font-fixing commit, that might be reversed undesirably if the font change goes wrong for any reason. Please update the commit message or split into a separate commit so that we're not secretly bringing along more than just 'Font'.

skins/standard/global.css
```diff
@@ -304,7 +306,6 @@ div#docslinks {
 
 /* tbody.file pre is for the Diff view of attachments. */
 .bz_comment_text, .uneditable_textarea, tbody.file pre {
-     font-family: monospace;
```
floatingatoll: Is this change to bz_comment_text necessary to bring Fira Sans to the User Story field?

skins/standard/global.css
```diff
@@ -1566,10 +1566,6 @@ body {
     font-size: small;
 }
 
-pre {
```
floatingatoll: This seems like it could have much wider effects and isn't obviously linked to the font size change. Have we vetted all `<pre>` uses to ensure that new alignment issues are not introduced by this change?

kyoshino:
This PR has stalled for a while. I've found another `<textarea>` using Courier, and will redesign the entire site based on the new design system which includes different font choices.

PR #151, opened 2017-07-14
Bug 1365983 - enter_bug.cgi should be updated to use "Data Platform and Tools"

Dipanks:


Dipanks:
Is there something else to be done in this?

extensions/BMO/template/en/default/global/choose-product.html.tmpl
```diff
@@ -118,7 +118,7 @@
       icon="localization.png"
     %]
     [% INCLUDE easyproduct
-      name="Cloud Services"
+      name="Data Pipeline and Tools"
```
dylanwh: I think the product is "Data Platform and Tools"

Dipanks:
@floatingatoll  I pushed this commit with the commit message. Is this okay?

@dylanwh I made the changes.

floatingatoll:
Yep, thanks! (re message)

PR #152, opened 2017-07-14
Bug 1381042 - new-bug doesn't properly engage the email generation machinery

SebastinSanty:


new_bug.cgi
```diff
@@ -59,6 +62,27 @@ if (lc($cgi->request_method) eq 'post') {
...
+         my $dep_sent = Bugzilla::BugMail::Send($dep, $recipients);
+         $dep_sent->{type} = 'dep';
+         $dep_sent->{id}   = $dep;
+         push(@all_mail_results, $dep_sent);
+     }
+
+     # Sending emails for any referenced bugs.
+     foreach my $ref_bug_id (uniq @{ $new_bug->{see_also_changes} || [] }) {
+         my $ref_sent = Bugzilla::BugMail::Send($ref_bug_id, $recipients);
+         $ref_sent->{id} = $ref_bug_id;
```
floatingatoll: Didn't we pass ID in already to the constructor one line prior? Is this a normal thing to do here?

dylanwh: I suspect it is from https://github.com/bugzilla/bugzilla/blob/master/post_bug.cgi#L215-L241.

@SebastinSanty -- I think we'll still collect $vars->{sentmail} but we need to stash it in a session and have show_bug display. I'll file a related bug for that.

dylanwh:
What's the difference between $bug->send_changes() and what post_bug.cgi does?
https://github.com/mozilla-bteam/bmo/blob/master/Bugzilla/WebService/Bug.pm#L783-L822

is send_changes() identical to what post_bug is doing here: 
https://github.com/bugzilla/bugzilla/blob/master/post_bug.cgi#L215-L235?

@dklawren?

dylanwh:
I'm about halfway done testing this, comparsing ->send_changes() to what this is doing.

PR #153, opened 2017-07-14
replace perltidyrc with on that specifies -pbp with two exceptions

dylanwh:
previously, the perltidyrc was written with the original recommendations of the
upstream bugzilla project. The style used there is... inconsistent at best. For
a while now, I've been using perltidy -pbp and telling others to do the same for
new code. However there are two exceptions to -pbp that are widely useful:

1. line length can be up to 120.
2. brace vertical tigntness is turned off.
   This means that there is always a newline after the { in a multi-line brace
   construct.

### DIFF
--- a/.perltidyrc
+++ b/.perltidyrc
@@ -1,16 +1,3 @@
---block-brace-tightness=0
---brace-tightness=1
---continuation-indentation=2
---indent-columns=4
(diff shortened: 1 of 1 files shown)


dylanwh:
the current code is very non-compliant.

PR #154, opened 2017-07-17
Bug 1381314- Link "Permissions" from user-profile to get_permissions.html

Dipanks:


Dipanks:
How to clear these dylan? rebase is not working probably.. Shall I open a new pull request?

emceeaich:
The pull request now only has one file and change in it. I'll add the PR so it works in splinter.

Dipanks:
@emceeaich, do I need to change anything?

extensions/UserProfile/template/en/default/pages/user_profile.html.tmpl
```diff
@@ -108,6 +108,13 @@
       [% END %]
     </td>
   </tr>
+[% ELSE %]
+  <tr>
+    <th>Get Permissions</th>
+    <td>
```
dylanwh: pretty sure there is a colspan="2" needed here, were you able to test this? need help with the vagrant vm?

Dipanks: I think I tested it before creating the pr. But I will test it once more and let you know in irc

Dipanks:
Is this okay? 
Or the 'Merge branch 'master'.....' will be a problem?
Do I have to rebase, in case if it is a problem?

PR #157, opened 2017-07-17
Bug 1380352 - Allow vertical resizing of autosized comments

wlach:
And if a comment is resized manually, then stop autosizing it (for that
session).

Requires a tweaked version of the autosize library.


wlach:
Yeah this needs more work. Adjusting now, will request review again when it's reading. Sorry for the premature ping.

wlach:
@dylanwh Ok this is ready for another look. I think the edge cases are genuinely working correctly now (tested on Chrome and Firefox).

floatingatoll:
We should definitely test Safari/Edge/ESR on this behavior as well, since
they're each on the "BMO supported browsers" list.

extensions/BugModal/web/autosize.min.js
```diff
@@ -1,6 +1,6 @@
 /*!
-	Autosize 3.0.20
+	Autosize 4.0.0
```
floatingatoll: The commit message says "Requires a tweaked version of the autosize library." However, no tweaks are documented in the comments here, nor is authorship updated below. Where is the commit reflecting the tweaks in pre-minimized form? How can we regenerate autosize.min.js including your tweaks if we need to (for example) merge Autosize 4.0.1 someday?

floatingatoll: If you're live-editing .min.js, you could instead commit 4.0.0.min.js *first* unaltered (as its own commit), so that your later commits to fix bug 1380352 include your diff against 4.0.0.min.js.

dylanwh: we're forking the autosize library due to extensive changes, so this is not so important. The version number is updated and it comes from the same github org, so that's good enough for me.

dylanwh:
I can't actually get this to work or resize

extensions/BugModal/web/autosize.min.js
wlach: I will add a short comment to the minified js indicating its origin

wlach:
Verified that the fix works on Safari as well. Haven't tested on ESR or Edge but I don't see why it wouldn't work. We're not using anything exotic here.

wlach:
@dylanwh Sorry about that, something happened that I'm not sure of that caused the wrong version of autosize to be committed. But I'm confident that the current PR works. Please give it a try.

wlach:
Ok, after (yet) more testing I noticed a few edge cases around manual resizing that still aren't working properly with Safari and Chrome, though Firefox now seems to be working well.

I can't really justify spending any more cycles on this, so I think I'm going to give up for now. That said, this WIP could be merged and would be better than what is in there now. Just keep it turned off by default. :)

PR #158, opened 2017-07-18
Bug 1380064 - Remove references to Mozilla skin to prevent problems when default preferences are changed

dylanwh:


template/en/default/global/header.html.tmpl
```diff
@@ -325,65 +324,14 @@
...
-      <a href="#" onclick="set_language('[% lang FILTER none %]');">
-       [%- lang FILTER html FILTER upper %]</a>
-    [% END %]
-    </li>
-  [% END %]
-  </ul>
-[% END %]
-</td></tr></table>
-
-[% PROCESS "global/common-links.html.tmpl" qs_suffix = "_top" %]
```
floatingatoll: I assume common-links.html.tmpl is still imported by all the skins here?

PR #163, opened 2017-07-18
Bug 1381869 - Use separate elasticsearch index for Bugzilla::User

dylanwh:
This patch removes the concept of a single, bugzilla-wide index in favor of a per-class index.
bugs and comments continue to use Bugzilla->params->{elasticsearch_index} but users
use Bugzilla->params->{elasticsearch_index} . "_users".

It is assured via the ChildObject trait (role) that comments will share the index with bugs, and we have kept the index for bugs/comments the same to avoid the multi-hour reindexing of production. Re-indexing users takes only five minutes.

Subsequent work on this will allow use to version the index names and use aliases, but I wanted to keep this patch small.

This patch also corrects a mistake: $indexer->put_mapping() should not have been a public method.


Bugzilla/Bug.pm
```diff
@@ -302,6 +302,73 @@ with 'Bugzilla::Elastic::Role::Object';
 
 sub ES_TYPE {'bug'}
 
+sub ES_INDEX { Bugzilla->params->{elasticsearch_index} }
+
+sub ES_SETTINGS {
```
floatingatoll: Were these patterns copied as-is from their prior location, or were changes made in support of conveying them over?

Bugzilla/Elastic/Role/Object.pm
```diff
@@ -9,7 +9,7 @@ package Bugzilla::Elastic::Role::Object;
 use 5.10.1;
 use Role::Tiny;
 
-requires qw(ES_TYPE ES_PROPERTIES es_document);
+requires qw(ES_TYPE ES_INDEX ES_SETTINGS ES_PROPERTIES es_document);
```
floatingatoll: To confirm, I assume this isn't meant to care about ES_PARENT_TYPE or _CLASS.

Bugzilla/Elastic/Indexer.pm
```diff
@@ -213,7 +155,8 @@ sub _select_updated_ids {
 sub bulk_load_ids {
     my ($self, $class, $ids) = @_;
 
-    $self->put_mapping($class);
+    $self->_create_index($class);
```
floatingatoll: This was unexpected - why did you add a _create_index here? Did it not work previously?

Bugzilla/Elastic/Indexer.pm
```diff
@@ -238,7 +181,6 @@ sub _bulk_load_ids {
     }
 
     my $total = 0;
-    use Time::HiRes;
```
floatingatoll: Can we remove all of these at once? Why is this particular ~fish~ instance deemed worthy of zapping?

Bugzilla/Elastic/Role/Object.pm
```diff
@@ -45,4 +45,6 @@ around 'es_document' => sub {
     return $doc;
 };
 
+
```
floatingatoll: Accidental or intentional?

Bugzilla/User.pm
```diff
@@ -126,6 +126,10 @@ use constant EXTRA_REQUIRED_FIELDS => qw(is_enabled);
 
 with 'Bugzilla::Elastic::Role::Object';
 
+sub ES_INDEX {
+    Bugzilla->params->{elasticsearch_index} . "_user";
```
floatingatoll: `. "_" . ES_TYPE` would be a bit less repetitive of the word 'User', which is already available in the class name and then human-specified both here at +130 and below at +133 in ES_TYPE.

scripts/bulk_index.pl
```diff
@@ -43,12 +43,10 @@ BEGIN
 }
 
 my $indexer = Bugzilla::Elastic::Indexer->new(
-    $debug_sql ? ( debug_sql => 1 ) : (),
+    $debug_sql    ? ( debug_sql => 1 )                      : (),
     $progress_bar ? ( progress_bar => 'Term::ProgressBar' ) : (),
 );
 
-$indexer->create_index;
```
floatingatoll: It seems like this removal pairs with an above code change that auto-creates indexes now?

Bugzilla/Elastic/Role/Object.pm
dylanwh: Yep! The ChildObject role is what requires ES_PARENT_CLASS. A role is basically an interface that can provide default implementations.

Bugzilla/Elastic/Role/Object.pm
dylanwh: accidental

Bugzilla/Elastic/Indexer.pm
dylanwh: Because bulk_load_ids() is public, it could be called when the index doesn't exist.

Bugzilla/Elastic/Indexer.pm
dylanwh: It shouldn't have been there to begin with -- but at the top level. Sorry for the unrelated change.

scripts/bulk_index.pl
dylanwh: yes, that's correct.

Bugzilla/User.pm
dylanwh: fixed in latest commit, good idea.

Bugzilla/Bug.pm
dylanwh: Pretty much copied over. looking at this, I think I should remove two of the analyzers that are only used in the users index... but that opens the can of worms about when the index settings don't match the code -- which should be addressed in a different bug.

floatingatoll: Do that in a separate commit for sure :)

Bugzilla/Elastic/Indexer.pm
floatingatoll: No worries, just log it in the commit message or make it a separate commit. I don't mind cleanup, I just mind *unlogged* cleanup :)

Bugzilla/Elastic/Indexer.pm
floatingatoll: It would be nice if the commit message mentioned this, since no other bug exists documenting the bug.

dylanwh:
That was merged without the proper commit message, so that commit is now gone. Re-issuing the PR.

PR #164, opened 2017-07-18
Bug 1381567 - wrap heredoc in an if

emceeaich:
Allow resolve_bugs.pl to run from command line.


dylanwh:
someone else has to review now, because I fixed the syntax nit. :-)

PR #165, opened 2017-07-19
Bug 1382085 - Add find_first_flag_type() to Bugzilla::Component

dylanwh:


Bugzilla/Component.pm
```diff
@@ -616,6 +616,16 @@ Component.pm represents a Product Component object.
 
  Returns:     Nothing.
 
+
+=item C<find_first_flag_type($target_type, $name).
+
+ Description: Changes the description of the component.
```
dklawren: Left over text from copy/paste. Change to match function.

dylanwh: fix on commit is not possible, so there will be one more approval needed

PR #166, opened 2017-07-19
Bug 1382225 - Missing code from PhabBugz extension such as http basic auth support and other minor improvements

dklawren:


extensions/PhabBugz/lib/Util.pm
```diff
@@ -53,23 +55,39 @@ sub get_revision_by_id {
...
+    ThrowUserError('invalid_phabricator_uri') unless $phab_base_uri;
+
+    my $revision_uri = $phab_base_uri . "D" . $revision_id;
+
+    # Check for previous attachment and return if one matches exact
+    my $review_attachment = first {
+        $_->contenttype eq PHAB_CONTENT_TYPE
+          && trim( $_->data ) eq trim($revision_uri)
+    }
+    @{ $bug->attachments };
```
dylanwh: This needs a comment -- and probably doesn't need the `@{ }` de-referencing.

extensions/PhabBugz/lib/Util.pm
```diff
@@ -53,23 +55,39 @@ sub get_revision_by_id {
...
+    return $review_attachment if defined $review_attachment;
+
+    # None present, so create new one
```
dylanwh: hmm, are we sure $dbh is always going to be the master db?

extensions/PhabBugz/lib/Util.pm
```diff
@@ -251,10 +269,10 @@ sub request {
         $ua->default_header('Content-Type' => 'application/x-www-form-urlencoded');
     }
 
-    my $phab_api_key  = $params->{phabricator_api_key};
-    my $phab_base_uri = $params->{phabricator_base_uri};
-    ThrowUserError('invalid_phabricator_uri') unless $phab_base_uri;
-    ThrowUserError('invalid_phabricator_api_key') unless $phab_api_key;
+    my $phab_api_key = $params->{phabricator_api_key}
+      || ThrowUserError('invalid_phabricator_api_key');
```
dylanwh: nit: The named operator would be better here. Seeing || makes me think you care about the return value.

extensions/PhabBugz/lib/Util.pm
```diff
@@ -265,11 +283,11 @@ sub request {
     ThrowCodeError('phabricator_api_error', { reason => $response->message })
         if $response->is_error;
 
-    my $result = decode_json($response->content);
-    if ($result->{error_code}) {
-        ThrowCodeError('phabricator_api_error',
-                       { code   => $result->{error_code},
-                         reason => $result->{error_info} });
+    my $result = eval { decode_json( $response->content ) };
+    if ( !$result || $@ ) {
```
dylanwh: Pretty sure you just want either $@ (although in rare pathological cases it can be cleared, because of DESTROY handlers) or defined $result.
To be super paranoid about someone passing 'null' as a json value...
```perl
my $result;
my $result_ok = eval { $result = decode_json( $response->content ); 1 };
if ( !$result_ok) {
```

extensions/PhabBugz/lib/WebService.pm
```diff
@@ -42,12 +43,23 @@ use constant PUBLIC_METHODS => qw(
 sub revision {
     my ($self, $params) = @_;
 
+    # Phabricator only supports sending credentials via HTTP Basic Auth
+    # so we exploit that function to pass in an API key as the password
+    # of basic auth. BMO does not support basic auth but does support
+    # use of API keys.
+    my $http_auth = delete $ENV{'HTTP_AUTHORIZATION'};
```
dylanwh: $cgi->http('Authorization'), directly using %ENV makes later porting more difficult/slow

extensions/PhabBugz/lib/Util.pm
```diff
@@ -53,28 +55,50 @@ sub get_revision_by_id {
...
+    my $revision_uri = $phab_base_uri . "D" . $revision_id;
+
+    # Check for previous attachment with same revision id.
+    # If one matches then return it instead. This is fine as
+    # BMO does not contain actual diff content.
+    my $review_attachment = first {
+        $_->contenttype eq PHAB_CONTENT_TYPE
+          && trim( $_->data ) eq trim($revision_uri)
+    }
+    @{ $bug->attachments };
```
dylanwh: This needs some reformatting... maybe private sub that checks against a revision?
```perl
 my $review_attachment = first { _is_phabricator_revision($_) } @{ $bug->attachments };
```

dklawren:
Thanks David. Dylan, mind giving a quick look over as well? I would feel better merging with two reviews due to the security nature of these changes.

Thanks
dkl

extensions/PhabBugz/lib/Constants.pm
```diff
@@ -0,0 +1,25 @@
...
+
+package Bugzilla::Extension::PhabBugz::Constants;
+
+use 5.10.1;
+use strict;
+use warnings;
+
+use base qw(Exporter);
+our @EXPORT = qw(
+  PHAB_AUTOMATION_USER
```
dylanwh: two-space indent is weird.

dklawren: So fixing the two space indent, everything else was ok?

extensions/PhabBugz/lib/WebService.pm
```diff
@@ -34,6 +35,7 @@ use Bugzilla::Extension::PhabBugz::Util qw(
     request
 );
 
+use MIME::Base64 qw(decode_base64);
 use Data::Dumper;
```
dylanwh: must've missed this before, but do we use Data::Dumper here or is this leftover from debugging?

PR #167, opened 2017-07-19
Bug 1381752 - Re-organize the layout

SebastinSanty:


SebastinSanty:
Hi @kyoshino, I heard about the two column layout similar to github and that sounds interesting. I'll try mocking up that design. A look at BUX header makes me feel that you are opting for a material design. Should I try material on new-bug too?

kyoshino:
I won't recommend entirely adopting Material Design especially on desktop, because people may confuse Bugzilla with Google apps. Still, it's worth learning from their well-designed guidelines.

Mimicking GitHub Issues' layout makes sense because many people are already familiar with it 😉 The current comment form is too wide anyway.

SebastinSanty:
screen shot 2017-07-21 at 9 27 05 pm

kyoshino:
Looks much better now 👍 I think the expander/title on each section (Labels & Status) isn't necessary because there are only a few fields. Other quick suggestions:

* Make the page title (Create New Bug) bigger
* Make the Submit button bigger
* Disable the Submit button until all the required fields (Summary, Product, Component, Version) are filled in
* Do we need the Labels section, by the way? I think Whiteboard isn't necessary, at least

dylanwh:
@kyoshino whiteboard is needed; the first users of this should be people that file Core bugs and currently keep multiple tabs open so they don't have to wait.

SebastinSanty:
`I think the expander/title on each section (Labels & Status) isn't necessary because there are only a few fields. `
Title is there, because other sections might be collapsed initially and hence a uniformity can be maintained across. It is taken from bug_modal.

* <del>Make the page title (Create New Bug) bigger</del> - Done
* <del>Make the Submit button bigger</del> - Done
* `Disable the Submit button until all the required fields (Summary, Product, Component, Version) are filled in` - I have added an alert instead. Disable would work, but I feel I have keep track of changes in those fields, which would include a bit more code.
* <del>Reverse the order of the Labels and Status sections, because the Status fields are required, while Keywords and Whiteboard are not</del> - Done
* `If the fields on the right side are too small (because Component names could be long), each field's label could be above the control instead of left` - I'll change them eventually as now, this is heavily based on `field.html.tmpl`. There are plans to move away from it though.

PR #168, opened 2017-07-19
Bug 1264093 - Security Bug Form

umohm1:


extensions/BMO/template/en/default/bug/create/create-firefox-security.html.tmpl
```diff
@@ -0,0 +1,100 @@
...
+            <br>
+
+            <input type="submit" id="commit" value="Submit Request">
+            <p>
+              [ <span class="required_star">*</span> <span class="required_explanation">
+              Required Field</span> ]
+            </p>
+          </form>
+
+          [% PROCESS global/footer.html.tmpl %]
```
dylanwh: The indentation is a little bit messed up in this file

PR #171, opened 2017-07-19
Bug 1381869 - Use separate elasticsearch index for Bugzilla::User

dylanwh:
RE-APPLY from PR #163.

This patch removes the concept of a single, bugzilla-wide index in favor of a
per-class index. bugs and comments continue to use
Bugzilla->params->{elasticsearch_index} but users use
Bugzilla->params->{elasticsearch_index} . "_user".

It is assured via the ChildObject trait (role) that comments will share the
index with bugs, and we have kept the index for bugs/comments the same to avoid
the multi-hour reindexing of production. Re-indexing users takes only five
minutes.

Subsequent work on this will allow use to version the index names and use
aliases, but I wanted to keep this patch small.

This patch also corrects some mistakes

1. $indexer->put_mapping() should not have been a public method.
2. Time::HiRes should be imported at the top of the file, not in a sub.


Bugzilla/User.pm
```diff
@@ -126,6 +126,11 @@ use constant EXTRA_REQUIRED_FIELDS => qw(is_enabled);
 
 with 'Bugzilla::Elastic::Role::Object';
 
+sub ES_INDEX {
+    my ($class) = @_;
+    sprintf("%s_%s", Bugzilla->params->{elasticsearch_index}, $class->ES_TYPE);
+}
+
 sub ES_TYPE { 'user' }
```
floatingatoll: Does it matter that you define ES_TYPE after ES_INDEX?

PR #180, opened 2017-07-22
Bug 1380531 - Write initial CircleCI configuration for BMO

dylanwh:
This is the first (working) iteration of a configuration that allows
CircleCI to build 'bmo' docker images that are runnable.

Some other changes were required for this:

1. The base image used in Dockerfile was changed from centos:6.7 to
   mozillabteam/bmo-base:slim.
   This is a new version of (github) mozilla-bteam/docker-bmo-base
   which is much smaller, and includes all the system RPMs and other
   bits that our bmo Dockerfile expects.
2. init.pl mistakenly had '--no-assets' which is not currently a checksetup.pl
   option.
3. 901-secure-mail-loop.t (which checks for an infinite loop) had
   to have its timeout increased because CircleCI sometimes is quite slow.
4. version.txt // __version__ now includes build information from CircleCI
   environmental variables
5. editorconfig knows about yml files now.
6. An error where we write out the default localconfig
   configuration when 'Bugzilla.pm' is not loaded was fixed.
   (in Bugzilla/Install/Localconfig.pm)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,10 +1,8 @@
-FROM centos:6.7
+FROM mozillabteam/bmo-base:slim
 MAINTAINER Dylan William Hardison <dylan@mozilla.com>
 
-RUN yum update -y && \
-    yum install -y perl perl-core mod_perl httpd wget tar openssl mysql-libs gd git && \
-    wget -q https://s3.amazonaws.com/moz-devservices-bmocartons/bmo/vendor.tar.gz && \
-    tar -C /opt -zxvf /vendor.tar.gz bmo/local/ bmo/LIBS.txt bmo/cpanfile bmo/cpanfile.snapshot && \
+RUN wget -q https://s3.amazonaws.com/moz-devservices-bmocartons/bmo/vendor.tar.gz && \
+    tar -C /opt -zxf /vendor.tar.gz bmo/local/ bmo/LIBS.txt bmo/cpanfile bmo/cpanfile.snapshot && \
     rm /vendor.tar.gz && \
     mkdir /opt/bmo/httpd && \
     ln -s /usr/lib64/httpd/modules /opt/bmo/httpd/modules && \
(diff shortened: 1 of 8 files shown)


Dockerfile
```diff
@@ -1,32 +1,16 @@
-FROM centos:6.7
+FROM mozillabteam/bmo-base:slim
```
ckolos: I think this repo should probably be added to circle as well since it's a core component of the BMO app as a whole.

ckolos: sorry, added to circleci as well as being auto-pushed to dockerhub

dylanwh: it is a docker hub automated build, it lives in mozilla-bteam/docker-bmo-base

dylanwh: https://hub.docker.com/r/mozillabteam/bmo-base/ and https://github.com/mozilla-bteam/docker-bmo-base

dylanwh: (pushing to docker-bmo-base will trigger the build on dockerhub)

Dockerfile
```diff
@@ -1,32 +1,16 @@
...
 MAINTAINER Dylan William Hardison <dylan@mozilla.com>
 
-RUN yum update -y && \
-    yum install -y perl perl-core mod_perl httpd wget tar openssl mysql-libs gd git && \
-    wget -q https://s3.amazonaws.com/moz-devservices-bmocartons/bmo/vendor.tar.gz && \
-    tar -C /opt -zxvf /vendor.tar.gz bmo/local/ bmo/LIBS.txt bmo/cpanfile bmo/cpanfile.snapshot && \
+RUN wget -q https://s3.amazonaws.com/moz-devservices-bmocartons/bmo/vendor.tar.gz && \
```
dylanwh: this is currently built by the carton_bundles repo, and will later be moved to circleci

dylanwh:
Yup, that is intended. What about the tag name? do you have a recommendation on generating the tag name? I feel the branch name is bad for that.

ckolos:
I'm okay with the image not being tagged with a specific name, i.e. mozilla/BMO:sha

PR #184, opened 2017-07-25
Bug 1383802 - Elasticsearch indexed documents lack an _id based on the object's primary key

dylanwh:


Bugzilla/Comment.pm
```diff
@@ -119,7 +119,7 @@ sub ES_SELECT_UPDATED_SQL {
 sub es_parent_id {
     my ($self) = @_;
 
-    return $self->bug_id,
+    return $self->ES_PARENT_CLASS->ES_TYPE . '_' . $self->bug_id,
```
floatingatoll: This code looks like a variation on a theme of `&es_id()` later in the patch. It might be worth a TODO comment about migrating it to use B:E:R:O:es_id() somehow instead of handcrafting, or not.

dylanwh: A comment must be needed, because $self here is a comment so it's not the same as $comment->es_id.

PR #186, opened 2017-07-27
Bug 1381566 - Add alternative query option to resolve_bugs.pl

umohm1:


scripts/resolve_bugs.pl
```diff
@@ -42,8 +44,15 @@
 $auto_user->{bless_groups} = [ Bugzilla::Group->get_all ];
 Bugzilla->set_user($auto_user);
 
-my $query = { resolution => '---', product => $product };
-$query->{component} = $component if defined $component;
+if ($buglist) {
+    my $uri = URI->new($buglist);
+    die("Invalid input") if $uri->host eq 'bugzilla.mozilla.org';
```
floatingatoll: I think this is a logic error - `eq` should be `ne` here, right?

I'd technically like to see 'bugzilla.mozilla.org' loaded from a Bugzilla::Config or something like that, rather than hard-coded, but I understand the expedience of doing it this way.

scripts/resolve_bugs.pl
```diff
@@ -42,8 +44,15 @@
 $auto_user->{bless_groups} = [ Bugzilla::Group->get_all ];
 Bugzilla->set_user($auto_user);
 
-my $query = { resolution => '---', product => $product };
-$query->{component} = $component if defined $component;
+if ($buglist) {
+    my $uri = URI->new($buglist);
+    die("Invalid input") if $uri->host ne 'bugzilla.mozilla.org';
```
dylanwh: Think about running this with the wrong url, and getting that error. The person running this would probably appreciate getting a more descriptive error message, such as "the buglist url must point to bugzilla.mozilla.org"

floatingatoll:
You may also wish to test $uri->protocol for 'https' and $uri->port for
'443'.

scripts/resolve_bugs.pl
```diff
@@ -10,7 +10,8 @@
 use strict;
 use warnings;
 use lib qw(. lib local/lib/perl5);
-
+use URI;
+use Test::URI;
```
dylanwh: I'm not sure what Test::URI is, but that's not needed.

scripts/resolve_bugs.pl
```diff
@@ -21,6 +22,10 @@
 
 my ($product, $component, $comment);
 my $resolution = 'WONTFIX';
+my $buglist;
+
+uri_protocol_ok( $uri, 'https');
```
dylanwh: Oh! I see. When I said test, I just meant like the existing code that does "die" with some sort of specific error message. This would be good if we were writing tests, but not so much for an interactive script

scripts/resolve_bugs.pl
```diff
@@ -42,8 +48,15 @@
 $auto_user->{bless_groups} = [ Bugzilla::Group->get_all ];
 Bugzilla->set_user($auto_user);
 
-my $query = { resolution => '---', product => $product };
-$query->{component} = $component if defined $component;
+if ($buglist) {
+    my $uri = URI->new($buglist);
+    die("The buglist url must match bugzilla.mozilla.org") if $uri->host ne 'bugzilla.mozilla.org';
```
dylanwh: just add addition die "..." if $uri->SOMETHING ne "something"

scripts/resolve_bugs.pl
```diff
@@ -42,8 +44,17 @@
...
-$query->{component} = $component if defined $component;
+if ($buglist) {
+    my $uri = URI->new($buglist);
+    die "The buglist url must match bugzilla.mozilla.org" if $uri->host ne 'bugzilla.mozilla.org';
+    die "Path must point to /buglist.cgi" if $uri->path ne '/buglist.cgi';
+    die "Protocol must match https" if $uri->protocol ne 'https';
+    $query = { $uri->query_form };
+}
+else {
+    $query = { resolution => '---',};
```
dylanwh: should include the product name too, { resolution => '---', product => $product }

scripts/resolve_bugs.pl
```diff
@@ -42,8 +44,17 @@
 $auto_user->{bless_groups} = [ Bugzilla::Group->get_all ];
 Bugzilla->set_user($auto_user);
```
dylanwh: need to declare $query on this line, e.g.
```perl
my $query;
```

PR #197, opened 2017-08-07
nit: only push to latest for the master branch

dylanwh:
I forgot to do this -- this makes it so only 'master' is pushed to 'latest'.

Later we will support other branches (and tags)


dylanwh:
the test failure is weird and looks like a circleci problem

PR #199, opened 2017-08-08
Bug 1388371 - Error: Invalid parameter bug_NNNNN passed to Bugzilla::User->visible_bugs: It must be numeric.

dylanwh:


Bugzilla/Elastic/Search.pm
```diff
@@ -102,7 +102,9 @@ sub data {
     my @fields = map { $REMAP_NAME{$_} // $_ } @{ $self->fields };
     my (@ids, %hits);
     foreach my $hit (@{ $result->{hits}{hits} }) {
-        push @ids, $hit->{_id};
+        my ($id) = $hit->{_id} =~ /^bug_([0-9]+)/;
+        die "invalid bug id: $hit->{_id}";
```
floatingatoll: die without if means +107 below is unreachable, r-

dylanwh: durp

dylanwh: yeah, I just got to that (after reloading the ES index on my dev box, which takes hours)

floatingatoll:
I can't see what changes I requested, but I feel like this is probably good to go. @dylanwh can you refresh this bug/issue or close them if it's already shipped in a different form?

PR #200, opened 2017-08-10
Bug 1389111 - Address complaints with autosized comments

wlach:
* Don't set max-height of comment if autosizing disabled
* Don't restrict resizing to vertical if autosizing is disabled


wlach:
@dylanwh this seems to work great in my local testing for both cases (autosizing disabled/enabled). Note that the autosizer will set the resize property to vertical by itself, so there's no need for us to put that in the CSS (perhaps we should not set it to anything at all, but I wanted to keep the set of changes here to a minimum).

PR #203, opened 2017-08-11
no bug - have the test runners start up the httpd

dylanwh:
This patch removes the background step that starts httpd, and instead has the entrypoint.pl spawn both prove and the httpd, in a way that makes them interdependent. If httpd exists early, prove is killed. When prove (which is, by the way, the test runner) exits, the httpd is shut down.


dylanwh:
My bad, this was never reporting failures because the $exitcode passed to on_finish() is the value of $?, which contains extra information. TL;DR is that you need to use POSIX::WEXITSTATUS() to get the real value.

dylanwh:
@SebastinSanty  or @floatingatoll -- this could use a review as tests will continue to fail randomly until it lands.

scripts/entrypoint.pl
```diff
@@ -174,8 +315,6 @@ sub write_localconfig {
         }
     }
 
-    unlink($filename);
```
floatingatoll: This change caught my eye, so I wanted to check in and make sure it's supposed to be here, but otherwise I'm not equipped to say whether this is a good patch or not — too much here is changed that I simply have no context for and can't test readily.

dylanwh: We no longer unlink, relying on the container to start without that file existing.
In the tests we have to ensure it is deleted before calling the entrypoint (and it is an obvious error if we forget). I'll document that behavior in the readme in a later PR.

dylanwh:
Just a note, all tests will continue to fail, randomly, until this is merged...
Note that all tests are passing on this PR.

It needs a review.

darkwing:
Worked for me!

PR #207, opened 2017-08-16
Bug 1391702 - Replace Bugzilla::User::validate_password() with calls to Data::Password::passwdqc

dylanwh:


Bugzilla/Config/Auth.pm
```diff
@@ -107,18 +107,41 @@ sub get_param_list {
...
+        {
+            name => 'passwdqc_match_length',
+            type => 't',
+            default => '4',
+        },
+
+        {
+            name => 'passwdqc_random_bits',
+            type => 't',
+            default => '47',
```
globau: these should use the `check_numeric` checker

Bugzilla/Config/Auth.pm
```diff
@@ -107,18 +107,41 @@ sub get_param_list {
...
-        {
             name    => 'password_check_on_login',
             type    => 'b',
             default => '1'
         },
+
+        {
+            name => 'passwdqc_min',
+            type => 't',
+            default => 'undef, 24, 11, 8, 7',
```
globau: this needs a checker to ensure the correct number of elements are provided, and that each element is either "undef" or an int.

template/en/default/admin/params/auth.html.tmpl
```diff
@@ -24,6 +24,79 @@
...
+
+[%  desc_passwdqc_max = BLOCK %]
+  The maximum allowed password length. This can be used to prevent users
+  from setting passwords that may be too long for some system services.
+  <p>
+  The value 8 is treated specially: with "max=8", passwords longer than
+  8 characters will not be rejected, but will be truncated to 8
+  characters for the strength checks and the user will be warned. This
+  is to be used with the traditional DES-based password hashes, which
+  truncate the password at 8 characters.
```
globau: this paragraph doesn't seem relevant to bugzilla, since it doesn't use oldschool hashes.  is it possible to disable the special handling of `max=8`?

Bugzilla/Install.pm
```diff
@@ -500,9 +500,14 @@ sub _prompt_for_password {
...
         my $pass2 = <STDIN>;
         chomp $pass2;
-        eval { validate_password($password, $pass2); };
-        if ($@) {
-            print "\n$@\n";
+        my $pwqc = Bugzilla->passwdqc;
+        my $ok = $pwqc->validate_password($password);
+        if (!$ok) {
+            print "\n", $pwqc->reason, "\n";
+            undef $password;
```
globau: it may be worth documenting the default password complexity requirements in the README, to ease the initial setup.

Bugzilla/WebService/Constants.pm
```diff
@@ -159,7 +158,6 @@ use constant WS_ERROR_CODE => {
     auth_cant_create_account    => 501,
     account_creation_disabled   => 501,
     account_creation_restricted => 501,
-    password_too_short    => 502,
```
globau: akin to 503, add a comment that 502 password_too_short no longer exists

qa/t/test_bmo_enter_new_bug.t
```diff
@@ -242,6 +242,7 @@ $sel->type_ok("comment", "--- Bug created by Selenium ---", "Enter bug descripti
 $sel->click_ok("commit", undef, "Submit bug data to post_bug.cgi");
 $sel->wait_for_page_to_load_ok(WAIT_TIME);
 $sel->is_text_present_ok('has been added to the database', 'Bug created');
+screenshot_page($sel, "/app/artifacts/power_bug_id.png");
```
globau: doesn't appear to be related to this change.

reset_password.cgi
```diff
@@ -49,7 +48,14 @@ if ($cgi->param('do_save')) {
...
     }
-    validate_password($password_1, $password_2);
+
+    my $pwqc = Bugzilla->passwdqc;
+    if (not $pwqc->validate_password($password_1)) {
+        ThrowUserError("password_insecure", { reason => $pwqc->reason });
+    }
+    elsif ($password_1 ne $password_2) {
+        ThrowUserError("password_mismatch");
+    }
```
globau: you should be checking that `$password_1 eq $password_2` _before_ checking the password complexity.

this same logic is repeated multiple times, so either apply the same fix everywhere, or move `password_mismatch` checking into a common fn.

template/en/default/admin/params/auth.html.tmpl
```diff
@@ -24,6 +24,79 @@
...
+  partially discounted.
+  <p>
+  The substring search is case-insensitive and is able to detect and
+  remove a common substring spelled backwards.
+[% END %]
+
+[%  desc_random_bits = BLOCK %]
+  The size of randomly-generated passphrases in bits (24 to 85), or 0 to
+  disable this feature.
+[% END %]
```
globau: i don't think this should be an admin setting, as `generate_password` is only called by `generate_test_data.pl`.  it isn't clear what disabling this feature would do.

Bugzilla/Config/Auth.pm
```diff
@@ -107,18 +107,41 @@ sub get_param_list {
         },
 
         {
-            name    => 'password_complexity',
```
globau: `password_complexity` is still used by `account/email/confirm-new.html.tmpl` and `global/password-features.html.tmpl`.

dylanwh:
Removing the UI is outside the scope of this bug, that's in bug 1390903 /  https://github.com/mozilla-bteam/bmo/pull/205
Both of these will land at the same time.

reset_password.cgi
dylanwh: No, it is intentional to check the password sameness second, as passwdqc also lets us set a max password. It is useful to avoid comparing huge strings.

Bugzilla/Config/Auth.pm
dylanwh: I'll skip removing that in this patch and move it to #205

template/en/default/admin/params/auth.html.tmpl
dylanwh: we'll be exposing generate_password() in a later patch, but random_bits will be always 85.

qa/t/test_bmo_enter_new_bug.t
dylanwh: debugging screenshot. :-)

dylanwh:
So almost ready for another review. IMHO repeating the password check code is preferable to having some opaque function like "valid_password" that sometimes takes one argument, sometimes two, and sometimes returns a boolean and at other times throws an exception (actually, there were two functions on that axis, but I digress)

Anyway, I will factor out the password validation *and* "did you type it correctly?" logic into a *verbosely* named function exported from Bugzilla::Util, something like:

```perl
sub assert_password_is_secure_and_typed_correctly {
my $pwqc = Bugzilla->passwdqc;
    if (not $pwqc->validate_password($pwd1)) {
        ThrowUserError('password_insecure', { reason => $pwqc->reason });
    }
    elsif ($pwd1 ne $pwd2) {
        ThrowUserError('password_mismatch');
    }
}
```

floatingatoll:
`if ($pwqc->insecure($pwd1)) { ThrowUserError('password_insecure'` seems like it would be simpler to read in code, fwiw.

dylanwh:
yes, but I'm also not about to change the API of a dependency. :-)

floatingatoll:
👍

dylanwh:
Note: leave-open until after tomorrow's push.

dylanwh:
I've added code here that changes the way we handle insecure passwords on login. Discussion to follow on the bug.

dylanwh:
looks like kang approved of this last change -- so @globau can re-approve?

PR #208, opened 2017-08-18
use mozillabteam/bmo-slim:20170818.1

dylanwh:
need to use a newer version of bmo-slim, which includes git. Without git, the "native" circleci is used which has a lot of bugs.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -7,7 +7,7 @@ version: 2
 
 defaults:
   bmo_slim_image: &bmo_slim_image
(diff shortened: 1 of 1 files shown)


dylanwh:
So things were failing, and I narrowed it down to this:

a screenshot of an error message

This was caused because we didn't have 'git' in the image, so circleci does its own thing:

another screenshot

That warning is apparently serious, as what happens is "git" they provide doesn't checkout the branch of the PR at all. Or at least not consistently.

floatingatoll:
In case it helps, https://twitter.com/sirrohara/status/898666398288760832

PR #210, opened 2017-08-20
Bug 1390112 - Option for filing security bugs: Part 2

dylanwh:


floatingatoll:
Could you instead filter the listed products to only those matching the selected security group?

PR #211, opened 2017-08-21
Bug 1392769 - Add other fields to new-bug

SebastinSanty:


dylanwh:
looks like a sanity test failure on this.

dylanwh:
It looks like user auto completion works for the mentor field, but not CC or assignee.

SebastinSanty:
Any updates to this?

PR #214, opened 2017-08-22
Bug 1390316 - Unused variable in PhabBugz script $phab_api_key

umohm1:


extensions/PhabBugz/lib/Util.pm
```diff
@@ -314,15 +314,8 @@ sub request {
         $ua->default_header('Content-Type' => 'application/x-www-form-urlencoded');
     }
 
-    my $phab_api_key = $params->{phabricator_api_key};
```
dylanwh: No changes to this file -- the variables here appear to all be used.

PR #218, opened 2017-08-23
Bug 1393145 - See if we can add a directive/rule to robots.txt to get…

umohm1:
… search engines to crawl more slowly


floatingatoll:
Or, replace it with "User-agent: *" so that it's clearly declared,
depending on which makes more sense in this context.

dylanwh:
@umohm1 or anyone -- this can be merged once tests pass.

PR #222, opened 2017-08-25
Bug 1394023 - Possible Suggestions based on Summary

SebastinSanty:


dylanwh:
if there are no matches, can the duplicates table not be shown?
Also can there be a link to "dismiss" (or close) it?

SebastinSanty:
I feel that the duplicates table should be shown as soon as three characters are filled in the summary box. This makes the users aware of the existence of duplicates table.

I'll put the closing button in my next commit.

PR #224, opened 2017-08-26
Bug 1393888 - Write httpd access handler to block by ip address stored in memcached

dylanwh:


dylanwh:
atoll is on PTO, and this is important -- hope you can take a look at it sooner than later. (it's more important than the password code)

Bugzilla/ModPerl/BlockIP.pm
```diff
@@ -0,0 +1,59 @@
...
+sub unblock_ip {
+    my ($class, $ip) = @_;
+    $MEMCACHED->delete("block_ip:$ip") if $MEMCACHED;
+}
+
+sub handler {
+    my $r = shift;
+    return Apache2::Const::OK if $r->uri =~ $STATIC_URI;
+
+    my $ip = $r->headers_in->{'X-Forwarded-For'} // $r->connection->remote_ip;
```
globau: `X-Forwarded-For` can be forged by the client; they could set it to a different value with each request and bypass this block.

assuming there's always only one proxy server in front of bmo, you should be able to split on `[\s,]+` and use the penultimate element as the ip.

scripts/block-ip.pl
```diff
@@ -0,0 +1,56 @@
+
+#!/usr/bin/perl
```
globau: `#!` should be the first line

Bugzilla/ModPerl/BlockIP.pm
dylanwh: Is that still the case? That would be a gross misconfiguration of Zeus.

scripts/block-ip.pl
dylanwh: going to blame vscode for that.

Bugzilla/ModPerl/BlockIP.pm
globau: if there's proxy servers between the client and zeus that should appear in `x-forwarded-for`; zeus should just be appending its ip on the list.

Bugzilla/ModPerl/BlockIP.pm
```diff
@@ -40,7 +40,13 @@ sub handler {
     my $r = shift;
     return Apache2::Const::OK if $r->uri =~ $STATIC_URI;
 
-    my $ip = $r->headers_in->{'X-Forwarded-For'} // $r->connection->remote_ip;
+    my $ip = $r->headers_in->{'X-Forwarded-For'};
+    if ($ip) {
+        $ip = (split(/\s*,\s*/ms, $ip))[-1];
```
globau: i'm not 100%, but i'm pretty sure you want the second-last ip address here; iirc the last one will always be zeus.

perhaps at this point it's better to use Bugzilla::Util::remote_ip(), which we know does the right thing with respect to x-forwarded-for and inbound proxies (it works backwards through x-forward-for until it finds the first non-proxy address).

dylanwh:
I think there is a misunderstanding of how X-Forwarded-For is used by load balancers.
For Amazon/ELBs, the last item is always the client IP.
And as configured, Zeus strips off X-Forwarded-For regardless.

All of the magic stuff that bugzilla does in remote_ip() is really a work-around for broken proxy implementations.

```http
GET /forward.cgi HTTP/1.1
Accept: */*
Accept-Encoding: gzip, deflate
Connection: keep-alive
Host: bugzilla-dev.allizom.org
User-Agent: HTTPie/0.9.9
X-Forwarded-For: 192.168.1.1

HTTP/1.1 200 OK
Connection: Keep-Alive
Content-Encoding: gzip
Content-Type: text/plain; charset=UTF-8
Date: Mon, 11 Sep 2017 15:52:12 GMT
Server: Apache
Set-Cookie: {omitted}; path=/; secure; HttpOnly
Set-Cookie: {omitted}; path=/; secure; HttpOnly
Strict-transport-security: max-age=31536000; includeSubDomains
Transfer-Encoding: chunked
Vary: Accept-Encoding
X-Backend-Server: web5.stage.bugs.scl3.mozilla.com
X-content-type-options: nosniff
X-frame-options: SAMEORIGIN
X-xss-protection: 1; mode=block

174.228.131.128
```

floatingatoll:
X-Forwarded-For should be explicitly trustable *in the scenario* where this
code is run in a trusted environment, with load balancers that strip and
set XFF themselves. I believe our Zeus config is such that we do exactly
this - *always* set a trusted XFF - so the implication that it cannot be
trusted applies only to some instances, not production BMO.

klibby:
To follow up on atoll's comment: Zeus is currently stomping on any pre-existing XFF headers, and setting it's own via a TrafficScript rule with http.setheader. This applies to all Bugzilla properties behind Zeus (dev, stage, bmoattachments, etc).

I cannot comment on what happens in the non-Zeus environment that's just around the corner, though.

PR #230, opened 2017-09-05
Bug 1396957 - Synchronize CC list on security bugs over to revisions as subscribers

dklawren:
- This change will update the subscriber list of the revision when assignee, qa_contact, or cc list members   change since any of those will allow access to the bug.
- I switched to using subscribers.set instead of add in the Conduit API call which will overwrite whatever the current subscriber list is. This works well except in the case where someone adds themself as a subscriber in Phabricator instead of as a CC on the bug.
- TODO: we need to check the group changes in the push connector data to see if the groups actually changed and if not, then just do the subscribers change only. Right now it will create a new custom policy even if groups have not changed. This pollutes the change history of the revision needlessly.

### DIFF
--- a/Bugzilla/Config/Common.pm
+++ b/Bugzilla/Config/Common.pm
@@ -30,6 +30,7 @@ use base qw(Exporter);
     check_bug_status check_smtp_auth check_theschwartz_available
     check_maxattachmentsize check_email
     check_comment_taggers_group
+    get_all_group_names
 );
 
 # Checking functions for the various values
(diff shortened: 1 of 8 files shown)


dklawren:
Fixed merge conflicts and added logging for subscriber only changes. As for the error it worked for me in the different use cases that did not include a group change.

1 Changing assignee only
2. Changing qa contact only
3. Changing cc only
4. Changing any combination of the above

We can work through it on monday.

PR #232, opened 2017-09-08
Bug 1398340 - Typo in 2fa enrollment flow: "enrolment" (missing an L)

dylanwh:


template/en/default/global/user-error.html.tmpl
```diff
@@ -1401,7 +1401,7 @@
 
   [% ELSIF error == "reset_password_denied" %]
     [% title = "Reset Password Denied" %]
-    You cannot reset your password without administrative permission. 
+    You cannot reset your password without administrative permission.
```
emceeaich: What's our policy on whitespace cleanup like this? Should it had been a separate PR? Since it's one line in one file it's not a blocker for me.

floatingatoll:
I've seen no policy, but I loathe mix-in cleanups because any revert later
unnecessarily rips out cleanups and/or creates merge conflicts with them.

PR #233, opened 2017-09-08
Bug 1397927 - Add comment to bug when a revision is accepted/denied with the users real names

dklawren:


floatingatoll:
I support using Github's wording here ("approved", "requested changes")
here as well.

dklawren:
I will use ("approved", "requested changes") as atoll suggested.

floatingatoll:
I meant to agree with the previous suggestion which wasn’t mine - if
there’s variance, go with theirs over mine :)

PR #235, opened 2017-09-11
Bug 1398889 - Add param 'silent_users' that never trigger sending bug…

dylanwh:
…mail

### DIFF
--- a/Bugzilla/User.pm
+++ b/Bugzilla/User.pm
@@ -2246,6 +2246,7 @@ sub wants_bug_mail {
     my $self = shift;
     my ($bug, $relationship, $fieldDiffs, $comments, $dep_mail, $changer) = @_;
 
+    # is_silent_user is true if the username is mentioned in the param `silent_users`
     return 0 if $changer && $changer->is_silent_user;
 
     # Make a list of the events which have happened during this bug change,
@@ -2420,8 +2421,8 @@ sub is_silent_user {
     my $self = shift;
 
     if (!exists $self->{'is_silent_user'}) {
-        my @helpers = split(/\s*,\s*/, Bugzilla->params->{'silent_users'});
-        $self->{'is_silent_user'} = (any { $self->login eq $_ } @helpers) ? 1 : 0;
+        my @users = split(/\s*,\s*/, Bugzilla->params->{'silent_users'});
+        $self->{'is_silent_user'} = (any { $self->login eq $_ } @users) ? 1 : 0;
     }
 
     return  $self->{'is_silent_user'};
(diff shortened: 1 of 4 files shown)


emceeaich:
UI looked fine on Mozilla-dev but couldn't test bugmail blocking there

Bugzilla/User.pm
```diff
@@ -2407,13 +2409,24 @@ sub is_insider {
 sub is_global_watcher {
     my $self = shift;
 
-    if (!defined $self->{'is_global_watcher'}) {
-        my @watchers = split(/[,;]+/, Bugzilla->params->{'globalwatchers'});
-        $self->{'is_global_watcher'} = scalar(grep { $_ eq $self->login } @watchers) ? 1 : 0;
+    if (!exists $self->{'is_global_watcher'}) {
+        my @watchers = split(/\s*,\s*/, Bugzilla->params->{'globalwatchers'});
```
dylanwh: note: it was never documented that a ; could be used, so I removed it. Also added optional whitespace

template/en/default/admin/params/mta.html.tmpl
```diff
@@ -73,6 +73,8 @@
                "Set to 0 to disable whining.",
 
   globalwatchers => "A comma-separated list of users who should receive a " _
-                    "copy of every notification mail the system sends." }
+                    "copy of every notification mail the system sends.",
+
+  silent_users => "A comma-separated list of users who never trigger sending email" }
```
globau: nit: missing period at end of sentence.

Bugzilla/User.pm
```diff
@@ -2407,13 +2409,24 @@ sub is_insider {
...
     }
     return  $self->{'is_global_watcher'};
 }
 
+sub is_silent_user {
+    my $self = shift;
+
+    if (!exists $self->{'is_silent_user'}) {
+        my @helpers = split(/\s*,\s*/, Bugzilla->params->{'silent_users'});
+        $self->{'is_silent_user'} = (any { $self->login eq $_ } @helpers) ? 1 : 0;
```
globau: `helpers` is a weird name to use here, instead of something like `users`.

Bugzilla/User.pm
```diff
@@ -2246,6 +2246,8 @@ sub wants_bug_mail {
     my $self = shift;
     my ($bug, $relationship, $fieldDiffs, $comments, $dep_mail, $changer) = @_;
 
+    return 0 if $changer && $changer->is_silent_user;
```
globau: i feel this needs a comment which points to the param.

Bugzilla/User.pm
dylanwh: woops. Originally it was "global_helpers" before I settled on silent_users

dylanwh:
Fixed. Also guess what! globalwatchers was split on a DIFFERENT regex! I updated BugMail.pm to be consistent.

PR #239, opened 2017-09-13
WIP: Bug 1399593 - Add sync group comments during new revision creation as well as push connector

darkwing:
Will open a bug in a moment.

Want to get dkl's opinion on this quick.

### DIFF
--- a/extensions/PhabBugz/lib/Util.pm
+++ b/extensions/PhabBugz/lib/Util.pm
@@ -11,7 +11,10 @@ use 5.10.1;
 use strict;
 use warnings;
 
+use Bugzilla::Bug;
+use Bugzilla::Constants;
 use Bugzilla::Error;
+use Bugzilla::User;
 use Bugzilla::Util qw(trim);
 use Bugzilla::Extension::PhabBugz::Constants;
 
@@ -23,6 +26,7 @@ use base qw(Exporter);
 
 our @EXPORT = qw(
     add_comment_to_revision
+    add_security_sync_comments
     create_revision_attachment
     create_private_revision_policy
     create_project
@@ -32,6 +36,7 @@ our @EXPORT = qw(
     get_members_by_bmo_id
     get_project_phid
     get_revisions_by_ids
+    get_security_sync_groups
     intersect
     is_attachment_phab_revision
     make_revision_private
@@ -380,6 +385,48 @@ sub request {
     }
 
     return $result;
+}
+
+sub get_security_sync_groups {
+    my $bug = shift;
+
+    my $phab_sync_groups = Bugzilla->params->{phabricator_sync_groups}
+        || ThrowUserError('invalid_phabricator_sync_groups');
+    my $sync_group_names = [ split('[,\s]+', $phab_sync_groups) ];
+
+    my $bug_groups = $bug->groups_in;
+    my $bug_group_names = [ map { $_->name } @$bug_groups ];
+
+    my @set_groups = intersect($bug_group_names, $sync_group_names);
+
+    return @set_groups;
+}
+
+sub add_security_sync_comments {
+    my ($revisions, $bug) = @_;
+
+    my $phab_error_message = 'Revision is being made private due to unknown Bugzilla groups.';
+
+    foreach my $revision (@$revisions) {
+        add_comment_to_revision( $revision->{phid}, $phab_error_message );
+    }
+
+    my $num_revisions = 0 + @$revisions;
+    my $bmo_error_message =
+    ( $num_revisions > 1
+    ? $num_revisions.' revisions were'
+    : 'One revision was' )
+    . ' made private due to unknown Bugzilla groups.';
+
+    my $user = Bugzilla::User->new( { name => PHAB_AUTOMATION_USER } );
+    $user->{groups} = [ Bugzilla::Group->get_all ];
+    $user->{bless_groups} = [ Bugzilla::Group->get_all ];
+    Bugzilla->set_user($user);
+
+    $bug->add_comment( $bmo_error_message, { isprivate => 0 } );
+
+    my $bug_changes = $bug->update();
+    $bug->send_changes($bug_changes);
 }
 
 1;
--- a/extensions/PhabBugz/lib/Util.pm
+++ b/extensions/PhabBugz/lib/Util.pm
@@ -83,6 +83,9 @@ sub create_revision_attachment {
     my $is_shadow_db = Bugzilla->is_shadow_db;
     Bugzilla->switch_to_main_db if $is_shadow_db;
 
+    my $old_user = Bugzilla->user;
+    _set_phab_user();
+
     my $dbh = Bugzilla->dbh;
     $dbh->bz_start_transaction;
 
@@ -106,6 +109,8 @@ sub create_revision_attachment {
 
     $dbh->bz_commit_transaction;
     Bugzilla->switch_to_shadow_db if $is_shadow_db;
+
+    Bugzilla->set_user($old_user);
 
     return $attachment;
 }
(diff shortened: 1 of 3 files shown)


extensions/PhabBugz/lib/WebService.pm
```diff
@@ -89,19 +90,13 @@ sub revision {
...
-        my @set_groups = intersect($bug_group_names, $sync_group_names);
+        my @set_groups = get_security_sync_groups($bug);
 
         # If bug privacy groups do not have any matching synchronized groups,
         # then leave revision private and it will have be dealt with manually.
         if (!@set_groups) {
-            ThrowUserError('invalid_phabricator_sync_groups');
+            #ThrowUserError('invalid_phabricator_sync_groups');
+            add_security_sync_comments(@revisions, $bug);
         }
```
darkwing: We can just let this fall through, right?  An admin policy doesn't hurt anything, right?  Or do we need to do something else?

darkwing:
OK, this should be good to look at @dklawren

extensions/PhabBugz/lib/Util.pm
```diff
@@ -382,4 +387,46 @@ sub request {
...
+sub add_security_sync_comments {
+    my ($revisions, $bug) = @_;
+
+    my $phab_error_message = 'Revision is being made private due to unknown Bugzilla groups.';
+
+    foreach my $revision (@$revisions) {
+        add_comment_to_revision( $revision->{phid}, $phab_error_message );
+    }
+
+    my $num_revisions = 0 + @$revisions;
```
dklawren: Use:

my $num_revisions = scalar @$revisions;

extensions/PhabBugz/lib/Util.pm
```diff
@@ -382,4 +387,46 @@ sub request {
...
+    my $bmo_error_message =
+    ( $num_revisions > 1
+    ? $num_revisions.' revisions were'
+    : 'One revision was' )
+    . ' made private due to unknown Bugzilla groups.';
+
+    my $user = Bugzilla::User->new( { name => PHAB_AUTOMATION_USER } );
+    $user->{groups} = [ Bugzilla::Group->get_all ];
+    $user->{bless_groups} = [ Bugzilla::Group->get_all ];
+    Bugzilla->set_user($user);
```
dklawren: Need to store old user when using set_user()

my $old_user = Bugzilla->user;
Bugzilla->set_user($phab_user);
[do stuff]
Bugzilla->set_user($old_user);

extensions/PhabBugz/lib/Util.pm
```diff
@@ -382,4 +387,46 @@ sub request {
...
+    }
+
+    my $num_revisions = 0 + @$revisions;
+    my $bmo_error_message =
+    ( $num_revisions > 1
+    ? $num_revisions.' revisions were'
+    : 'One revision was' )
+    . ' made private due to unknown Bugzilla groups.';
+
+    my $user = Bugzilla::User->new( { name => PHAB_AUTOMATION_USER } );
```
dklawren: You could even not do this if the user is already PHAB_AUTOMATION_USER in the case of an API call and not the push connector:

if (Bugzilla->user->name ne PHAB_AUTOMATION_USER) {

}

extensions/PhabBugz/lib/Util.pm
floatingatoll: You can just use bare `@num_revisions` below, since all uses are scalar context, and drop 414 `my $num_revisions` entirely.

`@num_revisions > 1`
`? @num_revisions.' revisions were'`

are both valid.

darkwing:
Updates made; I didn't do the user check as it seemed to complicated the code unnecessarily, IMO

extensions/PhabBugz/lib/WebService.pm
```diff
@@ -89,19 +91,12 @@ sub revision {
...
+        my @set_groups = get_security_sync_groups($bug);
 
         # If bug privacy groups do not have any matching synchronized groups,
         # then leave revision private and it will have be dealt with manually.
         if (!@set_groups) {
-            ThrowUserError('invalid_phabricator_sync_groups');
+            add_security_sync_comments(\@revisions, $bug);
         }
 
         my $policy_phid = create_private_revision_policy($bug, \@set_groups);
```
dklawren: So we are getting a failure here in that create_private_revision_policy looks up the project_ids for the bug group set. In this case, the list is empty and so we ThrowUserError(invalid_phabricator_sync_groups). This causes Dorrkeeper to requeue and we get a lot of looping comments.

For this I think we either need to 1) not set a policy at all leaving the current policy in place (most likely 'Admins only') and return or 2) set a custom policy to 'Admins Only' and subscribers here instead. We could do 2) to be sure I suppose in case it had gotten changed in Phab by a user.

In the push connector you use make_revision_private() instead of create_private_revision_policy so we need to match.

darkwing:
@dklawren Anything left here?

extensions/PhabBugz/lib/Util.pm
darkwing: Are the two `$user->{groups}` and `$user->{bless_groups}` required here at all?

darkwing:
*Also in create_revision_attachment() you will need to do Bugzilla->set_user() like you do in add_security_sync_comments() for the same reason as above.*

How much of the code would that need to wrap?  `my $attachment = Bugzilla::Attachment->create(` to `Bugzilla->switch_to_shadow_db if $is_shadow_db;`?  A bit confused about this comment.

extensions/PhabBugz/lib/WebService.pm
darkwing: 1.  Could we simply change the code to:

```
# Initialize this result to admin
my $result = 'admin';

# .....

# If bug privacy groups do not have any matching synchronized groups,
# then leave revision private and it will have be dealt with manually.
if (!@set_groups) {
    add_security_sync_comments(\@revisions, $bug);
}
else {
    my $policy_phid = create_private_revision_policy($bug, \@set_groups);
    my $subscribers = get_bug_role_phids($bug);
    $result = edit_revision_policy($revision_phid, $policy_phid, $subscribers);
}
```

...and then let it keep going?

2.

dklawren: 1. I thought about that and normally we could just do nothing. But we could also just set the policy to admin + subscribers. It would still be a custom policy. This would allow the assignee, qa contact, and cc members to see the revision even though it doesnt have any phab project (groups) associated with it. If we do nothing then the subscribers are not added and they would not be able to see the revision anyway.

2. Yeah maybe it doesnt matter here. I was saying because it would be a shortcut to making the revision private to admin/admin if @set_groups is empty. We use that in Push connector when a bugs groups are changed to a set of group which are not being synced. Which sounded very similar to what is occurring in WebService.pm as well.

dklawren:
> How much of the code would that need to wrap? my $attachment = Bugzilla::Attachment->create( to
> Bugzilla->switch_to_shadow_db if $is_shadow_db;? A bit confused about this comment.

Reason for this is because we still want to add an attachment to the bug even if phab-bot is not in the groups checked for the bug. If we do not wrap Attachment->create with the heightened permissions it may thrown an error.

extensions/PhabBugz/lib/Util.pm
dklawren: Actually the bless_groups may be omitted as it just makes it where the user can bless (add) users to the any group in BMO which we do not need to do. We should be okay with just 'groups'.

darkwing:
So just the `attachment.create` . anywhere else?

darkwing:
Should `$attachment->update($timestamp);` also be wrapped?

dklawren:
> Should $attachment->update($timestamp); also be wrapped?

Before the $dbh->bz_start_transaction and after $dbh->bz_commit_transaction is this case.

Some places we are not doing transactions so you would wrap around any attachment->update and bug->update calls.

dklawren:
```
  sub create_private_revision_policy {
!8    my ($bug, $groups, $set_admin) = @_;

      my $data = {
          objectType => 'DREV',
          default    => 'deny',
          policy     => [
              {
                  action => 'allow',
+                 rule   => 'PhabricatorSubscriptionsSubscribersPolicyRule',
+             }
+         ]
+     };
+
+     if (!$set_admin) {
+         my $project_phids = [];
+         foreach my $group (@$groups) {
+             my $phid = get_project_phid('bmo-' . $group);
+             push(@$project_phids, $phid) if $phid;
+         }
+
+         ThrowUserError('invalid_phabricator_sync_groups') unless @$project_phids;
+
+         push(@{ $data->{policy} },
+             {
+                 l
+                 action => 'allow',
                  rule   => 'PhabricatorProjectsPolicyRule',
                  value  => $project_phids,
!             }
+         );
+     }
+     else {
+         push(@{ $data->{policy} },
              {
                  action => 'allow',
!                 value  => 'admin',
              }
!         );
!     }
```

floatingatoll:
"{ 1 action" midway through this patch may cause trouble; is the 1 meant to
have a key => prefixed?

darkwing:
Even in the updates, at current state, still getting the looping messages.

dklawren:
I did not encounter the looping in the latest iteration. It does however loop if something is not configured properly on each end (api keys, etc.). Issue is that since BMO is the one adding the comments, we cannot check the failure count and prohibit adding additional comments for failure count > 1. So we should look into doing that as a separate bug.

PR #240, opened 2017-09-15
Fix cmd httpd

dylanwh:


scripts/entrypoint.pl
```diff
@@ -38,7 +38,8 @@
 sub cmd_httpd  {
     check_data_dir();
     wait_for_db();
-
+    run( 'perl', 'checksetup.pl', '--no-template' );
+    run( '/usr/sbin/httpd', '-DFOREGROUND', '-f', '/app/httpd/httpd.conf' );
```
floatingatoll: Starting Apache seems sensible enough.

scripts/entrypoint.pl
```diff
@@ -249,7 +250,7 @@ sub wait_for_db {
         $dbh = DBI->connect(
             $dsn,
             $c->{db_user},
-            $c->{db_pass}, 
+            $c->{db_pass},
```
floatingatoll: Whitespace nit without any mention in a commit message.

scripts/entrypoint.pl
```diff
@@ -38,7 +38,7 @@
 sub cmd_httpd  {
     check_data_dir();
     wait_for_db();
-
+    run( '/usr/sbin/httpd', '-DFOREGROUND', '-f', '/app/httpd/httpd.conf' );
```
floatingatoll: As before, httpd seems fine. I guess checksetup isn't needed?

dylanwh: Yep, looks like it

PR #241, opened 2017-09-16
Bug 1400419 - Direct GitHub-auth'd users that are required to use 2FA to issue a password reset

dylanwh:


dylanwh:
There are about 100 github users (at least) that will be impacted by the update on Monday. This change gives a password change button on the 2fa page (and allows locked out users to access token.cgi).

template/en/default/account/prefs/mfa.html.tmpl
```diff
@@ -16,11 +23,15 @@
...
 [% IF user.cryptpassword == '*' %]
-  <input type="hidden" name="mfa_action" id="mfa-action" value="">
+  <input type="hidden" name="mfa_action" id="mfa-action" value="" data-nopassword="true">
   <p>
     Two-factor Authentication is not available on your account because you are
     using an external authentication provider.
   </p>
+  <p>
+    Clicking "Reset Password" will cause [% terms.Bugzilla %] to send you an
+    email which will allow you to create a password.
```
globau: this message should describe why a "reset password" is required here, and what the implications will be.  in particular if mfa-required means they will no longer be able to authenticate with github that needs to be crystal clear here.

dylanwh:
Let's get @emceeaich's input on the revised text, duplicated below for ease of reading:

> Your account requires MFA to be configured before you can use Bugzilla.
>
> As part of this transition you will no longer be able to use GitHub to log in to Bugzilla; instead you will need to set a password for your existing account on Bugzilla and use that to login in future.
>
> To set a password on your account please click on "Reset Password"; this will email you instructions to complete this process.

floatingatoll:
What if they don’t have Secure Mail configured?

template/en/default/account/prefs/mfa.html.tmpl
```diff
@@ -16,10 +23,17 @@
   [% RETURN %]
 [% END %]
 [% IF user.cryptpassword == '*' %]
-  <input type="hidden" name="mfa_action" id="mfa-action" value="">
+  <input type="hidden" name="mfa_action" id="mfa-action" value="" data-nopassword="true">
+  <p>
+    Your account requires MFA to be configured before you can use [% terms.Bugzilla %]
```
emceeaich: Your account requires Multi-factor Authentication (MFA) to be configured before you can use [% terms.Bugzilla %].

emceeaich: MFA or 2FA here?

template/en/default/account/prefs/mfa.html.tmpl
```diff
@@ -16,10 +23,17 @@
...
+  </p>
+  <p>
+    As part of this transition you will no longer be able to use GitHub to log in to [% terms.Bugzilla %];
+    instead you will need to set a password for your existing account on Bugzilla and use that to login in future.
```
emceeaich: You will no longer be able to use GitHub to log in to [% terms.Bugzilla %]; instead you will need to set a password for your existing account on Bugzilla and use that to log in with your MFA application.

The MFA provider applications which work with [% terms.Bugzilla %] are: Duo, Google, Blah, Blah, Blah. 

The Mozilla wiki has <a href="https://wiki.mozilla.com/path/to/page">instructions on enabling your 2FA application</a>.

dylanwh:
@floatingatoll re: "What if they don’t have Secure Mail configured?"

SecureMail only encrypts password emails (or blanks them) if you're in a security group.
GitHub users are not allowed to be in those groups.

floatingatoll:
Huh. Okay. Thanks!

template/en/default/account/prefs/mfa.html.tmpl
dylanwh: That wording isn't needed. Once they change the password the regular 2FA instructions will be shown.

dylanwh: But do need to s/MFA/Two-Factor Authentication/

template/en/default/account/prefs/mfa.html.tmpl
```diff
@@ -16,10 +23,19 @@
...
-  <input type="hidden" name="mfa_action" id="mfa-action" value="">
+  <input type="hidden" name="mfa_action" id="mfa-action" value="" data-nopassword="true">
+  [% IF Bugzilla.request_cache.mfa_warning %]
+    <p>
+      Your account requires Two-Factor authentication to be configured before you can use [% terms.Bugzilla %]
+    </p>
+  [% END %]
+  <p>
+    Before you can enable two-factor authentication you must set a password on your account.
+    To do this on your account please click on "Reset Password";
```
globau: this is a bit clumsy with "on your account" repeated.
i suggest tweaking it later to read `To do this please click on "Reset Password";`.

PR #244, opened 2017-09-20
Bug 1401569 - typo in "enabled two-factor authentication"

globau:


dylanwh:
CI is failing to global DNS incident, circleci will rerun tests I think.

PR #252, opened 2017-09-27
Bug 1403344 - Add command to docker container for schema / data migra…

dylanwh:
…tions


dylanwh:
I had to actually generate a fake schema change to test this, very short change. 
It does work. :-)

Dockerfile
```diff
@@ -1,6 +1,5 @@
 FROM mozillabteam/bmo-slim:20170927.1
 
-ENV BUNDLE=https://s3.amazonaws.com/moz-devservices-bmocartons/bmo/vendor.tar.gz
```
floatingatoll: This change isn't documented as necessary anywhere in the PR or in the bug, but it's mixed in with a documented change below to add checksetup. Is this an intentional change?

dylanwh: I'll add it back, it isn't used by anything but it was left over from another session.

Dockerfile
```diff
@@ -1,5 +1,6 @@
 FROM mozillabteam/bmo-slim:20170927.1
 
+
```
globau: don't forget to remove this

PR #256, opened 2017-10-02
Bug 1405011 - Prevent code in mod_perl.pl from executing twice

dylanwh:
The code in mod_perl.pl is executed twice. mod_perl does this to "make sure graceful restarts work".

It isn't clear that this is useful, since we don't use graceful restarts (which are not recommended with apache / mod_perl anyway) and it costs performance.
But more importantly it makes sanitizing %ENV more complicated than it needs to be.

This patch uses a source filter which clears out the rest of the mod_perl.pl during the first apache restart (when restart_count < 2). It includes a warning message so that this won't be a weird to debug problem in the future. Although I doubt it will, since I'm getting the hell off of apache as soon as we're out of scl3.)

### DIFF
--- a/Bugzilla/ModPerl/StartupFix.pm
+++ b/Bugzilla/ModPerl/StartupFix.pm
@@ -1,0 +1,43 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+#
+# This Source Code Form is "Incompatible With Secondary Licenses", as
+# defined by the Mozilla Public License, v. 2.0.
+package Bugzilla::ModPerl::StartupFix;
+use 5.10.1;
+use strict;
+use warnings;
+
+use Filter::Util::Call;
+use Apache2::ServerUtil ();
+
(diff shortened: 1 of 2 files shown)


dylanwh:
I don't think this will bite us, and it's also a good change for bugzilla prod.

Bugzilla/ModPerl/StartupFix.pm
```diff
@@ -0,0 +1,43 @@
...
+            }
+            else {
+                $_ = "";
+            }
+        }
+    }
+    return $status;
+}
+
+1;
```
floatingatoll: I think Github is trying to tell me there's no newline at the end of this file, which is slightly odd.

Bugzilla/ModPerl/StartupFix.pm
```diff
@@ -0,0 +1,43 @@
...
+    my ($type) = @_;
+    my ($ref)  = {};
+    filter_add( bless $ref );
+}
+
+sub filter {
+    my ($self) = @_;
+    my ($status);
+    if ($status = filter_read() > 0) {
+        if (Apache2::ServerUtil::restart_count() < 2) {
```
floatingatoll: A code comment about why this is "2", and not "1" or "3" or "==", would be helpful for posterity here.

PR #260, opened 2017-10-06
no bug - document code and fix single-arg bless()

dylanwh:
Patch to address the documentation issue


dylanwh:
@floatingatoll hey, this simple change will make the linter happier, if you want to LGTM it.

PR #265, opened 2017-10-26
Bug 1379607 - Reimplement Google Analytics on bugzilla.mozilla.org

kyoshino:
* Modify CSP to allow GA to run. This doesn't require `unsafe-inline` especially on the home page, as I've avoided inline scripts
* Disable tracking of people in the core-security group
* Omit part of the page title of bugs in any (confidential) group as well as some user-related pages
* Log template names instead of real URLs so it's easier to distinguish some multiple section pages, e.g. search options and preferences
* Use the debugging GA script on test servers
* Anonymize IP address, rather than respecting Do Not Track which reduces the number of logs, for better privacy

This is blocked until the GA property ID is provided by Gareth.


floatingatoll:
Could you say more about this part: "Anonymize IP address, rather than respecting Do Not Track" ?

extensions/BMO/template/en/default/hook/global/header-additional_header.html.tmpl
```diff
@@ -52,3 +52,25 @@
     [% END %]
   [% END %]
 [% END %]
+
+[%# BMO - Google Analytics %]
+[% IF !user.in_group('core-security') %]
```
dylanwh: This actually needs to check the insider group param.
user.in_group(Param('insidergroup'))

kyoshino:
[Anonymizing IP address](https://support.google.com/analytics/answer/2763052) is a feature of GA itself. GA will still track visitors but it doesn't record their real IP address.

I believe that site analytics using 1st party cookies is not subject to Do Not Track, and the [previous implementation](https://github.com/mozilla-bteam/bmo/commit/e4d5d34) didn't respect DNT actually. But hmm, Mozilla's [Website Privacy Notice](https://www.mozilla.org/en-US/privacy/websites/) says GA needs to be disabled when DNT is used:

> If you have configured your browser to send a “Do Not Track” signal when accessing our websites, Mozilla will not utilize any of the tools described in the Metrics section.

Then I'll update the pull request to get it covered.

extensions/BMO/template/en/default/hook/global/header-additional_header.html.tmpl
kyoshino: Fixed

kyoshino:
Okay, so I've eventually made a separate extension with a new admin panel, making the tracking ID configurable.

extensions/GoogleAnalytics/template/en/default/hook/global/header-additional_header.html.tmpl
```diff
@@ -0,0 +1,28 @@
+[%# This Source Code Form is subject to the terms of the Mozilla Public
+  # License, v. 2.0. If a copy of the MPL was not distributed with this
+  # file, You can obtain one at http://mozilla.org/MPL/2.0/.
+  #
+  # This Source Code Form is "Incompatible With Secondary Licenses", as
+  # defined by the Mozilla Public License, v. 2.0.
+  #%]
+
+[% RETURN UNLESS Bugzilla.params.google_analytics_tracking_id && !user.in_group(Param('insidergroup')) %]
+
```
globau: we shouldn't load GA on provide bugs.  this probably means

`[% RETURN IF bug.defined && bug.groups_in.size %]`

extensions/GoogleAnalytics/template/en/default/hook/global/header-additional_header.html.tmpl
```diff
@@ -0,0 +1,28 @@
...
+  [%~ ELSIF template.name == 'pages/user_profile.html.tmpl' ~%]
+    User Profile
+  [%~ ELSE ~%]
+    [% title FILTER html %]
+  [%~ END ~%]
+  ">
+<script async src="https://www.google-analytics.com/analytics
+  [%~ IF Bugzilla.params.google_analytics_debug ~%]
+    _debug
+  [%~ END ~%]
```
globau: this can be rewritten as `[%~ "_debug" IF Bugzilla.params.google_analytics_debug ~%]

dylanwh:
@kyoshino I haven't seen anything beyond what glob noticed. The test failure is not related (just noise). Do you have time to make those changes or do you need me to step in and do it?

kyoshino:
@globau Thanks for your review! @dylanwh Updated the PR to address the comments. Also added [`collapse`](http://www.template-toolkit.org/docs/manual/Filters.html#section_collapse) to the title filter to make sure a proper page title is sent to GA.

dylanwh:
Getting an error:

```
Content Security Policy: The page’s settings blocked the loading of a resource at data:application/javascript;base64,KGZ1b
```

kyoshino:
On which page? I don't see any error on my vagrant box 🤔

extensions/GoogleAnalytics/web/js/dnt-helper.js
```diff
@@ -0,0 +1,54 @@
...
+* License, v. 2.0. If a copy of the MPL was not distributed with this
+* file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+// create namespace
+if (typeof Mozilla === 'undefined') {
+    var Mozilla = {};
+}
+
+/**
+ * Returns true or false based on whether doNotTack is enabled. It also takes into account the
```
emceeaich: Typo: doNotTrack

kyoshino: This file is just a copy from [Bedrock](https://github.com/mozilla/bedrock/blob/master/media/js/base/dnt-helper.js) so I'll fix it separately 😆

kyoshino: Fixed.

Bugzilla/CGI.pm
```diff
@@ -34,9 +34,9 @@ BEGIN {
 sub DEFAULT_CSP {
     my %policy = (
         default_src => [ 'self' ],
-        script_src  => [ 'self', 'unsafe-inline', 'unsafe-eval' ],
+        script_src  => [ 'self', 'unsafe-inline', 'unsafe-eval', 'https://www.google-analytics.com' ],
         child_src   => [ 'self', ],
-        img_src     => [ 'self', 'https://secure.gravatar.com' ],
+        img_src     => [ 'self', 'https://secure.gravatar.com', 'https://www.google-analytics.com' ],
```
reedloden: I highly recommend that you use XHR over images for GA.

See https://githubengineering.com/githubs-post-csp-journey/ for specifics as to the security risks of images vs. XHR.

kyoshino: Umm, but this is what (all?) other Mozilla properties are doing.

kyoshino: Also, because BMO stores/serves attachments on different hosts, the security risk could be mitigated, I guess.

dylanwh: Reed is correct, but we're already vulnerable to this with gravatar 's d=parameter. We should file a *separate* bug to fix both (lock-down img-src to only urls we control and use XHR for GA and an image proxy for gravatar)

kyoshino: Just for reference: BzDeck was using a [service worker](https://github.com/bzdeck/bzdeck/blob/master/webroot/static/scripts/workers/service-worker.js) to proxy Gravatar requests.

dylanwh: @kyoshino does that mean the origin of a service worker is different than a direct ? do we need img-src in that case? Regardless, this is off-topic -- I'll file a new bug for this deficiency

kyoshino: BzDeck's [CSP](https://github.com/bzdeck/bzdeck/blob/master/webroot/.htaccess#L60-L63) still had `https://secure.gravatar.com` for `img-src`. I thought that was necessary for some reason.

PR #267, opened 2017-10-31
 Bug 1393950 - Block users from signing into Phabricator unless they have MFA enabled

dylanwh:


dylanwh:
Sorry for the confusion.

PR #272, opened 2017-11-17
Bug 1417980 - Fix non-HTTPS links and outdated links where possible

kyoshino:
Fix [Bug 1417980 - Fix non-HTTPS links and outdated links where possible](https://bugzilla.mozilla.org/show_bug.cgi?id=1417980)

Note: I'll keep some historical broken links, like bonsai.mozilla.org or lxr.mozilla.org, that have no replacement.


Bugzilla/DB/Schema.pm
```diff
@@ -3179,6 +3179,6 @@ database-specific basis:
 
 L<Bugzilla::DB>
 
-L<http://www.bugzilla.org/docs/developer.html#sql-schema>
+L<https://www.bugzilla.org/docs/developer.html#sql-schema>
```
floatingatoll: Are all www.bugzilla.org//docs/ URLs meant to be re-written to readthedocs?

extensions/BMO/template/en/default/account/create.html.tmpl
```diff
@@ -83,31 +83,19 @@ function onSubmit() {
...
-      icon = "other"
-      name = "Support for other products"
-      url = "http://www.mozilla.org/projects/"
-      desc = "Support for products not listed here."
-  %]
-  [% INCLUDE product
       icon = "input"
       name = "Feedback"
-      url = "http://input.mozilla.com/feedback"
+      url = "https://input.mozilla.org/feedback"
```
floatingatoll: Is this still actually operational and in use?

extensions/BMO/template/en/default/bug/create/create-ipc.html.tmpl
```diff
@@ -167,7 +167,7 @@ $(document).ready(function() {
     </div>
 
     <p>
-    If you would like your snippet to include a link, please tag it by using the URL builder <a href="http://bit.ly/1JOgDr1">(http://bit.ly/1JOgDr1)</a> and following the steps below:
+    If you would like your snippet to include a link, please tag it by using the URL builder <a href="https://bit.ly/1JOgDr1">(https://bit.ly/1JOgDr1)</a> and following the steps below:
```
floatingatoll: I don't like these bit.ly links. We likely have no reason to use a URL shortener here, and if we do, we need to use mzl.la for it.

template/en/default/bug/create/user-message.html.tmpl
```diff
@@ -30,7 +30,7 @@
 [% PROCESS global/variables.none.tmpl %]
 
 Before reporting [% terms.abug %], please read the 
-<a href="page.cgi?id=bug-writing.html">
```
floatingatoll: Is this already a broken link, or is that page removed by this patch as well?

template/en/default/pages/release-notes3.html.tmpl
```diff
@@ -3312,11 +3312,11 @@
 <p>This gives [% terms.Bugzilla %] very advanced plugin support. You can
   hook templates, hook code, add new parameters, and use the XML-RPC
   interface. So we'd like to see some [% terms.Bugzilla %] plugins
-  written! Let us know on the <a href="http://bugzilla.org/cgi-bin/mj_wwwusr?func=lists-long-full&amp;extra=developers">developers&#64;bugzilla.org</a>
+  written! Let us know on the <a href="https://lists.bugzilla.org/cgi-bin/mj_wwwusr?func=lists-long-full&amp;extra=developers">developers&#64;bugzilla.org</a>
   mailing list if you write a plugin.</p>
 
 <p>If you need more hooks, please 
-  <a href="http://www.bugzilla.org/developers/reporting_bugs.html">File a b<!-- -->ug</a>!</p>
+  <a href="https://www.bugzilla.org/developers/reporting_bugs.html">File a b<!-- -->ug</a>!</p>
```
floatingatoll: `b<!-- -->ug` should be replaced by [% terms.bug %], right?

Bugzilla/DB/Schema.pm
kyoshino: This particular page is not available on readthedocs (yet)

extensions/BMO/template/en/default/account/create.html.tmpl
kyoshino: AFAIK input.m.o is no longer developed, but it's now redirected to a new feedback page.

extensions/BMO/template/en/default/bug/create/create-ipc.html.tmpl
kyoshino: I don't know why those bit.ly links exist, but it's probably out of scope.

template/en/default/bug/create/user-message.html.tmpl
kyoshino: Not broken so far, but has been moved to MDN. Used the same URL as other guideline links.

template/en/default/pages/release-notes3.html.tmpl
kyoshino: Wasn't aware of that, but it's out of scope I think.

dylanwh: Since it is linking to an extern system, using the comment hack is fine.

PR #273, opened 2017-11-17
Bug 1418204 - Update Firefox logo

kyoshino:
Fix [Bug 1418204 - Update Firefox logo](https://bugzilla.mozilla.org/show_bug.cgi?id=1418204)


dylanwh:
I'm assuming you tested this and the size change works out fine.

PR #280, opened 2017-12-01
Bug 1422329 - The phabricator conduit API method feed.query_id return data format has changed so the phabbugz_feed.pl daemon needs to be updated

dklawren:


extensions/PhabBugz/lib/Revision.pm
```diff
@@ -30,7 +30,11 @@ my $SearchResult = Dict[
         authorPHID        => Str,
         dateCreated       => Int,
         dateModified      => Int,
+        diffPHID          => Str,
         policy            => Dict[ view => Str, edit => Str ],
+        repositoryPHID    => Maybe[Str],
+        status            => Dict,
```
dylanwh: s/Dict/HashRef/ in this case. A dict is a hashref that must have specific keys.

PR #282, opened 2017-12-02
Bug 1328659 - Add support for utf8=utf8mb4 (switches to dynamic/compressed row format, and changes charset to utf8mb4)

dylanwh:


Bugzilla/Config/General.pm
```diff
@@ -33,8 +33,9 @@ use constant get_param_list => (
 
     {
         name    => 'utf8',
-        type    => 'b',
-        default => '0',
+        type    => 't',
+        choices => [ '1', 'utf8', 'utf8mb4' ],
+        default => 'utf8',
```
dklawren: Should be instead:

type => 's',
choices => [ 'utf8', 'utf8mb4' ],
default => 'utf8',

Bugzilla/DB/Mysql.pm
```diff
@@ -559,11 +575,13 @@ sub bz_setup_database {
...
     # TABLE_COLLATION IS NOT NULL prevents us from trying to convert views.
+    my $charset = $self->utf8_charset;
+    my $collate = $self->utf8_collate;
     my $non_utf8_tables = $self->selectrow_array(
         "SELECT 1 FROM information_schema.TABLES
           WHERE TABLE_SCHEMA = ? AND TABLE_COLLATION IS NOT NULL
-                AND TABLE_COLLATION NOT LIKE 'utf8%'
-          LIMIT 1", undef, $db_name);
+                AND TABLE_COLLATION != ?
+          LIMIT 1", undef, $db_name, $collate);
```
dylanwh: @dklawren this is one of those changes I talked about. Before it was a prefix match against 'utf8', so if it was utf8mb4_unicode_blah, it would still match. In prod it appears it is only utf8_general_ci

dklawren:
Checked out the latest changes and after restart of DB everything checked out. r=dkl

PR #286, opened 2017-12-08
Bug 1424155 - Write scripts to import/export attachments to disk

dylanwh:


dylanwh:
I realize everyone might be traveling, so whoever reviews this first wins a raffle ticket for a copy Netscape Navigator Gold, still in original shrink wrap.

scripts/archive-attachments.pl
```diff
@@ -0,0 +1,64 @@
...
+    my $hash   = sha256($data);
+    my $bug_id = $attachment->bug_id;
+    print pack 'NNNa32a*', $bug_id, $attach_id, length($data), sha256($data), $data;
+}
+
+__DATA__
+
+=head1 NAME
+
+archive-attachments.pl - save attachments to a dump file
```
globau: this script doesn't "archive"; it dumps/extracts and should be named accordingly.  in my mind archiving would include the act of removing the attachment data from the database after it has been dumped.

globau: please add a third script which processes the list of attachment-ids in the same way that deletes the attachdata.  this is safer than running a DELETE against the production database, and allows for bulk deletion at a later date regardless of the storage backend.

dylanwh:
@globau What measure should I take to make the attachment deletion script not be a foot gun?

dylanwh:
also, I'm renaming these attachment-{import,export,remove} (purge for the new one}

globau:
@dylanwh depends on how through you want to be i guess.

if you wanted to only allow deletion of exported attachments you could change reading attachment-ids from STDIN to specifying a list by filename.  when an attachment is exported update that line in the file, and only delete attachments which specify they have been exported successfully.

dylanwh:
I've made it so it is one script and three sub-commands.
Removing requires reading in an archive, which validates the archive.
I've added a checksum to the archive metadata as well.

dylanwh:
It did not handled properly already-deleted attachments.

PR #288, opened 2017-12-08
Bug 1424408 - "Sign in with GitHub" button triggers a bugzilla securi…

dylanwh:
…ty error, if I'm viewing a page with e.g. "&product=Core&component=SVG" in the URL

### DIFF
--- a/github.cgi
+++ b/github.cgi
@@ -44,7 +44,7 @@ if (lc($cgi->request_method) eq 'post') {
       unless $target_uri =~ /^\Q$urlbase\E/;
 
     ThrowCodeError("github_insecure_referer", { target_uri => $target_uri })
-      if $cgi->referer && $cgi->referer =~ /(reset_password\.cgi|token\.cgi|t=|token=|api_key=)/;
+      if $cgi->referer && $cgi->referer =~ /(?:reset_password\.cgi|token\.cgi|\bt=|token=|api_key=)/;
 
     if ($user->id) {
         print $cgi->redirect($target_uri);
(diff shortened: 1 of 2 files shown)


dylanwh:
This was matching all strings that cointain a t=, e.g. product=. \b makes it match only &t= as was originally intended.

github.cgi
```diff
@@ -44,7 +44,7 @@ if (lc($cgi->request_method) eq 'post') {
       unless $target_uri =~ /^\Q$urlbase\E/;
 
     ThrowCodeError("github_insecure_referer", { target_uri => $target_uri })
-      if $cgi->referer && $cgi->referer =~ /(reset_password\.cgi|token\.cgi|t=|token=|api_key=)/;
+      if $cgi->referer && $cgi->referer =~ /(?:reset_password\.cgi|token\.cgi|\bt=|token=|api_key=)/;
```
floatingatoll: Would `/\b(?:reset_password\.cgi|token\.cgi|t=|token=|api_key=)/` be a useful consideration here?

floatingatoll:
I hesitate to r+ this since it comes with no improvements to tests, but it does appear to be a sound quick fix for the issue described.

dylanwh:
If we add \b for everything then we need (?:Bugzilla_)?api_key but otherwise yeah

floatingatoll:
I mean, I could go into “why aren’t we URI-parsing”, but at some level it’s
not apparent to *me* how detailed this should be - so I defer to your
judgement.

dylanwh:
Note this still needs an approval.

PR #289, opened 2017-12-10
Bug 1376826 - New HTML Header for BMO

kyoshino:
Fix [Bug 1376826 - New HTML Header for BMO](https://bugzilla.mozilla.org/show_bug.cgi?id=1376826)

## Description

* Implement the new global navigation
* Make the header fixed-positioning with a comment focus workaround (partially solve [Bug 1330451](https://bugzilla.mozilla.org/show_bug.cgi?id=1330451))
* Use the modal bug page's dropdown menu script/style for the header menus (and fix the bugs)
* Improve part of the mini login form UX (to be overhauled later in [Bug 1420769](https://bugzilla.mozilla.org/show_bug.cgi?id=1420769))
* Remove the redundant footer navigation links
* Tested with Firefox, Chrome, Safari, Edge, IE11
* The design is tentative; we need a post-Sandstone style guide for buttons, menus, etc.

## Screenshots

screen shot 2017-12-09 at 22 58 34
screen shot 2017-12-09 at 22 58 16


kyoshino:
There are still test failures in test_bmo but seems unrelated 🤔

dylanwh:
Fixing test cases while reviewing it is definitely a thing I can do.

dylanwh:
When I turn off gravatar, it still loads the gravatar for the profile menu item.

dylanwh:
Nevermind, I got the two gravatar settings confused.

kyoshino:
Just checked the 2 preferences. I should probably use the latter. Will fix it.

* Show gravatar images when viewing bugs: `[% IF user.settings.show_gravatars.value == 'On' %]`
* Show my gravatar image to other users: `[% IF user.gravatar %]`

dylanwh:
Neither of those are quite right though. I'd keep it as-is.

kyoshino:
Some failures in test_selenium are valid as link labels have been changed, e.g. New to New Bug. Fixing them.

SebastinSanty:
Documentation doesn't link for me.

kyoshino:
Yeah I noticed that. It doesn't work unless you set the `docs_urlbase` in the admin settings page, which is https://bmo.readthedocs.org/en/latest/ on BMO.

kyoshino:
Okay **test_bmo** is now fixed. Previously the header's login form was inconsistent depending on the page, but I've solved the issue. That means the login page will have 2 login forms, one on the header and another on the body, hence the test failures. I've added `//*[@id="bugzilla-body"]` to the XPath expression to work around it.

kyoshino:
This is the last failure but hmm... how could this happen? 😕 
> Failed test 'is_text_present, QA-Selenium-TEST' at [test_show_all_products.t line 51](https://github.com/kyoshino/bmo/blob/07f322ebd92d9097f62081cee85b2c51ee601ec2/qa/t/test_show_all_products.t#L51).

dylanwh:
Time for a debugging screenshot!
Adding this at that point in the test will cause a screenshot artifact to be saved on the circleci build:

```perl
screenshot_page($sel, "/app/artifacts/what-is-going-on.png");
```

kyoshino:
A screenshot was captured and uploaded to somewhere in the cloud.

dylanwh:
I will look from my laptop in about 15 minutes

dylanwh:
https://2752-17650869-gh.circle-artifacts.com/0/app/artifacts/test_show_all_products.png

kyoshino:
Thanks! I definitely see a link labelled "QA-Selenium-TEST" on the screenshot. And the comment under the problematic line 51 reads 

> For some unknown reason, Selenium doesn't like hyphens in links.

So... I'm going to comment this out. Actually this test is redundant as line 54 checks the link again:

```perl
$sel->click_ok('//div[@id="choose_product"]//a[contains(@href, "QA-Selenium-TEST")]');
```

dylanwh:
The selenium test suite should be treated as legacy; new test should either look like test_bmo's selenium test or be totally new. It's possible to use /any/ software for writing the tests, as long as it can run in a docker container.

kyoshino:
Yay, all tests passed now 🎉

template/en/default/global/header.html.tmpl
```diff
@@ -244,96 +244,148 @@
...
+      <div class="dropdown">
+        <button type="button" id="header-tools-menu-button" class="dropdown-button minor" title="More tools…"
+                aria-label="More tools…" aria-expanded="false" aria-haspopup="true" aria-controls="header-tools-menu">
+          <span class="icon" aria-hidden="true"></span>
+        </button>
+        <ul class="dropdown-content" id="header-tools-menu" role="menu" style="display:none;">
+          [% IF user.showmybugslink %]
+            [% filtered_username = user.login FILTER uri %]
+            <li role="presentation">
+              <a href="[% Param('mybugstemplate').replace('%userid%', filtered_username) %]" role="menuitem" tabindex="-1">My [% terms.Bugs %]</a>
```
dylanwh: I realize this is probably in the existing code, but I feel that 'FILTER uri FILTER html' is often more correct.
Actually, is it even required that hrefs be uri-escaped, rather than html-escaped?

template/en/default/global/header.html.tmpl
```diff
@@ -244,96 +244,148 @@
...
+        </ul>
+      </div>
+    </nav>
+    [% Hook.process("badge") %]
+    [% IF user.id %]
+      <div id="header-accunt" class="dropdown">
+        <button type="button" id="header-account-menu-button" class="dropdown-button minor" title="Account"
+                aria-label="Account" aria-expanded="false" aria-haspopup="true" aria-controls="header-account-menu">
+          [% IF user.gravatar %]
+            <img src="[% user.gravatar FILTER none %]" width="32" height="32" alt="">
```
dylanwh: FILTER html here, not none.

template/en/default/global/useful-links.html.tmpl
```diff
@@ -26,10 +26,6 @@
 [% PROCESS global/variables.none.tmpl %]
 
 <ul id="useful-links">
-  <li id="links-actions">
-      [% PROCESS "global/common-links.html.tmpl" qs_suffix = "_bottom" %]
-  </li>
-
```
dylanwh: I have wanted to do this for so long! 👍 👍 👍

template/en/default/global/header.html.tmpl
kyoshino: I think [`FILTER uri`](http://www.template-toolkit.org/docs/manual/Filters.html#section_uri) does the right thing here.

template/en/default/global/header.html.tmpl
kyoshino: Or probably `FILTER uri` agian.

kyoshino:
Specified a viewport and media query so the header is still usable on mobile screen. Otherwise part of it may go out of view.

dylanwh:
One last thing, the caching headers in index.cgi should be removed.

kyoshino:
Now I see `Cache-Control: no-cache, max-age=300` on `index.cgi`.

PR #291, opened 2017-12-11
Bug 1424787 - Due Date on bug modal is 1 day ahead

kyoshino:
Fix [Bug 1424787 - Due Date on bug modal is 1 day ahead](https://bugzilla.mozilla.org/show_bug.cgi?id=1424787)

## Description

* Use the default `html` filter in the same way as the hidden `<input>` field for the due date.
* Enclose the value in the `<time>` element.
* I have added the custom field locally but can't find on bug pages, so not tested this yet actually.

### DIFF
--- a/extensions/BugModal/template/en/default/bug_modal/field.html.tmpl
+++ b/extensions/BugModal/template/en/default/bug_modal/field.html.tmpl
@@ -161,7 +161,7 @@ END;
 
           [% CASE constants.FIELD_TYPE_DATE %]
             [%# date %]
-            [% value FILTER time("%Y-%m-%d") %]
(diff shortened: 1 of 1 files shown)


dylanwh:
cf_due_date is limited to some products in extenions/BMO/lib/Data.pm.
I'm going to add a new product to the default data.

kyoshino:
Okay! I just quickly added bugzilla.mozilla.org::General to see the Due Date field. Can't reproduce the issue right now, but it should work, because the hidden `<input>` field on [line 293](https://github.com/kyoshino/bmo/blob/master/extensions/BugModal/template/en/default/bug_modal/field.html.tmpl#L293) only uses the `html` filter.

PR #287, opened 2017-12-08
Minor fix: Remove deprecation warning due to `include` in Ansible

SebastinSanty:

PR #292, opened 2017-12-12
Update Ansible to the latest during vagrant provision

SebastinSanty:


SebastinSanty:
Fixed what @kyoshino pointed out in #287

PR #294, opened 2017-12-12
Bug 1424940 - Support HTML5 datepicker

SebastinSanty:
Removing the datetimepicker plugin is left.


dylanwh:
The date format -- how can we specify that? @kyoshino any input?
I know bugzilla supports only a few specific format for these values.

kyoshino:
As per [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/date), "the actual value is always normalized to the format `yyyy-mm-dd`". So even the visible value is `12 / 12 / 2017` (looks like there's [some weirdness](https://bugzilla.mozilla.org/show_bug.cgi?id=1421910) in Firefox 57), the browser will send `2017-12-12` to the server, which should be fine with Bugzilla.

PR #295, opened 2017-12-13
Bug 1425119 - Release new-bug interface

SebastinSanty:
This PR discusses everything that needs to be done, in order to release new-bug interface to the public.


dylanwh:
This needs a bug to land.

SebastinSanty:
@dylanwh I have updated the PR, does it look good?

dylanwh:
oh yeah, probably split on /\s*,\s*/ rather than /, /

SebastinSanty:
Yep, a better solution. BTW would having a beta tag be better? We can have a feedback from users.

PR #298, opened 2017-12-16
add a developer mode to make frontend changes a bit easier

dylanwh:
1. assets are not concatenated
2. templates are not cached in memory or on disk

As a result, templates or js/css can be updated with rsync:

for instance:
```
rsync -vaz --chmod=a+rX template/ web:template/
```


dylanwh:
In light of the new /static/ path, we'll need to make this mode not send HTTP caching headers as well.

PR #299, opened 2017-12-17
Bug 1425783 - Split up mod_perl.pl into multiple modules

dylanwh:
split mod_perl.pl out into:

Bugzilla::ModPerl
Bugzilla::ModPerl::ResponseHandler
Bugzilla::ModPerl::CleanupHandler

The only behavioral change is that the memcached object used by the BlockIP access handler is explicitely set during startup -- after we're certain we have memcached configuration values.


floatingatoll:
A comment says:

> Loading Bugzilla.pm doesn't load this

Does it mean “these”?

(If not, an extra newline or two would help understand what it means.)

Bugzilla/ModPerl.pm
```diff
@@ -14,12 +14,93 @@ use File::Find ();
...
+    Bugzilla::Extension->load_all();
+
+    Bugzilla->preload_features();
+
+    # Force instantiation of template so Bugzilla::Template::PreloadProvider can do its magic.
+    Bugzilla->template;
+
+    Bugzilla::ModPerl::BlockIP->set_memcached( Bugzilla::Memcached->_new()->{memcached} );
+
+    # Have ModPerl::RegistryLoader pre-compile all CGI scripts.
```
dylanwh: This doesn't actually happen. We added this in [Bug 1214704](https://bugzilla.mozilla.org/show_bug.cgi?id=1214704) (which I reviewed), but it never actually loads the CGI until it gets a request.

Bugzilla/ModPerl.pm
```diff
@@ -14,12 +14,101 @@ use File::Find ();
 use Cwd ();
 use Carp ();
 
-# We don't need (or want) to use Bugzilla's template subclass.
-# it is easier to reason with the code without all the extra things Bugzilla::Template adds
-# (and there might be side-effects, since this code is loaded very early in the httpd startup)
-use Template ();
```
globau: `apache_config` still uses `Template`, so this should remain.

Bugzilla/ModPerl.pm
```diff
@@ -14,12 +14,101 @@ use File::Find ();
...
+    Apache2::SizeLimit->set_max_unshared_size($limit);
+
+    my $cgi_path = Bugzilla::Constants::bz_locations()->{'cgi_path'};
+
+    # Set up the configuration for the web server
+    my $server = Apache2::ServerUtil->server;
+    my $conf   = Bugzilla::ModPerl->apache_config($cgi_path);
+    $server->add_config( [ grep { length $_ } split( "\n", $conf ) ] );
+
+
```
globau: nit: too many blank lines

Bugzilla/ModPerl.pm
globau: it's actually from the original mod_perl implementation ([bug 345376](https://bugzilla.mozilla.org/show_bug.cgi?id=345376)).  bug 1214704 was a rollback.

Bugzilla/ModPerl.pm
```diff
@@ -14,12 +14,101 @@ use File::Find ();
...
+    # Pre-load all extensions
+    Bugzilla::Extension->load_all();
+
+    Bugzilla->preload_features();
+
+    # Force instantiation of template so Bugzilla::Template::PreloadProvider can do its magic.
+    Bugzilla->template;
+
+    Bugzilla::ModPerl::BlockIP->set_memcached( Bugzilla::Memcached->_new()->{memcached} );
```
globau: it's weird to use a private constructor and then reach into the object to get the memcached var.  i see that's how the old code worked, but as you're touching how and when this object is created it would be nice (ie. optional) to fix this now.

i'd prefer to see a public method added to `Bugzilla::Memcache` which returns the internal `{memcached}` object, and give `Bugzilla->memcached` the ability to return the non-metric'd version upon request.

Bugzilla/ModPerl/ResponseHandler.pm
```diff
@@ -0,0 +1,56 @@
...
+
+    # Prevent "use lib" from modifying @INC in the case where a .cgi file
+    # is being automatically recompiled by mod_perl when Apache is
+    # running. (This happens if a file changes while Apache is already
+    # running.)
+    no warnings 'redefine';
+    local *lib::import = sub {};
+    use warnings;
+
+
```
globau: nit: too many blank lines

Bugzilla/ModPerl.pm
dylanwh: I just made the constructor public. The metrics code can probably be removed in a subsequent PR, I don't think it would be easy to get running again given changes to elasticsearch and so on.

dylanwh:
I'm going to need this change for oauth2 support, and I'm making further changes to mod_perl.pl now, so this had to be brought back up to par.

dylanwh:
I'll start this from scratch again.

PR #300, opened 2017-12-18
Bug 1427621 - Support uploading multiple attachments from new-bug

SebastinSanty:


SebastinSanty:
This breaks the normal attachments, I'm fixing that. Otherwise, on the new-bug side, it is perfectly working. The design can be commented upon. r? @dylanwh @kyoshino

dylanwh:
Yep, you should change all the places where get_content_type is called.

dylanwh:
@SebastinSanty there needs to a bug on file for this -- then I'll land it.

dylanwh:
@SebastinSanty need to resolve the conflict, then I'll approve and land.

PR #301, opened 2017-12-18
Bug 905763 - Fix named anchors in various pages so that the Sandstone theme header can be set to a fixed position

kyoshino:
Fix [Bug 905763 - Fix named anchors in various pages so that the Sandstone theme header can be set to a fixed position](https://bugzilla.mozilla.org/show_bug.cgi?id=905763)

## Description

* Apply the fixed header + comment focus workaround (added in #289) to any page
* Enable [smooth scroll](https://developer.mozilla.org/en-US/docs/Web/CSS/scroll-behavior) for a better experience
* Use ECMAScript 6+, notably [arrow functions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions) 🎉

### DIFF
--- a/js/global.js
+++ b/js/global.js
@@ -190,3 +190,24 @@ $().ready(function() {
         $('.bz_autocomplete').attr('autocomplete', 'off');
     });
 });
+
+/**
+ * If the URL contains a hash like #c10, scroll down the page to show the
+ * element below the fixed global header. This workaround is required for
+ * comments on show_bug.cgi, components on describecomponents.cgi, etc.
+ */
+const scroll_element_into_view = () => {
+    if (location.hash) {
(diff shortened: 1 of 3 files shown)


js/global.js
```diff
@@ -190,3 +190,24 @@ $().ready(function() {
         $('.bz_autocomplete').attr('autocomplete', 'off');
     });
 });
+
+/**
+ * If the URL contains a hash like #c10, scroll down the page to show the
+ * element below the fixed global header. This workaround is required for
+ * comments on show_bug.cgi, components on describecomponents.cgi, etc.
+ */
+const scroll_element_into_view = () => {
```
dylanwh: this is so beautiful it makes me want to cry.

PR #307, opened 2017-12-20
Bug 1361890 - Remove asset concatenation

dylanwh:


vagrant_support/apache.j2
```diff
@@ -15,7 +15,7 @@ PerlConfigRequire /vagrant/mod_perl.pl
     <Directory "/vagrant">
         DirectoryIndex index.cgi
         Options Indexes FollowSymLinks ExecCGI
-        AllowOverride All
+        AllowOverride None
```
dylanwh: Without this change, the vagrant env will have double cache control headers

dylanwh:
For CSS, we use relative paths and those should still appear to be relative to the new prefix.
For JS, I think we can read the version from the DOM and do "/static/v" + VERSION + "/"

dylanwh:
Note that paths that do not begin with /static/* will never be cached.

PR #308, opened 2017-12-20
Bug 1426383 - Serve web fonts with proper HTTP headers

dylanwh:
This comes after #307


kyoshino:
I see the required `.htaccess` changes made in #307.

PR #312, opened 2017-12-21
 Bug 1426518 - Revisions can optionally not have a bug id so we need to make it optional in the type constraints of Revision.pm

dklawren:
Also had to make some other refactor to make this work properly so apologize for the extra length.

### DIFF
--- a/extensions/PhabBugz/lib/Util.pm
+++ b/extensions/PhabBugz/lib/Util.pm
@@ -423,11 +423,11 @@ sub get_attachment_revisions {
         }
 
         if (@revision_ids) {
-            @revisions = get_revisions_by_ids( \@revision_ids );
+            $revisions = get_revisions_by_ids( \@revision_ids );
         }
     }
 
-    return @revisions;
+    return @$revisions;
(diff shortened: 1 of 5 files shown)


extensions/PhabBugz/lib/Util.pm
```diff
@@ -423,11 +423,11 @@ sub get_attachment_revisions {
         }
 
         if (@revision_ids) {
-            @revisions = get_revisions_by_ids( \@revision_ids );
+            $revisions = get_revisions_by_ids( \@revision_ids );
         }
     }
 
-    return @revisions;
+    return @$revisions;
```
dylanwh: I haven't run this -- did the return value of get_revisions_by_ids() change in this patch? or was this wrong before?

dklawren: This was wrong before. get_revisions_by_ids returns reference to a list but code expects get_attachment_revisions to return a list and not a ref so I had to fix this. Luckily the inconsistent code will be going away.

PR #319, opened 2017-12-29
Allow request_cache to be constant-folded in Bugzilla.pm

dylanwh:
This saves quite a bit of time all over the place: It makes Bugzilla->user take 600µs instead of 2ms.

### DIFF
--- a/Bugzilla.pm
+++ b/Bugzilla.pm
@@ -80,6 +80,11 @@ use constant SHUTDOWNHTML_EXIT_SILENTLY => qw(
 # should search engines attempt to index the page again?
 use constant SHUTDOWNHTML_RETRY_AFTER => 3600;
 
+# This is identical to Install::Util::_cache so that things loaded
+# into Install::Util::_cache during installation can be read out
+# of request_cache later in installation.
+use constant request_cache => Bugzilla::Install::Util::_cache();
+
 #####################################################################
 # Global Code
 #####################################################################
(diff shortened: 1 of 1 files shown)


Bugzilla.pm
```diff
@@ -221,26 +226,26 @@ sub init_page {
...
     }
-    $_[0]->request_cache->{template}->{_is_main} = 1;
+    request_cache->{template}->{_is_main} = 1;
 
-    return $_[0]->request_cache->{template};
+    return request_cache->{template};
 }
 
 sub template_inner {
     my ($class, $lang) = @_;
```
staabm: $class no longer used? Could it be dropped?

Bugzilla.pm
```diff
@@ -221,26 +226,26 @@ sub init_page {
...
-    my $cache = $class->request_cache;
+    my $cache = request_cache;
     my $current_lang = $cache->{template_current_lang}->[0];
     $lang ||= $current_lang || '';
     return $cache->{"template_inner_$lang"} ||= Bugzilla::Template->create(language => $lang);
 }
 
 sub extensions {
     my ($class) = @_;
```
staabm: $class no longer used? Could it be dropped?

Bugzilla.pm
```diff
@@ -256,12 +261,12 @@ sub extensions {
 }
 
 sub cgi {
-    return $_[0]->request_cache->{cgi} ||= new Bugzilla::CGI();
+    return request_cache->{cgi} ||= new Bugzilla::CGI();
 }
 
 sub input_params {
     my ($class, $params) = @_;
```
staabm: $class no longer used? Could it be dropped?

Bugzilla.pm
```diff
@@ -289,32 +294,32 @@ sub get_param_with_override {
 }
 
 sub user {
-    return $_[0]->request_cache->{user} ||= new Bugzilla::User;
+    return request_cache->{user} ||= new Bugzilla::User;
 }
 
 sub set_user {
     my ($class, $user) = @_;
```
staabm: $class no longer used? Could it be dropped?

Bugzilla.pm
dylanwh: class is still used here.

staabm: Oh .. I am not usedto the notion of return-if, you are right

dylanwh:
Also, @staabm thanks for commenting on this PR! It's nice getting more eyes on the code.

staabm:
I am not used to this language, so maybe my comments are just shit. Found this class thing in even more places, didnt commented all though

dylanwh:
Well, you made me realize I could remove it from several places (which I have). It's not much of a performance difference but still the right thing to do.

PR #321, opened 2017-12-29
Use sereal for database schema

dylanwh:


dylanwh:
This needs code to trigger a schema save during the checksetup process.

PR #322, opened 2017-12-30
Redefine html_quote() in terms of HTML::Escape::escape_html()

dylanwh:
This is a lot faster. On show_bug, this is the difference between 11 ms and 584µs

This means we "lose" the Bi-Di stripping from [bug 319331](https://bugzilla.mozilla.org/show_bug.cgi?id=319331)


dylanwh:
@dklawren do you think this is okay? to lose the BI-Di stripping? I'm unfamiliar with that bugs that caused in the past.

PR #324, opened 2017-12-31
Bug 1426673 - The logout link cannot be found as what Sessions page says

kyoshino:
Fix [Bug 1426673 - The logout link cannot be found as what Sessions page says](https://bugzilla.mozilla.org/show_bug.cgi?id=1426673)

## Description

* Show the fallback image on the Account menu if loading of Gravatar is blocked
* Tested with the Ghostery extension and Firefox's built-in Tracking Protection (strict mode)

### DIFF
--- a/js/global.js
+++ b/js/global.js
@@ -192,6 +192,18 @@ $().ready(function() {
 });
 
 /**
+ * Check if Gravatar images on the page are successfully loaded, and if blocked
+ * (by any content blocker), replace them with the default/fallback image.
+ */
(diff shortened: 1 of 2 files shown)


kyoshino:
I have changed my course: simply detect blocked Gravatars and replace them with the fallback image. This works with not only the Account menu in question but also commenter avatars on `show_bug.cgi`. In the future we could use people's initial for fallback icons as seen in Gmail and BzDeck.

PR #331, opened 2018-01-03
Bug 1427656 - Remove ZeroClipboard helper

kyoshino:
Fix [Bug 1427656 - Remove ZeroClipboard helper](https://bugzilla.mozilla.org/show_bug.cgi?id=1427656)


Bugzilla/CGI.pm
```diff
@@ -62,7 +62,6 @@ sub SHOW_BUG_MODAL_CSP {
     my ($bug_id) = @_;
     my %policy = (
         script_src  => ['self', 'nonce', 'unsafe-inline', 'unsafe-eval', 'https://www.google-analytics.com' ],
-        object_src  => [Bugzilla->localconfig->{urlbase} . "extensions/BugModal/web/ZeroClipboard/ZeroClipboard.swf"],
```
dylanwh: Confirmed that this means we send object-src: none, so perfect.

PR #335, opened 2018-01-04
Bug 1428038 - Background color of security bugs is wrong again

dylanwh:


dylanwh:
Ignore the test failures for selenium, they're from the security bug fix.

PR #338, opened 2018-01-04
Bug 1428085 - Disallow non-BMP characters in some text fields on the show bug page

dylanwh:


extensions/BugModal/web/bug_modal.js
```diff
@@ -569,6 +569,18 @@ $(function() {
         });
     $('#mode-btn').prop('disabled', false);
 
+
+    let re = /[\uD800-\uDFFF]/;
+    let checker = function (event) {
+        if (re.test(this.value)) {
+            this.setCustomValidity("Field contains emoji");
```
kyoshino: When I input some emoji and submit the form, both `<input id="short_desc">` and `<textarea id="comment">` gets a red shadow and the submission fails, but I don't see any error message anywhere on neither Firefox or Chrome. I'm still figuring out why. Also, the message could probably be a bit clearer, like "Emoji is not yet supported in this field"

dylanwh: The error message shows up in a mouse-over, which I think is not enough. I swear I've used setCustomValidity this way before...

PR #341, opened 2018-01-04
Bug 1428166 - Move <meta charset> to start of <head>

garas:
The `<meta>` element declaring the encoding must be inside the `<head>` element and **within the first 1024 bytes** of the HTML as some browsers only look at those bytes before choosing an encoding.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta#attr-charset

### DIFF
--- a/template/en/default/global/header.html.tmpl
+++ b/template/en/default/global/header.html.tmpl
@@ -94,6 +94,10 @@
 <!DOCTYPE html>
 <html lang="en">
   <head>
+    [% IF Param('utf8') %]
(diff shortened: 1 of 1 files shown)


garas:
There are no "Take" button at "Assignee" field. Maybe I'm missing some permission.

I've added myself to CC.

dylanwh:
Ah, that's interesting. We're not displaying it properly for your type of account. 
I've set you as the assignee and I'll merge this as soon as tests finish.

It'll be live next week. Thanks for the patch!

PR #342, opened 2018-01-04
Bug 1428156 - BMO will mark a revision as public if it sees a new one created without a bug id associated with it

dklawren:


extensions/PhabBugz/lib/Feed.pm
```diff
@@ -107,7 +107,17 @@ sub feed_query {
...
+            if ($story_text =~ /\s+created\s+D\d+/) {
+                # If new revision and bug id was omitted, make revision public
+                $self->logger->debug("No bug associated with new revision. Marking public.");
+                $revision->set_policy('view', 'public');
+                $revision->set_policy('edit', 'users');
+                $revision->update();
+                $self->logger->info("SUCCESS");
+            }
+            else {
+                $self->logger->debug("SKIPPING: No bug associated with revision change");
```
dylanwh: While you probably don't log the whole story text, I can't help but feel if this breaks, we'd appreciate more details about what went wrong.

extensions/PhabBugz/lib/Feed.pm
```diff
@@ -84,7 +84,7 @@ sub feed_query {
         $self->logger->debug("STORY PHID: $story_phid");
         $self->logger->debug("AUTHOR PHID: $author_phid");
         $self->logger->debug("OBJECT PHID: $object_phid");
-        $self->logger->debug("STORY TEXT: $story_text");
+        $self->logger->info("STORY TEXT: $story_text");
```
dylanwh: And I assume this is single-line or else the logger handles multiple lines. LGTM.

dklawren: The story text is always a single line.

dklawren:
I have found some issue in our current logic that need to be addressed. 

1. When a bug id is changed on a revision from a private to a public one, we current do not change anything about the policies so the revision stays private. 
2. We currently do not know that a bug id has changed, only that there is one. Need to figure out how to tell in BMO if a revisions bug id is a change from a previous one.

I will do them as a separate bug.

dklawren:
@dylanwh As for this change it is not impacted. This addresses a very specific condition in that a revision is brand new and does not have a bug id. So we stop early, make it public and loop to the next story. The other issues I found only have to do with when a revisions bug id is changed from something valid to something else (or cleared) so this can be handled separately.

PR #343, opened 2018-01-04
Bug 1428194 - Add CI test that tests securemail

dylanwh:


dylanwh:
One more problem still elludes me, but pushed this up for safety's sake.

dylanwh:
Okay, so with admin@mozilla.test, we see:

2_bugs_found

But with admin@mozilla.bugs (the existing value)
2_bugs_found_good

wtf is going on

PR #347, opened 2018-01-05
Bug 1427800 - Wrong anchor scrolling with old UI

kyoshino:
Fix [Bug 1427800 - Wrong anchor scrolling with old UI](https://bugzilla.mozilla.org/show_bug.cgi?id=1427800)

## Description

* This function `scroll_element_into_view` is actually no longer unnecessary since #329 solved the overlapping issue, so I'll keep it to adjust the scroll position.
* For forward compatibility, I've added the support for [`<div role="feed">`](https://www.w3.org/TR/wai-aria-1.1/#feed) which could be the new scroll container of the comments.

### DIFF
--- a/extensions/BugModal/web/bug_modal.js
+++ b/extensions/BugModal/web/bug_modal.js
@@ -830,13 +830,15 @@ $(function() {
             // remove embedded links to attachment details
             reply_text = reply_text.replace(/(attachment\s+\d+)(\s+\[[^\[\n]+\])+/gi, '$1');
 
-            if ($('#comment').val() != reply_text) {
-                $('#comment').val($('#comment').val() + reply_text);
-            }
-            if (BUGZILLA.user.settings.autosize_comments) {
-                autosize.update($('#comment'));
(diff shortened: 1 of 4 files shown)


kyoshino:
I'm extending this function so it can be used to fix [Bug 1428377](https://bugzilla.mozilla.org/show_bug.cgi?id=1428377) as well.

kyoshino:
Added the fixes for Bug 1428377 and other minor issues.

kyoshino:
I'm still seeing a bounce when replying on a long page. Will figure out why.

kyoshino:
While merging #350 I just noticed that **Rebase and merge** is disabled on this repo. 🤔

kyoshino:
Looks like the bouncing issue is due to a bug in the autosize jQuery plug-in. I haven't been able to solve it yet but I think it's okay; it won't happen often. In the near future the comment box will be fixed at the bottom of the page for convenience so the issue will be gone anyway.

PR #348, opened 2018-01-06
Bug 1374113 - Add a link to "Bugs filed in last 30 days" or something similar in "User Statistics" section of "User Profile" page

gr8h:
Add a link to "Bugs filed in last 30 days" or something similar in "User Statistics" section of "User Profile" page.

### DIFF
--- a/extensions/UserProfile/template/en/default/pages/user_profile.html.tmpl
+++ b/extensions/UserProfile/template/en/default/pages/user_profile.html.tmpl
@@ -204,6 +204,20 @@
     [% "</a>" IF user.id %]
   </td>
 </tr>
+
+<tr>
+  <td>&nbsp;</td>
+  <th>[% terms.Bugs %] filed last 30 days</th>
+  <td class="numeric">
+    [% IF user.id %]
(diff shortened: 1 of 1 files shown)


dylanwh:
Thanks for this patch, @gr8h! I'll be taking a look at it now.

extensions/UserProfile/template/en/default/pages/user_profile.html.tmpl
```diff
@@ -204,6 +204,20 @@
...
+
+<tr>
+  <td>&nbsp;</td>
+  <th>[% terms.Bugs %] filed last 30 days</th>
+  <td class="numeric">
+    [% IF user.id %]
+      <a href="buglist.cgi?query_format=advanced&amp;emailtype1=exact&amp;emailreporter1=1&amp;email1=[% target.login FILTER uri %]&amp;chfield=[Bug creation]&amp;chfieldfrom=-30d"
+        target="_blank">
+    [% END %]
+    [% stats.bugs_filed || 0 FILTER html %]
```
dylanwh: The problem is that this number is the same as the total number of bugs filed. 

Take a look in /Users/dylan/src/mozilla/bmo/extensions/UserProfile/lib/Util.pm, around line 38.
You should be able to add a new stats entry, bugs_filed_30_days.

dylanwh: If you need some sql pointers, someone will probably be available in #bteam or #bugzilla on irc.mozilla.org.

PR #351, opened 2018-01-07
Bug 1428641 - Implement Requests quick look dropdown on global header

kyoshino:
Fix [Bug 1428641 - Implement Requests quick look dropdown on global header](https://bugzilla.mozilla.org/show_bug.cgi?id=1428641)

## Description

* Add a dropdown menu to the Requests badge that allows to check recent requests at a glance
* [Screenshot](https://twitter.com/BugzillaUX/status/949948104693837824)


template/en/default/request/queue.json.tmpl
```diff
@@ -0,0 +1,42 @@
...
+  # This Source Code Form is "Incompatible With Secondary Licenses", as
+  # defined by the Mozilla Public License, v. 2.0. #%]
+
+[% RAWPERL %]
+my @display_columns = (
+    "requester", "requestee",      "type",    "status",  "bug_id", "bug_summary",
+    "attach_id", "attach_summary", "ispatch", "created", "category"
+);
+my $requests    = $stash->get('requests');
+my $time_filter = $context->filter('time', [ '%Y-%m-%dT%H:%M:%SZ', 'UTC' ]);
```
dylanwh: Why UTC and not the user's timezone?

kyoshino: I just copy-pasted this from somewhere else but it actually doesn't matter because the frontend JavaScript code parses the string with `Date()` that can retrieve the local time easily. I think, for API response payloads, it's safe to always use UTC.

kyoshino:
It's from https://github.com/blueimp/JavaScript-MD5 (not the latest version, just copied from the BzDeck repo) and the license is MIT.

kyoshino:
I want to see the MD5 support in the [Web Crypto API](https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/digest) but they exclude it purposely because it's not safe.

PR #359, opened 2018-01-09
Bug 1429220 - Issues with the new fixed header implementation

kyoshino:
Fix [Bug 1429220 - Issues with the new fixed header implementation](https://bugzilla.mozilla.org/show_bug.cgi?id=1429220)

## Description

Scroll containers are `#header` and `#bugzilla-body` so `<body>` should not scroll. Actually `#header` should not scroll either. This issue will be solved once the site gets responsive design.

### DIFF
--- a/skins/standard/global.css
+++ b/skins/standard/global.css
@@ -37,6 +37,7 @@
 /* global (begin) */
     body {
         position: absolute;
+        overflow: hidden;
         margin: 0;
         width: 100%;
         height: 100%;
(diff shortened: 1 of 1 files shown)


dylanwh:
Is there a #bugzilla-body on the search page?
/me checks

kyoshino:
Yes, `<main id="bugzilla-body">` and `<header id="header">` exist on all the pages.

PR #365, opened 2018-01-10
Make bug modal usable at less than 900px

dylanwh:
Re-did this PR to focus on the sub-900 widths on bug modal

### DIFF
--- a/extensions/BugModal/web/bug_modal.css
+++ b/extensions/BugModal/web/bug_modal.css
@@ -989,3 +989,27 @@ div.ui-tooltip {
 #user-guide {
     padding-top: 5px;
 }
+
+
+/* even narrower displays */
+@media screen and (max-width: 900px) {
+    #wrapper {
+        position: initial !important;
+    }
+
+    #header .inner {
+        width: auto !important;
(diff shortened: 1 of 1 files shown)


dylanwh:
@purelogiq what about https://bugzilla-dev.allizom.org/show_bug.cgi?id=900 at a 800px width?

purelogiq:
@dylanwh At exactly 800px things look "fine" but the tables in the people and status column are stuck. This is noticble when you reduce the width even further.

At 800px
800e

At 720px


At less than 600, clipping is based on the length of the text:
800l

dylanwh:
I think you have cached CSS, can you do a shift-reload?

dylanwh:
At 800 it looks like this for me, which I consider "good enough":

screen shot 2018-01-10 at 14 33 24

purelogiq:
@dylanwh I tested in a private windows (which I assume doesn't cache). Your 800px and my 800px look about the same, no?

dylanwh:
Ah, yes, you're right. Anyway, an r+ is an r+. Thanks!

PR #368, opened 2018-01-12
Bug 1429621 - Add Saved Searches to My Dashboard

kyoshino:
Fix [Bug 1429621 - Add Saved Searches to My Dashboard](https://bugzilla.mozilla.org/show_bug.cgi?id=1429621)

## Description

Add the Saved Searches section to the bottom of the My Dashboard page so people can have a plain list of their searches as seen on the removed global footer.

## Screenshot

screen shot 2018-01-11 at 21 31 48

### DIFF
--- a/extensions/MyDashboard/web/styles/mydashboard.css
+++ b/extensions/MyDashboard/web/styles/mydashboard.css
@@ -20,9 +20,8 @@
 
 .query_heading {
     font-size: 18px;
-    font-weight: strong;
-    padding-bottom: 5px;
-    padding-top: 5px;
+    font-weight: 600;
+    margin: 5px 0;
     color: rgb(72, 72, 72);
 }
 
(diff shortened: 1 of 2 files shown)


extensions/MyDashboard/web/styles/mydashboard.css
```diff
@@ -71,3 +70,17 @@
...
+    align-items: center;
+    margin: 5px 0;
+}
+
+#saved_searches_container > header h2 {
+    margin: 0 10px 0 0;
+}
+
+#saved_searches_container li {
+    margin: 5px 0;
```
mnoorenberghe: I have over 30 saved searches so a vertical layout isn't great for displaying them. Can we use a horizontal layout more like the old footer so I don't have a large blank area causing more scrolling on the page

dylanwh: Maybe a responsive grid layout?

mnoorenberghe: That sounds like overkill for a temporary thing. The following would be good enough IMO:
```css
#saved_searches_container li {
    margin: 5px;
    display: inline;
}
```
image

dylanwh: Done in #451. Needs a bug filed.

PR #370, opened 2018-01-12
Bug 1429785 - In-page back navigation broken

kyoshino:
Fix [Bug 1429785 - In-page back navigation broken](https://bugzilla.mozilla.org/show_bug.cgi?id=1429785)

## Description

This mimics the browser's scroll position auto saving behaviour, so in-page navigation with the back and forward buttons will work as expected.

### DIFF
--- a/js/global.js
+++ b/js/global.js
@@ -192,13 +192,47 @@ $().ready(function() {
 });
 
 /**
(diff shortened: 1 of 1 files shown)


kyoshino:
@purelogiq Exactly 😺 Thanks for your review!

PR #371, opened 2018-01-12
Bug 1430259 - Update policy code in BMO PhabBugz extension to update custom policy if a private bugs groups have changed.

dklawren:


extensions/PhabBugz/lib/Policy.pm
```diff
@@ -0,0 +1,174 @@
...
+use strict;
+use warnings;
+
+use Bugzilla::Error;
+use Bugzilla::Util qw(trim);
+use Bugzilla::Extension::PhabBugz::Util qw(request);
+use Bugzilla::Extension::PhabBugz::Project;
+
+use List::Util qw(first);
+
```
dylanwh: 👍  new code using List::Util instead of ::MoreUtils is very good to see.

PR #372, opened 2018-01-14
Bug 1429344 - Review requests in requests dropdown should link to MozReview or GitHub instead of Bugzilla details page

kyoshino:
Fix [Bug 1429344 - Review requests in requests dropdown should link to MozReview or GitHub instead of Bugzilla details page](https://bugzilla.mozilla.org/show_bug.cgi?id=1429344)

## Description

* Retrieve `attachments.mimetype` from DB as `attach_mimetype`
* Retrieve `attachments.ispatch` from DB as `attach_ispatch` instead of `ispatch` for consistency (this field is not used in other templates so the change should be safe)
* Treat MozReview and Phabricator requests as patches
* Support GitHub pull requests and Google Docs as well
* Link directly to those external sites
* For regular patches, link to the Splinter Review page if available

### DIFF
--- a/request.cgi
+++ b/request.cgi
@@ -102,6 +102,7 @@ sub queue {
                 requesters.realname, requesters.login_name,
                 requestees.realname, requestees.login_name, COUNT(privs.group_id),
     " . $dbh->sql_date_format('flags.modification_date', '%Y.%m.%d %H:%i') . ",
+                attachments.mimetype,
                 attachments.ispatch,
                 bugs.bug_status,
                 bugs.priority,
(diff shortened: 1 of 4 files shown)


kyoshino:
Ah, one thing I've noticed is that the splinter review URL `page.cgi?id=splinter.html` is customizable by admin. To avoid any potential breakage in the future or upstream, it has to be exposed as a string like the extension list also added in this PR.

kyoshino:
Instead of an extension list, exposed the `splinter_base` param in the template so review links always work.

request.cgi
```diff
@@ -248,7 +249,7 @@ sub queue {
                 requesters.login_name, requestees.realname,
                 requestees.login_name, flags.modification_date, attachments.ispatch
                 cclist_accessible, bugs.reporter, bugs.reporter_accessible,
-                bugs.assigned_to, attachments.ispatch');
+                bugs.assigned_to, attachments.mimetype');
```
dylanwh: doesn't this need to also group by ispatch?

kyoshino: There's already `ispatch` 2 lines above. Looks like it's a duplicate, so just changing it to `mimetype`.

PR #373, opened 2018-01-14
Bug 1430495 - Make loading of Requests dropdown faster

kyoshino:
Fix [Bug 1430495 - Make loading of Requests dropdown faster](https://bugzilla.mozilla.org/show_bug.cgi?id=1430495)

## Description

* Use the same technique as Firefox's tab warming to load the Requests dropdown content faster.


dylanwh:
Note: please mark the bug as resolved. bonus points for setting an r? but if not I'll go and add an r+ to this patch before the push.

PR #374, opened 2018-01-15
Bug 1431294 - requests.cgi - allocate less memory

dylanwh:
`@data = @$result` makes a copy of an array. Instead of doing that, we just use the array ref (and rename `$result` to `$data`)


globau:
please file a bug for this change.

dylanwh:
@globau this has a bug now btw.

PR #376, opened 2018-01-15
Bug 1428270 - Unwrappable content in summary can cause top buttons to bleed out of main content box

kyoshino:
Fix [Bug 1428270 - Unwrappable content in summary can cause top buttons to bleed out of main content box](https://bugzilla.mozilla.org/show_bug.cgi?id=1428270)

## Description

* Add a simple template filter for inserting [`<wbr>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/wbr) elements to camel and/or snake case words, like what I'm doing on [fxsitecompat.com](https://github.com/fxsitecompat/www.fxsitecompat.com/blob/master/layouts/docs/single.html#L32)
* Use the filter for the bug summary on the bug details and search results
* It can also be used for other pages but those are good starting points!

## Screenshots

screen shot 2018-01-14 at 23 27 52
screen shot 2018-01-14 at 23 27 15

### DIFF
--- a/Bugzilla/Template.pm
+++ b/Bugzilla/Template.pm
@@ -683,6 +683,18 @@ sub create {
                 return $var;
             },
 
+            # Insert `<wbr>` HTML tags to camel and snake case words in the
+            # given string so a long bug summary, for example, will be wrapped
+            # in a preferred manner rather than overflowing or expanding the
+            # parent element. Examples:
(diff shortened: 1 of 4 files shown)


kyoshino:
Ugh, true, the double slash in the regex is wrong. Why I use it in my fxsitecompat code... Fixing.

PR #377, opened 2018-01-16
Bug 1430367 - "preconnect" to google-analytics domain for improved pe…

staabm:
…rformance

implements https://bugzilla.mozilla.org/show_bug.cgi?id=1430367

be warned: I dont know perl at all and I wasnt able to test this locally.
This change might e.g. contain syntax errors I am not aware of..


staabm:
per https://developer.mozilla.org/en-US/docs/Web/HTML/Link_types
the actual benefit is...

> Provides a hint to the browser suggesting that it open a connection to the linked web site in advance, without disclosing any private information or downloading any content, so that when the link is followed the linked content can be fetched more quickly.

staabm:
like that?

PR #381, opened 2018-01-18
Bug 1430905 - Remove legacy phabbugz code that is no longer needed

dklawren:


darkwing:
I think this might need a bit of cleanup or a rebase?

dklawren:
Sorry forgot to mention that you need to first apply https://github.com/mozilla-bteam/bmo/pull/371 before this one as this one is dependent on that. If you want to review 371 first that would be better I think.

dkl

PR #383, opened 2018-01-22
no bug - fix the entrypoint arbitrary command execution

dylanwh:


dylanwh:
@Micheletto so you don't have to wait on CI, mozillabteam/bmo:latest has been updated (docker hash 5f4c4946914c)

PR #384, opened 2018-01-25
I found this a bit hard to review, so I've tried to tidy a bit

dylanwh:
I had wanted to do this with the other code but it was too big. Policy
is a nice small use case where we can see about making it tidy.

If these changes *work* then I've understood part of this code
correctly.

### DIFF
--- a/extensions/PhabBugz/lib/Policy.pm
+++ b/extensions/PhabBugz/lib/Policy.pm
@@ -8,11 +8,9 @@
 package Bugzilla::Extension::PhabBugz::Policy;
 
 use 5.10.1;
-use strict;
-use warnings;
+use Moo;
 
 use Bugzilla::Error;
-use Bugzilla::Util qw(trim);
 use Bugzilla::Extension::PhabBugz::Util qw(request);
 use Bugzilla::Extension::PhabBugz::Project;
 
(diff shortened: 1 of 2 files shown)


extensions/PhabBugz/lib/Policy.pm
```diff
@@ -21,24 +19,30 @@ use List::Util qw(first);
...
             value  => Maybe[ArrayRef[Str]]
         ]
     ]
-];
+);
+
+has 'rule_projects' => (
+    is => 'lazy',
+    isa => ArrayRef[Str],
+);
```
dklawren: For reasons I am not quite sure of right now, I could not get Revision.pm to work this way either. Something is different in how Moo pre-inits the variables and then when I get the data from Phabricator API, it cannot just simply copy the data struct into $self the way I do it in the current way. I would need to have to through each key/value pair in the API data and manually set the value for its corresponding Moo variable. Is there a way to do this more simply when loading the object with data? Using this code I get the following error right away:

bmo.feed_1          | [Fri Jan 26 20:07:02 2018] ERROR: Can't call method "rule_projects" on unblessed reference at /app/extensions/PhabBugz/lib/Feed.pm line 196.
bmo.feed_1          |  at /app/extensions/PhabBugz/lib/Feed.pm line 196.
bmo.feed_1          |   Bugzilla::Extension::PhabBugz::Feed::process_revision_change(Bugzilla::Extension::PhabBugz::Feed=HASH(0x5dd1228), "PHID-DREV-o674lixaeyxbfcr3e6oc", "conduit created D8: test.") called at /app/extensions/PhabBugz/lib/Feed.pm line 110
bmo.feed_1          |   Bugzilla::Extension::PhabBugz::Feed::feed_query(Bugzilla::Extension::PhabBugz::Feed=HASH(0x5dd1228)) called at /app/extensions/PhabBugz/lib/Feed.pm line 44
bmo.feed_1          |   eval {...} called at /app/extensions/PhabBugz/lib/Feed.pm line 42
bmo.feed_1          |   Bugzilla::Extension::PhabBugz::Feed::start(Bugzilla::Extension::PhabBugz::Feed=HASH(0x5dd1228)) called at /app/extensions/PhabBugz/lib/Daemon.pm line 97
bmo.feed_1          |   Bugzilla::Extension::PhabBugz::Daemon::gd_run(Bugzilla::Extension::PhabBugz::Daemon=HASH(0x1c72068)) called at local/lib/perl5/Daemon/Generic.pm line 192
bmo.feed_1          |   Daemon::Generic::new("Bugzilla::Extension::PhabBugz::Daemon") called at local/lib/perl5/Daemon/Generic.pm line 37
bmo.feed_1          |   Daemon::Generic::newdaemon() called at /app/extensions/PhabBugz/lib/Daemon.pm line 25
bmo.feed_1          |   Bugzilla::Extension::PhabBugz::Daemon::start("Bugzilla::Extension::PhabBugz::Daemon") called at extensions/PhabBugz/bin/phabbugz_feed.pl line 22

dklawren: Hmm. Nevermind my last comment. I see that you are doing new_from_query() which calls $class->new($result). This was not so easy in Revision.pm at the time as the API call has the data values nested under fields and not in the top level of the hash. So it became more code to do it with Moo that just using old-school accessors/setters.

In this case $result has the proper data from the API call, but  $class->new($result) is not returning a proper object to the caller. If I dump $current_policy in Feed.pm it is not blessed.

extensions/PhabBugz/lib/Policy.pm
```diff
@@ -75,32 +79,17 @@ my $SearchResult = Dict[
...
-
-sub _load {
-    my ($params) = @_;
     my $result = request('policy.query', $params);
     if (exists $result->{result}{data} && @{ $result->{result}{data} }) {
         return $result->{result}->{data}->[0];
     }
-    return $result;
+    return $class->new($result);
 }
```
dklawren: Ok this works fine after the below change. We were returning too early in new_from_query().

sub new_from_query {
    my ($class, $params) = @_;
    my $result = request('policy.query', $params);
    if (exists $result->{result}{data} && @{ $result->{result}{data} }) {
        return $class->new($result->{result}->{data}->[0]);
    }
    return undef;
}

PR #387, opened 2018-01-30
Bug 1434064 - Refactor Project.pm to use Moo for better type checking

dklawren:


dylanwh:
Oh convenient -- I guess this code was already in the other one. So that's good then, right?

PR #389, opened 2018-02-02
Changes version.json build setting to match dockerflow specifications.

Micheletto:
See: https://github.com/mozilla-services/Dockerflow/blob/master/circle.yml#L18-L24


Micheletto:
> To confirm, this is just changing the field value to be a URL string instead of a number?

That is correct.  The image mirroring function we use expects that value to be a URL.
The Circle docs for environment variables are here: https://circleci.com/docs/1.0/environment-variables/

Micheletto:
> This won't work.

Addressed in [56360fc](https://github.com/mozilla-bteam/bmo/pull/389/commits/56360fcbad43d6ec2137c19f5cc5116a0e66ac56)

PR #392, opened 2018-02-12
Bug 1433573 - Display the short URL link even for queries without any results

globau:
fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1433573


dylanwh:
@globau what bug was this?

PR #396, opened 2018-02-13
Bug 1437238 - Create override parameters for mailer configuration.

dylanwh:


dylanwh:
Since breaking this in prod would be quite scary, I've added @globau as a reviewer. I've made this change in a bit of a hurry as well, so more eyes == better.

Bugzilla/Install/Localconfig.pm
```diff
@@ -136,8 +136,8 @@ use constant LOCALCONFIG_VARS => (
         name    => 'param_override',
         default => {
             inbound_proxies     => undef,
-            memcached_servers   => undef,
-            memcached_namespace => undef,
+            use_mailer_queue    => undef,
+            mail_delivery_method => undef,
```
globau: nit: fix => alignment

globau:
`template/en/default/admin/admin.html.tmpl` will still show the queue functions if jobqueue is enabled in params but disabled via the override.  i'm not sure if that's something that was accidentally left out of the updates or if you're ok with that likely-to-never-happen problem (i'm ok with ignoring this issue)

PR #401, opened 2018-02-18
Bug 1439266 - Add PSGI support to BMO

dylanwh:


dylanwh:
@upasana-me before diving into error handling, could you help me debug this perhaps?

On this branch, I can a running plack-based httpd with these steps:

```
perl checksetup.pl
edit localconfig # to set db_* params to a local mysql
perl app.psgi
```

and a lot of it seems to work, but after a few requests it dies. With no error message, nothing!

And then when I try to go to http://localhost:5000/page.cgi?id=sitemap/sitemap.xml

nothing happens.

I've tried using the Plack::Middleware::Debug, I've tried setting a `$SIG{__DIE__}`, and somehow it is just exiting without telling me why. :(

dylanwh:
I fixed this and some more problems!

So the way this was being done was wrong -- Bugzilla->init_page() expects to be called inside a CGI environment.

Also it is important to load CGI::Compile first because it overrides exit().

PR #403, opened 2018-02-19
Add missing ntp package

globau:
creating a new vagrant dev environment failed due to the ntp package not being installed.


dylanwh:
Interesting, ntp is included in the centos 6 images for virtualbox and vmware_fusion

globau:
evidently parallels didn't think to include it 😞

PR #405, opened 2018-02-19
Bug 1437383: Create User.pm PhabBugz class for loading of a user object from phabricator

dklawren:
This is the User.pm module only for now. I have patches that add support for using the new module for loading of phabricator users instead of the utility methods in Util.pm. I felt review would be easier if I made smaller PRs for each of those instead of one large PR. Please review and once landed, I will create PRs for other parts of the code to utilize the new class module.

dkl

### DIFF
--- a/extensions/PhabBugz/lib/User.pm
+++ b/extensions/PhabBugz/lib/User.pm
@@ -108,8 +108,8 @@ sub BUILDARGS {
 
 sub new_from_query {
     my ( $class, $params ) = @_;
-    my ($user) = $class->match($params);
-    return $user;
+    my $matches = $class->match($params);
+    return $matches->[0];
 }
 
 sub match {
--- a/extensions/PhabBugz/lib/User.pm
+++ b/extensions/PhabBugz/lib/User.pm
@@ -14,6 +14,7 @@ use Bugzilla::User;
 
 use Bugzilla::Extension::PhabBugz::Util qw(request);
 
+use List::Util qw(first);
 use Types::Standard -all;
 use Type::Utils;
 
@@ -32,6 +33,7 @@ has 'roles'           => ( is => 'ro', isa => ArrayRef [Str] );
 has 'view_policy'     => ( is => 'ro', isa => Str );
 has 'edit_policy'     => ( is => 'ro', isa => Str );
 has 'bugzilla_id'     => ( is => 'ro', isa => Maybe [Int] );
+has 'bugzilla_user'   => ( is => 'lazy' );
 
 sub BUILDARGS {
     my ( $class, $params ) = @_;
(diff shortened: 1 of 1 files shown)


extensions/PhabBugz/lib/User.pm
```diff
@@ -0,0 +1,203 @@
...
+    $params->{roles}           = $params->{fields}->{roles};
+    $params->{view_policy}     = $params->{fields}->{policy}->{view};
+    $params->{edit_policy}     = $params->{fields}->{policy}->{edit};
+
+    delete $params->{fields};
+
+    if ( my $external_accounts =
+        $params->{attachments}{'external-accounts'}{'external-accounts'} )
+    {
+        foreach my $account (@$external_accounts) {
```
dylanwh:
```perl
my $external_accounts = $params->{attachments}{'external-accounts'}{'external-accounts'} 
my $bug_user = first { $_->{type} ne 'bmo' } @$external_accounts;
$params->{bug_user_id} = $bug_user->{id};
```

extensions/PhabBugz/lib/User.pm
```diff
@@ -0,0 +1,203 @@
...
+
+use 5.10.1;
+use Moo;
+
+use Bugzilla::User;
+
+use Bugzilla::Extension::PhabBugz::Util qw(request);
+
+use Types::Standard -all;
+use Type::Utils;
```
dylanwh:
```perl
use List::Util qw(first);
```

extensions/PhabBugz/lib/User.pm
```diff
@@ -0,0 +1,203 @@
...
+        }
+    }
+
+    return $phab_users;
+}
+
+#################
+#   Accessors   #
+#################
+
```
dylanwh:
```perl
has 'bugzilla_user' => ( is => 'lazy' );
sub _build_bugzilla_user {
    my ($self) = @_;
   return undef if !$self->bugzilla_id;
   return Bugzilla::User->new( { id => $self->bugzilla_id, cache => 1 } );
}
```

extensions/PhabBugz/lib/User.pm
```diff
@@ -0,0 +1,203 @@
...
+}
+
+sub get_phab_bugzilla_ids {
+    my ( $class, $params ) = @_;
+
+    my $memcache = Bugzilla->memcached;
+
+    # Try to find the values in memcache first
+    my @results;
+    my @bugzilla_ids = @{ $params->{ids} };
```
dylanwh: How big do we expect `@bugzilla_ids` to get? splice() is very slow for large arrays.

dklawren:
Suggested changes made. Please review.

zalun:
Seems finr for me.

PR #407, opened 2018-02-21
Bug 1439797 - Enable reporting-only CSP by default

dylanwh:


psiinon:
hey, I think you could enable strict CSP for these end points, there are only a few errors

dylanwh:
@emceeaich Pretty much each script needs a nonce when CSP is turned on.
I'm certain I caught all of them, but even if I didn't the CSP policy is report-only, so we'll see it in the error console.

After we move to the cloud, we'll turn on the report-uri and receive reports about all the places we need to fix, including those without a script_nonce.

PR #408, opened 2018-02-21
Bug 1441475 - Added rel tags to links with target=_blank

psiinon:


psiinon:
Raised bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1441475
r? @dylanwh

PR #411, opened 2018-02-23
Bug 1438206 - Process SES email bounces properly

globau:
Adds a /ses/ endpoint, protected by http basic auth, for handling ses
notifications.

SES must be configured with SNS topics for bounce and complaint that
both https post to the /ses/ endpoint.

Credentials for this endpoint must be provided in localconfig.

### DIFF
--- a/ses/index.cgi
+++ b/ses/index.cgi
@@ -1,0 +1,192 @@
+#!/usr/bin/perl
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+#
+# This Source Code Form is "Incompatible With Secondary Licenses", as
+# defined by the Mozilla Public License, v. 2.0.
+
+use 5.10.1;
+use strict;
+use warnings;
+
+use lib qw(.. ../lib ../local/lib/perl5);
+
+use Bugzilla ();
+use Bugzilla::Mailer qw( MessageToMTA );
+use Bugzilla::User ();
+use Bugzilla::Util qw( remote_ip );
+use JSON::XS qw( decode_json encode_json );
+use LWP::UserAgent ();
+use Try::Tiny qw( try catch );
+
+main();
+
+sub main {
+    my $message = decode_json_wrapper(Bugzilla->cgi->param('POSTDATA')) // return;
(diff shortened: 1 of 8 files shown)


ses/index.cgi
```diff
@@ -0,0 +1,192 @@
...
+use strict;
+use warnings;
+
+use lib qw(.. ../lib ../local/lib/perl5);
+
+use Bugzilla ();
+use Bugzilla::Mailer qw( MessageToMTA );
+use Bugzilla::User ();
+use Bugzilla::Util qw( remote_ip );
+use JSON::XS qw( decode_json encode_json );
```
dylanwh: s/JSON::XS/JSON::MaybeXS/

ses/index.cgi
```diff
@@ -0,0 +1,192 @@
...
+use LWP::UserAgent ();
+use Try::Tiny qw( try catch );
+
```
dylanwh: Since this is in sub directory, the error pages will never work properly. 
This isn't strictly needed, but it would be the "right" thing to do and probably good to do:

```perl
Bugzilla->error_mode(ERROR_MODE_DIE);
try {
    main();
} catch {
    warn "Error in SES handler: $_";
    respond(500 => 'Internal Server Error');
};
```

template/en/default/email/ses-complaint.txt.tmpl
```diff
@@ -0,0 +1,31 @@
...
+
+From: [% Param('mailfrom') %]
+To: [% Param('maintainer') %]
+Subject: [% terms.Bugzilla %]: SES Complaint: [% email %]: [% reason %]
+X-Bugzilla-Type: admin
+
+SES Complaint received for [% email %]: [% reason %]
+
+[% IF user %]
+[% Param('urlbase') %]/editusers.cgi?action=edit&userid=[% user.id %]
```
dylanwh: There is no urlbsse param now. Use [% Bugzilla.localconfig.urlbase %]

ses/index.cgi
globau: i'm not invoking it as an object so i don't need the syntax sugar that `JSON::MaybeXS` provides.

dylanwh: I suppose it'll get caught up when i do a global find and replace.

template/en/default/email/ses-complaint.txt.tmpl
globau: oops.

will fix, but i'll use raw `urlbase` as that's what most templates are using.

globau:
@dylanwh ready for re-review

PR #412, opened 2018-02-23
Add cereal for reliable sequencing of log statements

dylanwh:
In #402 I've added log4perl, which lets me log to a socket.

This PR just adds cereal, a server to receive log messages on a socket, and some better process management to run it and httpd at the same time.


dylanwh:
Ah, missed one thing.

Note: This alone is *not* enough to guarantee that apache's logs would be non-corrupted, even if they could be made in json form.

The problem is that the apache ErrorLog runs only a single instance of netcat.
This doesn't seem to be how ErrorLog "|..." works in scl3 production, an is weird. But I'm not interested in fixing that. I'd rather just use gollum and log4perl.

dylanwh:
I'm going to close this PR and redo this as a series of easier to review chunks.

PR #413, opened 2018-02-26
Bug 1441244 - prevent compounding error messages in tests

dylanwh:
These changes make it so errors in sanity tests don't fail without a database present.

To test this, simply try running 001compile.t without having run Makefile.PL first. You should get a reasonable error (about MYMETA.json not existing, I think).

@upasana-me can you test this? I think without the patch you get the error you reported.

### DIFF
--- a/Bugzilla/Error.pm
+++ b/Bugzilla/Error.pm
@@ -38,15 +38,14 @@ sub _in_eval {
 
 sub _throw_error {
     my ($name, $error, $vars) = @_;
-    my $dbh = Bugzilla->dbh;
     $vars ||= {};
-
     $vars->{error} = $error;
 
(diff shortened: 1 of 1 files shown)


mohawk2:
It looks to me as though the stated purpose is achieved, and the rollback is only attempted if `$dbh` is valid. @dylanwh Can you confirm the function[s] shouldn't immediately return if that's not true?

dylanwh:
@mohawk2: I can confirm that. Honestly those db rollbacks are *ancient* -- they come from a time before mysql had transactions as such, and these Error* functions used to perform table unlocking.

mohawk2:
In that case it's fine by me!

PR #415, opened 2018-02-26
Bug 1437646 - Part 3 - Refactor entrypoint to use Bugzilla::DaemonControl

dylanwh:


conf/httpd.conf
```diff
@@ -50,7 +50,18 @@ TypesConfig /etc/mime.types
 DefaultType text/plain
 MIMEMagicFile conf/magic
 HostnameLookups Off
-ErrorLog /dev/stderr
+<IfDefine NETCAT_LOGS>
+    ErrorLog "|nc localhost 5880"
+    <IfDefine ACCESS_LOGS>
+        TransferLog "|nc localhost 5880"
```
globau: i'd feel better if this was `/usr/bin/nc`

dylanwh: Done.

conf/httpd.conf
```diff
@@ -50,7 +50,18 @@ TypesConfig /etc/mime.types
 DefaultType text/plain
 MIMEMagicFile conf/magic
 HostnameLookups Off
-ErrorLog /dev/stderr
+<IfDefine NETCAT_LOGS>
+    ErrorLog "|/usr/bin/nc localhost 5880"
+    <IfDefine ACCESS_LOGS>
+        TransferLog "|/usr/bin/nc localhost 5880"
```
globau: these should probably be `${LOGGING_PORT}` instead of `5800`.  (i think this will work; untested).

dylanwh:
Assuming tests pass, @dklawren or @Micheletto can you re-approve verifying I added the LOGGING_PORT to the /usr/bin/nc lines as requested?

Micheletto:
> Assuming tests pass, @dklawren or @Micheletto can you re-approve verifying I added the LOGGING_PORT to the /usr/bin/nc lines as requested?

LGTM

dylanwh:
Testing found another bug -- I was using needs_any() instead of wait_any().

## wait_any
Returns a new Future instance that will indicate it is ready once any of the sub future objects given to it indicate that they are ready, either by success or failure. Any remaining component futures that are not yet ready will be cancelled. Its result will be the result of the first component future that was ready; either success or failure. Any component futures that are cancelled are ignored, apart from the final component left; at which point the result will be a failure.

## needs_any
Returns a new Future instance that will indicate it is ready once any of the sub future objects given to it indicate that they have completed successfully, or when all of them indicate that they have failed. If any sub future succeeds, then this will succeed immediately, and the remaining subs not yet ready will be cancelled. Any component futures that are cancelled are ignored, apart from the final component left; at which point the result will be a failure.

dylanwh:
Dear pull request.

please pass your tests now.

I would like to move on to the next thing.

love, dylan

PR #416, opened 2018-02-26
Bug 1437646 - Part 2 - Add Bugzilla::DaemonControl

dylanwh:
This PR adds a module that exports functions which return [Futures](https://metacpan.org/pod/Future) that either start daemons (`run_httpd()`, `run_cereal()`),
check for running services (`assert_httpd()`, `assert_database()`, `assert_selenium()`).

It also has functions that return callbacks to assist writing functions similar to the above run_() functions.

### DIFF
--- a/Bugzilla/DaemonControl.pm
+++ b/Bugzilla/DaemonControl.pm
@@ -10,6 +10,7 @@ use 5.10.1;
 use strict;
 use warnings;
 
+use Bugzilla::Constants qw(bz_locations);
 use Cwd qw(realpath);
 use English qw(-no_match_vars $PROGRAM_NAME);
 use File::Spec::Functions qw(catfile);
@@ -19,10 +20,9 @@ use IO::Async::Loop;
 use IO::Async::Process;
 use IO::Async::Protocol::LineStream;
 use IO::Async::Signal;
+use IO::Socket;
 use LWP::Simple qw(get);
 use POSIX qw(setsid WEXITSTATUS);
-use WWW::Selenium::Util qw(server_is_running);
-use Bugzilla::Constants qw(bz_locations);
 
 use base qw(Exporter);
 
@@ -161,9 +161,10 @@ sub assert_selenium {
     my $repeat = repeat {
         $loop->delay_future(after => 1)->then(
             sub {
-                my $ok = server_is_running() ? 1 : 0;
-                warn "no selenium\n" unless $ok;
-                Future->wrap($ok)
+                my $host = $ENV{SRC_HOST} // 'localhost';
+                my $port = $ENV{SRC_PORT} // 4444;
+                my $sock = IO::Socket::INET->new( PeerAddr => $host, PeerPort => $port );
+                Future->wrap($sock ? 1 : 0);
             },
         );
     } until => sub { shift->get };
--- a/Bugzilla/DaemonControl.pm
+++ b/Bugzilla/DaemonControl.pm
@@ -33,12 +33,10 @@ our @EXPORT_OK = qw(
 );
 
 our %EXPORT_TAGS = (
-    all => \@EXPORT_OK,
-    run => [grep { /^run_/ } @EXPORT_OK],
+    all   => \@EXPORT_OK,
+    run   => [grep { /^run_/ } @EXPORT_OK],
     utils => [qw(catch_signal on_exception on_finish)],
 );
-
-our $LOGGING_PORT = $ENV{LOGGING_PORT} // 5880;
 
 use constant HTTPD_BIN     => '/usr/sbin/httpd';
 use constant HTTPD_CONFIG  => realpath(catfile( bz_locations->{confdir}, 'httpd.conf' ));
@@ -67,6 +65,8 @@ sub catch_signal {
 
 sub cereal {
     local $PROGRAM_NAME = "cereal";
+    $ENV{LOGGING_PORT} //= 5880;
+
     my $loop = IO::Async::Loop->new;
     my $on_stream = sub {
         my ($stream) = @_;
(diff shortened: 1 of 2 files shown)


Bugzilla/DaemonControl.pm
```diff
@@ -0,0 +1,317 @@
...
+use Future::Utils qw(repeat try_repeat);
+use Future;
+use IO::Async::Loop;
+use IO::Async::Process;
+use IO::Async::Protocol::LineStream;
+use IO::Async::Signal;
+use LWP::Simple qw(get);
+use POSIX qw(setsid WEXITSTATUS);
+use WWW::Selenium::Util qw(server_is_running);
+use Bugzilla::Constants qw(bz_locations);
```
globau: there's quite a few modules that we now directly depend on that aren't in `Makefile.PL`.

Bugzilla/DaemonControl.pm
```diff
@@ -0,0 +1,317 @@
...
+use File::Spec::Functions qw(catfile);
+use Future::Utils qw(repeat try_repeat);
+use Future;
+use IO::Async::Loop;
+use IO::Async::Process;
+use IO::Async::Protocol::LineStream;
+use IO::Async::Signal;
+use LWP::Simple qw(get);
+use POSIX qw(setsid WEXITSTATUS);
+use WWW::Selenium::Util qw(server_is_running);
```
globau: is it possible to have different sets of packages for prod vs test?  i don't think selenium should be part of the production deployment requirements; instead it can be loaded at runtime during test.

Bugzilla/DaemonControl.pm
```diff
@@ -0,0 +1,317 @@
...
+    };
+}
+
+sub on_exception {
+    my ( $name, $f ) = @_;
+    return sub {
+        my ( $self, $exception, $errno, $exitcode ) = @_;
+
+        if ( length $exception ) {
+            $f->fail( "$name died with the exception $exception " . "(errno was $errno)\n" );
```
globau: nit: useless string concatenation

Bugzilla/DaemonControl.pm
```diff
@@ -0,0 +1,317 @@
...
+The C<run_> and C<assert_> functions return Futures, see L<Future> for details
+on that. But if you've used Promises in the javascript, Futures are the same concept.
+
+=head1 FUNCTIONS
+
+Nothing is exported by default, but you can request C<:all> for that.
+You can also just get the run_* functions with C<:run>.
+
+=head2 run_httpd()
+
```
globau: should probably also mention that it starts httpd

Bugzilla/DaemonControl.pm
```diff
@@ -0,0 +1,317 @@
...
+    my $loop   = IO::Async::Loop->new;
+    my $exit_f = $loop->new_future;
+    my $cereal = IO::Async::Process->new(
+        code         => \&cereal,
+        on_finish    => on_finish($exit_f),
+        on_exception => on_exception( "cereal", $exit_f ),
+    );
+    $exit_f->on_cancel( sub { $cereal->kill('TERM') } );
+    $loop->add($cereal);
+    catch_signal('USR1')->get;
```
globau: is there a small race condition here?
if the USR1 signal is sent before the signal handler is setup this would kill the process

Bugzilla/DaemonControl.pm
```diff
@@ -0,0 +1,317 @@
...
+
+    return $signal_f;
+}
+
+sub cereal {
+    local $PROGRAM_NAME = "cereal";
+    my $loop = IO::Async::Loop->new;
+    $loop->listen(
+        host => '127.0.0.1',
+        service => '5880',
```
globau: the bind port should be configurable

dylanwh: @globau: Is a package global okay?

Bugzilla/DaemonControl.pm
dylanwh: I'm not sure if the process starts at that time, but there is a clear way to rewrite this with futures, so I'll do this:

```perl
my $signal_f = catch_signal('USR1');
 $loop->add($cereal);
$signal_f->get;
```

Bugzilla/DaemonControl.pm
globau: my concern here is around when this code gets upstreamed non-bmo installs may need to bind to a different port.  within that context i don't think that needing to edit code to set the bind port is correct.

however if this is a misplaced concern feel free to keep it hardcoded here.

Bugzilla/DaemonControl.pm
```diff
@@ -0,0 +1,327 @@
...
+use File::Spec::Functions qw(catfile);
+use Future::Utils qw(repeat try_repeat);
+use Future;
+use IO::Async::Loop;
+use IO::Async::Process;
+use IO::Async::Protocol::LineStream;
+use IO::Async::Signal;
+use IO::Socket;
+use LWP::Simple qw(get);
+use POSIX qw(setsid WEXITSTATUS);
```
globau: there's still modules missing from Makefile.PL – `Future`, and `IO::Async`.

dylanwh: Future is a dep of io async and we list io-async as part lf elasticsearch. We don’t have optional dependencies in actuality anymore so this is fine.

globau: sorry, but i don't agree.

you're relying on modules that you depend on to pull in modules that you use directly.  this relies on those modules always using those modules, and removes clarity around which modules are required.

if you use a module directly it should be documented as a dependency.

dylanwh: The chance of IO::Async not depending on Future is 0%. They're tightly integrated.

dylanwh: I will however make sure this code checks for feature('bmo')

globau: as per irc, let's not block this any longer.  we can (and should!) add the deps later.

PR #419, opened 2018-02-27
Bug 1440239 - PhabBugz - Assign a secure-revision.

zalun:
Phabricator's Herald has a "Secure Email" rule and Phabricator sends a
restricted email notification if a revision belongs to the newly created
"secure-revision" group project.

With this patch Bugzilla will automatically assign a secure revision to
the "secure-revision" group project.

Additionally:
The get_project_phid is modified and is now caching the response in
memcached.

### DIFF
--- a/extensions/PhabBugz/lib/Project.pm
+++ b/extensions/PhabBugz/lib/Project.pm
@@ -232,6 +232,20 @@ sub update {
         }
     }
 
+    if ($self->{add_projects}) {
+        push(@{ $data->{transactions} }, {
+            type => 'projects.add',
+            value => $self->{add_projects}
+        });
+    }
+
+    if ($self->{remove_projects}) {
+        push(@{ $data->{transactions} }, {
+            type => 'projects.remove',
+            value => $self->{remove_projects}
+        });
+    }
+
     my $result = request( 'project.edit', $data );
 
     return $result;
@@ -265,6 +279,24 @@ sub remove_member {
     $self->{remove_members} ||= [];
     my $member_phid = blessed $member ? $member->phab_phid : $member;
     push( @{ $self->{remove_members} }, $member_phid );
+}
+
+sub add_project {
+    my ($self, $project) = @_;
+    $self->{add_projects} ||= [];
+    my $project_phid = blessed $project ? $project->phid : $project;
+    $project_phid = get_project_phid($project);
+    return undef unless $project_phid;
+    push(@{ $self->{add_projects} }, $project_phid);
+}
+
+sub remove_project {
+    my ($self, $project) = @_;
+    $self->{add_projects} ||= [];
+    my $project_phid = blessed $project ? $project->phid : $project;
+    $project_phid = get_project_phid($project);
+    return undef unless $project_phid;
+    push(@{ $self->{remove_projects} }, $project_phid);
 }
 
 sub set_members {
--- a/extensions/PhabBugz/lib/Project.pm
+++ b/extensions/PhabBugz/lib/Project.pm
@@ -281,24 +281,6 @@ sub remove_member {
     push( @{ $self->{remove_members} }, $member_phid );
 }
 
-sub add_project {
-    my ($self, $project) = @_;
-    $self->{add_projects} ||= [];
-    my $project_phid = blessed $project ? $project->phid : $project;
-    $project_phid = get_project_phid($project);
-    return undef unless $project_phid;
-    push(@{ $self->{add_projects} }, $project_phid);
-}
-
-sub remove_project {
-    my ($self, $project) = @_;
-    $self->{add_projects} ||= [];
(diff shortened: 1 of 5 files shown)


extensions/PhabBugz/lib/Util.pm
```diff
@@ -273,19 +273,26 @@ sub add_comment_to_revision {
...
 
-    my $data = {
-        queryKey => 'all',
-        constraints => {
-            name => $project
-        }
-    };
+    # Check memcache
+    my $project_phid = $memcache->get({ key => "phab_project_phid_" . $project });
+    if (not length $project_phid) {
```
dklawren: Fine to just do here:

if (!$project_phid) {

extensions/PhabBugz/lib/Revision.pm
```diff
@@ -389,4 +404,20 @@ sub set_policy {
...
+
+sub remove_project {
+    my ( $self, $project ) = @_;
+    $self->{remove_projects} ||= [];
+    my $project_phid = blessed $project ? $project->phid : $project;
+    return undef unless $project_phid;
+    push @{ $self->{remove_projects} }, $project_phid;
+}
+
+1;
```
dklawren: You were asking in IRC before, the 1; at the end is because Perl modules need to always return a true value when a calling file loads (require or use) the module.

extensions/PhabBugz/lib/Project.pm
```diff
@@ -267,6 +281,24 @@ sub remove_member {
...
+    $project_phid = get_project_phid($project);
+    return undef unless $project_phid;
+    push(@{ $self->{add_projects} }, $project_phid);
+}
+
+sub remove_project {
+    my ($self, $project) = @_;
+    $self->{add_projects} ||= [];
+    my $project_phid = blessed $project ? $project->phid : $project;
+    $project_phid = get_project_phid($project);
```
dklawren: remove this line as it is redundant and will fail if $project is an object or a PHID.

extensions/PhabBugz/lib/Project.pm
```diff
@@ -267,6 +281,24 @@ sub remove_member {
     push( @{ $self->{remove_members} }, $member_phid );
 }
 
+sub add_project {
+    my ($self, $project) = @_;
+    $self->{add_projects} ||= [];
+    my $project_phid = blessed $project ? $project->phid : $project;
+    $project_phid = get_project_phid($project);
```
dklawren: remove this line as it is redundant and will fail if $project is an object or a PHID.

extensions/Push/lib/Connector/Phabricator.pm
```diff
@@ -106,6 +106,11 @@ sub send {
               $bug->id
             ));
             make_revision_public($revision_phid);
+            # remove "secure-revision" tag
+            my $rev_obj = Bugzilla::Extension::PhabBugz::Revision->new_from_query({ phids => [ $revision_id ] });
+            my $secure_project_phid = get_project_phid('secure-revision');
```
dklawren: These two lines can be moved up to line 101 so they are not duplicated.

extensions/Push/lib/Connector/Phabricator.pm
```diff
@@ -106,6 +106,11 @@ sub send {
...
+            $revision->remove_project($secure_project_phid);
+            $rev_obj->update();
```
dklawren: Also move $revision->update() to the bottom of the foreach() loop.

You may also want to create a $revision_updated boolean that you set in the if else blocks and call

$revision_obj->update() if $revision_updated;

This way we do not make an empty call to differential.revision.edit if no change was made.

extensions/PhabBugz/lib/Util.pm
```diff
@@ -273,19 +273,26 @@ sub add_comment_to_revision {
...
+    my $memcache = Bugzilla->memcached;
 
-    my $data = {
-        queryKey => 'all',
-        constraints => {
-            name => $project
-        }
-    };
+    # Check memcache
+    my $project_phid = $memcache->get({ key => "phab_project_phid_" . $project });
```
dylanwh: question about this: how long do you expect it to be in the cache? 
also, possibly you want to use get_config() for this key.

extensions/PhabBugz/lib/Revision.pm
dylanwh: If we had a new enough version of perl, like one from 5 years ago, I think package block syntax returns a true value.

```perl
package Foo::Bar {


}
```
some day...

extensions/PhabBugz/lib/Util.pm
zalun: PHID does not change. What is the policy?

zalun: As log as possible - PHID does not change.

PR #421, opened 2018-02-27
Bug 1439993 - Remove COMPILE_DIR => setting from Bugzilla::Template

globau:
Remove COMPILE_DIR => setting from Bugzilla::Template when effective group != webservergroup to prevent filesystem permission errors


Bugzilla/Util.pm
```diff
@@ -254,6 +256,11 @@ sub i_am_webservice {
            || $usage_mode == USAGE_MODE_REST;
 }
 
+sub is_webserver_group {
+    state $web_server_gid = getgrnam(Bugzilla->localconfig->{webservergroup});
```
dylanwh: nit:
Bugzilla->localconfig->{webservergroup} can be '' (though, probably only on windows: the default is the first $EGID)
and the return value of getgrnam() can be undefined.

PR #423, opened 2018-02-27
Bug 1441732 - Improve missing module error in Bugzilla::Extensions and catch more compile errors in tests

dylanwh:
Dear reviewer, to test this I recommend installing the [circleci command line program](https://circleci.com/docs/2.0/local-jobs/#installing-the-cli-locally)
and removing asuch as *extensions/BMO/lib/Data.pm*. With this patch, `circleci --build test_sanity` should fail with a message that indicates that the file you removed is removed. Without this patch, the error will be quite different.

Also verify you seen an error with "docker build ." 

**Note**: Some extension modules are not loaded at startup, so  *extensions/BMO/lib/Util.pm* being removed won't be caught.

The purpose of this patch is to improve the situation when someone is writing extension code, to prevent hard-to-debug errors.


mohawk2:
This looks fine to me: no cleanup if running with `perl -c`. However, why the slightly different command lines for `.circleci/config.yml` and `Dockerfile`?

dylanwh:
@mohawk2 : Ah yes, those two should be identical. Fixed.

mohawk2:
I'd have thought the `-M` one was better as shorter, and you're not really actually compile-only-ing which is `-c`. However, so long as they are the same, this is fine by me.

mohawk2:
This still looks fine. I see a couple of whitespace diffs which may not be intentional. Also would it be better to rebase rather than adding merge commits?

PR #428, opened 2018-03-01
Bug 1442099 - Add memcached tracing to help debug weirdness in cloud env

dylanwh:
This adds TRACE-level logging to _set() and _get(), to help @Micheletto debug something in the new env. 

By default this will have almost no impact -- TRACE is not logged by default. Of course it's not free, but this is fine I think. :)

screen shot 2018-02-28 at 22 26 50


dylanwh:
Just add:

```
log4perl.logger.Bugzilla.Memcached = TRACE, Socket
```

to the end of log4perl-json.conf or add a new file using a docker volume mount
and set LOG4PERL_CONFIG_FILE.

PR #432, opened 2018-03-02
Bug 1442650 - Cookies dont always have httponly flag set (add debugging)

dylanwh:
Bug 1442650 - Cookies dont always have httponly flag set (add debugging)


dylanwh:
This branch helped uncover a weird bug, but we don't need the code.

PR #438, opened 2018-03-02
Bug 1441181 - Step 5 - Add logging to jobqueue

dylanwh:
This is just logging statements, and overloading the debug() method which is implemented by TheSchwartz. The local assignment tells log4perl to ignore the function itself and report from whence the debug() function is called.


dylanwh:
Replaced with #458

PR #440, opened 2018-03-02
Bug 1441181 - Step 7 - Fix bug in catch_signal()

dylanwh:
we need to remove the signal watcher after the future is marked as done.
It's not legal to call ->done() on a future more than once.


mohawk2:
While I agree with the logic of this, I would be much happier with a test that fails without this. @dylanwh ?

dylanwh:
@mohawk2 thanks for the review. I wrote a test case -- and I verified it fails on master but passes in this branch.

dylanwh:
github is confused about test_sanity -- it actually passed on the most recent run. Once the current run finishes I'll re-run it.

PR #443, opened 2018-03-05
Bug 1443049 - is_interactive() must be declared before log4perl config is loaded

dylanwh:
So this function is called as a log filter in the conf/log4perl-*.conf files. So it needs to be defined before those files are loaded.

### DIFF
--- a/Bugzilla/Logging.pm
+++ b/Bugzilla/Logging.pm
@@ -18,7 +18,7 @@ use English qw(-no_match_vars $PROGRAM_NAME);
 
 sub is_interactive {
     state $is_tty = -t STDOUT || -t STDIN;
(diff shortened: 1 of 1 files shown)


dylanwh:
@mohawk2:

So what's going on here is that the initial code that got landed had a typo, ENV instead of INC.
What the is_interactive() is trying to do is determine if sending logs to STDOUT is save to do or not.
It's safe to call Bugzilla->usage_mode only if Bugzilla.pm is loaded.

Thinking about it, probably checking $ENV{SERVER_SOFTWARE} is probably more sensible, because that will be true for code running in a CGI and nothing else.

is_interactive() is called from the log4perl config file: https://github.com/mozilla-bteam/bmo/blob/master/conf/log4perl-syslog.conf

PR #444, opened 2018-03-05
Bug 1443058 - Backport 1087400 to bmo - CGI 4.05 throws tons of "CGI::param called in list context" warnings

dylanwh:
This does look bad, but we've had this code in upstream bugzilla for a while. And the warnings tend to be very annoying. Currently this is keeping us on an older version of CGI.pm

### DIFF
--- a/Bugzilla/CGI.pm
+++ b/Bugzilla/CGI.pm
@@ -597,6 +597,8 @@ sub header {
 sub param {
     my $self = shift;
 
+    local $CGI::LIST_CONTEXT_WARN = 0;
+
     # When we are just requesting the value of a parameter...
     if (scalar(@_) == 1) {
         my @result = $self->SUPER::param(@_);
--- a/Bugzilla/CGI.pm
+++ b/Bugzilla/CGI.pm
@@ -11,6 +11,7 @@ use 5.10.1;
 use strict;
 use warnings;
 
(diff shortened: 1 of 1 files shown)


mohawk2:
Having been asked to comment on this, these are my thoughts. Just adding something to suppress warnings is highly misguided given the warnings are there for very, very good reasons, so I disagree with that.

Instead, a good way forward as suggested by @dylanwh is: capture the caller context (ie `wantarray`), then call `SUPER::param` with the same context, which will trigger warnings from those calls. With the correct `$Carp::CarpLevel`, the warning will even be given at the correct location for easy fixing!

mohawk2:
That change is an improvement, but I think you would want to use not just a one-time flag for having warned, but a per-file, per-line set of nested hashes of having warned. A bit more code, but a lot more sound. Next stage would be to fix all those occurrences, then delete this warning-suppression :-)

mohawk2:
Looks great now. For extra syntactic beauty `$has_warned` could be a simple hash instead of a hash-ref. However this looks good now and I approve it (and would also approve if it were changed to a simple hash)!

dylanwh:
@mohawk2 thanks! But you can't use 'state' on non-scalars, annoyingly.

mohawk2:
Annoyingly indeed! I wonder what they were thinking.

PR #445, opened 2018-03-05
Bug 1443057 - Backport 1136125 to bmo - Reduce memory usage of Bugzilla::CGI->send_cookie()

dylanwh:
This is just removing some dumb code that turns an array into a hash using (shift, shift) and a then a hash into an array using unshift... Originally [Bug 1136125](https://bugzilla.mozilla.org/show_bug.cgi?id=1136125).

### DIFF
--- a/Bugzilla/CGI.pm
+++ b/Bugzilla/CGI.pm
@@ -661,16 +661,7 @@ sub should_set {
 # pass them around to all of the callers. Instead, store them locally here,
 # and then output as required from |header|.
 sub send_cookie {
-    my $self = shift;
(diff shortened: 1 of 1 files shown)


mohawk2:
I'm a member of the "Perl community" and I agree this change should have no substantive negative effect. Semantically, it just simplifies the turning of `@_` into `%paramhash` but with much less code, and without an intermediate variable. Therefore, it achieves the objective of saving memory, without harming functionality.

dylanwh:
Good enough for me. :-D

PR #451, opened 2018-03-07
Bug 1444627 - saved searches as inline list

dylanwh:


extensions/MyDashboard/web/styles/mydashboard.css
```diff
@@ -83,4 +83,5 @@
 
 #saved_searches_container li {
     margin: 5px 0;
```
mnoorenberghe: Thanks! You should also make the margin 5px on every side by deleting the " 0"

PR #464, opened 2018-03-10
The default workflow should be the one we use

dylanwh:
The workflow that is created by default should be the one we create by default


Bugzilla/DB.pm
```diff
@@ -49,14 +49,12 @@ use constant ISOLATION_LEVEL => 'REPEATABLE READ';
...
-    priority     => ["Highest", "High", "Normal", "Low", "Lowest", "---"],
-    op_sys       => ["All","Windows","Mac OS","Linux","Other"],
-    rep_platform => ["All","PC","Macintosh","Other"],
-    bug_status   => ["UNCONFIRMED","CONFIRMED","IN_PROGRESS","RESOLVED",
-                     "VERIFIED"],
-    resolution   => ["","FIXED","INVALID","WONTFIX", "DUPLICATE","WORKSFORME"],
+    bug_severity => [qw(blocker critical major normal minor  trivial enhancement  )],
+    priority     => [qw(-- P1 P2 P3 P4 P5)],
+    op_sys       => [ "Unspecified", "All", "Windows", "Mac OS", "Linux", "Other" ],
+    rep_platform => [ "Unspecified", "All", "PC", "Macintosh", "Other" ],
```
zalun: Is there a reason why that one is not using qw?

dylanwh: No, actually. That would be easier to read too.

PR #466, opened 2018-03-12
Bug 1445041 - if memcached server does not end with a port, append :11211

dylanwh:


Ryuno-Ki:
@dylanwh I am not an Ops person, but after reading [ Some notes on memcached DDoS ](http://blog.erratasec.com/2018/03/some-notes-on-memcached-ddos.html) by an security engineer, I was wondering, whether it would make BMO vulnerable to that amplification attack vector. After all, changing the port is a way to avoid the most naive way of attackers to find potential victims … (Like scanning the web for port 11211 only).

If it would turn it vulnerable, I can open a ticket on BMO.

dylanwh:
Having memcached accessible to anything other than the application it is serving would be a severe security problem. it's kind of amazing that so many memcached servers are on public interfaces.

PR #478, opened 2018-03-16
Bug 1440829 - Bugzilla comment for Phabricator commit should include entire commit message, not just first line

dklawren:
…entire commit message, not just first line

### DIFF
--- a/extensions/PhabBugz/lib/Feed.pm
+++ b/extensions/PhabBugz/lib/Feed.pm
@@ -237,7 +237,7 @@ sub process_revision_change {
 
     my ($timestamp) = Bugzilla->dbh->selectrow_array("SELECT NOW()");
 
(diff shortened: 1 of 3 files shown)


extensions/PhabBugz/lib/Feed.pm
```diff
@@ -237,7 +237,7 @@ sub process_revision_change {
 
     my ($timestamp) = Bugzilla->dbh->selectrow_array("SELECT NOW()");
 
-    my $attachment = create_revision_attachment($bug, $revision->id, $revision->title, $timestamp);
+    my $attachment = create_revision_attachment($bug, $revision, $timestamp);
```
dylanwh: suggestion: maybe this should be passed a hash, so you can use named parameters?

PR #481, opened 2018-03-19
add Bugzilla->urlbase method that returns a URI object

dylanwh:


Bugzilla/Template.pm
```diff
@@ -928,7 +928,7 @@ sub create {
             'sudoer' => sub { return Bugzilla->sudoer; },
 
             # Allow templates to access the "corect" URLBase value
-            'urlbase' => sub { return Bugzilla->localconfig->{urlbase}; },
+            'urlbase' => sub { Bugzilla->urlbase },
```
globau: do we really need to create a new `URI` object each time a template requests `urlbase`?  at it stands that involves doing extra work unnecessarily.

i think it would be better create a new method that provides a `URI urlbase` object should a template need it (eg. `urlbase_uri`)

dylanwh: I'm worried about mutability here -- you can mutate uri objects at any point and there is no API to "freeze" them.

globau: you're replacing a hash lookup with a `URI` object creation (which involves parsing the url), and then stringification to render it within a template.  that comes at a cost which is questionable given no templates currently use `urlbase` as a `URI` object.

i suggest instead leaving the `urlbase` sub as-is, and instead creating a new `urlbase_uri` template sub that returns `Bugzilla->urlbase`.

dylanwh: I suppose we can do that, though all the calls to urlbase add up to about 1ms.
(each is about five microseconds).

PR #483, opened 2018-03-19
Bug 1447027 - Change default mail method for vagrant and update README

purelogiq:
Updates the default mail_delivery_method to be Sendmail for vagrant
development. This allows developers to connect to the imap server
running at bmo-web.vm:143 to view mail using a real mail client.

The default method remains the same for docker users, 'Test'.

Both methods are described in the README in a new section.

Preview the RST here: https://github.com/purelogiq/bmo/tree/plq/test-mail-readme#id12


purelogiq:
@dylanwh Fixed!

PR #486, opened 2018-03-20
fake people are always active

dylanwh:


extensions/Review/Extension.pm
```diff
@@ -93,6 +93,8 @@ sub _reviewers_objs {
 sub _user_is_active {
     my ($self) = @_;
 
+    # never consider .bugs or .tld addresses as inactive.
+    return 1 if $self->login =~ /bugs$/ || $self->login =~ /\.tld$/;
```
purelogiq: Are we missing a `\.` in the bugs regex or is that intended?

dylanwh: unfortunately that's how it checks in other parts of the code. Perhaps I should fix it in those places and here...

purelogiq: I'll r+ this then and let you fix that at your discretion :)

PR #489, opened 2018-03-21
add shadowdb username and password in localconfig param override

purelogiq:
Adds a localconfig option to specify a username and password to use when connecting to the shadow db. If not provided, the default dbuser an dbpass will be used.
Also fixes a typo with setting the shadowport and shadowsock.

### DIFF
--- a/Bugzilla/DB.pm
+++ b/Bugzilla/DB.pm
@@ -111,8 +111,13 @@ sub connect_shadow {
     my $connect_params = dclone(Bugzilla->localconfig);
     $connect_params->{db_host} = Bugzilla->get_param_with_override('shadowdbhost');
     $connect_params->{db_name} = Bugzilla->get_param_with_override('shadowdb');
-    $connect_params->{db_port} = Bugzilla->get_param_with_override('shadowport');
-    $connect_params->{db_sock} = Bugzilla->get_param_with_override('shadowsock');
+    $connect_params->{db_port} = Bugzilla->get_param_with_override('shadowdbport');
+    $connect_params->{db_sock} = Bugzilla->get_param_with_override('shadowdbsock');
+
+    if ( Bugzilla->localconfig->{'shadowdb_user'} && Bugzilla->localconfig->{'shadowdb_pass'} ) {
+        $connect_params->{db_user} = Bugzilla->localconfig->{'shadowdb_user'};
+        $connect_params->{db_pass} = Bugzilla->localconfig->{'shadowdb_pass'};
+    }
 
     return _connect($connect_params);
 }
--- a/Bugzilla/Install/Localconfig.pm
+++ b/Bugzilla/Install/Localconfig.pm
@@ -175,6 +175,14 @@ use constant LOCALCONFIG_VARS => (
         name => 'inbound_proxies',
         default => _migrate_param( 'inbound_proxies', '' ),
     },
+    {
+        name => 'shadowdb_user',
+        default => '',
+    },
+    {
+        name => 'shadowdb_pass',
+        default => '',
+    }
 );
 
 
--- a/template/en/default/setup/strings.txt.pl
+++ b/template/en/default/setup/strings.txt.pl
@@ -269,6 +269,12 @@ END
 This is the max amount of unshared memory the apache process is allowed to use
 before Apache::SizeLimit kills it. This is only applicable when run under mod_perl.
 EOT
+    localconfig_shadowdb_user => <<EOT,
+The username used to authenticate to the shadow db. 
+EOT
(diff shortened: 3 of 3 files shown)


purelogiq:
Some questions to finish this up:

1. I've added the shadowdb user/pass to the localconfig param override; does this mean that there must be a configuration option in the prefrences too? I.e. in here: https://github.com/mozilla-bteam/bmo/blob/master/Bugzilla/Config/ShadowDB.pm?

2. In `strings.txt.pl` (https://github.com/mozilla-bteam/bmo/blob/master/template/en/default/setup/strings.txt.pl) we describe each localconfig variable. Since the shadowdb settings are nested within the `param_override` field (https://github.com/mozilla-bteam/bmo/blob/master/Bugzilla/Install/Localconfig.pm#L136) which already has a entry in `strings.txt.pl` is it necessary to describe the nested fields? If so what would be the correct key in `strings.txt.pl` to use for a nested field, e.g. `localconfig_param_override_shadowdbuser`?

Bugzilla/DB.pm
```diff
@@ -111,8 +111,11 @@ sub connect_shadow {
     my $connect_params = dclone(Bugzilla->localconfig);
     $connect_params->{db_host} = Bugzilla->get_param_with_override('shadowdbhost');
     $connect_params->{db_name} = Bugzilla->get_param_with_override('shadowdb');
-    $connect_params->{db_port} = Bugzilla->get_param_with_override('shadowport');
-    $connect_params->{db_sock} = Bugzilla->get_param_with_override('shadowsock');
+    $connect_params->{db_port} = Bugzilla->get_param_with_override('shadowdbport');
```
purelogiq: notice that `shadowport` and `shadowsock` below have been changed to `shadowdbport` and `shadowdbsock`. I assume this was a typo from before, these new values match what is defined in the localconfig.

Bugzilla/DB.pm
```diff
@@ -111,8 +111,11 @@ sub connect_shadow {
...
+    $connect_params->{db_sock} = Bugzilla->get_param_with_override('shadowdbsock');
+    $connect_params->{db_user} = Bugzilla->get_param_with_override('shadowdbuser');
```
dylanwh: Sorry I didn't mention in the bug -- don't use the get_param_with_override stuff.
That exists because the original shadow_ params exist in the data/params file, but cloudops can only modify the localconfig file.

Just use `Bugzilla->localconfig->{KEY}` for the new entries.

Bugzilla/Install/Localconfig.pm
```diff
@@ -141,6 +141,8 @@ use constant LOCALCONFIG_VARS => (
             shadowdbhost         => undef,
             shadowdbport         => undef,
             shadowdbsock         => undef,
+            shadowdbuser         => undef,
```
dylanwh: instead of having these in the override thing, just add them directly above (e.g. the same as "urlbase" or "apache_size_limit"

Bugzilla/DB.pm
```diff
@@ -111,8 +111,13 @@ sub connect_shadow {
     my $connect_params = dclone(Bugzilla->localconfig);
     $connect_params->{db_host} = Bugzilla->get_param_with_override('shadowdbhost');
     $connect_params->{db_name} = Bugzilla->get_param_with_override('shadowdb');
-    $connect_params->{db_port} = Bugzilla->get_param_with_override('shadowport');
-    $connect_params->{db_sock} = Bugzilla->get_param_with_override('shadowsock');
+    $connect_params->{db_port} = Bugzilla->get_param_with_override('shadowdbport');
+    $connect_params->{db_sock} = Bugzilla->get_param_with_override('shadowdbsock');
+
+    if ( Bugzilla->localconfig->{'shadowdb_user'} && Bugzilla->localconfig->{'shadowdb_pass'} ) {
```
purelogiq: Let me know if there are any line length requirements, perltidy didn't seem to mind the length.

purelogiq:
Updated and confirmed that bugzilla still installs correctly

Bugzilla/DB.pm
dylanwh: It's fine, I think it's 120 chars.

Bugzilla/DB.pm
dylanwh: looks like sock got deleted here?

purelogiq:
@dylanwh What's the norm for merging things? When they're accepted, or "at the right time".

dylanwh:
Squash and merge but make sure the commit message is from the bug via the "Copy Summary" button on bmo. (remove the stuff github adds to the end too)

PR #490, opened 2018-03-21
Bug 1200695 - API-key-creation emails should reflect if the action was a result of auth delegation

purelogiq:
When a user logs in to MozReview or Phabricator for the first time an api key is created for them. Because the api key is created in the background, the user may be alarmed when they next check their email and see that an api key was created on their Bugzilla account.

This PR updates our email to inform the user if one of our review tools was responsible for creating the api key.

This was tested with the auth delegation test shim in https://bugzilla.mozilla.org/show_bug.cgi?id=1447028:
bmo-api-email

### DIFF
--- a/template/en/default/email/new-api-key.txt.tmpl
+++ b/template/en/default/email/new-api-key.txt.tmpl
@@ -26,9 +26,17 @@ or update the key at the following URL:
 
 [%+ urlbase %]userprefs.cgi?tab=apikey
 
+[% IF new_key.app_id == Param('mozreview_app_id') %]
+This API key was automatically created by MozReview. If you did not recently log in to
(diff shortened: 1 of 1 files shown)


template/en/default/email/new-api-key.txt.tmpl
```diff
@@ -26,9 +26,17 @@ or update the key at the following URL:
 
 [%+ urlbase %]userprefs.cgi?tab=apikey
 
+[% IF new_key.app_id == Param('mozreview_app_id') %]
```
dylanwh: @markrcote What do you think of the wording?

purelogiq:
@dylanwh Fixed!

PR #492, opened 2018-03-22
Bug 1446974 - Revert changes of the subscribers list on secure revisions

dklawren:


dklawren:
Sorry about the lack of information. Basically for revisions that are private because of being linked to a private bug in BMO, we want the bug data to be the authority and the revision to just mirror the data. So the bugs cc list should be where new people are added or people are removed. The subscriber list of the revision in Phab will be kept in sync. We are also creating a warning banner on the revision stating that subscriber changes will be overridden by bug data so making membership changes in the bug is preferred. If the bug is public, then the revision is public and changes can be made independently of each other and no syncing occurs.

PR #502, opened 2018-03-28
Bug 1449413 - Refactor circleci container building stuff

dylanwh:
Mana updated with a draft of the procedure.


.circleci/config.yml
```diff
@@ -146,7 +144,7 @@ jobs:
           /app/scripts/entrypoint.pl load_test_data --legacy
       - run:
           command: |
-            /app/scripts/entrypoint.pl test_selenium | tee artifacts/$CIRCLE_JOB.txt
+            false "#/app/scripts/entrypoint.pl test_selenium | tee artifacts/$CIRCLE_JOB.txt"
```
globau: this doesn't look correct

dylanwh: Needed to stub it out to verify how circleci works; the local circleci client doesn't entirely behave the same way. :)

PR #512, opened 2018-04-02
Bug 1450679 - Replace custom Sentry integration with Logging

dylanwh:
- Throw{User,Code}Errors are logged unless running under `ERROR_MODE_DIE`.
   Basically if the user (either from the API or from the web interface) sees an error, we log the error.
- We don't inject a `__WARN__` or `__DIE__` handler here -- that is out of scope
- Logging the entire MDC structure as MozLog 'Fields' seems to be problematic for further work I want to do to use log4perl's sentry module, so instead we have a single MDC field accessible as `Bugzilla::Logging->fields`.
- The error name and the variables are logged, but we don't log the message. The message isn't important as we can figure it out from the error name.
- I've resisted the temptation to refactor _throw_error(), which currently builds the log message twice in some cases. This is pre-existing and not important to fix right now.


Bugzilla/Error.pm
```diff
@@ -214,7 +201,6 @@ sub ThrowTemplateError {
     $vars->{'template_error_msg'} = $template_err;
     $vars->{'error'} = "template_error";
 
-    sentry_handle_error('error', $template_err);
     $vars->{'template_error_msg'} =~ s/ at \S+ line \d+\.\s*$//;
```
globau: errors here should be logged too

PR #513, opened 2018-04-03
Bug 1450791 - SES handler needs to support both "event" and "notification" messages

globau:
Resolves issues with handling complaint messages in production.

### DIFF
--- a/ses/index.cgi
+++ b/ses/index.cgi
@@ -13,7 +13,7 @@ use warnings;
 use lib qw(.. ../lib ../local/lib/perl5);
 
 use Bugzilla ();
-use Bugzilla::Logging;
+use Bugzilla::Logging qw( FATAL WARN );
 use Bugzilla::Constants qw( ERROR_MODE_DIE );
 use Bugzilla::Mailer qw( MessageToMTA );
 use Bugzilla::User ();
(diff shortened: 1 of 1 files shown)


ses/index.cgi
```diff
@@ -13,7 +13,7 @@ use warnings;
 use lib qw(.. ../lib ../local/lib/perl5);
 
 use Bugzilla ();
-use Bugzilla::Logging;
+use Bugzilla::Logging qw( FATAL WARN );
```
dylanwh: This is not correct. Use the bare import for this.
I'll add a warning about this later.

ses/index.cgi
```diff
@@ -42,12 +42,13 @@ sub main {
     elsif ( $message_type eq 'Notification' ) {
         my $notification = decode_json_wrapper( $message->{Message} ) // return;
 
-        my $notification_type = $notification->{notificationType} // '';
+        my $notification_type = $notification->{eventType} // $notification->{notificationType} // '';
```
dylanwh: I'd rather check for the existence of the key here and error out if it is neither (with logging)

ses/index.cgi
```diff
@@ -42,12 +42,13 @@ sub main {
...
         if ( $notification_type eq '' ) {
             my $keys = join ', ', keys %$notification;
-            WARN("No notificationType in notification (keys: $keys)");
+            WARN("No notification type in message (keys: $keys)");
+            respond( 200 => 'OK' );
```
dylanwh: is 200 correct here?

globau: hrm, 400 is probably better, and won't trigger a resend.

PR #514, opened 2018-04-03
[WIP] Bug 1449282 - add jobqueue_status api

purelogiq:
## In Progress 

Adds a new REST api endpoint that reports the status of jobs
in the jobqueue.

bmo-job-queue

### DIFF
--- a/Bugzilla/WebService/Bugzilla.pm
+++ b/Bugzilla/WebService/Bugzilla.pm
@@ -14,6 +14,7 @@ use warnings;
 use base qw(Bugzilla::WebService);
 use Bugzilla::Constants;
 use Bugzilla::Util qw(datetime_from);
+use Try::Tiny qw( try catch );
 
 use DateTime;
 
@@ -28,6 +29,7 @@ use constant READ_ONLY => qw(
     timezone
     time
     version
+    jobqueue_status
 );
 
 use constant PUBLIC_METHODS => qw(
@@ -35,6 +37,7 @@ use constant PUBLIC_METHODS => qw(
     time
     timezone
     version
+    jobqueue_status
 );
 
 sub version {
@@ -78,6 +81,36 @@ sub time {
         tz_offset     => $self->type('string', '+0000'),
         tz_short_name => $self->type('string', 'UTC'),
     };
+}
+
+sub jobqueue_status {
+    my ( $self, $params ) = @_;
+    
+    Bugzilla->login(LOGIN_REQUIRED);
+
+    my $dbh = Bugzilla->dbh;
+    my $query = qq{
+        SELECT
+            COUNT(*) AS total,
+            COALESCE(
+                (SELECT COUNT(*)
+                    FROM ts_error
+                    WHERE ts_error.jobid = j.jobid
+                ) 
+            , 0) AS errors
+        FROM ts_job j
+            INNER JOIN ts_funcmap f
+                ON f.funcid = j.funcid;
+    };
+
+    my $status = {};
+    try {
+        $status = $dbh->selectall_arrayref($query, { Slice => {} });
+    } catch {
+        ThrowCodeError( "jobqueue_status_failed", { message => "Unable to report job queue status at this time." } );
+    }
+
+    return $status;
 }
 
 1;
--- a/docs/en/rst/api/core/v1/bugzilla.rst
+++ b/docs/en/rst/api/core/v1/bugzilla.rst
@@ -300,3 +300,32 @@ name             type    description
 ===============  ======  ====================================================
 last_audit_time  string  The maximum of the at_time from the audit_log.
 ===============  ======  ====================================================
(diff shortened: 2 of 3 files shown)


purelogiq:
Questions for @dylanwh or @dklawren 

- If I log a message or cause a runtime error where can i view the logs inside the vagrant container? I haven't found anything in /var/logs/httpd/error_log.

- In the bug, dkl also included a count of 'deferred' jobs. Is this still wanted and if so can someone point me to the right database table that I could pull the info from?

- The SQL query here isn't very complicated, but, it would be simpler as 2. Do we prioritize making 1 request with a complex query to save connection overhead, or making several simple queries for readability.

- I've also added an inline comment on line 110.

Bugzilla/WebService/Bugzilla.pm
```diff
@@ -80,6 +83,36 @@ sub time {
...
+        FROM ts_job j
+            INNER JOIN ts_funcmap f
+                ON f.funcid = j.funcid;
+    };
+
+    my $status = {};
+    try {
+        $status = $dbh->selectall_arrayref($query, { Slice => {} });
+    } catch {
+        ThrowCodeError( "jobqueue_status_failed", { message => "Unable to report job queue status at this time." } );
```
purelogiq: - Is there a better error code to use? Here I'm using a string instead of one of the error codes defined in Bugzilla::Constants because I couldn't find one that fit.
- This ThrowCodeError doesn't seem to work as expected. When I trigger an error with an invalid SQL query, I end up getting `{}` in the response instead of the normal json error response. Could I be doing something wrong here?

purelogiq: Actually I see now that this error code is causing a test failure.

docs/en/rst/api/core/v1/bugzilla.rst
```diff
@@ -300,3 +300,32 @@ name             type    description
...
+   {
+     "total": 12,
+     "errors": 0
+   }
+
+===============  =======  ====================================================
+name             type     description
+===============  =======  ====================================================
+total            integer  The total number of jobs in the job queue.
+errors           integer  The number of errors produced by jobs in the queue.
```
purelogiq: - I think this is right? Or is it 'the number of jobs that have errored out'? The db query makes it seem like there can be more than one error per job.

Bugzilla/WebService/Bugzilla.pm
```diff
@@ -14,6 +14,7 @@ use warnings;
 use base qw(Bugzilla::WebService);
 use Bugzilla::Constants;
 use Bugzilla::Util qw(datetime_from);
+use Try::Tiny qw( try catch );
```
dylanwh: For Try::Tiny, using the default imports is fine.
```perl
use Try::Tiny;
```

Bugzilla/WebService/Bugzilla.pm
```diff
@@ -80,6 +83,36 @@ sub time {
     };
 }
 
+sub jobqueue_status {
+    my ( $self, $params ) = @_;
+    
+    Bugzilla->login(LOGIN_REQUIRED);
+
+    my $dbh = Bugzilla->dbh;
+    my $query = qq{
```
dylanwh: use `q{` in place of `qq{` here since you don't interpolate any variables into the string.

Bugzilla/WebService/Bugzilla.pm
dylanwh: This interesting! I have no idea what is going on.

dylanwh: Perhaps `jobqueue_status_error`.
Also `use Bugzilla::Logging` and use ERROR($_); to log the error.

purelogiq:
Fixed:
- return hash instead of array. ({total: 0, ...}, not [{total: 0, ...})
- add ; to fix try/catch block
- import bugzilla::error to use ThrowCodeError
- add 'jobqueue_status_error' to Bugzilla::Constants and code-error
template.

Bugzilla/WebService/Bugzilla.pm
```diff
@@ -80,6 +85,37 @@ sub time {
...
+                ) 
+            , 0) AS errors
+        FROM ts_job j
+            INNER JOIN ts_funcmap f
+                ON f.funcid = j.funcid;
+    };
+
+    my $status;
+    try {
+        $status = $dbh->selectrow_hashref($query);
```
dylanwh: One more bit: All values that come out of mysql are strings by default, so you need to convince perl that they're numbers.

This should be sufficient:

```perl
        $status->{errors} = 0 + $status->{errors};
        $status->{total}  = 0 + $status->{total};
```

purelogiq: Nice catch, tested this and it works great.

PR #515, opened 2018-04-03
Add data-no-csrf to 'safe' forms

psiinon:
This is to fix https://bugzilla.mozilla.org/show_bug.cgi?id=1446431
The data-no-csrf attribute is used to signify that a form is 'safe' (ie doesnt actually make any permanent changes) and so doesnt need an anti-csrf token.
If any of the forms I've changed do actually make permanent changes then please flag them to me :)
They may be other forms that need this attribute but hopefully this is a good start.

### DIFF
--- a/template/en/default/bug/choose.html.tmpl
+++ b/template/en/default/bug/choose.html.tmpl
@@ -24,7 +24,7 @@
    title = "Search by $terms.bug number"
  %]
 
-<form method="get" action="show_bug.cgi">
+<form method="get" action="show_bug.cgi" data-no-csrf>
   <p>
     You may find a single [% terms.bug %] by entering its [% terms.bug %] id here:
     <input name="id" size="6">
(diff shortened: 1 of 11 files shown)


purelogiq:
@psiinon Is there a way for me to easily run the baseline scan myself, or a place where I can view the forms that get flagged?

psiinon:
@purelogiq yeah, details are here: https://github.com/zaproxy/zaproxy/wiki/ZAP-Baseline-Scan
You'll need to use one of the report options to see which forms its complaining about.
I've specified a config file with the contents:

```
rules.cookie.ignorelist=DEFAULTFORMAT,Bugzilla_login_request_cookie
rules.csrf.ignorelist=github_secret,Bugzilla_login_token,token_top,token
rules.csrf.ignore.attname=data-no-csrf
```
We use the 'live' ZAP Docker image, but the weekly one should be fine too.
Or you can just spider the app using ZAP locally, which is what I do ;)
Let me know if you have any problems.

PR #519, opened 2018-04-05
Bug 1451573 Remove links to Firefox Marketplace forms

emceeaich:


emceeaich:
Ah, I need to do this from my local clone of the repo.

PR #520, opened 2018-04-05
Bug 1451573, Remove marketplace forms

emceeaich:


emceeaich:
Haha, yes. I'll dig into those. Thanks.

PR #525, opened 2018-04-10
Bug 1451416 - Bugzilla sometimes sends emails to accounts when it shouldn't

dylanwh:
to populate the badhosts bloomfilter for testing:

```bash
perl scripts/bloomfilter-populate.pl badhosts < list-of-hostnames.txt
```

### DIFF
--- a/Bugzilla/Mailer.pm
+++ b/Bugzilla/Mailer.pm
@@ -191,7 +191,7 @@ sub MessageToMTA {
         die qq{Unable to parse "To:" address - $to} unless @recipients;
         die qq{Did not expect more than one "To:" address in $to} if @recipients > 1;
         my $badhosts = Bugzilla::Bloomfilter->lookup("badhosts") or die "No badhosts bloomfilter";
-        if ($badhosts->test($to_address->host)) {
+        if ($badhosts->test($recipients[0]->host)) {
             WARN("Attempted to send email to address in badhosts: $to");
             $email->header_set(to => '');
         }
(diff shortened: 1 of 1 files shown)


Bugzilla/Mailer.pm
```diff
@@ -182,6 +185,21 @@ sub MessageToMTA {
     Bugzilla::Hook::process('mailer_before_send',
                             { email => $email, mailer_args => \@args });
 
+    try {
+        my $to         = $email->header('to') or die 'Unable to find "To:" address';
+        my @recipients = Email::Address->parse($to);
+        die qq{Unable to parse "To:" address - $to} unless @recipients;
+        foreach my $to_address (@recipients) {
+            my $badhosts   = Bugzilla::Bloomfilter->lookup("badhosts") or die "No badhosts bloomfilter";
```
purelogiq: Is this something that must be done in the inner loop? Or can we move this outside?

dylanwh: It's unlikely to ever be a loop of more than one iteration.
Perhaps it would be more clear to make it an error if `@recipients > 1`?

purelogiq: @dylanwh So this code is run several times with each person to be emailed, I see. I thought for a minute that `@recipients` variable would have a list of all people receiving the email, but, yeah if it's like a normal email To field then there'll likely be just one address. I'll approve this and leave it to you if you want to clarify that.

purelogiq:
@dylanwh r+ as soon as that test error is fixed :)

dylanwh:
should be passing all tests now @purelogiq.
fingers crossed.

PR #526, opened 2018-04-10
Bug 1453126 - Bugzilla::Bloomfilter should encourage preserving the input files for its filters

dylanwh:


dylanwh:
@purelogiq  So is this good now?

PR #531, opened 2018-04-11
Bug 1451960 - Improve opengraph metadata for access-denied bugs

gijsk:
@dylanwh does this look OK? :-)


purelogiq:
@gijsk Looks good at first glance. I'll pull it down and test it out :)

PR #533, opened 2018-04-11
Bug 1451599 - Move guidelines checkbox on signup page to the left

purelogiq:
On the right the checkbox would be in a weird place if text had
overflowed on a smaller screen. Being on the left is also standard
practice.

## Before
before-checkbox

## After
after-checkbox


dylanwh:
I kinda think both the checkbox and the button should be on the right.

Buttons on the left are unusual, right?

PR #535, opened 2018-04-11
Bug 1452241 - Improve feed error handling and logging

dklawren:


zalun:
Is `USERS` making it obvious the log comes from `Feed.pm`?

dylanwh:
To answer @zalun's question -- yeah, there are context for the loggers.

TRACE, DEBUG, etc are just shortcuts.
DEBUG is:

```perl
sub DEBUG {
    my $logger = Log::Log4perl->get_logger(__PACKAGE__);
    $logger->DEBUG(@_);
}
```

and __PACKAGE__ is whatever "package ..." says at the top of the file.

In the logging formats, this is %c (for category) or in the mozlog JSON format this is the 'Type' field.

Now with that said, nothing prevents the code from having multiple loggers.
I don't think there is anything wrong with the "USERS:" or "FEED:" prefixes,
but perhaps each of the sync functions should have its own category?

Loggers are cheap:
my $feed_log = Log::Log4perl->get_logger("feed_sync");
$feed_log->debug("hello, word");

dklawren:
Only problem is that is I do:

my $feed_log = Log::Log4perl->get_logger("feed_sync");
$feed_log->debug("hello, word");

I would lose the use of the DEBUG(), INFO(), etc. shortcut functions.

I am fine with just leaving the prefixes that I have in the latest patch if that is OK,

dkl

dylanwh:
Yep! I wanted to let you know what is possible, it isn't needed though

extensions/PhabBugz/lib/Feed.pm
```diff
@@ -111,15 +111,15 @@ sub feed_query {
         my $object_phid = $story_data->{objectPHID};
         my $story_text  = $story_data->{text};
 
-        DEBUG("STORY ID: $story_id");
-        DEBUG("STORY PHID: $story_phid");
-        DEBUG("AUTHOR PHID: $author_phid");
-        DEBUG("OBJECT PHID: $object_phid");
-        INFO("STORY TEXT: $story_text");
+        TRACE("FEED: STORY ID - $story_id");
```
dylanwh: One thing, you could put info into Bugzilla::Logging->fields->{ITEM},
if you wanted to associate information with each log entry.

This means when you log an INFO()  or something in production, the information would be available in the 'Fields' JSON field.

dklawren: I like that idea so the new patch does that. Much cleaner IMO.

dklawren:
Going to deploy this to bugzilla-dev.mozilla.org for further testing.

PR #541, opened 2018-04-13
Bug 1452984 - double-check new accounts with BMO to catch and notify of username squatting

dklawren:
Currently in the patch, I am just using WARN to note when a possible squatting has occurred. I will discuss in the bug if we want to go further and email someone or disable the account.


extensions/PhabBugz/lib/Feed.pm
```diff
@@ -472,8 +494,10 @@ sub process_new_user {
     # the first value of each row should be the bug id
     my @bug_ids = map { shift @$_ } @$data;
 
+    INFO("USERS: Updating subscriber values for old private bugs");
+
     foreach my $bug_id (@bug_ids) {
-        DEBUG("Processing bug $bug_id");
+        INFO("USERS: Processing bug $bug_id");
```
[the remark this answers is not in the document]
dylanwh: I think both of those are interesting / useful for ops, so INFO is the right level.
Perhaps the logging on line 500 should be TRACE or DEBUG but the line 497 is definitely INFO-worthy.

extensions/PhabBugz/lib/Feed.pm
```diff
@@ -482,21 +506,21 @@ sub process_new_user {
 
         foreach my $attachment (@attachments) {
             my ($revision_id) = ($attachment->filename =~ PHAB_ATTACHMENT_PATTERN);
-            DEBUG("Processing revision D$revision_id");
+            INFO("USERS: Processing revision D$revision_id");
```
dylanwh: This could be DEBUG

extensions/PhabBugz/lib/Feed.pm
```diff
@@ -420,7 +420,7 @@ sub process_new_user {
     my $phab_user = Bugzilla::Extension::PhabBugz::User->new($user_data);
 
     if (!$phab_user->bugzilla_id) {
-        DEBUG("SKIPPING: No bugzilla id associated with user");
+        DEBUG("USERS: SKIPPING - No bugzilla id associated with user");
```
dylanwh: Is this debugging, or would you want to know in production if a phabricator user had no bmo id?

extensions/PhabBugz/template/en/default/admin/email/squatter-alert.txt.tmpl
```diff
@@ -0,0 +1,34 @@
...
+  # file, You can obtain one at http://mozilla.org/MPL/2.0/.
+  #
+  # This Source Code Form is "Incompatible With Secondary Licenses", as
+  # defined by the Mozilla Public License, v. 2.0.
+  #%]
+
+[% PROCESS global/variables.none.tmpl %]
+
+From: [% Param('mailfrom') %]
+To: moz-phabricator-admin@mozilla.com
```
globau: as per irc this should be `phabricator-admin@mozilla.com`

extensions/PhabBugz/lib/Feed.pm
```diff
@@ -518,6 +520,58 @@ sub process_new_user {
     # Pre setup before querying DB
     my $old_user = set_phab_user();
 
+    # CHECK AND WARN FOR POSSIBLE USERNAME SQUATTING
+    INFO("Checking for username squatters");
+    my $dbh     = Bugzilla->dbh;
+    my $regexp  = $dbh->quote( ":?:\Q" . $phab_user->name . "\E[[:>:]]" );
```
dylanwh: \Q doesn't work here.

```
2> my $phab_user = "...."
$res[0] = '....'

3>  ":?:\Q" . $phab_user . "\E[[:>:]]"
Useless use of \E at reply input line 1.
$res[1] = ':?:....[[:>:]]'
```

extensions/PhabBugz/lib/Feed.pm
```diff
@@ -520,6 +521,58 @@ sub process_new_user {
...
+        SELECT userid, login_name, realname
+          FROM profiles
+         WHERE userid != ? AND " . $dbh->sql_regexp( 'realname', $regexp ),
+        { Slice => {} },
+        $bug_user->id );
+    if (@$results) {
+        # The email client will display the Date: header in the desired timezone,
+        # so we can always use UTC here.
+        my $timestamp = Bugzilla->dbh->selectrow_array('SELECT LOCALTIMESTAMP(0)');
+        $timestamp = format_time($timestamp, '%a, %d %b %Y %T %z', 'UTC');
```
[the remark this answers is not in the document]
dklawren: IMO it would be better to leave it for those not using an AWS environment since as you stated AWS will just overwrite it anyway.

extensions/PhabBugz/lib/Feed.pm
```diff
@@ -520,6 +521,58 @@ sub process_new_user {
...
+                  . "phab user login: "
+                  . $phab_user->name
+                  . " phab user realname: "
+                  . $phab_user->realname
+                  . " bugzilla user id: "
+                  . $row->{userid}
+                  . " bugzilla login: "
+                  . $row->{login_name}
+                  . " bugzilla realname: "
+                  . $row->{realname} );
```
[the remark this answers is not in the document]
dklawren: I blame perltidy for this. Will fix formatting.

extensions/PhabBugz/lib/Feed.pm
```diff
@@ -520,6 +521,53 @@ sub process_new_user {
...
+        my $timestamp = Bugzilla->dbh->selectrow_array('SELECT LOCALTIMESTAMP(0)');
+        $timestamp = format_time($timestamp, '%a, %d %b %Y %T %z', 'UTC');
+
+        foreach my $row (@$results) {
+            WARN( "Possible username squatter: "
+                  . "phab user login: " . $phab_user->name
+                  . " phab user realname: " . $phab_user->realname
+                  . " bugzilla user id: " . $row->{userid}
+                  . " bugzilla login: " . $row->{login_name}
+                  . " bugzilla realname: " . $row->{realname} );
```
dylanwh: The off-side formatting looks weird too. WARN behaves like warn and print, so you can use a list arguments.

```perl
WARN(
    'Possible username squatter: ',
    'phab user login: ' . $phab_user->name,
    ' phab user realname: ' . $phab_user->realname,
    ' bugzilla user id: ' . $row->{userid},
    ' bugzilla login: ' . $row->{login_name},
    ' bugzilla realname: ' . $row->{realname},
);
```

dylanwh: Also perltidy/critic will complain about double quotes that don't need to be double quotes.

PR #542, opened 2018-04-15
Bug 1454250 - landfill.bugzilla.org - Software error at "bugzilla-4.4-branch"

SuriyaaKudoIsc:
This pull request fixes the [**landfill.bugzilla.org software error bug** *at Bugzilla Documentation*](https://bugzilla.mozilla.org/show_bug.cgi?id=1454250). 🔧 

I changed the version `4.4` to the latest version `5.0`.


dylanwh:
The right thing here would be to update the link to a running copy of bugzilla 6, "harmony" which we don't have yet.

PR #543, opened 2018-04-16
Bug 1438205 - Preserve comments in progress across page reloads

purelogiq:
### How it works
- When you input (type or copy paste) into the comment box, we save the text of the box into localStorage.
- When the bug page loads, or tab is switched to (if 2 tabs with the same bug are open) we populate the comment box with the saved text, if it exists.
- When you save a bug we clear the localStorage.
- This solves the problem mentioned in https://bugzilla.mozilla.org/show_bug.cgi?id=1438205 and just in general allows people to start typing a comment, close the page, and come back to it tomorrow.

### DIFF
--- a/extensions/BugModal/web/bug_modal.js
+++ b/extensions/BugModal/web/bug_modal.js
@@ -97,8 +97,8 @@ $(function() {
     }
 
     function saveBugComment(text) {
-        if(text.length < 1) return clearSavedBugComment();
-        if(text.length >  65535) return;
+        if (text.length < 1) return clearSavedBugComment();
+        if (text.length >  65535) return;
         let key = `bug-modal-saved-comment-${BUGZILLA.bug_id}`;
         let value = {
             text: text,
@@ -116,7 +116,7 @@ $(function() {
         expireSavedComments();
         let key = `bug-modal-saved-comment-${BUGZILLA.bug_id}`;
         let value = JSON.parse(localStorage.getItem(key));
-        if(value){
+        if (value){
             let commentBox = document.querySelector("textarea#comment");
             commentBox.value = value['text'];
             if (BUGZILLA.user.settings.autosize_comments) {
@@ -128,13 +128,13 @@ $(function() {
     function expireSavedComments() {
         const AGE_THRESHOLD = 7 * 24 * 60 * 60 * 1000; // 7 days in milliseconds.
         let expiredKeys = [];
-        for(let i = 0; i < localStorage.length; i++) {
+        for (let i = 0; i < localStorage.length; i++) {
             let key = localStorage.key(i);
-            if(key.match(/^bug-modal-saved-comment-/)) {
+            if (key.match(/^bug-modal-saved-comment-/)) {
                 let value = JSON.parse(localStorage.getItem(key));
                 let savedAt = value['savedAt'] || 0;
                 let age = Date.now() - savedAt;
-                if(age < 0 || age > AGE_THRESHOLD) {
+                if (age < 0 || age > AGE_THRESHOLD) {
                     expiredKeys.push(key);
                 }
             }
--- a/extensions/BugModal/web/bug_modal.js
+++ b/extensions/BugModal/web/bug_modal.js
@@ -1331,6 +1331,7 @@ $(function() {
     // finally switch to edit mode if we navigate back to a page that was editing
     $(window).on('pageshow', restoreEditMode);
     $(window).on('pageshow', restoreSavedBugComment);
+    $(window).on('focus', restoreSavedBugComment);
     restoreEditMode();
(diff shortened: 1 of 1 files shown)


purelogiq:
### Questions:

- In general for UI bmo work, we don't have to worry about super old IE browsers right? For browsers that we expect to be running the JS, can we assume >IE10 or something?

- Is this file, `bug_modal.js` the right place to put this stuff? It seems to be where all the other JS for the show bug page lives.

- How do you want to handle comment expiration, if at all? If the user hasn't submitted a saved comment in like 30 days, should we just remove it or leave it saved forever?

- Overall feedback appreciated :)

dylanwh:
### Questions:

- Q: _In general for UI bmo work, we don't have to worry about super old IE browsers right? For browsers that we expect to be running the JS, can we assume >IE10 or something?_
  - A: This is covered here: https://wiki.mozilla.org/BMO#Browser_Support_For_BMO

- Q: _Is this file, `bug_modal.js` the right place to put this stuff? It seems to be where all the other JS for the show bug page lives._
  - A: __Yes. bug_modal.js, and for styles bug_modal.css__
  - This is all very messy right now, I'll develop a plan for making this cleaner.

- Q: _How do you want to handle comment expiration, if at all? If the user hasn't submitted a saved comment in like 30 days, should we just remove it or leave it saved forever?_
  - A: >7 days, expire.
  - Also, if possible remove old saved comments for unvisited bugs too.

purelogiq:
## Updated:

- Refactored the core functions to be vanilla ES6. Only the event handlers near the bottom are left with jQuery style to be consistent in that part of the file.

- Added expiration for saved comments older than a week. This happens once on page load and will clear out all comments that have expired.

- Added some safe guards around saving 0-length or comments > 65535 characters. 

- Tested with large comments. **Result:** Pasting a comment that is 65535 characters causes about a quarter-second jitter, but, afterwards editing the large comment is smooth as usual.

- If the user has the autosize option enabled, we respect that and resize when updating the textbox from localstorage.

@dylanwh FYI autosize seems to be really wonky with large comments (it jitters between the top and bottom of the textbox when typing), even before this patch and on production. Since the vagrant vm has it enabled by default but production doesn't, for a long time I thought it was because of my changes. On my vm I've made disabled 'Expand the comment box dynamically' and enabled 'Zoom textareas large when in use' in General Preferences - do you want to make that the default for the vagrant vm?

dylanwh:
> do you want to make that the default for the vagrant vm?
Yes, and perhaps that code needs to be removed in a future bug?

extensions/BugModal/web/bug_modal.js
```diff
@@ -96,6 +96,54 @@ $(function() {
...
+            }
+        }
+    }
+
+    function expireSavedComments() {
+        const AGE_THRESHOLD = 7 * 24 * 60 * 60 * 1000; // 7 days in milliseconds.
+        let expiredKeys = [];
+        for(let i = 0; i < localStorage.length; i++) {
+            let key = localStorage.key(i);
+            if(key.match(/^bug-modal-saved-comment-/)) {
```
dylanwh: space here

extensions/BugModal/web/bug_modal.js
```diff
@@ -96,6 +96,54 @@ $(function() {
...
+            if (BUGZILLA.user.settings.autosize_comments) {
+                autosize.update(commentBox);
+            }
+        }
+    }
+
+    function expireSavedComments() {
+        const AGE_THRESHOLD = 7 * 24 * 60 * 60 * 1000; // 7 days in milliseconds.
+        let expiredKeys = [];
+        for(let i = 0; i < localStorage.length; i++) {
```
dylanwh: space here

extensions/BugModal/web/bug_modal.js
```diff
@@ -96,6 +96,54 @@ $(function() {
...
+    function clearSavedBugComment() {
+        let key = `bug-modal-saved-comment-${BUGZILLA.bug_id}`;
+        localStorage.removeItem(key);
+    }
+
+    function restoreSavedBugComment() {
+        expireSavedComments();
+        let key = `bug-modal-saved-comment-${BUGZILLA.bug_id}`;
+        let value = JSON.parse(localStorage.getItem(key));
+        if(value){
```
dylanwh: space here

extensions/BugModal/web/bug_modal.js
```diff
@@ -96,6 +96,54 @@ $(function() {
         $('#editing').val('');
     }
 
+    function saveBugComment(text) {
+        if(text.length < 1) return clearSavedBugComment();
```
dylanwh: space here and the next line

PR #545, opened 2018-04-23
Bug 1373280 - Highlight private comments in new bug modal UI

purelogiq:
Prior to this commit, the only indication that a comment was private was the small "Private" tag in the comment header. This commit restores the red font color to private comments and adds a small dashed border for those that have trouble seeing the color red.

private2


purelogiq:
This was way easier that I thought - sorry I took so long!

PR #547, opened 2018-04-24
Fix 935517 - allow users to put links in their profiles

gideon425:
submitted for feedback

### DIFF
--- a/extensions/UserProfile/Extension.pm
+++ b/extensions/UserProfile/Extension.pm
@@ -13,6 +13,7 @@ use warnings;
 
 use base qw(Bugzilla::Extension);
 
+use Bugzilla::Error;
 use Bugzilla::Constants;
 use Bugzilla::Extension::UserProfile::Util;
 use Bugzilla::Install::Filesystem;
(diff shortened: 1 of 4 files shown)


extensions/UserProfile/Extension.pm
```diff
@@ -543,6 +606,7 @@ sub install_update_db {
     my $dbh = Bugzilla->dbh;
     $dbh->bz_add_column('profiles', 'last_activity_ts', { TYPE => 'DATETIME' });
     $dbh->bz_add_column('profiles', 'last_statistics_ts', { TYPE => 'DATETIME' });
+    $dbh->bz_add_column('profiles', 'linksinfo', { TYPE => 'VARCHAR(600)' });
```
dylanwh: can VARCHAR be 600 on mysql??

extensions/UserProfile/Extension.pm
```diff
@@ -34,6 +35,65 @@ BEGIN {
...
+sub _getlinks {
+    my $self = shift;
+    my $value = $self->getlinksinfo();
+
+    my $state = 1;
+
+    my $hash;
+    my $arr = [ ];
+
+    for my $elem (split (/\n/, $value)) {
```
dylanwh: This is clever!

However I think we want only a single URL on the profile, and that the URL should be to "Mozillians"

extensions/UserProfile/Extension.pm
```diff
@@ -34,6 +35,65 @@ BEGIN {
...
+    my ($self, $args) = @_;
+    my %args = %{ $args };
+    my ($invocant, $validators) = @args{qw(class validators)};
+
+    if ($invocant->isa('Bugzilla::User')) {
+
+        $validators->{'linksinfo'} = sub {
+            my ($self, $value) = @_;
+
+            if ($value =~ m/\A([^'"]*)\z/) {
```
dylanwh: the validation should ensure the URL points to mozillians.org

PR #548, opened 2018-04-24
Bug 1454647 - Mirror all BMO groups as Phabricator projects and keep them in sync

dklawren:


zalun:
Makes sense.

extensions/PhabBugz/lib/Feed.pm
```diff
@@ -204,21 +205,17 @@ sub user_query {
 sub group_query {
     my ($self) = @_;
 
+    Bugzilla::Logging->fields->{type} = 'GROUPS';
```
dylanwh: You only want this for the duration of the function, right?
if so:
```perl
local  Bugzilla::Logging->fields->{type} = 'GROUPS';
```

PR #549, opened 2018-04-25
Bug 1456529 - Support SameSite attribute on session cookies

dylanwh:


dylanwh:
@psiinon hey, browsers that don't speak SameSite are fine with the attribute, right?

psiinon:
@dylanwh I believe so yes. 
https://caniuse.com/#feat=same-site-cookie-attribute says its ok too

PR #557, opened 2018-04-26
Bug 1275545 Use Sereal for Cache::Memcached::Fast r=dylan

is2ei:
I changed it to use "Sereal".


dylanwh:
This seems to fail with the error:

```
Serialize method is not specified at /app/local/lib/perl5/x86_64-linux-thread-multi/Cache/Memcached/Fast.pm line 590.
```

is2ei:
@dylanwh Thanks Dylan!
I fixed the error. Could you check it?

PR #562, opened 2018-04-30
Bug 1440086 - Refactor PhabBugz extension code to use new User.pm module for better type checking

dklawren:


extensions/PhabBugz/lib/Feed.pm
```diff
@@ -255,11 +258,10 @@ sub group_query {
...
 
-        if ( my @group_members = get_group_members($group) ) {
-            INFO("Setting group members for " . $project->name);
-            $project->set_members( \@group_members );
-            $project->update();
-        }
+        INFO("Setting group members for " . $project->name);
+        my @group_members = get_group_members($group);
+        $project->set_members( \@group_members );
+        $project->update();
```
globau: i assume this is a no-op when `@group_members` is empty

dklawren: Actually not really. If there are no members of a group in BMO, then there will be no members in the related project in Phabricator. This is by design.

extensions/PhabBugz/lib/Revision.pm
```diff
@@ -335,19 +334,13 @@ sub _build_subscribers {
...
-    return [] unless @phids;
-
-    my @subscribers;
-    foreach my $user (@$users) {
-        my $subscriber =
-          Bugzilla::User->new( { id => $user->{id}, cache => 1 } );
-        $subscriber->{phab_phid} = $user->{phid};
-        push @subscribers, $subscriber;
-    }
+    my $users = Bugzilla::Extension::PhabBugz::User->math(
```
zalun: shouldn't it be match?

PR #563, opened 2018-05-02
no bug - fix typo in meta referer directive

dylanwh:


floatingatoll:
This looks good to me, but I don't have reviewer access.

dylanwh:
I don't see harm in showing that traffic came from bugzilla. At least not a major harm, and the benefit is traffic stats that show we're popular.

PR #565, opened 2018-05-02
Bug 1413328 - Use tct (tocotrienol on npm) to encrypt bugmail

dylanwh:


Bugzilla/TCT.pm
```diff
@@ -0,0 +1,100 @@
...
+use 5.10.1;
+use Moo;
+
+use Bugzilla::DaemonControl qw( on_finish on_exception );
+use File::Temp;
+use Future::Utils qw(call);
+use Future;
+use IO::Async::Process;
+
+use constant TCT_BIN => '/usr/local/bin/tct';
```
globau: this belongs in `localconfig`, with all the other paths to binaries/scripts.

Bugzilla/TCT.pm
```diff
@@ -0,0 +1,100 @@
...
+}
+
+sub _build_is_valid {
+    my ($self) = @_;
+
+    my $loop = IO::Async::Loop->new;
+    my $exit_f = $loop->new_future;
+    my ($stderr, $stdout);
+    my $process = IO::Async::Process->new(
+        command => [TCT_BIN, 'check', '-k', $self->public_key_file ],
```
globau: does tct require ascii armoured keys?

the `Crypt::OpenPGP` code explicitly checks for ascii armour, but i don't see a similar check here.
the `textarea` in the UI coupled with the lack of `BINMODE` on the tempfile means a binary key is unlikely to work, but it may be a good idea to add a check for `BEGIN PGP MESSAGE` or similar.

Bugzilla/TCT.pm
```diff
@@ -0,0 +1,100 @@
...
+    $self->is_valid->then(
+        sub {
+            my ($is_valid) = @_;
+            call {
+                die 'invalid public key!' unless $is_valid;
+
+                my $output;
+                my $loop = IO::Async::Loop->new;
+                my $exit_f = $loop->new_future;
+                my @command = ( 'tct', 'encrypt', '-k', $self->public_key_file );
```
globau: `'tct'` should be `TCT_BIN`

Bugzilla/TCT.pm
dylanwh: And here I realize that every path listed in localconfig might now actually work in the new environment, when using LOCALCONFIG_ENV=1

Bugzilla/TCT.pm
dylanwh: non-ascii-armored keys will be rejected. tocotrienol will never support binary format.
The secure mail code wouldn't have ever worked with non-armored keys as well.

dylanwh:
- updated bmo-slim to version that will contain tct
- moved tct class to SecureMail extension
- added stub documentation
- researched the point about non-ascii files (details below)

# Non-Armored files
It seems nobody has ever done this, and it probably would never work. Bugzilla form fields are assumed to be (valid) utf8, and a binary gpg key wouldn't always be valid utf8. Given the properties of encryption (e.g. that it should have obvious distribution of bytes) I bet it is likely that most binary keys are not valid utf8.

As for File::Temp, File::Temp->new defaults to the binary mode.
This is meaningless on unix, but if not for that it would break on windows.
So even if the public key was binary, it would still be safe to write it out.
This is because a string containing binary data would have code points only from 0 to 255.
A unicode string would have code points above 255, and that is when you need to use `$fh->binmode(":encoding(utf-8)")`

PR #569, opened 2018-05-07
Bug 1456877 - Add a wrapper around libcmark_gfm to Bugzilla

dylanwh:


purelogiq:
@dylanwh On my reviews I normally try to test out the feature so I have some idea of what it does. In this case I think maybe I can play around with the GFM module in re.pl? Can you explain/demo the expected usage of these new modules?

dylanwh:
This PR is still in progress of course, I think I should add pod docs to at least `::Parser`.
However for testing you could write a script to use the module, or use the re.pl.
Usage is below.

```perl
use 5.10.1;
use strict;
use warnings;
use lib '.';
use Bugzilla::Markdown::GFM;

my $parser = Bugzilla::Markdown::GFM->new({ extensions => [qw( autolink tagfilter table strikethrough)]});
say $parser->render_html(<<'MARKDOWN');
markdown text here
MARKDOWN
```

purelogiq:
Cool @dylanwh, was able to parse some markdown in the re.pl with a slight modification to the `my $parser = Bugzilla::Markdown::GFM->new({...}) ` line you provided - I think it should be `Bugzilla::Markdown::GFM::Parser->new({...})`.

I also really feel my lack of expertise when reading through this code, what things would you be looking out for with a change like this that uses FFI a lot?

purelogiq:
Closed in favor of https://github.com/mozilla-bteam/bmo/pull/598

PR #571, opened 2018-05-09
Bug 1456886 - Client side link previews

purelogiq:
## Adds client side link previews
- Adds a new backend comment render function (creativiely named `quoteUrls2`) which does renders comments with an anonymous user. This allows us to cache the result for all users.
  - Extension hooks work as expected (e.g. CVE and hg.m.o autolinking).
  - Rendered public bugs/attachment links will work as before.
  - Rendered private bugs/attachment links will behave as though the user is unable to view that resource. Mainly, the title of the bug/attachment won't be there.

- Adds client side code which fetches data for the bug and attachment links in all comments on a bug page. 
  - It fetches this data as the logged in user, so if the user is able to see a bug/attachment then the link tooltip will be added.

## Notice
@dylanwh, right now the frontend code fetches data for _all_ bug/attachment links on the page, this include public ones which technically have already been rendered and had their tooltip added, since I didn't change get_bug_link and get_attachment_link to output plain links. 

### The reason I didn't change get_bug/attachment_link:
- Didn't want to duplicate that code since passing an anonymous user has the same effect.
- For bug links specifically, on production BMO bugs which you cannot see but are closed are strikethrough'd, but, if we go the frontend only route then the users will not know the open/closed status of bugs they can't see which I think would be confusing, especially in the context of some comments and what people have come to expect using bmo. Same thing for obsolete attachments.

### The reason the frontend code queries all the bugs/attachments instead of just the private ones:
I think we should filter out to only bugs with no tooltip and private bug links. Not doing that currently because the impact is minimal (I batch requests and request only the subset of data needed using the `include_fields` param) and because I didn't know which route you wanted to take.
- If left to me I think we should add a class to private links generated by get_bug/attachment_links when the anonymous user is unable to view that link. That would let the frontend code filter by things which have no class (i.e. markdown links) and things with this class (i.e. private links).

### DIFF
--- a/Bugzilla/Template.pm
+++ b/Bugzilla/Template.pm
@@ -173,7 +173,7 @@ sub quoteUrls {
     my @hook_regexes;
     Bugzilla::Hook::process('bug_format_comment',
         { text => \$text, bug => $bug, regexes => \@hook_regexes,
-          comment => $comment, user => $user });
+          comment => $comment, user => undef });
 
     foreach my $re (@hook_regexes) {
         my ($match, $replace) = @$re{qw(match replace)};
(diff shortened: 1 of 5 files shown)


purelogiq:
@dylanwh My plan of attack - feedback appreciated :)

# Discovery Step

Here we'll find all the bugs which we'll need info for so that we can make a
one batch request to BMO and have all the bug data we'll need client side.

- Query Select all the comments
- Regex match for bug id text ("bug 123"), bug link text ("http://bmo-web.vm/show_bug.cgi?id=2"),
  or a anchor tag linking to a bug (\<a href="http://bmo-web.vm/show_bug.cgi?id=2">...) needed to
  support markdown specifically.
- Add each bug's id to a set of ids to query the REST API with.
- Query the REST API and keep the data.


# Enhance Step

Here we'll actually go through and linkify bugs and add the preview hover

- Create a evenhandler function which accepts a bug id. When triggered it will display a hoverbox with the 
text being the bug summary for that bug id, pulled from our local bug data.
- Iterate through all the comments again.
- Regex match again for the same stuff.
- When a bug reference is found create a anchor tag with the same text and link to the bug.
- Search our local data for the bug id. If found, add onhover eventlistener with our function we made to the newly created anchor tag.
- Replace the matched regex with our new a tag html

purelogiq:
And that's just for bug id's...maybe we'll do just that before moving on to attachments?

dylanwh:
I think those steps make sense but you've jumped forward a bit. The next step is filling out quoteUrls2().

I'd start by copying:
https://github.com/mozilla-bteam/bmo/blob/master/Bugzilla/Template.pm#L141-L261

and replacing the bug_link (https://github.com/mozilla-bteam/bmo/blob/master/Bugzilla/Template.pm#L144) with a function that just returns a static string (like sub { "BUG" })

Then we need to figure out where else the "user" info is used.

See here: https://github.com/mozilla-bteam/bmo/blob/master/Bugzilla/Template.pm#L174-L176

For figuring this stuff out, I usually make a class to shadow the object (code below),
and instead of passing $user to the hook, pass `Bugzilla::Shadow->new(shadow_target => $user)`.
Then render some comments, and you'll get logs on how the hooks are using the `$user` object.
The goal there would be to make the hooks /not/ call the user object.

```perl
package Bugzilla::Shadow;
use Moo;
use Bugzilla::Logging;

has 'shadow_target' => ( is => 'ro', required => 1);

sub AUTOLOAD {
   my $self = shift;
   our $AUTOLOAD;
   my $method = [ split(/::/, $AUTOLOAD) ][-1];
   my ($pkg, $file, $line) = caller;
   DEBUG(sprintf "Call %s->%s() from %s", ref $self->shadow_target, $method, "$file:$line");
   return $self->shadow_target->$method(@_);
}

sub DESTROY { } # so this doesn't get called on the target

1;
```

dylanwh:
Next you'd take https://github.com/mozilla-bteam/bmo/blob/master/Bugzilla/Template.pm#L224-L228
but instead of calling get_attachment_link() you'd make a new function that (to start) just returns "attachment" or something like that

purelogiq:
### Interactions with quoteUrl + markdown parser
- if quoteUrls linkify's either a bug/attachment or autolinks url-looking text, the markdown will leave that section alone. Examples:
  - `[bug 1](http://bmo.com/bug/1)`
    - Resulting html: `[<a href="http://bmo.com/bug/1">bug 1</a>](<a href="http://bmo.com/bug/1">http://bmo.com/bug/1</a>)`

  - ` [not a linking word](http://mozila.org)`
    - Resulting html: `[not a linking word](<a href="http://mozilla.org">http://mozilla.org</a>)`

- **Recommendation**: I recommend we just disable the autolinking of url-looking text in quoteUrls. Markdown already autolinks url-looking text anyway.
  - The first example above will still break (the `[ ]( )` stuff will be displayed to the reader instead of as a link). I think this is fine because that will let people know that they shouldn't be manually linking bug numbers. We can also add some client side warning for this. Either that or we modify the regex to not pick up this pattern.
  - The second example will work properly (it will render a proper link with the text "not a linking word"). This is because markdown knows when it should autolink vs use the text specified by the user.

Bugzilla/Markdown/GFM.pm
```diff
@@ -0,0 +1,81 @@
+package Bugzilla::Markdown::GFM;
```
purelogiq: Ignore these markdown files in this PR

purelogiq:
Updated with some prelimary frontend stuff

Bugzilla/Template.pm
```diff
@@ -260,6 +264,132 @@ sub quoteUrls {
     return $text;
 }
 
+sub quoteUrls2 {
```
purelogiq: Most of this is left the same as the original `quoteUrls`. Just mainly change the bug link/attachment link functions to the new ones and commented out the autolinking code.

extensions/BugModal/web/bug_modal.js
```diff
@@ -1328,6 +1328,131 @@ $(function() {
...
+        const getHrefIdParam = anchor => (new URL(anchor.href)).searchParams.get("id");
+
+        const filterUnique = (value, index, array) => value && array.indexOf(value) === index;
+
+        const findLinkElements = pathname => {
+            return (
+                Array
+                .from(document.querySelectorAll('.comment-text a'))
+                .filter(anchor => {
+                    // TODO better way to get current bmo instance other than window.location?
```
purelogiq: 🤔

extensions/BugModal/web/bug_modal.js
```diff
@@ -1328,6 +1328,131 @@ $(function() {
...
+            })
+            .then(responseData => {
+                if(responseData['bugs'][BUGZILLA.bug_id]) {
+                    responseData['bugs'][BUGZILLA.bug_id].forEach(attachment => {
+                        attachmentData[attachment['id']] = attachment
+                    });
+                }
+            })
+            .then(() => {
+                // Individually fetch attachment data for each attachment id that wasn't found above.
```
purelogiq: Found this out the hard way. This code here tries to limit the requests to bmo by querying by the current bug first (which returns all of that bug's attachments at once) which I imagine would cover most of the attachment links on a bug page. For those other attachment links, we have to query them one by one to see if the user has permission to view it :/

extensions/BugModal/web/bug_modal.js
```diff
@@ -1328,6 +1328,131 @@ $(function() {
...
+                attachmentLinks.forEach(attachmentLink => {
+                    let data = attachmentData[attachmentLink.id];
+                    if(data) {
+                        attachmentLink.element.setAttribute("title",  data.summary);
+                        $(attachmentLink.element).tooltip({
+                            position: { my: "left top+8", at: "left bottom", collision: "flipfit" },
+                            show: { effect: 'none' },
+                            hide: { effect: 'none' }
+                        });
+                        // TODO if it's a patch append a link to the diff viewer.
```
purelogiq: Help! The perl get_attachment_links code had access on the backend to check if the diff viewer is enabled. Can I replicate that on the frontend somehow (hmm I guess I could pass in an html attr when generating the link...) OR can we just assume that the diff viewer is always enabled for now?

Bugzilla/Template.pm
```diff
@@ -260,6 +260,129 @@ sub quoteUrls {
...
+    my ($text, $bug, $comment) = @_;
+    return $text unless $text;
+    my $anon_user = new Bugzilla::User;
```
purelogiq: Instead of forking get_bug_link and get_attachment_link, I pass an anonymous user to those methods. This will make this quoteUrls2 consistent for caching while at the same time preventing duplication of those methods.

Bugzilla/Template.pm
```diff
@@ -260,6 +260,129 @@ sub quoteUrls {
...
+    }
+
+    # Provide tooltips for full bug links (Bug 74355)
+    my $urlbase_re = '(' . quotemeta(Bugzilla->localconfig->{urlbase}) . ')';
+    $text =~ s~\b(${urlbase_re}\Qshow_bug.cgi?id=\E([0-9]+)(\#c([0-9]+))?)\b
+              ~($things[$count++] = get_bug_link->($3, $1, { comment_num => $5, user => $anon_user })) &&
+               ("\x{FDD2}" . ($count-1) . "\x{FDD3}")
+              ~egox;
+
+    # In the future we want to disable this in favor of Markdown's autolink.
```
purelogiq: Notice comment here :)

extensions/BugModal/web/bug_modal.js
```diff
@@ -1328,12 +1328,161 @@ $(function() {
...
+        const getHrefIdParam = anchor => (new URL(anchor.href)).searchParams.get("id");
+        const filterUnique = (value, index, array) => value && array.indexOf(value) === index;
+
+        const findLinkElements = pathname => {
+            return (
+                Array
+                .from(document.querySelectorAll('.comment-text a'))
+                .filter(anchor => {
+                    return (
+                        anchor.hostname === window.location.hostname &&
```
purelogiq: Is there a better way to get the current bmo instance hostname?

Bugzilla/Template.pm
dylanwh: The indirect object syntax is a style that isn't recommended for new code,
so ``Bugzilla::User->new` is preferred.

are you sure this is a good way forward? I guess this means non-js clients at least get some tooltips for public bugs... and it doesn't hurt cache-ability.

extensions/BugModal/web/bug_modal.js
dylanwh: Look at js_BUGZILLA in template/en/default/global/header.html.tmpl,
you can add Bugzilla->localconfig->{urlbase} to that structure.

Bugzilla/Template.pm
```diff
@@ -260,6 +260,129 @@ sub quoteUrls {
...
+    $text =~ s~\b($safe_protocols)
+              ~($tmp = html_quote($1)) &&
+               ($things[$count++] = "<a rel=\"nofollow\" href=\"$tmp\">$tmp</a>") &&
+               ("\x{FDD2}" . ($count-1) . "\x{FDD3}")
+              ~egox;
+
+    # We have to quote now, otherwise the html itself is escaped
+    # THIS MEANS THAT A LITERAL ", <, >, ' MUST BE ESCAPED FOR A MATCH
+
+    $text = html_quote($text);
```
dylanwh: HTML::Escape (which we already depend on) might be a nice change here.
it's decently fast: https://metacpan.org/pod/HTML::Escape#BENCHMARK

kyoshino:
This API bug is probably a headache: [Bug 1169040 - Cannot retrieve multiple bugs/users when one or more entries are private or not found](https://bugzilla.mozilla.org/show_bug.cgi?id=1169040)

purelogiq:
Closed in favor of https://github.com/mozilla-bteam/bmo/pull/604

PR #577, opened 2018-05-15
Fix 935517 - allow a user to supply a mozillians link

gideon425:


gideon425:
I see that this has been closed, perhaps due to it being considered not important.

Is there a way to get it included?  If not, is there a way to determine what issue
to work on that might be considered important?

PR #579, opened 2018-05-18
Bug 1462685 - Use Phabricators Draft functionality to allow sending of initial revision email after BMO has updated the policies

dklawren:
To test this you will need to perform some administrative steps as admin on the phabricator instance.

1. More Applications -> Harbormaster -> Manage Build Plans -> Create Build Plan
2. Description: "Wait Until Bugzilla Updates Revision"
3. Create
4. Add Build Step
5. Make HTTP Request
6. URI: "http://bmo.test/rest/phabbugz/build_target/${buildable.revision}/${target.phid}"
7. HTTP Method: "POST"
8. Create New Credential
9. Name: "Phabricator Bot Credential"
10. Login/Username: "phab-bot@bmo.tld"
11. Password: <phab-bot@bmo.tld api key from bmo.test>
12. Create Credential
13. When Complete: "Wait for message"
14. Phabricator -> Herald -> Create Herald Rule
15. New Rule for: "Differential Revisions" and Continue
16. Rule Type: "Global" and Continue
17. Rule Name: "Wait Until Bugzilla Updates Revision"
18. For conditions: When "all of" these conditions are met: "is newly created" "is true"
19. For actions: "Run build plans" "Wait Until Bugzilla Updates Revision"
20. Save Rule
21. You will need to enable the phabricator.show-prototypes config value to true. This will need ot be done from a shell of the running phabricator container.
'/app/phabricator/bin/config set phabricator.show-prototypes true'

To test:

1, Create a test private bug at http://bmo.test and record the bug id
2. In Phabricator create a new revision either through the UI or with arc using the bug id.
3. When the revision is created it should be in a "Draft" state shown by the Draft icon at the top left next to Custom Policy.
4, The "Wait Until Bugzilla Updates Revision" should show as "Building".
5. Once Bugzilla updates the revision with new policies, Bugzilla will submit a "pass" status for the running build.
6. Once that is complete the revision should enter the "Needs Review" state and email should be delivered normally.

### DIFF
--- a/extensions/PhabBugz/lib/WebService.pm
+++ b/extensions/PhabBugz/lib/WebService.pm
@@ -20,7 +20,7 @@ use Bugzilla::Constants;
 use Bugzilla::Error;
 use Bugzilla::Extension::Push::Util qw(is_public);
 use Bugzilla::User;
-use Bugzilla::Util qw(detaint_natural datetime_from time_ago);
+use Bugzilla::Util qw(detaint_natural datetime_from time_ago trick_taint);
 use Bugzilla::WebService::Constants;
 
 use Bugzilla::Extension::PhabBugz::Constants;
@@ -54,6 +54,7 @@ use constant PUBLIC_METHODS => qw(
     revision
     update_reviewer_statuses
     needs_review
+    set_build_target
 );
 
 sub revision {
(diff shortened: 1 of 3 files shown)


extensions/PhabBugz/lib/WebService.pm
```diff
@@ -20,7 +20,7 @@ use Bugzilla::Constants;
 use Bugzilla::Error;
 use Bugzilla::Extension::Push::Util qw(is_public);
 use Bugzilla::User;
-use Bugzilla::Util qw(detaint_natural datetime_from time_ago);
+use Bugzilla::Util qw(detaint_natural datetime_from time_ago trick_taint);
```
dylanwh: nit: `use Taint::Util qw(untaint);`  in new code
directly. It's faster.

extensions/PhabBugz/lib/WebService.pm
```diff
@@ -398,8 +399,64 @@ sub _phabricator_precheck {
...
+    # so we exploit that function to pass in an API key as the password
+    # of basic auth. BMO does not support basic auth but does support
+    # use of API keys.
+    my $http_auth = Bugzilla->cgi->http('Authorization');
+    $http_auth =~ s/^Basic\s+//;
+    $http_auth = decode_base64($http_auth);
+    my ($login, $api_key) = split(':', $http_auth);
+    $params->{'Bugzilla_login'}   = $login;
+    $params->{'Bugzilla_api_key'} = $api_key;
+
```
dylanwh: I thought we were already doing this higher in the auth stack?

dylanwh:
@smacleod 's first point is irrelevant; if bmo is down is a major event and a disruption of a majority of systems that touch the firefox repos. However we should discuss making sure this is a robust transaction.

extensions/PhabBugz/lib/WebService.pm
dklawren: Not that I can find. We used to do it for another PhabBugz API method revision() but that is no longer being used.

dklawren:
@smacleod 

-     What happens if bugzilla goes down?

Same thing that happens now if BMO is down. The revision will stay in the secure state until BMO comes back up and Feed daemon catches backup.

-     If an intermittent network failure causes the herald rule web request to fail will that revision just be stuck?

 If the build plan fails to store the build target in the BMO DB, the build plan will fail and the revision will stay in Draft mode. The author, or an admin, can restart the build plan as many times as necessary to finally get through to BMO.

-     It seems like there might be a race here, where we could send the hearld rule update before the feed story for the revision is actually handled. If the story is created on the feed while we're already processing stories, we may just hit the build plan after.

Well if this happens, it really doesn't matter since even if the revision is looked at more than once, the same changes do not happen more than once. But since Draft revisions do not show up in the Phabricator Feed, BMO will only process the build targets it knows about from the API call to BMO from the build plan. When the Draft revision exits Draft mode, the next feed story will be that the author has requested review and not that the revision has been created as it happens today.
- 

I'm not sure we want Phabricator actually making a real web request in the build plan. Maybe it should just make a local web request, or can we have a wait condition without the web request. There's a conduit API for posting a story into the feed, maybe we should just post a "Draft waiting for bugzilla" story or something and have the normal feed code pick it up and react?

dklawren:
- I'm not sure we want Phabricator actually making a real web request in the build plan. Maybe it should just make a local web request, or can we have a wait condition without the web request. There's a conduit API for posting a story into the feed, maybe we should just post a "Draft waiting for bugzilla" story or something and have the normal feed code pick it up and react?

I am not sure we have have a wait state without some sort of HTTP request from the way the UI is setup but I can look at that some more. Injecting a new feed story is also something I can look at.

dklawren:
@smacleod 
I have asked upstream if I am missing something but there seems to be no way to provide a JSON data structure to the conduit API from a build step in Harbormaster. Right now it seems like you can only make simple calls using variables in the URL query string. So we would have to write a doorkeeper extension to add the feed story when a new revision is created. I am not even sure if Doorkeeper would even see an event since the draft revision send out zero notifications until accepted for review.

dklawren:
@smacleod said this was OK to merge with a better fix to be done at a later time.

PR #585, opened 2018-05-23
Bug 1461379 - use DBIx::Connector to manage database connections

dylanwh:
This should make it so there is only two connections to the DB per process.

This is related to mozilla/infosec-risk-management-bugzilla#13


dylanwh:
@claudijd can you summarize which API (rest path) is triggering the mysql connection errror?
I suspect this patch will fix it, but I need to understand how many times / how fast we're connecting here.

PR #586, opened 2018-05-23
Bug 1461379 - refactor Bugzilla::DB to not subclass DBI

dylanwh:
I'd like to use DBIx::Connector (in #585), so first let's remove the really weird stuff we do about subclassing DBI::db.

### DIFF
--- a/Bugzilla/DB.pm
+++ b/Bugzilla/DB.pm
@@ -8,13 +8,21 @@
 package Bugzilla::DB;
 
 use 5.10.1;
-use strict;
-use warnings;
+use Moo;
 
 use DBI;
 
-# Inherit the DB class from DBI::db.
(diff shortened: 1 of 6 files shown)


Bugzilla/DB.pm
```diff
@@ -1274,20 +1280,24 @@ sub db_new {
             $attributes->{$key} = $override_attrs->{$key};
         }
     }
+    my $class = ref $self;
+    if ($class->can('on_dbi_connected')) {
+        $attributes->{Callbacks} = {
+            connected => sub { $class->on_dbi_connected(@_); return },
```
globau: nice

dylanwh:
I just want to make sure.. @globau and @dklawren, you're comfortable with this change??

I guess it has been on dev for a while now...

PR #588, opened 2018-05-25
WIP - script to undo bug history for inactive bug changes

dylanwh:


dylanwh:
Okay. @dklawren, I'm going to email you credentials to the DB I'm testing this against.
I'm very interested in if there are parts of the database this is not updating. 

The original INACTIVE changes were done by a modified version of `scripts/resolve_bugs.pl`
and the goal of this script is to make it so it looks like `scripts/resolve_bugs.pl` was never run, *unless* the bug was updated since `scripts/resolve_bugs.pl` was run.

scripts/undo.pl
```diff
@@ -0,0 +1,186 @@
...
+}
+
+sub get_changes {
+    my ($where, @bind) = @_;
+
+    my $sql = qq{
+        SELECT
+            BA.id AS change_id,
+            BA.bug_id,
+            FD.name field_name,
```
dklawren: nit: 
FD.name AS field_name

scripts/undo.pl
```diff
@@ -0,0 +1,186 @@
...
+            die "cannot find previous last updated time" unless $previous_last_ts;
+            my $action = delete $action{$bug_id}{$delta_ts};
+            if (keys %{ $action{$bug_id}{$delta_ts}}) {
+                die "skipping because more than one change\n";
+            }
+            elsif (!$action) {
+                die "skipping because most recent change newer than automation change\n";
+            }
+            $action->{change}{delta_ts} = { replace => $delta_ts, with => $previous_last_ts };
+            $action->{change}{lastdiffed} = { replace => $delta_ts, with => $previous_last_ts };
```
dklawren: As you also mentioned in IRC, these need to actually be updated in the DB for the bugs table.

dklawren: oh wait nm. i see that they ARE updated.

PR #598, opened 2018-06-05
Bug 1456877 - Add a wrapper around libcmark_gfm to Bugzilla

purelogiq:
Finish up of #569, with improvements. Adds tests, some usage docs, and a helper method in Bugzilla.pm to easily get the markdown parser.


purelogiq:
@dklawren As dylan requested, could you do a look over review of this? I don't think we should merge it yet till dylan's back but would like some quick feedback, thank you :)

PR #602, opened 2018-06-14
Bug 1468818 - Re-introduce is_markdown to the Bugzilla::Comment model

dylanwh:


Bugzilla/Install/DB.pm
```diff
@@ -710,6 +710,9 @@ sub update_table_definitions {
     # 2014-07-27 LpSolit@gmail.com - Bug 1044561
     _fix_user_api_keys_indexes();
 
+    $dbh->bz_add_column('longdescs', 'is_markdown',
```
purelogiq: Looks like most of the changes in this file have a comment referencing the bug, should we do that here too?

Bugzilla/Comment.pm
```diff
@@ -45,6 +45,7 @@ use constant DB_COLUMNS => qw(
     already_wrapped
     type
     extra_data
+    is_markdown
```
purelogiq: Does BMO still keep up with POD docs? If so we can add some docs for the `is_markdown` field though it is self descriptive :)

PR #603, opened 2018-06-14
Bug 1468848 - Change presentation of show_bug.cgi urls to be /bug/ID or /bug/ALIAS.

dylanwh:


extensions/BMO/template/en/default/hook/global/header-additional_header.html.tmpl
```diff
@@ -21,7 +21,7 @@
 
 <link rel="shortcut icon" href="extensions/BMO/web/images/favicon.ico">
 [% IF bug %]
-<link id="shorturl" rev="canonical" href="https://bugzil.la/[% bug.bug_id FILTER uri %]">
+<link id="shorturl" rev="canonical" href="[% Bugzilla.localconfig.urlbase FILTER none %]bug/[% bug.bug_id FILTER uri %]">
```
kyoshino: I should have flagged this earlier but looks like the [canonical URL](https://support.google.com/webmasters/answer/139066?hl=en) has to be `rel="canonical"`, not `rev`. 

Also, the base URL has to always be the production URL even on the dev environment, otherwise URLs on [bugzilla-dev.allizom.org](https://bugzilla-dev.allizom.org/) will show up in Google search results. But it requires another field in the config?

kyoshino: Actually [bugzilla-dev.allizom.org](https://www.google.com/search?q=site%3Abugzilla-dev.allizom.org), [bugzilla.allizom.org](https://www.google.com/search?q=site%3Abugzilla.allizom.org) and other non-canonical URLs have already been cached by Google.

kyoshino:
Looks like many links will be broken because they are relative URLs like `<a href="script.cgi">`. We have to rewrite everything to use absolute URLs like  `<a href="/script.cgi">`. Hmm. 😕

extensions/BMO/template/en/default/hook/global/header-additional_header.html.tmpl
kyoshino: The Canonical URL can be found only on `show_bug.cgi` but it has to be everywhere. It will go beyond the scope of this bug though.

kyoshino:
Another issue: A comment number in the URL like `show_bug.cgi?id=1#2` will be lost while rewriting.

PR #606, opened 2018-06-16
Bug 1457900 - When restricting a revision to a bugzilla group we should tag the revision with the project

dklawren:


dklawren:
Conflicts resolved.

dylanwh:
Sorry, I somehow edited your comment instead of commenting. Weird UX. ;)

to make sure I'm following: `Revsion->projects` is gone (replaced by projects_raw) and the actual list of the (new) project objects is `Revision->rule_projects`?

And there are no more cases of `$revsion->projects` in the code.

PR #613, opened 2018-06-20
Bug 1469911 - Make user autocompletion faster

dylanwh:
This optimizes the common case of username lookup to be fast at the expense of slightly different behavior.

For bare words consisting of ASCII characters, a prefix match against login_name and the new nickname field is attempted.
For words containing @, only the login_name is considered.
For words beginning with :, only the nickname is considered.

Multiple words, or non-ASCII characters will use a fulltext search against the realname field.
This means 'Mark Cote' will correctly find @markrcote. 

The time savings of this on bugzilla-dev appears to be about 600ms

Prod:
screen shot 2018-07-03 at 10 41 57

Dev:
screen shot 2018-07-03 at 10 41 46

### DIFF
--- a/Bugzilla/DB.pm
+++ b/Bugzilla/DB.pm
@@ -350,6 +350,11 @@ sub import {
     $Exporter::ExportLevel-- if $is_exporter;
 }
 
+sub sql_prefix_match {
+    my ($self, $column, $str) = @_;
+    return "$column LIKE " . $self->quote("$str%");
+}
+
 sub sql_istrcmp {
     my ($self, $left, $right, $op) = @_;
     $op ||= "=";
--- a/Bugzilla/DB.pm
+++ b/Bugzilla/DB.pm
@@ -353,7 +353,7 @@ sub import {
 sub sql_prefix_match {
     my ($self, $column, $str) = @_;
     my $must_escape = $str =~ s/([_%!])/!$1/g;
-    my $escape      = $must_escape ? q/ESCAPE '!'/ : '');
+    my $escape      = $must_escape ? q/ESCAPE '!'/ : '';
     my $quoted_str  = $self->quote("$str%");
     return "$column LIKE $quoted_str $escape";
 }
--- a/Bugzilla/DB.pm
+++ b/Bugzilla/DB.pm
@@ -353,9 +353,9 @@ sub import {
 sub sql_prefix_match {
     my ($self, $column, $str) = @_;
     my $must_escape = $str =~ s/([_%!])/!$1/g;
-    my $escape      = $must_escape ? q/ESCAPE '!'/ : q//);
-    my $quoted_str  = $self->quote(qq/$str%/)
-    return qq/$column LIKE $quoted_str $escape/;
+    my $escape      = $must_escape ? q/ESCAPE '!'/ : '');
+    my $quoted_str  = $self->quote("$str%");
+    return "$column LIKE $quoted_str $escape";
 }
 
 sub sql_istrcmp {
--- a/Bugzilla/DB.pm
+++ b/Bugzilla/DB.pm
@@ -352,8 +352,10 @@ sub import {
 
 sub sql_prefix_match {
     my ($self, $column, $str) = @_;
-    my $escape = $str =~ s/([_%!])/!$1/gs;
-    return "$column LIKE " . $self->quote("$str%") . ( $escape ? " ESCAPE '!'" : '');
+    my $must_escape = $str =~ s/([_%!])/!$1/g;
+    my $escape      = $must_escape ? q/ESCAPE '!'/ : q//);
+    my $quoted_str  = $self->quote(qq/$str%/)
+    return qq/$column LIKE $quoted_str $escape/;
 }
 
 sub sql_istrcmp {
--- a/Bugzilla/DB.pm
+++ b/Bugzilla/DB.pm
@@ -352,7 +352,8 @@ sub import {
 
 sub sql_prefix_match {
     my ($self, $column, $str) = @_;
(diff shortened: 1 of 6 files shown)


Bugzilla/User.pm
```diff
@@ -726,12 +745,8 @@ sub nick {
...
 
     return "" unless $self->id;
-
-    if (!defined $self->{nick}) {
-        $self->{nick} = (split(/@/, $self->login, 2))[0];
-    }
-
-    return $self->{nick};
+    return $self->{nickname} if $self->{nickname};
+    return $self->{nick} //= (split(/@/, $self->login, 2))[0];
```
dylanwh: For right now, we use the old system.
We're not setting the nickname field to this because it doesn't make sense for ':dkl' to match dkl@mozilla.com if ':dkl' is not in the rename. Also it means the _migrate_nicknames() can complete fast, as only about 3k users use ircnicks like this.

Bugzilla/User.pm
```diff
@@ -478,10 +482,25 @@ sub set_login {
...
+    return $nick;
+}
+
 sub set_name {
     my ($self, $name) = @_;
     $self->set('realname', $name);
     delete $self->{identity};
+    $self->set('nickname', _generate_nickname($name, $self->login));
+}
+
```
dylanwh: 'set_nick' because nick() is the accessor (with some historical-enforced caveats).
Future work might be making the column names and method names *sane* but that it is outside the scope of this enhancement.

Bugzilla/DB.pm
```diff
@@ -350,6 +350,11 @@ sub import {
     $Exporter::ExportLevel-- if $is_exporter;
 }
 
+sub sql_prefix_match {
+    my ($self, $column, $str) = @_;
+    return "$column LIKE " . $self->quote("$str%");
```
globau: this won't escape `LIKE` wildcards in `$str`: `\`, `%`, and `_`.
`_` in particular as there's nicks that contain underscores.

note escaping is different between mysql and sqlite (and possibly other dbs).
here's what i do for sqlite (where the calling code uses placeholders):
```
sub like_value {
    my ($field, $value) = @_;
    my $condition = $field . ' LIKE ?';
    if ($value =~ s/(?=[\\%_])/\\/g) {
        $condition .= ' ESCAPE \'\\\'';
    }
    return ($condition, '%' . $value . '%');
}
```

Bugzilla/User.pm
```diff
@@ -22,6 +22,7 @@ use Bugzilla::Field;
 use Bugzilla::Group;
 use Bugzilla::Hook;
 use Bugzilla::BugUserLastVisit;
+use Bugzilla::Logging;
```
globau: logging doesn't appear to be used here

Bugzilla/WebService/User.pm
```diff
@@ -135,6 +137,61 @@ sub create {
...
+
+    untaint($params->{match});
+    my $s = $params->{match};
+    trim($s);
+
+    my $dbh = Bugzilla->dbh;
+    my @select = ('realname AS real_name', 'login_name AS name');
+    my $order  = 'last_seen_date DESC';
+    my $where;
+    state $have_mysql = $dbh->isa('Bugzilla::DB::Mysql');
```
globau: it's a bit weird to have db-specific paths outside of the `Bugzilla::DB` / `Bugzilla::DB::Mysql` classing

Bugzilla/WebService/User.pm
```diff
@@ -135,6 +137,61 @@ sub create {
...
+
+    ThrowCodeError('params_required', { function => 'User.suggest_users', params => ['match'] })
+      unless defined $params->{match};
+
+    ThrowUserError('user_access_by_match_denied')
+      unless Bugzilla->user->id;
+
+    untaint($params->{match});
+    my $s = $params->{match};
+    trim($s);
```
globau: probably makes sense to return early if `$s` is an empty string; otherwise it'll all users.

Bugzilla/WebService/User.pm
```diff
@@ -135,6 +137,61 @@ sub create {
...
+        else {
+            $where = join(' OR ',
+                $dbh->sql_prefix_match(nickname => $s),
+                $dbh->sql_prefix_match(login_name => $s),
+            );
+        }
+    }
+    $where = "($where) AND is_enabled = 1";
+
+    my $sql = "SELECT " . join(", ", @select) . " FROM profiles WHERE $where ORDER BY $order";
```
globau: a `LIMIT` should be placed on these results

Bugzilla/WebService/User.pm
dylanwh: I'm going to accept this weirdness for now, because I don't feel like looking up the postgres version just yet (but I'd rather not add to the non-mysql sql that someone will need to fix in harmony)

Bugzilla/WebService/User.pm
dylanwh: woops, yeah, no more than 25.

Bugzilla/WebService/User.pm
```diff
@@ -135,6 +137,62 @@ sub create {
...
+        else {
+            $where = join(' OR ',
+                $dbh->sql_prefix_match(nickname => $s),
+                $dbh->sql_prefix_match(login_name => $s),
+            );
+        }
+    }
+    $where = "($where) AND is_enabled = 1";
+
+    my $sql = "SELECT " . join(", ", @select) . " FROM profiles WHERE $where ORDER BY $order LIMIT 25";
```
globau: if you want to keep oracle happy you'll need to use `$dbh->sql_limit()` instead of `LIMIT`

PR #616, opened 2018-06-21
Bug 1469881 - Patches posted by Phabricator to Bugzilla don't list the patch author

dklawren:


extensions/PhabBugz/lib/Util.pm
```diff
@@ -59,6 +59,14 @@ sub create_revision_attachment {
         ($timestamp) = Bugzilla->dbh->selectrow_array("SELECT NOW()");
     }
 
+    # If submitter, then switch to that user when creating attachment
+    my $old_user;
```
dylanwh:
```perl
try {

File: extensions/PhabBugz/lib/Util.pm [hunk 2]
```diff
@@ -76,6 +84,8 @@ sub create_revision_attachment {
     $bug->add_comment($revision->summary, { type       => CMT_ATTACHMENT_CREATED,
                                             extra_data => $attachment->id });
```
dylanwh:
```perl
} catch {
   die $_;
} finally {
```

extensions/PhabBugz/lib/Util.pm
```diff
@@ -76,6 +84,8 @@ sub create_revision_attachment {
...
 
+    Bugzilla->set_user($old_user) if $old_user;
+
```
dylanwh:
```perl
};
```

dklawren:
It will be tested by phabricator QA process. Yes.

PR #618, opened 2018-06-22
Bug 1470485 - Create new policies using PhabricatorProjectsAllPolicyRule instead of PhabricatorProjectsPolicyRule

dklawren:
…all list projects when setting up security policies for revisions

### DIFF
--- a/extensions/PhabBugz/lib/Policy.pm
+++ b/extensions/PhabBugz/lib/Policy.pm
@@ -116,7 +116,7 @@ sub create {
 
         push @{ $data->{policy} }, {
             action => 'allow',
-            rule   => 'PhabricatorProjectsPolicyRule',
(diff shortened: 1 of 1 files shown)


extensions/PhabBugz/lib/Policy.pm
```diff
@@ -135,7 +135,7 @@ sub _build_rule_projects {
     my ($self) = @_;
 
     return [] unless $self->rules;
-    my $rule = first { $_->{rule} eq 'PhabricatorProjectsPolicyRule'} @{ $self->rules };
+    my $rule = first { $_->{rule} =~ /PhabricatorProjects(All)?PolicyRule/ } @{ $self->rules };
```
dylanwh: nit: (?:All)? since you're capturing the value.

PR #621, opened 2018-06-25
Bug 1456878 - Support markdown comments

purelogiq:
### Each commit has a pretty-ok commit message to help understand what changed in that commit!

This PR adds support for markdown comments.

# Caveats! Please see!
- Midair collision page renders the inbound comment as plaintext.
- **FIXED** ~~User story field renders plaintext (can change if we want to, I just don't know how this field is used in practice).~~
- Markdown links don't have the 'nofollow' ref on them.
- **FIXED** ~~`> blockquotes` have the `>` even on a continued line (e.g. it'll look like `> line 1 > line 1 continued`). This is a bug caused by calling `html_quote` before `render_markdown` but changing that may mean changing the regex which I **really** don't want to do...thoughts?~~
- **FIXED** ~~At the moment it is only missing a little banner to inform people that all comments are now Markdown.~~

### DIFF
--- a/Bugzilla/Template.pm
+++ b/Bugzilla/Template.pm
@@ -282,11 +282,17 @@ sub get_attachment_link {
     if ($attachment) {
         my $title = "";
         my $className = "";
+        my $linkClass = "";
+
         if ($user->can_see_bug($attachment->bug_id)
             && (!$attachment->isprivate || $user->is_insider))
         {
             $title = $attachment->description;
         }
+        else{
+            $linkClass = "bz_private_link";
+        }
+
         if ($attachment->isobsolete) {
             $className = "bz_obsolete";
         }
@@ -307,7 +313,7 @@ sub get_attachment_link {
 
         # Whitespace matters here because these links are in <pre> tags.
         return qq|<span class="$className">|
-               . qq|<a href="${linkval}" name="attach_${attachid}" title="$title">$link_text</a>|
+               . qq|<a href="${linkval}" class="$linkClass" name="attach_${attachid}" title="$title">$link_text</a>|
                . qq| <a href="${linkval}&amp;action=edit" title="$title">[details]</a>|
                . qq|${patchlink}|
                . qq|</span>|;
--- a/Bugzilla/Template.pm
+++ b/Bugzilla/Template.pm
@@ -130,17 +130,19 @@ sub get_format {
     };
 }
 
-# This routine quoteUrls contains inspirations from the HTML::FromText CPAN
+# This routine renderComment contains inspirations from the HTML::FromText CPAN
 # module by Gareth Rees <garethr@cre.canon.co.uk>.  It has been heavily hacked,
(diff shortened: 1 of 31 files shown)


purelogiq:
Video: https://drive.google.com/file/d/14ug9xXUGn8ks3eP42mlXhl9fwy8_Ok6r/view?usp=sharing

Bugzilla/Template.pm
```diff
@@ -130,17 +130,19 @@ sub get_format {
...
 # module by Gareth Rees <garethr@cre.canon.co.uk>.  It has been heavily hacked,
 # all that is really recognizable from the original is bits of the regular
 # expressions.
 # This has been rewritten to be faster, mainly by substituting 'as we go'.
 # If you want to modify this routine, read the comments carefully
+# Renamed from 'quoteUrls' to 'renderComment' after markdown support was added.
 
-sub quoteUrls {
-    my ($text, $bug, $comment, $user, $bug_link_func) = @_;
+sub renderComment {
```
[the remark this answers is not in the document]
dylanwh: Set $skip_markdown to 1 if !$comment->is_markdown, and simplify the other if statements.

purelogiq:
# Pictures!

### New and improved blockquotes
double-block-quote


### Markdown is here notice
I went with a minimal approach. I think this is something we can just leave up. I've linked to the github docs since we use GFM.

On the show bug page
markdown-tip-comment

On the new bug page (same for new attachment)
markdown-tip-new-bug

template/en/default/email/bugmail.html.tmpl
```diff
@@ -38,7 +38,7 @@
               on [% "$terms.bug $bug.id" FILTER bug_link(bug, { full_url => 1, user => to_user }) FILTER none %]
               from [% INCLUDE global/user.html.tmpl user = to_user, who = comment.author %]</b>
           [% END %]
-        <pre>[% comment.body_full({ wrap => 1 }) FILTER quoteUrls(bug, comment, to_user) %]</pre>
+        <pre>[% comment.body_full({ wrap => 1 }) FILTER renderComment(bug, comment) %]</pre>
```
dklawren: Don't we want markdown to render in HTML emails? The &lt;pre> tag would not allow this.

purelogiq: Thanks for catching this! I fixed it in the `extensions/BMO/template/en/default/email/bugmail.html.tmpl` but forgot to fix it here.

purelogiq:
Looks better!

## Before (mono space, low line height)
before-font

## After (san serif, default line height)
even-after

purelogiq:
even-even-after

purelogiq:
@kyoshino True true, just wanted to get this out before the push to -dev :)

dylanwh:
@kyoshino, @purelogiq I'm going to be doing some last-ditch testing but I think with three approvals this is good. Communication around this feature should be done so as to not surprise people too much.

PR #605, opened 2018-06-15
Bug 1469023 - Show "new changes since (datetime)" indicator that links to unread changes/comments

kyoshino:
## Description

Add a "new changes" indicator to the modal bug page, and automatically scroll down to those changes for faster browsing, as seen on Slack.

## Bug 

[Bug 1469023 - Show "new changes since (datetime)" indicator that links to unread changes/comments](https://bugzilla.mozilla.org/show_bug.cgi?id=1469023)

### link above the header

screen shot 2018-06-16 at 10 03 38

### separator above the new change(s)

screen shot 2018-06-16 at 10 04 03

PR #638, opened 2018-06-28
Bug 1471966 - Blue "new changes since" bar disappears too quickly

kyoshino:
Regressed from #620 even before #605 is live 😞 

The `IntersectionObserver`'s root should be the scroll container, and it's `<main id="bugzilla-body">` in this case. The `offsetParent` property could change when CSS is updated, so just specify the exact element instead.


kyoshino:
Sorry it's *my* mistake. I just have to make code more robust.

PR #646, opened 2018-07-04
Bug 1472954 - Implement one-click component watching on bug modal and component description pages

kyoshino:
## Description

* Implement REST API methods for the Component Watching extension
* Add Watch/Unwatch buttons to the bug page’s Product and Component fields

## Bug 

[Bug 1472954 - Implement one-click component watching on bug modal](https://bugzilla.mozilla.org/show_bug.cgi?id=1472954)

## Screenshot

https://twitter.com/BugzillaUX/status/1014225230636593152


kyoshino:
I’ll make changes to the component description pages as well, but the Perl bits here are ready for review.

extensions/ComponentWatching/Extension.pm
```diff
@@ -498,6 +501,10 @@ sub _getWatches {
         push @watches, \%watch;
     }
 
+    if ($watch_id) {
+        return @watches->[0] || {};
```
dklawren: Using an array as a reference is deprecated at /app/extensions/ComponentWatching/Extension.pm line 505.

return $watches[0] || {};

extensions/ComponentWatching/Extension.pm
```diff
@@ -467,15 +467,18 @@ sub bugmail_relationships {
 #
 
 sub _getWatches {
-    my ($user) = @_;
+    my ($user, $watch_id) = @_;
     my $dbh = Bugzilla->dbh;
 
+    $watch_id = ($watch_id =~ /^(\d+)$/) ? $1 : undef;
```
dklawren: Also missed this error in the log:

Use of uninitialized value $watch_id in pattern match (m//) at /app/extensions/ComponentWatching/Extension.pm line 473.

Change to:

$watch_id = (defined $watch_id && $watch_id =~ /^(\d+)$/) ? $1 : undef;

kyoshino: @dklawren Thanks! Will fix them and finish the frontend stuff later this week.

kyoshino:
The frontend implementation is also complete. Watch/Unwatch buttons are now available on component description pages like this:

screen shot 2018-07-17 at 00 07 29

kyoshino:
Frontend highlights:

* Replace the BMO extension's component description page override with a simple hook, because the essential difference is only whether the default assignees and QA contacts are displayed
* Redesign the non-extension component description page with the table layout removed
* Move the `bugzilla_ajax` function from `bug_modal.js` to `global.js` temporarily, which will be refactored soon in [Bug 1380437](https://bugzilla.mozilla.org/show_bug.cgi?id=1380437)
* Implement Google Analytics event tracking for the Watch/Unwatch buttons

dylanwh:
@kyoshino So what would you say about the thought this makes the show bug page more noisy?
@purelogiq too. If you've both considered that and think it doesn't add too much visual noise, r+ from me.

purelogiq:
@dylanwh I think you should get some users' opinions on the slack channels :)

kyoshino:
UI-wise, I don’t like it for sure. My alternative plan (from [BzDeck](https://github.com/bzdeck/bzdeck/issues/80)) is adding a popup like this. I’ll check if this can be quickly implemented.

screen shot 2018-07-17 at 12 50 18 pm

PR #648, opened 2018-07-05
Bug 1473726 - WebExtensions bugs missing crash reports when viewed

dylanwh:
Also includes removing at least two unused products.


dylanwh:
@globau can probably speak to why %cf_visible_in_products exists.

globau:
> Why is this needed? Is there a bug in the custom field configuration screen in admin view?

originally tracking flags were actual bugzilla custom fields (`cf_` prefixed).  @dklawren made a herculean effort to change how these flags were stored while still presenting them as normal custom fields to the rest of the bugzilla code.

my guess is `cf_visible_in_products` is either something that was missed, or was deemed too complicated to do within the limited time constraints, and it doesn't occur often enough to justify the engineering effort to address it.

PR #649, opened 2018-07-06
Bug 1473798 - Add and remove members of a phabricator project instead of setting exact list

dklawren:


extensions/PhabBugz/lib/Feed.pm
```diff
@@ -323,10 +323,20 @@ sub group_query {
...
+        my @current_member_phids = uniq map { $_->phid } @{ $project->members };
+        my ($added, $removed)    = diff_arrays(\@set_member_phids, \@current_member_phids);
+
+        INFO('Adding members: ' . join( ',', @$added ));
+        $project->add_member( $_ ) foreach @$added;
+        INFO('Removing members: ' . join( ',', @$removed ));
+        $project->remove_member( $_  ) foreach @$removed;
+
+        my $result = $project->update();
+        Bugzilla::Logging->fields->{api_result} = $result;
```
dylanwh: nit:
```perl
local Bugzilla::Logging->fields->{api_result} = $result;
INFO("Project " . $project->name . " updated");
```

and then the delete is not needed.

PR #654, opened 2018-07-13
misc changes that I've done for mojo but can land earlier

dylanwh:


globau:
these look sane; ping me on irc when there's a bug.

Bugzilla/Constants.pm
```diff
@@ -647,9 +648,8 @@ sub _bz_locations {
     # Calling dirname twice like that won't work on VMS or AmigaOS
     # but I doubt anybody runs Bugzilla on those.
```
globau: while it pains me to remove a reference to amigaos from bugzilla's source, this comment is no longer relevant.

PR #656, opened 2018-07-13
Bug 1475713 - Change the group that can delete attachments from hardcoded admin to can_delete_attachments

dklawren:


dylanwh:
This is a dupe of #493.

PR #657, opened 2018-07-16
Bug 1476288 - Replace moz_nick with (new, revised) nick and also attempt to disallow duplicate nicks

dylanwh:


Bugzilla/User.pm
```diff
@@ -482,19 +482,23 @@ sub set_login {
...
-    my ($name, $login) = @_;
+    my ($name, $login, $id) = @_;
     my ($nick) = extract_nicks($name);
     if (!$nick) {
         $nick = "";
     }
+    my ($count) = Bugzilla->dbh->selectrow_array('SELECT COUNT(*) FROM profiles WHERE nickname = ? AND userid != ?', undef, $nick, $id);
+    if ($count) {
+        $nick = "";
+    }
```
globau: probably doesn't matter now that we've done the migration, but this will run faster if you skip the db query when `!$id`.

PR #664, opened 2018-07-20
add unit tests for phabbugz

dylanwh:


dylanwh:
With a bit of mocking, I have tests that now fail for this problem.


```
ok 1 - use Bugzilla::Extension::PhabBugz::Feed;
ok 2 - use Bugzilla::Extension::PhabBugz::Util;
ok 3 - Bugzilla::Extension::PhabBugz::Feed->can('group_query')
ok 4 - no-op
ok 5 - remove foo
ok 6 - add foo
ok 7 - it is an array ref
not ok 8 - The class (or class-like) '0' isa 'Bugzilla::Extension::PhabBugz::Revision'
#   Failed test 'The class (or class-like) '0' isa 'Bugzilla::Extension::PhabBugz::Revision''
#   at t/phabbugz.t line 192.
#     The class (or class-like) '0' isn't a 'Bugzilla::Extension::PhabBugz::Revision'
not ok 9 - update revision
#   Failed test 'update revision'
#   at t/phabbugz.t line 193.
1..9
# Looks like you failed 2 tests of 9.
```

dylanwh:
screen shot 2018-07-20 at 18 14 04

Perl critic actually found the bug.
This if should have an else branch that is fatal, and callers should handle the situation where no revision is found.

zalun:
I think the best would be to make the failing ones skip and file bugs to fix them.

dylanwh:
We're going to commit this because it the bug it finds needs to be fixed before the next release.

PR #667, opened 2018-07-23
add skeleton for datadog monitoring

dylanwh:


extensions/Push/lib/Push.pm
```diff
@@ -50,15 +38,52 @@ sub start {
...
+    $pushd_loop->add($main_timer);
+    $main_timer->start;
+    $pushd_loop->run;
+}
+
+sub heartbeat {
+    my ($self) = @_;
+    my $dd = Bugzilla->datadog('pushd');
+
+    $dd->gauge('total', $self->_sql_count('SELECT COUNT(*) FROM push'));
```
dylanwh: Not sure about the naming of the metric. 'total' will get sent to datadog as 'pushd.total' currently.
@mars-f

extensions/Push/lib/Push.pm
```diff
@@ -50,15 +38,52 @@ sub start {
...
+
+    foreach my $connector ($connectors->list) {
+        if ($connector->enabled) {
+            $dd->gauge("${name}_backlog", $self->_sql_count('SELECT COUNT(*) FROM push_backlog WHERE connector = ?', $name);
```
dylanwh: Same here, "push.Phabricator_backlog". Can datadog names contain multiple dogs?

extensions/PhabBugz/lib/Feed.pm
```diff
@@ -44,6 +44,20 @@ has 'is_daemon' => ( is => 'rw', default => 0 );
 sub start {
     my ($self) = @_;
 
+    my $loop = IO::Async::Loop->new;
+    if ( Bugzilla->datadog ) {
+        my $dog_timer = IO::Async::Timer::Periodic->new(
+            interval   => 600,       # ten minutes, FIXME.
+            reschedule => 'drift',
+            on_tick    => sub {
+                # FIXME. Do something with Bugzilla->datadog
```
globau: FIXMEs need to be fixed.

extensions/Push/lib/Push.pm
```diff
@@ -50,15 +38,52 @@ sub start {
...
         }
-        sleep(POLL_INTERVAL_SECONDS);
     }
 }
 
+sub _sql_count {
+    my ($self, $sql, @bind) = @_;
+    my ($count) = Bugzilla->dbh->selectrow_array($sql, undef, @bind);
+    return $count;
```
globau: this seems unnecessary; there's nothing specific to this sub related to "counting", and `selectrow_array` called in a scalar context will return the desired result.

ie. replace
```
$dd->gauge('total', $self->_sql_count('SELECT COUNT(*) FROM push'));
```
with
```
$dd->gauge('total', Bugzilla->dbh->selectrow_array('SELECT COUNT(*) FROM push'));
```

extensions/Push/lib/Push.pm
mars-f: You want your metric names to be as loose hierarchical namespace with each level separated by dots.  The name `total` is a bit vague for people unfamiliar with how the metric is collected.  `pushd.queue_size` would be clearer.

If there are other pushd components besides the queue that you want to measure, or if you want to measure the queue in multiple ways, then you could make the `queue` level explicit with `pushd.queue.size`. (Personally, I prefer dots instead of underscores - I like how they read.)

If your code already prefixes the `bugzilla` namespace for you then the final metric name will be `bugzilla.pushd.queue.size`.

In prod we already have metrics like `bugzilla.jobqueue.total_errors`.

extensions/Push/lib/Push.pm
mars-f: Yes they can have dots, see the comment above.  Our convention uses snake_case.  This metric could be named `pushd.phabricator.backlog`.

In the docs we discussed measuring the age of the first queue item, so you might want to add a level to the namespace to hold both `backlog.size` and `backlog.first_item_time_enqueued`.

dylanwh:
@mars-f  took this over

dylanwh:
This was taken up in #708

PR #672, opened 2018-07-25
Bug 1320977 - Add review/feedback/needinfo request counts and block statuses to /rest/user and /rest/user/suggest responses

kyoshino:
Fix [Bug 1320977 - Add review/feedback/needinfo request counts and block statuses to /rest/user and /rest/user/suggest responses](https://bugzilla.mozilla.org/show_bug.cgi?id=1320977)

## Example

`/rest/user/suggest` will look like this:

```json
{
   "users" : [
      {
         "id" : 1,
         "name" : "vagrant@bmo-web.vm",
         "real_name" : "Vagrant Nakamoto",
         "requests" : {
            "feedback" : {
               "blocked" : false,
               "pending" : 0
            },
            "needinfo" : {
               "blocked" : true,
               "pending" : 0
            },
            "review" : {
               "blocked" : false,
               "pending" : 1
            }
         }
      }
   ]
}
```

### DIFF
--- a/Bugzilla/WebService/User.pm
+++ b/Bugzilla/WebService/User.pm
@@ -154,7 +154,7 @@ sub suggest {
     return { users => [] } if length($s) < 3;
 
     my $dbh = Bugzilla->dbh;
-    my @select = ('realname AS real_name', 'login_name AS name');
+    my @select = ('userid AS id', 'realname AS real_name', 'login_name AS name');
     my $order  = 'last_seen_date DESC';
     my $where;
     state $have_mysql = $dbh->isa('Bugzilla::DB::Mysql');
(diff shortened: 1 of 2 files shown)


dylanwh:
I'll do some benchmarks. I think we can avoid using $webservice->type() for some speed gains.
The only downside is XMLRPC won't work but we don't care about that.

For `$webservice->type('boolean', $val)` do `$val ? JSON::MaybeXS->true : JSON::MaybeXS->false`
For `$webservice->type('int', $val)` do int($val)

dylanwh:
This would have added 8ms:
# spent 8.16ms (158µs+8.00) within Bugzilla::Extension::Review::webservice_user_get

screen shot 2018-07-27 at 12 49 50

But I made it better:

# spent 3.87ms (127µs+3.74) within Bugzilla::Extension::Review::webservice_user_get

screen shot 2018-07-27 at 12 49 33

I'll post those changes shortly.

PR #674, opened 2018-07-26
Bug 1482475 - Add extensive testing framework

dylanwh:
This doesn't test everything I want to -- but it adds the right framework for writing tests more easily.


extensions/PhabBugz/lib/Feed.pm
```diff
@@ -382,7 +382,7 @@ sub process_revision_change {
 
     # Pre setup before making changes
     my $old_user = set_phab_user();
-    my $bug = Bugzilla::Bug->new({ id => $revision->bug_id, cache => 1 });
+    my $bug = $revision->bug;
```
dylanwh: @dklawren Is this a good / safe thing to fix here?

extensions/PhabBugz/lib/Feed.pm
```diff
@@ -688,7 +688,7 @@ sub process_new_user {
         # that are connected to revisions
         f11 => 'attachments.filename',
         o11 => 'regexp',
-        v11 => '^phabricator-D[[:digit:]]+-url[[.period.]]txt$',
+        v11 => '^phabricator-D[[:digit:]]+-url.txt$',
```
dylanwh: [[.period.]] is a mysql-ism that won't work with sqlite. Technically the regex is more open now, but I think it won't hurt.

extensions/PhabBugz/lib/Feed.pm
dklawren: Yeah pretty much the same thing.

extensions/PhabBugz/lib/Feed.pm
dklawren: works for me

PR #675, opened 2018-07-26
Bug 1478540 - Update User.pm to load more than 100 users by using the paging functionality of Conduit API

dklawren:


extensions/PhabBugz/lib/User.pm
```diff
@@ -131,13 +131,17 @@ sub match {
...
-        foreach my $user ( @{ $result->{result}{data} } ) {
-            push @$phab_users, $class->new($user);
+    while (1) {
+        my $result = request( 'user.search', $data );
+        if ( exists $result->{result}{data} && @{ $result->{result}{data} } ) {
+            foreach my $user ( @{ $result->{result}{data} } ) {
+                push @$phab_users, $class->new($user);
+            }
         }
+        last if !$result->{cursur}->{after};
```
dylanwh: cursur is misspelled, according to the doc you linked to it is cursor.

Rather than a while loop with 'last', I'd prefer seeing a do { } while $result->{cursor}{after}

PR #676, opened 2018-07-27
Bug 1419636 - Make Google Analytics use beacon/XHR instead of img tag

kyoshino:
## Description

* Set GA's [`transport`](https://developers.google.com/analytics/devguides/collection/analyticsjs/field-reference#transport) option to use [`navigator.sendBeacon`](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/sendBeacon) by default
* Replace the `image-src` CSP directive with `connect-src` to allow beacon or fallback XHR

## Bug

[Bug 1419636 - Make Google Analytics use beacon/XHR instead of img tag](https://bugzilla.mozilla.org/show_bug.cgi?id=1419636)


dylanwh:
@kyoshino: Can I assume you ran this code on your dev vm?

kyoshino:
Yup, tested with the debug version of analytics.js, and it works well.

PR #678, opened 2018-07-27
Bug 1320977 - performance tweaks

dylanwh:
In #672 performance regressed a bit, so let's win that back here.
The theme of this change is to load bugzilla classes earlier. It seems to save about 40ms per API call.

Before:

screen shot 2018-07-27 at 14 54 13

After
screen shot 2018-07-27 at 14 54 02

### DIFF
--- a/mod_perl.pl
+++ b/mod_perl.pl
@@ -67,6 +67,9 @@ use Bugzilla::Install::Requirements ();
 use Bugzilla::Util ();
 use Bugzilla::RNG ();
 use Bugzilla::ModPerl ();
+use Mojo::Loader qw(find_modules);
+use Module::Runtime qw(require_module);
+use Bugzilla::WebService::Server::REST;
 
 # Make warnings go to the virtual host's log and not the main
 # server log.
@@ -102,6 +105,10 @@ if ($ENV{LOCALCONFIG_ENV}) {
 Bugzilla::Extension->load_all();
 
 Bugzilla->preload_features();
+
+require_module($_) for find_modules('Bugzilla::User::Setting');
+
+Bugzilla::WebService::Server::REST->preload;
 
 # Force instantiation of template so Bugzilla::Template::PreloadProvider can do its magic.
 Bugzilla->preload_templates;
--- a/Bugzilla/User/Setting.pm
+++ b/Bugzilla/User/Setting.pm
@@ -13,8 +13,6 @@ use strict;
 use warnings;
 
 use base qw(Exporter);
-
-
 # Module stuff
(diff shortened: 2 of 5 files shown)


dylanwh:
@purelogiq can you review this?
Maybe verify the profiling numbers?
(see https://www.youtube.com/watch?v=1hrdVxI0uFM for how, or ask me in IRC)

mod_perl.pl
```diff
@@ -67,6 +67,9 @@ BEGIN
 use Bugzilla::Util ();
 use Bugzilla::RNG ();
 use Bugzilla::ModPerl ();
+use Mojo::Loader qw(find_modules);
```
purelogiq: Interesting use of mojo in a not-yet-mojo app :)

purelogiq:
@dylanwh Will verify the profiling numbers Monday, missed your comment there.

dylanwh:
cool, I'll wait to merge. if it's not fast enough-er I can find another slow part.

purelogiq:
@dylanwh First off, awesome stuff with the flamegraphs and profiling. When I compared the profile on master vs this branch I saw that `webservice_user_get` is faster in this PR but the overall request is the same or even slower. Can you verify that I'm doing this right?

Vid on google drive: https://drive.google.com/file/d/194cHcyvlZpyP5JvXUr1eyY42tCmp0DlA/view

dylanwh:
Are you restarting httpd between tests?

But actually, looking at the video this is mostly expected -- the difference disappears after the first request because the optimization here is just loading some modules earlier.

However "first requests" are quite common -- because it's perl worker-process and there are hundreds of those, and they get restarted somewhat frequently.

dylanwh:
Looks like the remaining balance of time is in using Safe->reval(). I wanted to fix that in #321 but it's a little tricky.

purelogiq:
@dylanwh Retested and still getting the same results :/

Changing $cpt->reval($serialized) in DB/Schema.pm as below and removing the Safe import: 
(**_preface I have no clue what I'm doing in this area of the code :). No idea if/why I need a 'my' inside eval. No clue of the implications of changing this, use at your own risk!_**)
```
my $stuff= eval("my $serialized") || die "Unable to restore cached schema: " . $@;
$thawed_hash = $stuff;
```
results in a pretty good speed up:

### BEFORE:
with-safe-flames


### AFTER:
without-safe-flames

purelogiq:
@dylanwh I'm still good on r+ing this because I've been testing post-warmup and the pre-warmup speed improvement like you mentioned maybe be worth it.

dylanwh:
Yup, I'll take it, and prioritize #321 later.

PR #679, opened 2018-07-27
Bug 1478897 - ensure phabbugs doesn't fail outright when encountering invalid bug ids

dklawren:
… invalid bug ids


dklawren:
Updated for review

PR #680, opened 2018-07-30
Bug 1479523 - Disable one-click component watching for logged out users

kyoshino:
## Description

The new one-click component watching functionality added in #646 should be disabled properly if the user is not logged in. This checks if their API key is available, and if not, simply remove the Watch buttons.

## Bug

[Bug 1479523 - Disable one-click component watching for logged out users](https://bugzilla.mozilla.org/show_bug.cgi?id=1479523)


kyoshino:
It might be better to check `BUGZILLA.user.id` here but the property is not exposed on all the pages including the component description pages. I’ll fix it later in a separate bug.

PR #665, opened 2018-07-22
Bug 1380437 - Implement REST API utility methods in global.js to replace bugzilla_ajax(), eliminate JSON-RPC usage

kyoshino:
## Description

* Implement the `Bugzilla.API` class offering REST API utility methods that replace `bugzilla_ajax()` originally from `bug_modal.js`
* Use the new, simple utility methods for all the REST API calls
* Replace all the JSON-RPC calls with REST API 🎉 
* Generate an API token on all pages so it can be used for new features on the global header (but the token auth support will be removed soon; see the Notes below)

## Notes

* Tested everything locally except for the PhabBugz review list on My Dashboard, which requires API access
* The `Bugzilla_api_token` param, now appended in `Bugzilla.API._init()`, will be [removed soon](https://bugzilla.mozilla.org/show_bug.cgi?id=1477163) in favour of login cookies

## Bug

[Bug 1380437 - Implement REST API utility methods in global.js to replace bugzilla_ajax(), eliminate JSON-RPC usage](https://bugzilla.mozilla.org/show_bug.cgi?id=1380437)

PR #690, opened 2018-08-01
Bug 1477163 - REST API authentication fails with X-Bugzilla-Token header

kyoshino:
## Description

* Add support for auth methods that are supposed to work:
    * `X-Bugzilla-Token` HTTP request header
    * ~`token` and `Bugzilla_token` URL query params~
* Use the `X-Bugzilla-Token` header in `bugzilla_ajax()` (under refactoring in #665)
* The token support could be deprecated but can’t be removed now, as mentioned in the bug, because it’s available via the [API](https://bmo.readthedocs.io/en/latest/api/core/v1/user.html#login).

## Bug

[Bug 1477163 - REST API authentication fails with X-Bugzilla-Token header or token/Bugzilla_token query param](https://bugzilla.mozilla.org/show_bug.cgi?id=1477163)

### DIFF
--- a/Bugzilla/Auth/Login/Cookie.pm
+++ b/Bugzilla/Auth/Login/Cookie.pm
@@ -33,7 +33,6 @@ sub get_login_info {
     my ($self) = @_;
     my $cgi = Bugzilla->cgi;
     my $dbh = Bugzilla->dbh;
-    my $input = Bugzilla->input_params;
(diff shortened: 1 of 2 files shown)


kyoshino:
Closing this PR as it turned out that there were 2 different types of tokens, and this change would break something.

PR #691, opened 2018-08-01
Bug 1480042 - Request queue dropdown badge shows empty names for some display names

kyoshino:
## Description

* Use people’s nicknames as display names in the Requests dropdown menu, rather than erroneously trying to extract their real names. So **Kohei Yoshino [:kohei]** will be **kohei** instead of **Kohei Yoshino**.
* Use the local part of their email address If colon-prefixed nickname could not be found, just like modal bug page’s readable presentation.

## Bug

[Bug 1480042 - Request queue dropdown badge shows empty names for some display names](https://bugzilla.mozilla.org/show_bug.cgi?id=1480042)


kyoshino:
Okay, fixing it now.

kyoshino:
Updated. The ad-hoc JSON API now returns `nick` and `gravatar` from `Bugzilla::User`, which will be displayed in the dropdown. The HTML and CSV versions of `request.cgi` will continue showing `NAME <EMAIL>` or `EMAIL` for now.

PR #701, opened 2018-08-03
Bug 1480891 my dashboard does not show the revision id and title for phabricator review requests

dklawren:
simplified what was there by just adding a new Revision column with the ID as a link directly to the revision in Phab instead of adding a new row. If you hover the link, the revision summary is displayed.

### DIFF
--- a/extensions/MyDashboard/web/js/flags.js
+++ b/extensions/MyDashboard/web/js/flags.js
@@ -143,10 +143,23 @@ $(function () {
                 Y.Escape.html(o.data.author_name) + '</span>';
         };
 
-        var phabRevisionFormatter = function(o) {
-            return '<a href="' + o.data.url + '" target="_blank" ' +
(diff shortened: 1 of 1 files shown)


extensions/MyDashboard/web/js/flags.js
```diff
@@ -143,24 +143,10 @@ $(function () {
...
-                '</a></td></tr>',
-                'before');
-
-            o.cell.set('text', o.data.status == 'added' ? 'pending' : o.data.status);
-
-            return false;
+        var phabRevisionFormatter = function(o) {
+            return '<a href="' + o.data.url + '" target="_blank" ' +
+                'title="' + Y.Escape.html(o.data.title) + '">D' +
+                o.data.id + '</a>';
```
globau: the `title` field can be very long.  iirc this resulted in a dysfunctional table.
it's probably better to fix the two-row output rather than dropping it (it was working correctly at the time it was landed).

PR #712, opened 2018-08-10
Bug 602313 follow up - Icons on new attachment uploader are broken

kyoshino:
Font Awesome was one of my local fonts which is not available on BMO 😓 Just use Material Icons for now. It doesn’t have media format icons, but I don’t want to add another icon font file.


dylanwh:
I didn't even notice. 😭

PR #714, opened 2018-08-11
support bugzilla syntax inside text segments of paragraph tags

dylanwh:


purelogiq:
@dylanwh Just want to advocate for my small PR #716 again (which in my humble opinion solves the 3 problems - utf8 encoding, double linking bug, nofollow).

I know that this is still a WIP, but, some potential bugs that I see in this PR:
- We have the same problem here in that the markdown parser sees a bug link and converts it to anchor tag and then quoteUrls sees that link inside the anchor inside the paragraph and creates a new anchor. Even worse, this applies to all links (even non-bmo links) since quoteUrls doesn't know to not autolink stuff that markdown has autolinked.
- This PR lacks the change that forces all new lines to render as new lines, instead of having to manually put 2 blank spaces at the end of each line. (GitHub forces new lines and it really improves quality of life for users imo).

I personally think that having to understand the old quoteUrls and its interaction with the new markdown stuff makes things harder instead of easier. While the concept of parsing the dom and calling quoteUrls on paragraph text is simpler, in many ways I think the implementation in #716 is easier to understand and maintain, but, perhaps I am being shortsighted :)

PR #713, opened 2018-08-10
Bug 1482474 - Unicode characters in comments are garbled

dylanwh:

PR #716, opened 2018-08-13
Bug 1482447 - nofollow external markdown links and fixes

purelogiq:
Reorders the flow of renderComment to fix 4 things:

- Garbage unicode characters (e.g emoji) are rendered after going
through the markdown parser. This is fixed now. [Bug 1482474](https://bugzilla.mozilla.org/show_bug.cgi?id=1482474)

- The ability to nofollow external user created markdown links with
simple regex matching.

- Fixes a bug where full bug links would be linkified by both
renderComment and the markdown autolinker, causing an invalid anchor tag
to be rendered on the page.

- Fixes a "bug" where users are unable to use full bug links as the link
in a markdown link tag e.g. `[my bug](http://bmo.test/show_bug.cgi?id=1)`.
Users can now do this if they want. The tooltip will be added by the
frontend.

This PR was inspired by #711 and #713 :)

### DIFF
--- a/Bugzilla/Template.pm
+++ b/Bugzilla/Template.pm
@@ -40,6 +40,7 @@ use IO::Dir;
 use List::MoreUtils qw(firstidx);
 use Scalar::Util qw(blessed);
 use JSON::XS qw(encode_json);
+use Encode;
 
 use parent qw(Template);
 
@@ -195,11 +196,10 @@ sub renderComment {
 
     # Provide tooltips for full bug links (Bug 74355)
     my $urlbase_re = '(' . quotemeta(Bugzilla->localconfig->{urlbase}) . ')';
-    $text =~ s~\b(${urlbase_re}\Qshow_bug.cgi?id=\E([0-9]+)(\#c([0-9]+))?)\b
+    $text =~ s~(?<!\]\()\b(${urlbase_re}\Qshow_bug.cgi?id=\E([0-9]+)(\#c([0-9]+))?)\b
               ~($things[$count++] = $bug_link_func->($3, $1, { comment_num => $5, user => $anon_user })) &&
                ("\x{FDD2}" . ($count-1) . "\x{FDD3}")
               ~egox;
-
 
     if ($skip_markdown) {
         # non-mailto protocols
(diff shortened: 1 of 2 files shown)


Bugzilla/Template.pm
```diff
@@ -195,12 +196,11 @@ sub renderComment {
 
     # Provide tooltips for full bug links (Bug 74355)
     my $urlbase_re = '(' . quotemeta(Bugzilla->localconfig->{urlbase}) . ')';
-    $text =~ s~\b(${urlbase_re}\Qshow_bug.cgi?id=\E([0-9]+)(\#c([0-9]+))?)\b
+    $text =~ s~(?<!\]\()\b(${urlbase_re}\Qshow_bug.cgi?id=\E([0-9]+)(\#c([0-9]+))?)\b
```
purelogiq: Here we check that this full bug link is not being used in a markdown link [ ]( ) tag. If it is being used in one, we leave it alone, the markdown parser will use the full link to generate the corresponding anchor tag.

With the other reorder in this PR and _without_ this change, an anchor tag would be rendered with invalid unicode characters (because the markdown parser will use the `"\x{FDD2}" . ($count-1) . "\x{FDD3}"` as part of the link but it outputs invalid unicode that does not match in line 280 of this function where we replace it with the correct link)

Bugzilla/Template.pm
```diff
@@ -227,10 +227,20 @@ sub renderComment {
...
         # As a preference, we opt into all new line breaks being rendered as a new line.
         $text =~ s/(\r?\n)/  $1/g;
+
+        # Render text as markdown
+        $text = decode('UTF-8', Bugzilla->markdown_parser->render_html($text));
+
+        # At this point all links are plain links created by markdown, i.e. they have no additional
+        # html attributes and so our regex should match. Here we nofollow any non-bmo link.
+        my $urlbase = Bugzilla->localconfig->{urlbase};
+        $text =~ s/<a href="(?!$urlbase)/<a rel="nofollow" href="/g;
```
purelogiq: I hope I'm not being naive about this - since we escape all html tags a few lines before this, I think it is safe to assume that any links at this point were created by the markdown parser above, which should output a plain link that matches this regex.

Bugzilla/Template.pm
```diff
@@ -227,10 +227,20 @@ sub renderComment {
...
         # This prevents things like inline styles on anchor tags, which otherwise would be valid.
+        # Note that this does not affect the html regex replacements that happens before this
+        # if block. See the comment about encoding hacks near the beginning of the function.
         $text =~ s/([<])/&lt;/g;
 
         # As a preference, we opt into all new line breaks being rendered as a new line.
         $text =~ s/(\r?\n)/  $1/g;
+
+        # Render text as markdown
+        $text = decode('UTF-8', Bugzilla->markdown_parser->render_html($text));
```
purelogiq: @kyoshino discovered a problem here: https://github.com/mozilla-bteam/bmo/pull/713#discussion_r209383992.

The root of the problem is that the markdown parser (with the autolink extension enabled) doesn't respect existing anchor tags containing links. Try this out in `re.pl`.
```
$link = '<a href="https://mozilla.org">https://mozilla.org</a>';
$result = Bugzilla->markdown_parser->render_html($link);
// <p><a href="https://mozilla.org"><a href="https://mozilla.org">https://mozilla.org</a></a></p>
```
As you see, the output is an anchor nested within an anchor. This is invalid html and browsers interpt this by extracting one of the anchors and making them sibling elements, leaving one with an empty body.

When `https://mozilla.org` was found in the text before this PR, our full bug link regex had already converted it to a valid anchor tag with the text being the link itself, and then the markdown parser converted the text of the anchor tag into another anchor tag :(

In this PR, by moving the rendering of markdown before the encoding substitution stuff (line 280) we fix this problem. This is because the markdown parser doesn't have a chance to see the full link (since it is replaced with the encoding substitution). It is like we're telling the parser that we'll take care of this link ;)

Bugzilla/Template.pm
purelogiq: This nofollows both autolinkifyied links and manual [ ]() markdown links.

PR #717, opened 2018-08-14
Bug 1482145 - PhabBot changes are showing up as from the wrong user

dylanwh:


dylanwh:
re: perltidy changes, I've seen some weirdness on windows. I'll have to take a look.

it appears that yours is ignoring the (longer) line-length limit.

PR #718, opened 2018-08-14
add more type checking to phabbugz code

dylanwh:


extensions/PhabBugz/lib/Project.pm
```diff
@@ -142,12 +146,20 @@ sub BUILDARGS {
...
+            view_policy => Str,
+            edit_policy => Str,
+            join_policy => Str,
+        ]
+    );
+    my ( $class, $params ) = $check->(@_);
 
-    my $description = $params->{description} || 'Need description';
+    my $name        = $params->{name};
+    my $description = $params->{description};
```
purelogiq: The old code had `|| 'Need description'`, just incase you overlooked that.

extensions/PhabBugz/lib/Project.pm
```diff
@@ -142,12 +146,20 @@ sub BUILDARGS {
...
+            description => Str,
+            view_policy => Str,
+            edit_policy => Str,
+            join_policy => Str,
+        ]
+    );
+    my ( $class, $params ) = $check->(@_);
 
-    my $description = $params->{description} || 'Need description';
+    my $name        = $params->{name};
```
purelogiq: The old code called trim on the name, just incase you overlooked that.

dylanwh: That is a good catch, I'll update it to still call trim. The other change here isn't needed because the type checking prevents those cases from happening.

PR #720, opened 2018-08-17
Bug 1482145 - add a scope_guard option to set_user()

dylanwh:


purelogiq:
@dylanwh If there's no hurry on this, I'd be more comfortable with @dklawren reviewing when he's back Monday. I see that it's mainly just the use of Scope::Guard, but, I'm not familiar with the intricacies of the phabbuz code.

dylanwh:
That's okay.

PR #726, opened 2018-08-24
Bug 1482338 - "Search Bugs" field has "cursor: pointer" style while it's text field

kyoshino:
## Description

Fix the cursor on the global header’s searchbar, that should be an I-beam instead of hand/pointer.

## Bug

[Bug 1482338 - “Search Bugs” field has “cursor: pointer” style while it’s text field](https://bugzilla.mozilla.org/show_bug.cgi?id=1482338)

### DIFF
--- a/skins/standard/global.css
+++ b/skins/standard/global.css
@@ -262,6 +262,7 @@
         box-shadow: none;
         font-size: 14px !important;
         line-height: 32px;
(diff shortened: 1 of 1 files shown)


kyoshino:
@dylanwh This one is just one-liner.

PR #730, opened 2018-08-26
Allow customizing the name of the "nobody" user (nobody@mozilla.org)

dylanwh:


kyoshino:
I think we can close this and instead convert the Assignee field to **Assignees** field that accepts zero, one or multiple assignees like GitHub.

PR #736, opened 2018-09-04
Bug 1484892 - Modify EditComments extension to let anyone use it conditionally and support inline editing

kyoshino:
## Description

* The expected UX and screenshots are in a [bug comment](https://bugzilla.mozilla.org/show_bug.cgi?id=1484477#c13)
* Add API methods that allow
    * users to save edited comment immediately
    * insiders to show/hide a revision with one click
* Hide all existing admin-edited revisions by default, which may contain sensitive info
* Redesign the buttons on the comment header as seen in the screenshot

## Bug

[Bug 1484892 - Modify EditComments extension to let anyone use it conditionally and support inline editing](https://bugzilla.mozilla.org/show_bug.cgi?id=1484892)


Bugzilla/Template.pm
```diff
@@ -888,7 +888,9 @@ sub create {
             },
 
             json_encode => sub {
-                return encode_json($_[0]);
+                my $str = encode_json($_[0]);
```
dylanwh: I'm sure this works, but using ::is_utf8() is wrong (it doesn't mean anything for the purposes of encoding).

dylanwh: @kyoshino:
I suggest this change: https://gist.github.com/dylanwh/001e58d22765d4b500eac97e504167f4

Mojo::JSON is defined as returning bytes (not characters) and Encode::decode will take bytes and turn them into characters (assuming the bytes are utf8). Anything that is invalid (e.g. not utf8) will be replaced with the replacement character.

kyoshino: I copied this from somewhere else. Without this, `js_BUGZILLA.string.InlineCommentEditor` will contain garbage characters.  Let’s use Mojo then 😺

dylanwh:
@kyoshino Can you make a second PR that only includes the schema change? We'll want to land that first, then this.

PR #751, opened 2018-09-18
no bug - bump Alien::libcmark_gfm

dylanwh:


dylanwh:
building a new base image for this too

PR #755, opened 2018-09-20
Bug 1492926 - Handle DBIx::Connectors more appropriately

dylanwh:
This is a bigger change than I anticipated, because the way we cached
DBIx::Connector objects was bad.

Now we cache the Bugzilla::DB instances in connect_main() and connect_shadow().
This is for maintaining a 1:1 mapping of Bugzilla::DB objects and
DBIx::Connector objects. This is important because we want be able to inspect
Bugzilla::DB->bz_in_transactions() from the 'connected' event.

Note that we weaken the lexical variable $self in _build_connector()
because it is referenced by the callback passed to DBI.
Without this there would be a memory cycle and stuff would never be freed.

(tested my understanding in this gist:
https://gist.github.com/dylanwh/646574a027f7b7d92cb7586676da7468)

### DIFF
--- a/Bugzilla/DB.pm
+++ b/Bugzilla/DB.pm
@@ -157,6 +157,7 @@ sub connect_shadow {
 
 sub connect_main {
     state $main_dbh = _connect(Bugzilla->localconfig);
+    return $main_dbh;
 }
 
 sub _connect {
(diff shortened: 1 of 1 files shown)


dylanwh:
Some assumptions I had when writing this: 

1. connect_main() / connect_shadow() are called on every request.
2. Both functions always return the same objects (that is, two different objects) for the life of the *process*
3. if a reconnection happens during a transaction, the CodeError not_in_transaction will be thrown

PR #758, opened 2018-09-23
Bug 1493490 - Remove navigator.buildID usage from Bugzilla Helper

kyoshino:
## Description

* Remove one usage of the non-standard `navigator.buildID` DOM property, which will be removed from Firefox in the near future
* Also remove another Build Identifier field based on the browser’s UA string, which isn’t useful because it has been frozen at `Gecko/20100101`.

## Bug

[Bug 1493490 - Remove navigator.buildID usage from Bugzilla Helper](https://bugzilla.mozilla.org/show_bug.cgi?id=1493490)

### DIFF
--- a/extensions/GuidedBugEntry/template/en/default/bug/create/comment-guided.txt.tmpl
+++ b/extensions/GuidedBugEntry/template/en/default/bug/create/comment-guided.txt.tmpl
@@ -1,8 +1,6 @@
 [% USE Bugzilla %]
 [% cgi = Bugzilla.cgi %]
 User Agent: [% cgi.param('user_agent') %]
-[% IF cgi.param('build_id') %]
-Build ID: [% cgi.param('build_id') %][% END %]
 [% IF cgi.param('firefox_for_android') %]
 Firefox for Android
 [% END %]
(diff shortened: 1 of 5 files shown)


cpeterso:
Looks like your PR also removes trailing whitespace characters from unrelated lines of code. That's harmless but the code reviewer will have to work a little harder to identify the important changes. :)

kyoshino:
Yeah, I’m removing trailing whitespaces according to the [VS Code project settings](https://github.com/mozilla-bteam/bmo/blob/master/.vscode/settings.json) but it’s probably fine, because GitHub allows to hide whitespace space changes while showing diff 😄 

screen shot 2018-09-23 at 4 43 24 am

cpeterso:
Maybe we don't need to remove any of this code. Firefox <= 63 will still support buildID until October 2018 and Firefox ESR 60 until [September 2019](https://wiki.mozilla.org/Release_Management/Calendar).

Should we continue to include buildID in bug reports from these users? I don't know how often buildID in bug reports is actually useful, but there is no harm in including it when it is available. We could just postpone this cleanup PR until October 2019. 😄

kyoshino:
I *think* it’s fairly safe to remove `navigator.buildID` from Bugzilla Helper as explained in [my bug comments](https://bugzilla.mozilla.org/show_bug.cgi?id=1493490#c2).

PR #774, opened 2018-09-28
Add new faster products API

dylanwh:
This adds a new products API, and also serves as documentation for new APIs
and provides a way of validating against API keys.


dylanwh:
All the new stuff actually needs to be under the quantum namespace -- that's Bugzilla::Quantum is the root namespace of the mojo stuff. 

Analogous to how we pass a route object to a route object to CGI->load_all, we can do that for products. I'm not sure if controllers can be sub-namespaced but I'll experiment with Bugzilla::Quantum::API::Products (not sure how that will work with the -> to('controller#action') syntax)

PR #775, opened 2018-09-28
cleanup routes with helpers and shortcuts

dylanwh:


dylanwh:
All of this is in #773

PR #778, opened 2018-10-01
Bug 1472565 - Allow to instantly search recently visited bugs, ID/aliases, keywords, components in the quick search bar

kyoshino:
## Description

Enhance the quick search bar on the global navigation so it will be able to show/filter the user’s recently visited bugs, saved searches, components and keywords as well as [possible duplicates](https://bmo.readthedocs.io/en/latest/api/core/v1/bug.html#possible-duplicates) as you type. See [this screenshot](https://twitter.com/BugzillaUX/status/1013919072252096512) in action.

## Bug

[Bug 1472565 - Allow to instantly search recently visited bugs, possible duplicates, keywords, components in the quick search bar](https://bugzilla.mozilla.org/show_bug.cgi?id=1472565)

### DIFF
--- a/js/dropdown.js
+++ b/js/dropdown.js
@@ -39,7 +39,7 @@ $(function() {
                 if ($content.is(':visible')) {
                     e.preventDefault();
                     e.stopPropagation();
-                    var $items = $content.find('[role="menuitem"], [role="option"]');
+                    var $items = $content.find('[role="menuitem"], [role="option"]').filter(':visible');
                     // if none active select the first or last
                     var $link = $items.filter('.active');
                     if ($link.length == 0) {
@@ -103,6 +103,7 @@ $(function() {
         var $div     = $(this);
         var $button  = $div.find('.dropdown-button');
         var $content = $div.find('.dropdown-content');
+        var is_input = $button.eq('input');
         $button.click(function(e) {
             // Do not handle non-primary click.
             if (e.button != 0 || $content.hasClass('hover-display')) {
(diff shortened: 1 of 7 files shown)


kyoshino:
Again, the Selenium tests are struggling 😕

kyoshino:
@dylanwh we still have to fix the tests by [upgrading Firefox](https://github.com/mozilla-bteam/bmo-systems/pull/8).

dylanwh:
There is no easy path forward for the selenium tests. How about we just remove them?
@dklawren

dklawren:
> There is no easy path forward for the selenium tests. How about we just remove them?
> @dklawren

I am hesitant to just remove selenium tests wholesale as they have been good for us in the past to catch random issues with new features even if the tests are not written to explicitly test those features. Sometimes new code will cause one or more of the standard pages to not load properly which are caught by the UI tests. Can we not just fix bug them by upgrading the dependencies as @kyoshino has made a PR for?

dylanwh:
It's not possible to use Test::Selenium with modern selenium, and we're stuck on firefox 47 there. :-/

dklawren:
> It's not possible to use Test::Selenium with modern selenium, and we're stuck on firefox 47 there. :-/

Or we convert our current scripts to Webdriver which I have wanted to do for quite some time but it is not a trivial undertaking.

kyoshino:
I’m also worried that removing Selenium tests may potentially lead to security holes. I was constantly adding tests while contributing to [Bedrock](https://github.com/mozilla/bedrock) so would rather like to add/see more (modernized) tests in Bugzilla too.

kyoshino:
I wanted to ship this earlier (with Bugzilla 6.0) so implemented a workaround for test failures on Firefox 47.

kyoshino:
I’ll remove the Firefox 47 workaround once #1273 is merged.

PR #782, opened 2018-10-01
Bug 1209148 - Stop using short-life token for internal REST API auth, which is problematic especially on My Dashboard, and rely on login cookie instead

kyoshino:
## Description

* Internal `Bugzilla_api_token` is problematic and redundant. The existing login cookie can be used instead for API auth.
* Public `Bugzilla_token` is a different thing, and it should not be affected by this change.

## Bug

[Bug 1209148 - Stop using short-life token for internal REST API auth, which is problematic especially on My Dashboard, and rely on login cookie instead](https://bugzilla.mozilla.org/show_bug.cgi?id=1209148)

### DIFF
--- a/Bugzilla/Auth/Login/Cookie.pm
+++ b/Bugzilla/Auth/Login/Cookie.pm
@@ -33,7 +33,7 @@ sub get_login_info {
   my ($self) = @_;
   my $cgi    = Bugzilla->cgi;
   my $dbh    = Bugzilla->dbh;
-  my ($user_id, $login_cookie, $is_internal);
+  my ($user_id, $login_cookie);
 
   if (!Bugzilla->request_cache->{auth_no_automatic_login}) {
     $login_cookie = $cgi->cookie("Bugzilla_logincookie");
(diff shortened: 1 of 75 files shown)


kyoshino:
So... I tried, and looks like this is working. The real culprit is `$login_cookie = ''` in `Cookie.pm`.

kyoshino:
`/rest/bug/attachment/${id}` I want to use for [Bug 1496207](https://bugzilla.mozilla.org/show_bug.cgi?id=1496207) doesn't work when the `Bugzilla_api_token` param is set in the query string. It's a PUT request. I would like to this PR get reviewed and merged so the problem should be gone.

> The requested method 'Bugzilla::Attachment::set_Bugzilla_api_token' was not found.

kyoshino:
I just encountered the invalid token error on My Dashboard  :neutral_face:

invalid token error

kyoshino:
This is basically a removal so no code review needed. Just want to make sure this works as expected and no security issues introduced.

floatingatoll:
@kyoshino What circumstances can activate this code block today?

https://github.com/mozilla-bteam/bmo/pull/782/files#diff-e432c9f5510c145bfbb4a9eb373e5209L107
```
            # forbid logging in with a cookie if only api-keys are allowed
            if (i_am_webservice() && !$is_internal) {
```

kyoshino:
I meant: no review for *newly added* code is required...

floatingatoll:
(I'd planned to ask that question in any case, but thank you for clarifying all the same!)

kyoshino:
Maybe @floatingatoll is right. That particular part should not be removed after all. I'm reverting it and only remove the `is_internal` check. But hmm, what happens then?

dylanwh:
I think api_key only might just have been for preventing CSRF attacks, which is now mitigated by SameSite cookies.

kyoshino:
So can we remove the `api_key_only` setting also?

kyoshino:
This PR needs to be updated once #917 is merged.

PR #786, opened 2018-10-02
Bug 1495869 - Crash graph not found after mojo migration

dylanwh:


kyoshino:
I cannot merged this though because the tests are not passing.

PR #793, opened 2018-10-03
reformat attachments.cgi

dylanwh:
I'm going to have to fix the utf8 encoding problem on attachment.cgi, ahead of that let's reformat it


kyoshino:
By the way, what’s the max chars per line? Still 80 I guess? I use 120 for JavaScript as per GitHub’s recommendation. Not sure about the manner in the Perl world.

dylanwh:
it's currently 80. I found that with 120 it became difficult to diff things. Where's github's recommendation written?

kyoshino:
It thought it was written on https://github.com/styleguide but can’t find it now…

PR #798, opened 2018-10-04
add some type checking

dylanwh:


Bugzilla/WebService/Bug.pm
```diff
@@ -128,7 +129,15 @@ BEGIN {
 ###########
 
 sub fields {
-    my ($self, $params) = validate(@_, 'ids', 'names');
+    state $check = compile(
+        Object,
+        Dict[
+            ids   => Optional[ArrayRef[Int]],
```
dylanwh: All the places we called validate, ArrayRef[Int] needs a coercion from Int
(and ArrayRef[Str] from Str)

PR #801, opened 2018-10-04
added basic phabbugz skeleton

dylanwh:


extensions/PhabBugz/Extension.pm
```diff
@@ -18,6 +18,26 @@ use Bugzilla::Extension::PhabBugz::Feed;
 
 our $VERSION = '0.01';
 
+sub mojo_after_dispatch {
+    my ($c) = @_;
+    my $stash = $c->stash;
+
+    return unless $stash->{controller} eq 'CGI' && $stash->{action} eq 'rest_cgi';
+    return unless $stash->{PATH_INFO} =~ m{^phabbugz/build_target};
+    my $datadog = Bugzilla->datadog('bugzilla.phabbugz') or return;
```
mars-f: Do we want to place all API requests/errors/duration monitoring under a `bugzilla.api.*` namespace?

extensions/PhabBugz/Extension.pm
```diff
@@ -18,6 +18,26 @@ use Bugzilla::Extension::PhabBugz::Feed;
...
+
+    if ($res->code == 200) {
```
mars-f: Will this catch HTTP 5XX errors?

Can responses return other 2XX codes?

extensions/PhabBugz/Extension.pm
mars-f: Do we want to place all API requests/errors/duration monitoring under a `bugzilla.api.*` namespace?

extensions/PhabBugz/Extension.pm
```diff
@@ -18,6 +18,26 @@ use Bugzilla::Extension::PhabBugz::Feed;
...
+        $datadog->increment('build_target_ok');
```
mars-f: It might help to be consistent naming this like the Datadog [ELB](https://docs.datadoghq.com/integrations/amazon_elb/#data-collected) and [Nginx](https://docs.datadoghq.com/integrations/nginx/#data-collected) integration metrics and use `build_target_ok.request_count`.

Perhaps not as relevant for this one operation, but that naming scheme allows us to group other counts under the space, like 2XX and 5XX HTTP code counts.

mars-f: I just realized this is named differently than the operation.  This is the metric name I'm proposing:

`bugzilla.api.phabbugz.set_build_target.request_count`

PR #802, opened 2018-10-05
Bug 1496207 - Allow to request uplift of multiple patches at once by adding checkboxes for other patches

kyoshino:
Allow to make an uplift request of multiple patches. This is done by sending API requests.

## Bugzilla link

[Bug 1496207 - Allow to request uplift of multiple patches at once by adding checkboxes for other patches](https://bugzilla.mozilla.org/show_bug.cgi?id=1496207)

### DIFF
--- a/extensions/BMO/template/en/default/hook/flag/type_comment-form.html.tmpl
+++ b/extensions/BMO/template/en/default/hook/flag/type_comment-form.html.tmpl
@@ -6,6 +6,8 @@
   # defined by the Mozilla Public License, v. 2.0.
   #%]
 
+<meta name="extra-patch-types" content="text/x-phabricator-request">
+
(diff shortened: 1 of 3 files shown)


kyoshino:
I’m no longer seeing a REST API error complaining about `Bugzilla_api_token` (dunno when it’s fixed) so removing the blocked label. This will successfully update the flags for other patches in background, but the problem is, only the first update will be notified via email probably due to the timestamp issue. Will try to find its workaround.

kyoshino:
Solved the email notification issue by submitting the form via XHR first. A bit hacky but it works.

kyoshino:
Yeah, this should still work with regular patches, but we are [removing Splinter anyway](https://bugzilla.mozilla.org/show_bug.cgi?id=1514775)…

PR #804, opened 2018-10-05
Bug 1482644 - Improve "Too many requests" page with an explaination

dylanwh:
This is just a proposal, I think we'd need higher approval to use recaptcha.

Also I'd check for the do-not-track header and not load recaptcha if found.

### DIFF
--- a/Bugzilla/Quantum/Plugin/BlockIP.pm
+++ b/Bugzilla/Quantum/Plugin/BlockIP.pm
@@ -35,7 +35,7 @@ sub _before_routes {
     $c->block_ip($ip);
     $c->res->code(429);
     $c->res->message('Too Many Requests');
-    $c->render(handler => 'bugzilla', template => 'global/ip-blocked');
+    $c->render(handler => 'bugzilla', template => 'global/ip-blocked', block_timeout => BLOCK_TIMEOUT);
   }
 }
 
--- a/Bugzilla/Quantum/Plugin/BlockIP.pm
+++ b/Bugzilla/Quantum/Plugin/BlockIP.pm
@@ -35,8 +35,7 @@ sub _before_routes {
     $c->block_ip($ip);
     $c->res->code(429);
     $c->res->message('Too Many Requests');
-    $c->res->body('Too Many Requests');
-    $c->finish;
+    $c->render(handler => 'bugzilla', template => 'global/ip-blocked');
   }
 }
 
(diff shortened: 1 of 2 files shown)


kyoshino:
Legal has already given the [go-ahead for reCaptcha](https://bugzilla.mozilla.org/show_bug.cgi?id=1447990) so it should be okay. Will take a look tomorrow.

dylanwh:
@kyoshino I'm thinking I want to land  this, but the copy is terrible. Can you help me come up with good copy for the ip-blocked template?

kyoshino:
Okay, I googled a bit. This is what Google itself says when you’re blocked:

> We’re sorry, but you have sent too many requests to us recently. Please try again later.

So we could say:

> We’re sorry, but you have sent too many requests to us recently. Please solve the captcha to get your IP address unblocked.

kyoshino:
I don’t know how to test this but some comments:

* https://www.google.com has to be added to the CSP whitelist.
* Where does `do_not_track` come from? Do we need it?
* A similar error message could be returned when API requests are blocked:

> We’re sorry, but you have sent too many requests to us recently. Please solve the captcha at https://bugzilla.mozilla.org/page.cgi?id=ip-blocked.html to get your IP address unblocked.

dylanwh:
> I don’t know how to test this but some comments:
> 
> * https://www.google.com has to be added to the CSP whitelist.
> * Where does `do_not_track` come from? Do we need it?
> * A similar error message could be returned when API requests are blocked:
> 
> > We’re sorry, but you have sent too many requests to us recently. Please solve the captcha at https://bugzilla.mozilla.org/page.cgi?id=ip-blocked.html to get your IP address unblocked.

This unblocking will only work for clients that send cookies, though. It uses a signed cookie and only the requests with that cookie are allowed. Of course people that use api keys are allowed anyway.

kyoshino:
Then, for web requests:

> We’re sorry, but you have sent too many requests to us recently. Please enable cookies in your browser and solve the captcha below to get your IP address unblocked.

And for API:

> We’re sorry, but you have sent too many requests to us recently. Please authenticate further requests with an API key generated at https://bugzilla.mozilla.org/userprefs.cgi?tab=apikey

Like this?

dylanwh:
Per @markrcote, it might be better to just use some service for blocking abuse. This patch now just provides a template and no recaptcha. The recaptcha code is preserved and will be used for something else possibly.

PR #808, opened 2018-10-05
qa contact should be more optional

dylanwh:


globau:

PR #809, opened 2018-10-05
Bug 1496832 - Add monitoring for feed daemon becoming unresponsive

mars-f:
Use Datadog to count the number of times that jobs are run by the
phabbugz job scheduler.  This should catch problems if the process
responsible for running the jobs dies.


dylanwh:
This and #807  should be part of the same bug.

PR #814, opened 2018-10-08
Bug 623384 - Use Module::Runtime instead of eval { require } or eval "use"

dylanwh:


dklawren:
Should we also be using this in all places where we 'require $something'? Such as:

Bugzilla/Auth/Verify/Stack.pm:        require $module;
Bugzilla/Auth/Login/Stack.pm:        require $module;
Bugzilla/Extension.pm:            my $name = require $config_file;
Bugzilla/Extension.pm:        my $name = require $extension_file;
Bugzilla/WebService/Server/XMLRPC.pm:    require $file;
Bugzilla/Migrate.pm:    require $module;

PR #817, opened 2018-10-08
Bug 1497077 - Convert links to absolute path

kyoshino:
## Description

* Convert all links, image/iframe sources, form actions using a relative path to an absolute path. That’s basically prepending `/` on BMO. 
* Using `<base href>` is simple but problematic as shown in #812.
* Some links are already prefixed with `urlbase` so these are now with `basepath` where possible (links in email are not the case.)

## Bug

[Bug 1497077 - Convert links to absolute path](https://bugzilla.mozilla.org/show_bug.cgi?id=1497077)


dylanwh:
The host name is actually a random string

kyoshino:
Ahh, I see. Then I have to replace them with `//a[contains(@href, ...`

PR #822, opened 2018-10-09
Bug 1495741 - memory issues: use local flag cache

dylanwh:
This saves a large amount of memory by avoiding the duplication of the serialized flag format.

### DIFF
--- a/Bugzilla/WebService/Product.pm
+++ b/Bugzilla/WebService/Product.pm
@@ -64,6 +64,7 @@ sub get_accessible_products {
 }
 
 # Get a list of actual products, based on list of ids or names
+our %FLAG_CACHE;
 sub get {
     my ($self, $params) = validate(@_, 'ids', 'names', 'type');
     my $user = Bugzilla->user;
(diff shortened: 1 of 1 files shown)


Bugzilla/WebService/Product.pm
```diff
@@ -227,11 +229,11 @@ sub _component_to_hash {
         $field_data->{flag_types} = {
             bug =>
                 [map {
-                    $self->_flag_type_to_hash($_)
```
dklawren: Crazy that we never even passed a $params to $self->_flag_to_hash() so the filter() call was a waste of time :)

Bugzilla/WebService/Product.pm
```diff
@@ -134,6 +135,7 @@ sub get {
     }
 
     # Now create a result entry for each.
+    local %FLAG_CACHE = ();
```
dklawren: Could we just do my %FLAG_CACHE here and not need the outer our %FLAG_CACHE? Seems like it would give the same benefit.

dylanwh: I don't think you can localize an entire lexical hash, so it has to be a package var.
Lookup time is slower but it should still be faster overall.

PR #824, opened 2018-10-09
Backport bug 767623 and bug 964113

dylanwh:


Bugzilla/Token.pm
```diff
@@ -20,7 +20,6 @@ use Bugzilla::User;
 use Date::Format;
 use Date::Parse;
 use File::Basename;
-use Digest::MD5 qw(md5_hex);
```
dklawren: Let's also remove this line in:

robots.cgi
Bugzilla/Template.pm

as they are not even used in those files.

dylanwh: I'll handle that later.

PR #826, opened 2018-10-10
Bug 1495741 - memory issues: Avoid copying stuff in the webservice layer so much

dylanwh:


kyoshino:
@dylanwh Non-ASCII chars in API response are garbled after this PR gets merged. Can you take a look? In this screenshot, hyphens in flag names are actually non-breaking hyphens.

screenshot_2018-10-13 my dashboard

dylanwh:
Oh, I think I can fix this. Wheee

PR #845, opened 2018-10-16
no bug - silence mojo warning

dylanwh:
In perl 5.12, length(undef) was made no longer a warning.
In order to reduce memory usage, mojo undefines the outgoing buffer.
This means on perl <= 5.12, a warning is printed frequently (multiple times
per request).

Mojo support for 5.10 is waning, so for the moment we patch that code to silence
this warning.


dylanwh:
One additional 'no warnings' should cover that other function.

PR #847, opened 2018-10-16
Bug 1499477 - Feature request: link from user profile to editusers.

dklawren:


dylanwh:
If this works that is.. astoundingly small hack!

dklawren:
> If this works that is.. astoundingly small hack!

Works for me :)

dklawren:
Good point. I have merged the two.

extensions/UserProfile/template/en/default/pages/user_profile.html.tmpl
```diff
@@ -57,26 +57,12 @@
       &nbsp;
     [% END %]
   </td>
-  <th>Name</th>
+  <th>User</th>
   <td colspan="2">
-    [% target.name || target.address.user FILTER html %]
+    [% INCLUDE bug_modal/user.html.tmpl u = target %]
```
kyoshino: This looks better now 😃 One more thing I should’ve noticed earlier: the **Profile** menu item in the dropdown links to the same user profile page, which might be confusing. Is there any way to remove it?

dklawren: Yeah it was hard coded. I made a change that will allow the profile link to be hidden.

PR #848, opened 2018-10-18
allow anonymous logins from github

dylanwh:
It's longer than two expressions because I determined this is less fragile than adding the github domain to the badhosts file.

### DIFF
--- a/extensions/Needinfo/Extension.pm
+++ b/extensions/Needinfo/Extension.pm
@@ -25,7 +25,7 @@ BEGIN {
 }
 
 sub _user_needinfo_blocked {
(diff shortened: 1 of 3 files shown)


extensions/Needinfo/Extension.pm
```diff
@@ -25,6 +25,7 @@ BEGIN {
 }
 
 sub _user_needinfo_blocked {
+    return 1 if $_[0]->login =~ /\@users\.noreply\.github\.com$/;
```
globau: this is a weird bit of hardcoding, especially without comment.

it should probably be a hook.

dylanwh: a comment, yes, but this is a hot code path.

globau:

PR #851, opened 2018-10-21
Bug 1502181 - MockParams should set user/verify authentication classes to match production

dklawren:
See bug for description


dylanwh:
The problem you found is that we don't work properly if GitHubAuth is not enabled.
I've fixed this in PR dklawren/bmo#1 which you can merge into this.

The second issue -- about the login cookie, this also doesn't work in production BMO! Open a private browsing window and navigate to /login, the first time you visit there will be no value.

dylanwh:
screen shot 2018-10-24 at 20 03 28

PR #854, opened 2018-10-22
Bug 1501893 - Tell robots to leave the graphs alone, and leave a trap

dylanwh:


template/en/default/robots.txt.tmpl
```diff
@@ -18,6 +18,8 @@ Disallow: [% basepath FILTER none %]show_bug.cgi*format=multiple*
 
 Allow: [% basepath FILTER none %]describecomponents.cgi
 Allow: [% basepath FILTER none %]describekeywords.cgi
+Disallow: [% basepath FILTER none %]showdependency*
+Disallow: [% basepath FILTER none %]banmyipforever.cgi
```
jens1o: where is this defined?

dylanwh: It doesn't exist, it's meant as bait for badly behaved bots.

jens1o: ah, so you can see in the logs when some bots misbehave?

PR #860, opened 2018-10-25
Bug 1501849 - Speed up IP blocked page: pre-render template

dylanwh:
By moving BlockIP plugin first, it is ensured to execute first.
This is a slight speed gain. The primary speed gain is by caching
the content of the template.

Before: 429 Too Many Requests (0.031951s, 31.298/s)
After:  429 Too Many Requests (0.000182s, 5494.505/s)


dylanwh:
WHY ARE THE SELENIUM TESTS STILL FAILING AAAARGH

PR #862, opened 2018-10-25
Bug 1501966 - Warn when there are outdated products or components in the teams list

purelogiq:
warnteamlist

Some additional changed added by the linter.

### DIFF
--- a/Bugzilla/Report/SecurityRisk.pm
+++ b/Bugzilla/Report/SecurityRisk.pm
@@ -153,14 +153,14 @@ sub _build_products {
 sub _build_missing_products {
   my ($self) = @_;
   my $dbh = Bugzilla->dbh;
-  my $products = join ', ', map { $dbh->quote($_) } @{$self->products};
+  my @products = map { $dbh->quote($_) } @{$self->products};
   my $query = qq{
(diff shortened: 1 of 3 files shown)


Bugzilla/Report/SecurityRisk.pm
```diff
@@ -146,6 +150,65 @@ sub _build_products {
...
+  my ($self) = @_;
+  my $dbh = Bugzilla->dbh;
+  my $products = join ', ', map { $dbh->quote($_) } @{$self->products};
+  my $query = qq{
+        SELECT
+            name
+        FROM
+            products
+         WHERE
+            products.name IN ($products)
```
dylanwh: Need to use $dbh->sql_in() here, for portability.

Note you can embed code in a string literal with:
```perl
"@{[ ]}"
```

which is sort of a hack but one I've used a few times recently.

purelogiq: Fixed! (Note to self, the `[ ]` inside is really necessary!)

purelogiq: (really fixed now)

PR #866, opened 2018-10-26
Bug 1493253 - Embed crash count table to bug pages

kyoshino:
Embed a crash data table on Firefox bugs as per a request from the Release Management team.

* Copy the [original WebExtension code](https://github.com/mozilla/crash-stop-addon/blob/master/webextension/content.js), contained in `window.addEventListener('DOMContentLoaded')`.
* Embed the script on both modal and legacy `show_bug.cgi`
* Add CSP for the product-details API and iframe

## Bug

[Bug 1493253 - Embed crash count table to bug pages](https://bugzilla.mozilla.org/show_bug.cgi?id=1493253)

### DIFF
--- a/Bugzilla/CGI.pm
+++ b/Bugzilla/CGI.pm
@@ -47,6 +47,9 @@ sub DEFAULT_CSP {
     connect_src => [
       'self',
 
+      # This is for extensions/BMO/web/js/firefox-crash-table.js
+      'https://product-details.mozilla.org',
+
       # This is for extensions/GoogleAnalytics using beacon or XHR
       'https://www.google-analytics.com',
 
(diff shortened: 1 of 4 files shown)


kyoshino:
@dylanwh can we deploy this to the dev server so we can test in advance? I see no CSP errors locally but the height of the `<iframe>` is somehow wrong. Want to make sure this works fine with actual bugs.

kyoshino:
The crash tables on -dev look good, no height issues found. But we cannot deploy it now because of a bug with buildhub as per @calixte. I’m marking this blocked.

kyoshino:
@calixteman Is buildhub still broken? If not, we can merge this now.

kyoshino:
The buildhub issue has been solved. I gonna merge this now.

PR #737, opened 2018-09-10
 WIP: Bug 1489706 - Add section to show_bug displaying the status of associated revisions

dklawren:
TODO:

* show Phabricator icons in place of plain text status for revision and individual reviewers
* Figure out better security model. Currently if a revision is not public, I just replace the title with (secured).

PR #883, opened 2018-11-01
Bug 1489706 - Add section to show_bug displaying the status of associated revisions

dklawren:
screenshot_2018-11-01 1 - conduit test bug


ExE-Boss:
Moved from https://github.com/mozilla-bteam/bmo/pull/737

dklawren:
Will look at using Material icons for review and revision statuses with a mouseover giving the full status text.

ExE-Boss:
Or use emoji like what Phabricator does, which only has an overhead of 4 bytes, whereas Material icons will require sending large-ish SVG files over the net.

kyoshino:
Material Icons are already used on Bugzilla as a font so the size doesn't matter. But looks like icons used on Phabricator are Font Awesome 4.7, which will be an addition to the site anyway even if sliced.

kyoshino:
... like this:
screen shot 2018-11-02 at 12 38 18 am

markrcote:
Check that accepted-prior-diff is displayed properly.

dklawren:
> Material Icons are already used on Bugzilla as a font so the size doesn't matter. But looks like icons used on Phabricator are Font Awesome 4.7, which will be an addition to the site anyway. There are [7 types of statuses](https://github.com/phacility/phabricator/blob/master/src/applications/differential/constants/DifferentialRevisionStatus.php#L110-L181) = 7 icons so we can slice the font to reduce the size.

@kyoshino are you saying that we will be adding FontAwesome to BMO at some point or should I use Material icons which already are being used in BMO? Or are you saying we can just import the icons from FontAwesome that we need and leave out the rest therefore reducing size?

kyoshino:
I think, if we want to use icons, we should have the same icons that are used on Phabricator. I’m attaching the sliced Font Awesome (2 KB) on the bug.

kyoshino:
The font files could be put under `/extensions/PhabBugz/web/fonts/`, I think. And the CSS will look like this (not yet tested though):

```css
@font-face {
  font-family: FontAwesome-DifferentialStatus;
  font-style: normal;
  font-weight: normal;
  src: url(../fonts/FontAwesome-DifferentialStatus.woff2?v=4.7) format('woff2'),
       url(../fonts/FontAwesome-DifferentialStatus.woff?v=4.7) format('woff');
}

[class^="revision-status-"]::before {
  font-family: FontAwesome-DifferentialStatus;
}

.revision-status-needs-review::before {
  content: '\F121';
}

.revision-status-needs-revision::before {
  content: '\F021';
}

.revision-status-changes-planned::before {
  content: '\F025';
}

.revision-status-accepted::before {
  content: '\F00C';
}

.revision-status-published::before {
  content: '\F046';
}

.revision-status-abandoned::before {
  content: '\F072';
}

.revision-status-draft::before {
  content: '\F110';
}
```

dklawren:
So I figured out my issue with the icons not displaying properly. I was not including the font-family directive in for my CSS rules for the containers with the icon content. So that is better. But some icons are still not displaying properly and some are so I am not sure what is going on there. Maybe @kyoshino can take a look and see if there is something obvious with the CSS or something that can be improved. Thanks.

ExE-Boss:
Did you make sure that all the icons are present in the font file and that you are using the correct code points?

extensions/PhabBugz/Extension.pm
```diff
@@ -14,10 +14,30 @@ use warnings;
...
+use Bugzilla::Extension::PhabBugz::Constants;
 
 our $VERSION = '0.01';
 
+sub template_before_process {
+    my ( $self, $args ) = @_;
+    my $file = $args->{'file'};
+    my $vars = $args->{'vars'};
+
+    return
```
dylanwh: I don't think we should support this on the old interface.
also perhaps two 'return unless ...' statements here would be better than one?

dklawren:
@kyoshino If I use the full FontAwesome woff files, the icons display properly for all cases. Is there possibly something wrong or missing in the splice fonts you gave me?

ExE-Boss:
> Is there possibly something wrong or missing in the splice fonts you gave me?
 
As I said in&nbsp;https://github.com/mozilla-bteam/bmo/pull/883#issuecomment-435707603, probably&nbsp;yes.

kyoshino:
@dklawren so several icons were missing. Please use an [updated slice attached to the bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1489706#c7).

PR #888, opened 2018-11-08
WIP Bug 1482451 - Instrument PhabBugz API requests

mars-f:


extensions/PhabBugz/t/monitor-rest-api.t
```diff
@@ -0,0 +1,82 @@
...
+use Bugzilla;
+
+my @calls;
+
+my $Datadog = mock 'DataDog::DogStatsd' => (
+    add_constructor => [
+        'fake_new' => 'hash',
+    ],
+    override => [
+        increment => sub { push @calls, [@_] }
```
dylanwh: I think we only want to record the first argument passed to increment.
The first arg is just going to be `self`.

extensions/PhabBugz/t/monitor-rest-api.t
```diff
@@ -0,0 +1,82 @@
...
+# Mojo::Test loads the application and provides methods for
+# testing requests without having to run a server.
+my $t = Test::Mojo->new('Bugzilla::Quantum');
+
+# Phabricator tells BMO that the Phabricator revision's permissions need to
+# be reconciled with the BMO permissions by POSTing a callback URL.
+$t->post_ok('/rest/phabbugz/build_target/1234/PHID-HMBT-1234' => { 'X-Bugzilla-API-Key' => $api_key });
+$t->status_is(200);
+
+is(['some.metric'], @calls, "Expected one metrics call.");
```
dylanwh: You've been bitten by the non-intuitiveness of perl variables.
Basically anytime you see a @array or %hash variable, pretend it's like `*foo` prefix in python -- because that's basically what happens. if @calls was (1, 2) then
`is(['some.metric'], @calls)` is the same as `is(['some.metric'], 1, 2)`

To avoid having to deal with this, most of the time you just only use the $-flavored variables and store references in them. which is what I've done in my proposed fixes.

dylanwh:
oh yeah, I didn't mention in the review, but you can merge mars-f/bmo/pull/2 for the fixes I made.

PR #904, opened 2018-11-14
filter new orange factor bot email

purelogiq:
Fixes filtering the orangefactor bot, before only the old treeherder bot was being filtered. (I guess this has been a bug for years)

### DIFF
--- a/extensions/BugModal/Extension.pm
+++ b/extensions/BugModal/Extension.pm
@@ -289,10 +289,11 @@ sub template_before_process {
     $vars->{tracking_flags_table} = \@tracking_table;
 
     # for the "view -> hide treeherder comments" menu item
-    my $treeherder_id = Bugzilla->treeherder_user->id;
+    my @treeherder_ids = map { $_->id } @{Bugzilla->treeherder_users};
     foreach my $change_set (@{ $bug->activity_stream }) {
-        if ($change_set->{comment} && $change_set->{comment}->author->id == $treeherder_id) {
-            $vars->{treeherder} = Bugzilla->treeherder_user;
+        if ($change_set->{comment} &&
+              grep { $change_set->{comment}->author->id == $_ } @treeherder_ids){
+            $vars->{treeherder_user_ids} = \@treeherder_ids;
             last;
         }
     }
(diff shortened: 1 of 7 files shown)


extensions/BugModal/lib/ActivityStream.pm
```diff
@@ -172,10 +172,11 @@ sub _add_comments_to_stream {
         next if $comment->body eq '' && ($comment->work_time - 0) != 0 && $user->is_timetracker;
 
         # treeherder is so spammy we hide its comments by default
-        if ($author_id == $treeherder_id) {
+        if (grep { $_ == $author_id  } @treeherder_ids) {
```
dylanwh: import any from `List::Util` and use *any* instead of grep, please.

purelogiq:
Video: https://drive.google.com/file/d/1SGVDudSzO42mazxVeOabBj4lH9Yls5_8/view?usp=sharing

extensions/BugModal/Extension.pm
```diff
@@ -289,10 +289,11 @@ sub template_before_process {
     $vars->{tracking_flags_table} = \@tracking_table;
 
     # for the "view -> hide treeherder comments" menu item
-    my $treeherder_id = Bugzilla->treeherder_user->id;
+    my @treeherder_ids = map { $_->id } @{Bugzilla->treeherder_users};
     foreach my $change_set (@{ $bug->activity_stream }) {
-        if ($change_set->{comment} && $change_set->{comment}->author->id == $treeherder_id) {
-            $vars->{treeherder} = Bugzilla->treeherder_user;
+        if ($change_set->{comment} &&
+              grep { $change_set->{comment}->author->id == $_ } @treeherder_ids){
```
dylanwh: import any from `List::Util` and use *any* instead of grep, please.

extensions/BugModal/web/comments.js
```diff
@@ -159,33 +159,39 @@ $(function() {
     $('#view-toggle-cc')
         .click(function() {
             var that = $(this);
-            var item = $('.context-menu-item.hover');
             if (that.data('shown') === '1') {
                 that.data('shown', '0');
-                item.text('Show CC Changes');
+                that.text('Show CC Changes');
```
dylanwh: What did this fix?

purelogiq: When you clicked on the button before it wouldn't toggle the text, it would be stuck on 'Show CC Changes'. Same thing with the Show/Hide Treeherder changes, thought I'd might as well make the quick fix while I was nearby

dylanwh: Make sure that's in the commit message (not in the summary, just in the body).

(as for the summary, I'm fine with "no bug - correctly collapse orange factor bot comments")

PR #909, opened 2018-11-15
Bug 1484892 - Modify EditComments extension to let anyone use it conditionally and support inline editing

dylanwh:
This is originally #736 but can't land until the schema change is in production


dylanwh:
Double check that I created this PR correctly.

PR #926, opened 2018-11-20
no bug - Include correct fontawesome file for PhabBugz

dklawren:


dklawren:
Also removed the full fontawesome font files i used as debugging and somehow snuck in there

PR #931, opened 2018-11-20
Bug 1508201 - Add framework for running background tasks

dylanwh:


Bugzilla/Job/RunTask.pm
```diff
@@ -0,0 +1,72 @@
...
+        encoding     => 'quoted-printable',
+      },
+      body_str => $html,
+    ),
+  );
+
+  my $email = Email::MIME->create(
+    header_str => [
+      From              => Bugzilla->params->{'mailfrom'},
+      To                => Bugzilla->params->{report_secbugs_emails},
```
globau: i suspect this should be the task's user, not `report_secbugs_emails`

dylanwh: woops.

dylanwh: fixed

PR #936, opened 2018-11-21
Bug 1456878 - Support markdown comments

purelogiq:
Adds support for markdown comments, notable features:
- Markdown rendering with the GFM library
- All new comments are treated as markdown
- new lines treated as new lines (hardbreaks extension on); this makes it so that you don't need 2 spaces a the end of a line to make a new line and is what github does too.
- BMO link previews for bug shorthand ~~as well as custom markdown links~~ (link previews on markdown links left for another day).
  - quoteUrls is used on the actual text nodes after markdown html is rendered.
- Works with comment editing
- Quoting a markdown comment will fetch the raw markdown
- Does *not* attempt to update the non-Bug Modal ui with markdown rendering, as requested in https://bugzilla.mozilla.org/show_bug.cgi?id=1456878#c9
- Adds a 'use_markdown' feature flag to the Advanced parameters. Things work as expected when on vs off.


vagrant_support/checksetup_answers.j2
```diff
@@ -10,7 +10,7 @@ $answer{'create_htaccess'}      = '1';
 $answer{'db_check'}             = 1;
 $answer{'db_driver'}            = 'mysql';
 $answer{'db_host'}              = '{{DB_IP}}';
-$answer{'db_name'}              = 'bugs_bmo';
+$answer{'db_name'}              = 'bugs_bmo_full';
```
purelogiq: ignore me

Bugzilla.pm
```diff
@@ -761,10 +761,9 @@ sub check_rate_limit {
 }
 
 sub markdown_parser {
```
dylanwh: looks like this should be renamed 'markdown', per the test that is failing:
`Can't locate object method "markdown" via package "Bugzilla" at t/markdown.t line 21.`

Bugzilla/Markdown.pm
```diff
@@ -0,0 +1,76 @@
...
+        extensions => [qw( autolink tagfilter table strikethrough )],
+      }
+    );
+  }
+  else {
+    return undef;
+  }
+}
+
+sub render_html {
```
purelogiq: Finally managed to figure out why quoteUrls wasn't being called at all, keen to get feedback here @dylanwh, thanks!

t/markdown.t
```diff
@@ -39,29 +46,4 @@ is(
   'Strikethrough extension'
 );
 
-my $table_markdown = <<'MARKDOWN';
-| Col1 | Col2 |
-| ---- |:----:|
-| val1 | val2 |
-MARKDOWN
-
-my $table_html = <<'HTML';
```
purelogiq: Have absolutely no idea why this happens, but this table test seems to try to connect to the database?! Thus causing test_santiy to fail. Is there's some magic that causes a plain html string to make bmo want to connect to the database? Either way I feel like removing this test is not the end of the world.

purelogiq:
Very keen to get feedback on email rendering too, thanks!

extensions/BugModal/web/bug_modal.css
```diff
@@ -781,7 +820,7 @@ body.platform-Win32 .comment-text, body.platform-Win64 .comment-text {
     clear: both;
     width: 100%;
     box-sizing: border-box !important;
-    margin: 0 0 1em;
+    margin: 0 0 0.5em;
```
dylanwh: @purelogiq is this change from my merge?

purelogiq: I don't think so, I remember intentionally changing that because of there being more default spacing with <p> tags vs <pre> tags. Thinking of reverting this?

Bugzilla/Bug.pm
```diff
@@ -995,6 +995,7 @@ sub create {
 
   # We now have a bug id so we can fill this out
   $creation_comment->{'bug_id'} = $bug->id;
+  $creation_comment->{'is_markdown'} = 1;
```
dylanwh: This is where a check for Bugzilla->params->{use_markdown} would be used

Bugzilla/Bug.pm
```diff
@@ -2727,7 +2728,8 @@ sub set_all {
       $params->{'comment'}->{'body'},
       {
         isprivate => $params->{'comment'}->{'is_private'},
-        work_time => $params->{'work_time'}
+        work_time => $params->{'work_time'},
+        is_markdown => 1
```
dylanwh: This is where a check for Bugzilla->params->{use_markdown} would be used

Bugzilla/Template.pm
```diff
@@ -739,6 +739,22 @@ sub create {
         1
       ],
 
+      renderMarkdown => [
+        sub {
+            my ($context, $bug, $comment, $user) = @_;
+            return sub {
+                my $text = shift;
+                if($comment && $comment->is_markdown) {
```
dylanwh: Add && Bugzilla->params->{use_markdown} so we can force the old behavior if needed

Bugzilla/WebService/Bug.pm
```diff
@@ -370,7 +370,7 @@ sub render_comment {
   Bugzilla->switch_to_shadow_db();
   my $bug = $params->{id} ? Bugzilla::Bug->check($params->{id}) : undef;
 
-  my $html = Bugzilla::Template::quoteUrls($params->{text}, $bug);
+  my $html = Bugzilla->markdown->render_html($params->{text}, $bug);
```
dylanwh: Add a conditional based on Bugzilla->params->{use_markdown}

extensions/BMO/template/en/default/email/bugmail.html.tmpl
```diff
@@ -50,7 +50,12 @@
               at [% comment.creation_ts FILTER time(undef, to_user.timezone) %]
             </b>
           [% END %]
-          <pre class="comment" style="font-size: initial">[% comment.body_full({ wrap => 1 }) FILTER quoteUrls(bug, comment) %]</pre>
+          [% IF comment.is_markdown %]
```
dylanwh: [% IF  comment.is_markdown  AND Param('use_markdown') %]

extensions/BugModal/template/en/default/bug_modal/activity_stream.html.tmpl
```diff
@@ -261,16 +261,25 @@
 [% END %]
 
 [% BLOCK comment_body %]
-  <pre class="comment-text [%= "bz_private" IF comment.is_private %]" id="ct-[% comment.count FILTER none %]"
-    [% IF comment.collapsed +%] style="display:none"[% END ~%]
+  [% IF comment.is_markdown %]
```
dylanwh: [% IF comment.is_markdown AND Param('use_markdown') %]

extensions/BugModal/web/bug_modal.css
dylanwh: Nah, just checking if I screwed up the merge

dylanwh:
Testing an input: <input type="text" name="foo">

Bugzilla/Markdown.pm
```diff
@@ -0,0 +1,73 @@
...
+  my ($self, $markdown, $bug, $comment, $user) = @_;
+  my $parser             = $self->markdown_parser;
+  my $bugzilla_shorthand = $self->bugzilla_shorthand;
+  my @valid_text_tags    = ('p', 'li', 'td');
+
+  if ($parser) {
+    my $html = decode('UTF-8', $parser->render_html($markdown));
+    my $dom = Mojo::DOM->new($html);
```
dylanwh: For now, let's strip img tags (because they could be used as a DDoS, e.g. a few hundred img tags sent to a bug page could trigger an ip block),
and input, button, form, ... any other dangerous ones tagfilter doesn't already fix?
Note github appears to silently remove input too.

```perl
my @bad_tags = qw( form button input );
$dom->find( join(', ', @bad_tags) )->map('remove')
```

purelogiq:
Additional styles were added to support the attachment comments being markdown (so that the preview renders properly)

template/en/default/global/header.html.tmpl
```diff
@@ -120,6 +120,7 @@
             },
             constant => {
                 COMMENT_COLS => constants.COMMENT_COLS,
+                URL_BASE => urlbase,
```
dylanwh: I think you can avoid adding this and use BUGZILLA.config.basepath

purelogiq: This was used for the smartlinking stuff, which was removed anyway.

PR #937, opened 2018-11-21
Use immutable http directive for immutable assets

staabm:
This directive saves re-requesting immutable resources on browser hard reloads

See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control


staabm:
Thx for the review. Do you want me to drop the .htaccess with this PR or should I just revert my changes to it?

dylanwh:
@staabm delete the .htaccess too. It has no purpose.

staabm:
Deleted as requested.

dylanwh:
@staabm was there a bug for this?

staabm:
Not to my knowledge

dylanwh:
Ah, please file a bug for this change then, I'll land it.
Component: General,
product: bugzilla.mozilla.org

thanks!

staabm:
Filed 
https://bugzilla.mozilla.org/show_bug.cgi?id=1510021

PR #939, opened 2018-11-22
Bug 1509308 - Date for cf_last_resolved in history results should be UTC

kyoshino:
Return date/time (and double) values in an appropriate data type instead of string. If the value is empty, make sure an empty string will be returned. Otherwise, return a UTC date in the ISO 8601 format.

## Bugzilla link

[Bug 1509308 - Date for cf_last_resolved in history results should be UTC](https://bugzilla.mozilla.org/show_bug.cgi?id=1509308)


kyoshino:
This PR just fixes the date/time format, not the reported legacy time zone issue.

kyoshino:
Adding a conversion script.

PR #941, opened 2018-11-27
Bug 1500441 - Make site-wide announcement dismissable

dylanwh:


Bugzilla/Quantum.pm
```diff
@@ -12,6 +12,7 @@ use Mojo::Base 'Mojolicious';
 use CGI::Compile;
 use utf8;
 use Encode;
+use FileHandle;    # this is for compat back to 5.10
```
dylanwh: This wasn't related to this commit, except without it the vagrant vm doesn't work for me.

dylanwh:
@kyoshino this is ready for review again. Or maybe you can assign to @purelogiq ?

dylanwh:
There is no user session at the moment, unfortunately. Not that we could tack this on. Note this is also client-side in behavior.

PR #942, opened 2018-11-27
Bug 1510109 - Implement per-product new bug comment templates

kyoshino:
* Allow Bugzilla admins to define a per-product new bug comment template on the Edit Product page. [Screenshot](https://screenshots.firefox.com/8EvjYUvtfAdZucfm/bmo-web.vm)
* Prefill the comment box on the New Bug page with the per-product template if defined.

## Bugzilla link

[Bug 1510109 - Implement per-product new bug comment templates](https://bugzilla.mozilla.org/show_bug.cgi?id=1510109)

### DIFF
--- a/Bugzilla/Product.pm
+++ b/Bugzilla/Product.pm
@@ -55,6 +55,7 @@ use constant UPDATE_COLUMNS => qw(
   defaultmilestone
   isactive
   allows_unconfirmed
+  bug_description_template
 );
 
 use constant VALIDATORS => {
@@ -521,6 +522,7 @@ sub set_description        { $_[0]->set('description',        $_[1]); }
 sub set_default_milestone  { $_[0]->set('defaultmilestone',   $_[1]); }
 sub set_is_active          { $_[0]->set('isactive',           $_[1]); }
 sub set_allows_unconfirmed { $_[0]->set('allows_unconfirmed', $_[1]); }
+sub set_bug_description_template { $_[0]->set('bug_description_template', $_[1]); }
 
 sub set_group_controls {
   my ($self, $group, $settings) = @_;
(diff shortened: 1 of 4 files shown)


Bugzilla/Product.pm
```diff
@@ -47,6 +47,7 @@ use constant DB_COLUMNS => qw(
    isactive
    defaultmilestone
    allows_unconfirmed
+   bug_comment_template
```
dylanwh: I'm concerned this will increase memory usage.
I think we can lazy-load this like the securemail extension with the public_key
field.

First remove it from DB_COLUMNS and then:

```perl
# We want to lazy-load the public_key.
sub bug_comment_template {
    my $self = shift;
    if (!exists $self->{public_key}) {
        ($self->{bug_comment_template}) = Bugzilla->dbh->selectrow_array(
            "SELECT bug_comment_template FROM products WHERE id = ?",
            undef,
            $self->id
        );
    }
    return $self->{bug_comment_template};
}
```

kyoshino:
Fixed the test. Renamed the column with lazy-load implemented.

dylanwh:
Since this is a schema change, normally we'd need to split it into two, but it's possible to make this mostly work with the existing schema.

Add `use Try::Tiny;` to the top, and then:

```perl
sub bug_description_template {
  my $self = shift;
  if (!exists $self->{'bug_description_template'}) {
    try {
      $self->{'bug_description_template'}
        = Bugzilla->dbh->selectrow_array(
        'SELECT bug_description_template FROM products WHERE id = ?',
        undef, $self->id);
    }
    catch {
      if (/Unknown column 'bug_description_template'/) {
        $self->{'bug_description_template'} = '';
      }
      else {
        die $_;
      }
    };
  }
  return $self->{'bug_description_template'};
}
```

kyoshino:
The schema change has been merged and deployed today.

PR #945, opened 2018-11-27
Bug 1508385 - Remove links to input.mozilla.org from Guided Bug Entry flow

dklawren:


template/en/default/bug/create/create-guided.html.tmpl
```diff
@@ -235,10 +235,7 @@ function PutDescription() {
     <td valign="top">
       <input type="text" size="80" name="bug_file_loc" placeholder="https://">
       <p>
-        URL that demonstrates the problem you are seeing (optional).<br>
-        <b>IMPORTANT</b>: if the problem is with a broken web page, you need
-        to report it
-        <a href="https://input.mozilla.org/feedback">a different way</a>.
```
emceeaich: I wonder if this link should be changed to webCompat? @miketaylr?

PR #946, opened 2018-11-27
make user suggest faster by doing less work

dylanwh:
I don't think we need to load the request info for every user suggestion,
so this adds a fast_mode option that avoids calling the hook.

### DIFF
--- a/Bugzilla/WebService/User.pm
+++ b/Bugzilla/WebService/User.pm
@@ -197,8 +197,10 @@ sub suggest {
         }
     } @$results;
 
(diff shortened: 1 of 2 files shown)


kyoshino:
The requests info was added for #684 and it has never been used. Maybe we just have to add the `extra` filter to [`filter_wants()`](https://github.com/mozilla-bteam/bmo/blob/master/extensions/Review/Extension.pm#L1077) like #896 (and use `include_fields=_default,_extra` in #684)?

PR #951, opened 2018-11-28
Bug 1510770 - Clicking Edit Comment Button before clicking Edit Bug blocks editing of comment

dylanwh:


extensions/EditComments/template/en/default/hook/global/header-start.html.tmpl
```diff
@@ -7,7 +7,7 @@
   #%]
 
 [%
-  RETURN UNLESS template.name == 'bug/show-modal.html.tmpl';
+  RETURN UNLESS bugs && bugs.0.id && bugs.size == 1;
```
kyoshino: Tested this locally but not working. Maybe it just requires an additional template name like [Needinfo’s template](https://github.com/mozilla-bteam/bmo/blob/master/extensions/Needinfo/template/en/default/hook/global/header-start.html.tmpl)? This works for me:

```suggestion
  RETURN UNLESS template.name == 'bug/show-modal.html.tmpl'
    || template.name == 'bug/process/results.html.tmpl';
```

PR #959, opened 2018-11-30
Bug 1511466 - The /api/user/profile API should provide the user's MFA status

dylanwh:


Bugzilla/Quantum/API.pm
```diff
@@ -21,6 +22,8 @@ sub user_profile {
         login  => $user->login,
         nick   => $user->nick,
         groups => [map { $_->name } @{$user->groups}],
+        mfa    => $user->mfa,
```
dklawren: We should lowercase this value before returning.

PR #960, opened 2018-12-04
Bug 1512024 - Bugzilla sets two different Referrer Policies

dylanwh:


kyoshino:
This solves the inconsistency but I think the `<meta name="referrer">` tag can be removed anyway if the HTTP header can be set.

PR #971, opened 2018-12-10
Bug 1513087 - User preferences page is totally empty with Template v2.28

dylanwh:
With the latest version of template toolkit, the use preferences pages are empty.

This seems like a bug-fix upstream, rather than a regression. The reason is we're calling tabs = tabs.import() in many extensions..

It doesn't appear that hash.import() returns the hash, and this was working because the context of the template was localized.

https://metacpan.org/pod/distribution/Template-Toolkit/lib/Template/Manual/VMethods.pod#import


mohawk2:
This change looks fine. Would love to approve, but no auth ;-)

PR #973, opened 2018-12-11
Bug 1508201 - Allow bulk edits to happen in the background, asynchronously

dylanwh:
Updated with (I think) all review changes applied.

### DIFF
--- a/process_bug.cgi
+++ b/process_bug.cgi
@@ -276,6 +276,10 @@ my %set_all_fields;
 if (not $bulk_edit) {
   $set_all_fields{other_bugs} = \@bug_objects;
 }
+else {
+  require Bugzilla::Object::Lazy;
+  $set_all_fields{other_bugs} = [ map { Bugzilla::Object::Lazy->new(id => $_) } @bug_ids ];
+}
 foreach my $field_name (@set_fields) {
   if (should_set($field_name, 1)) {
     my $param_name = $field_translation{$field_name} || $field_name;
--- a/process_bug.cgi
+++ b/process_bug.cgi
@@ -76,10 +76,10 @@ if (Bugzilla->params->{disable_bug_updates}) {
 my @bug_objects;
 my @bug_ids;
 
-# $bulk_edit is whether or not we use the bulk editing code,
+# $async_bulk_edit is whether or not we use the bulk editing code,
 # $background is if the background=1 parameter was passed.
-my $bulk_edit = $cgi->param('bulk_edit') // 0;
-$cgi->delete('bulk_edit') if $bulk_edit;
+my $async_bulk_edit = $cgi->param('async_bulk_edit') // 0;
+$cgi->delete('async_bulk_edit') if $async_bulk_edit;
 
 if (defined $cgi->param('id')) {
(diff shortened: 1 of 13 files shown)


dklawren:
diff --git a/t/bulk-edit.t b/t/bulk-edit.t                                                                                                                                                                            index b6aebec10..a97fa7f98 100644                                                                                                                                                                                     --- a/t/bulk-edit.t                                                                                                                                                                                                   +++ b/t/bulk-edit.t                                                                                                                                                                                                   @@ -1,4 +1,4 @@                                                                                                                                                                                                       -#!/usr/bin/perl                                                                                                                                                                                                      +#!/usr/bin/env perl                                                                                                                                                                                                   # This Source Code Form is subject to the terms of the Mozilla Public                                                                                                                                                 # License, v. 2.0. If a copy of the MPL was not distributed with this                                                                                                                                                 # file, You can obtain one at http://mozilla.org/MPL/2.0/.

template/en/default/task/created.html.tmpl
```diff
@@ -0,0 +1,14 @@
...
+
+[% PROCESS global/header.html.tmpl
+   title = task.subject,
+   header = task.subject,
+   style_urls = [ 'skins/standard/index.css' ]
+   no_yui = 1
+%]
+
+<div id="page-index">
+<p>Scheduled [% task.proper_name FILTER html %] task. You'll recieve an email when it is done.</p>.
```
dklawren: Rogue period at the end.

dylanwh:
> process_bug.cgi does not check that user is in group bulk_edit that I can tell and allows anyone to bulk edit. But the template does check. Which policy do we want?
It did, but it was probably hard to see. Now it's `bz_can_async_bulk_edit`

process_bug.cgi
```diff
@@ -73,23 +74,36 @@ if (Bugzilla->params->{disable_bug_updates}) {
...
-  foreach my $i ($cgi->param()) {
-    if ($i =~ /^id_([1-9][0-9]*)/) {
-      my $id = $1;
-      push(@bug_objects, Bugzilla::Bug->check($id));
-    }
+  @bug_ids = map { /^id_([1-9][0-9]*)/ ? $1 : () } $cgi->param;
+
+  # Make sure there are bugs to process.
+  ThrowUserError("no_bugs_chosen", {action => 'modify'}) unless @bug_ids;
+  $async_bulk_edit = 1 if @bug_ids > 10;
```
dklawren: Just noticed this. Seems that even if the user does not check the bulk edit checkbox on buglist.cgi, it will still enable it if @bug_ids > 10 which negates the need for a checkbox altogether.

Maybe this should be instead:

$async_bulk_edit = 0 if @bug_ids < 10; 

Or am I missing something?

dylanwh: right now, doing > 10 edits is likely to fail, so it's meant to make them more likely to succeed. Eventually bz_async_bulk_edit should be the same as editbugs. Does that make sense?

dklawren: Sounds fine. But is still sort of misleading from a user standpoint. Maybe we should add some text or tooltip that states it is automatic if more than 10 bugs?

PR #976, opened 2018-12-13
Bug 1506725 - Update form.doc to direct reporters to GitHub

dklawren:


dklawren:
> Do we actually need the "we have moved" message? In Bug 1513644 I'm thinking of directly redirecting 
> these URLs to new destinations. I haven't checked how to do it in Mojo though.

In https://bugzilla.mozilla.org/show_bug.cgi?id=1506725#c1 it was requested not to be a redirect.

dkl

jmswisher:
How can I help move this forward? Or at least revert the change that was made? The current state is a regression, because submissions to the "developer.mozilla.org" product are getting the "MDN Web Docs feedback has moved" page, which is incorrect. Nothing is supposed to happen to that product; the requested change was for "Developer Documentation" only.

dylanwh:
We can backout the change. @kyoshino can we do that tonight/tomorrow?

jmswisher:
Any idea when this change will be pushed? We're getting dev.mo bugs filed on GitHub, which is the wrong place for them. Thanks!

dylanwh:
this will go out on Monday -- which is when the push freeze is over. Push freezes are unusual for us and normally stuff gets out pretty fast. Sorry for the trouble!

PR #981, opened 2018-12-17
Bug 1514765 - fix new bug comment preview

purelogiq:
We forgot to correctly switch between a div and pre tag on the new
bug comment preview, based on whether markdown is enabled or not.
This was causing plaintext comments to lose their formatting (only when previewed).


dylanwh:
@purelogiq this only failed because it failed to find a bug in the commit messages. An edge case that happens whenever master is as current as production.

PR #983, opened 2018-12-18
Bug 1515144 - Prevent Google Analytics script from being loaded when DNT:1 so that Tracking Protection doesn't have to block it

kyoshino:
* Stop loading any of analytics-related scripts when DNT is on
* Rely on the `DNT` HTTP header instead of a helper script since we don't support legacy browsers
* Tested locally with the debug version of GA

## Bugzilla link

[Bug 1515144 - Prevent Google Analytics script from being loaded when DNT:1 so that Tracking Protection doesn't have to block it](https://bugzilla.mozilla.org/show_bug.cgi?id=1515144)


kyoshino:
By the way, `c.req.headers.dnt` didn't work so I used the legacy way: `Bugzilla.cgi.http('dnt')`. Maybe a [Mojo plug-in](https://metacpan.org/pod/Mojolicious::Plugin::TtRenderer)  is required?

PR #984, opened 2018-12-18
Bug 1511490 - BMO's oauth tokens should be use jwt

dklawren:


dylanwh:
Problem:

```
DBD::mysql::db do failed: Cannot add foreign key constraint [for Statement "ALTER TABLE oauth2_refresh_token ADD
     CONSTRAINT fk_oauth2_refresh_token_client_id_oauth2_client_id FOREIGN KEY (client_id)
     REFERENCES oauth2_client(id)
      ON UPDATE CASCADE ON DELETE CASCADE, ADD
     CONSTRAINT fk_486882066ec536cce17ab98bcfd22f2a FOREIGN KEY (access_token_id)
     REFERENCES oauth2_access_token(access_token)
      ON UPDATE CASCADE ON DELETE CASCADE"] at /vagrant/Bugzilla/DB.pm line 55.
vagrant /vagrant $
```

t/mojo-oauth2.t
```diff
@@ -13,6 +13,7 @@ use lib qw( . lib local/lib/perl5 );
 BEGIN {
   $ENV{LOG4PERL_CONFIG_FILE}     = 'log4perl-t.conf';
   $ENV{BUGZILLA_DISABLE_HOSTAGE} = 1;
+  $ENV{MOJO_TEST}                = 1;
```
dylanwh: This is similar to `BUGZILLA_UNSAFE_AUTH_DELEGATION`, I wonder if shouldn't rename both to something like `BUGZILLA_ALLOW_UNSAFE_HTTP`?

dylanwh: s/UNSAFE/INSECURE/

dylanwh:
@dklawren want this to ride out to -dev this cycle or wait?

t/mojo-oauth2.t
dklawren: Problem is since we are using Test::Mojo, we are just passing routes in the test suite and not fully qualified URLs. So we can't even check for any schemes from what I can tell.

PR #988, opened 2018-12-20
added is_new to Bugzilla.user object

spiro:
Bug 989476 - added is_new method to Bugzilla.user object


spiro:
@dylanwh please ignore commits before last one. I had some problems with git, hence started with fresh code again.
Also suggest changes on commit (if needed).

PR #989, opened 2018-12-20
Bug 1508389 - Change links for WebCompat reporting

dklawren:


dklawren:
Originally we were going to remove references to Core::Triaged in bug 1507265 but I am fine with doing some of the work here as well. New patch coming up.

PR #994, opened 2018-12-27
Bug 1516188 : Redesigned the logic of rowspan of image

spiro:


spiro:
So here goes the logic:

now "offset" => for button "Edit your preferences". this button is set to take rowspan of 2 (same as previous version)
Also one extra rowspan value is added if the user is logged in (same as previous version).
Hence maximum value will be 8.

PR #995, opened 2018-12-27

spiro:


spiro:
So here goes the logic:
"edit_preferences_offset" => if the user is logged in or not.

PR #996, opened 2018-12-27
Bug 1516188 : Redesigned the logic of rowspan of image

spiro:

### DIFF
--- a/extensions/UserProfile/template/en/default/pages/user_profile.html.tmpl
+++ b/extensions/UserProfile/template/en/default/pages/user_profile.html.tmpl
@@ -49,7 +49,7 @@
   [% permissions_offset = target.in_group("canconfirm") == 1 ? 1 : 0 %]
   [% creation_offset = target.can('creation_ts') == 1 ? 1 : 0 %]
   [% edit_preferences_offset = user.id ? 1 : 0 %]
-   <td rowspan="[% 3 + edit_preferences_offset + permissions_offset + creation_offset %]" id="gravatar-container">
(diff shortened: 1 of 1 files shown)


spiro:
So here goes the logic:
"edit_preferences_offset" => if the user is logged in or not.

extensions/UserProfile/template/en/default/pages/user_profile.html.tmpl
```diff
@@ -46,8 +46,10 @@
 [% END %]
 
 <tr>
-  [% offset = user.id == target.id ? 2 : 0 %]
-  <td rowspan="[% user.id ? 6 + offset : 5 + offset %]" id="gravatar-container">
+  [% permissions_offset = target.in_group("canconfirm") == 1 ? 1 : 0 %]
+  [% creation_offset = target.can('creation_ts') == 1 ? 1 : 0 %]
+  [% edit_preferences_offset = user.id ? 1 : 0 %]
+   <td rowspan="[% 3 + edit_preferences_offset + permissions_offset + creation_offset %]" id="gravatar-container">
```
purelogiq: This works great! One thing: on this line it is good practice to add `FILTER html` to the end of your template variable (e.g. `[% 3 + edit_preferences_offset + permissions_offset + creation_offset FILTER html %]`). FILTER html just escapes html tags before putting the result in the template. Although it is clearly just an integer, it protects us in case someone changes things in the future.

spiro:
@purelogiq  I have updated the patch. Kindly review it.

spiro:
@purelogiq  can you change status of bug on bugzilla too?

purelogiq:
Done! :) Thanks for the patch!

PR #997, opened 2019-01-02
remove index.html stuff

dylanwh:
index.html is never used, so we should remove that option.
This also removes the useless google verification html file too.


dylanwh:
No that's from some other google integration before my time

PR #1000, opened 2019-01-02
[WIP] Merge bugzilla-todos into BMO

dylanwh:
@kyoshino -- can I use something like document.domain to get the hostname + schema of where a .js file was loaded? With that + insuring XMLHTTPRequest (ugh) sends credentials, people can continue using bugzilla-todos at https://bugzilla.mozilla.org/todos without hitting rate limits.


dylanwh:
@emceeaich I'm thinking this might be important for some people. Talk to mikeratcliffe to get an idea of numbers.

Finishing this up is largely a matter of adding credentials to its outgoing requests.

MikeRatcliffe:
A huge proportion of Mozilla employees I know use bugzilla-todos as their main dashboard.

Some have stopped using bugzilla-todos because when we moved to phabricator it stopped showing pending reviews so it would be awesome if we could find a fix for that.

globau:

PR #1006, opened 2019-01-08
use the same font size and line height as github for markdown comments

dylanwh:


dylanwh:
How about adopting the 21px line-height of github as well?

kyoshino:
Sounds good (and looks good).

PR #1007, opened 2019-01-08
Bug 1518381 - lightbox is no longer used when clicking on an attachment's link in markdown comments

dylanwh:
The link to the attachment is under a paragraph tag if markdown is enabled.


purelogiq:
Oops, this doesn't fix https://bugzilla.mozilla.org/show_bug.cgi?id=1518218

PR #1008, opened 2019-01-08
Bug 1518380 - reduce the margins of blockquotes in html comments

dylanwh:


extensions/BugModal/web/bug_modal.css
```diff
@@ -729,6 +729,15 @@ div.comment-text blockquote {
     padding: 0.5em 10px;
 }
 
+
+div.comment-text blockquote p:first-child {
```
purelogiq: Just giving my two cents here: I think removing the `p` specificity would be good. Some blockquotes start/end with H1, or other tags and those would still have the big margins. E.g. `div.comment-text blockquote :first-child`.
This change successfully reduces top/bottom margins on the blockquotes, an oversight on my part!

extensions/BugModal/web/bug_modal.css
```diff
@@ -729,6 +729,15 @@ div.comment-text blockquote {
...
+    margin-top: 0;
+}
+
+div.comment-text blockquote p:last-child {
```
purelogiq: Here as well.

extensions/BugModal/web/bug_modal.css
kyoshino: Agree with @purelogiq. It would be:

```suggestion
div.comment-text blockquote > :first-child {
```

extensions/BugModal/web/bug_modal.css
kyoshino:
```suggestion
div.comment-text blockquote > :last-child {
```

PR #1009, opened 2019-01-08
Bug 1518266 - Font of comment emails is rendered too large since markdown has been enabled

dylanwh:
I verified this change looks ok in apple mail, I'll try to get it into gmail as well. Do you happen to have a working thunderbird setup?

### DIFF
--- a/extensions/BMO/template/en/default/email/bugmail.html.tmpl
+++ b/extensions/BMO/template/en/default/email/bugmail.html.tmpl
@@ -13,9 +13,6 @@
 <html>
 <head>
   <base href="[% urlbase FILTER html %]">
-  <style>
-    .comment { font-size: initial !important; }
-  </style>
 </head>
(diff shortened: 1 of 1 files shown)


template/en/default/bug/comment.html.tmpl
```diff
@@ -37,7 +37,7 @@
   <div id="comment_preview" class="bz_default_hidden bz_comment">
     <div id="comment_preview_loading" class="bz_default_hidden">Generating Preview...</div>
     <div id="comment_preview_error" class="bz_default_hidden"></div>
-    [% IF comment.is_markdown AND Param('use_markdown') %]
+    [% IF ( comment.is_markdown OR NOT comment.id ) AND Param('use_markdown') %]
```
purelogiq: Rebase? Or does github do that nowadays

dylanwh:
Oh jeeze, that's fixed now.

PR #1011, opened 2019-01-08
Bug 1518218 - The attachment icon is misplaced

dylanwh:
As a quick fix, we will hide the icon until something better can be done.


extensions/BugModal/web/bug_modal.css
```diff
@@ -980,6 +980,10 @@ div.ui-tooltip {
     vertical-align: sub;
 }
 
+a.lightbox-icon {
```
kyoshino: You mean:
```suggestion
a.lightbox-icon.markdown {
```

PR #1018, opened 2019-01-08
Bug 1510996 - Abandoned phabricator revisions should be hidden by default

dklawren:


extensions/PhabBugz/template/en/default/phabricator/table.html.tmpl
```diff
@@ -26,4 +26,10 @@
         <span class="phabricator-load-error-string"></span></td>
     </tr>
   </tbody>
+  <tr>
```
kyoshino: This `<tr>` has to be inside `<tbody>` and probably hidden when there are no abandoned revisions? I honestly don’t know how to test this locally 😅

dklawren: I put the checkbox row now in a <tbody> of its own and hide it by default. I also do not show the row if there are no abandoned revisions as you suggested. If the JS and template changes check out OK, then I can put this branch on bugzilla-dev and you can test it functionally there.

kyoshino:
`.phabricator.js.swp` and a blank line in `phabricator.css` have to be removed.

dklawren:
> 
> 
> `.phabricator.js.swp` and a blank line in `phabricator.css` have to be removed.

Oops. Will fix :)

PR #1019, opened 2019-01-09
Bug 1509329 - Do not display revisions that have moved out of the bug, but note the move in the bug history

dklawren:


dklawren:
This has been tested successfully on bugzilla-dev with the latest build. Just need rubber stamp from @dylanwh and this can go out.

PR #1020, opened 2019-01-09
only change the line-height of paragraphs

dylanwh:
https://bugzilla.mozilla.org/show_bug.cgi?id=1518238#c0 Looks pretty weird, we should only use the 
21px line height on paragraphs. Also, I think setting a max-width on paragraphs may make sense?


purelogiq:
LGTM

PR #1034, opened 2019-01-14
Bug 1519659 - Need to reload the page before being able to edit

kyoshino:
The page displayed once a new bug is filed uses a different template. The strings for the Edit Comments extension have to be included in the template just like #951. Tested locally.

## Bugzilla link

[Bug 1519659 - Need to reload the page before being able to edit](https://bugzilla.mozilla.org/show_bug.cgi?id=1519659)


kyoshino:
Hmm, a test failed due to a timeout? I'm force-pushing to rerun.

PR #1048, opened 2019-01-17
Bug 1518268 - Re-style all markdown content, consistently

purelogiq:
Contains all the changes in #1027 plus fixes
- Windows users fonts were all monospace
- Preview for long comments on bug page flowed outside of the box

- @dylanwh / @kyoshino I realized that when editing an old non-markdown comment, the comment preview is rendered by /rest/bug/comment/render which always chooses to render as markdown (since it doesn't have the comment object, so it can't know whether you're editing a non-markdown comment or not). Do you guys want me to modify the api to allow a 'as_plaintext' flag or something? Dunno if that will have any repercussions down the line. Or we could pass in the comment id.

### DIFF
--- a/extensions/BugModal/template/en/default/bug_modal/activity_stream.html.tmpl
+++ b/extensions/BugModal/template/en/default/bug_modal/activity_stream.html.tmpl
@@ -267,7 +267,7 @@
     [% comment_tag = 'pre' %]
   [% END %]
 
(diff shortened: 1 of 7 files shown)


dylanwh:
The API problem is a different bug

purelogiq:
@dylanwh Do we want to land this just yet (also is there a bug number for these changes?) Would love to have @kyoshino to take a look, but, let me know!

dylanwh:
We'll commit and send this to -dev for sign-off / tweaks from kohei

PR #1052, opened 2019-01-18
no bug - Pass 'name' to active_custom_fields if include_fields is defined

dklawren:


Bugzilla/WebService/Bug.pm
```diff
@@ -1522,11 +1522,18 @@ sub _bug_to_hash {
...
     product   => $bug->product_obj,
     component => $bug->component_obj,
     bug_id    => $bug->id
-  });
+  };
+  if (defined $params->{include_fields} && !defined $params->{exclude_fields}) {
+    $field_params->{name}
+      = ref $params->{include_fields}
+      ? $params->{include_fields}
+      : [split(/[\s,]+/, $params->{include_fields})];
```
dylanwh: I think this is always an arrayref

Bugzilla.pm
```diff
@@ -685,8 +685,10 @@ sub active_custom_fields {
     $cache_id .= ':noext' if $params->{skip_extensions};
   }
   if (!$can_cache || !exists request_cache->{$cache_id}) {
+    my $field_params = {custom => 1, obsolete => 0, skip_extensions => 1};
+    $field_params->{name} = $params->{name} if defined $params->{name};
     my $fields
-      = Bugzilla::Field->match({custom => 1, obsolete => 0, skip_extensions => 1});
+      = Bugzilla::Field->match($field_params);
```
dylanwh: we can use PerlX::Maybe here to make this cleaner.

PR #1061, opened 2019-01-23
Bug 1512815 - Make Bugzilla->active_custom_wants() 10x faster when using specific include_fields

dylanwh:
|             | real             | user           | sys             |
|   ----   | ----            | ----             | ----- |
| Before | 0m23.677s | 0m11.010s |  0m0.709s |
| After |  0m2.361s     | 0m1.624s  |  0m0.172s  |

### DIFF
--- a/Bugzilla.pm
+++ b/Bugzilla.pm
@@ -702,7 +702,7 @@ sub active_custom_fields {
     }
     my $fields = Bugzilla::Field->match($match_params);
     Bugzilla::Hook::process('active_custom_fields',
-      {fields => \$fields, params => $params});
+      {fields => \$fields, params => $params, wants => $wants});
     request_cache->{$cache_id} = $fields if $can_cache;
     return @$fields;
   }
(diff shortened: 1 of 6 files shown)


Bugzilla.pm
```diff
@@ -676,29 +676,43 @@ sub fields {
...
+    if ($wants) {
+      if ($wants->exclude->{custom}) {
+        return ();
+      }
+      elsif ($wants->exclude->{extensions}) {
+        $exclude_extensions = 1;
+      }
+      elsif ($wants->is_specific) {
+        my $names = [ grep { /^cf_/ } $wants->includes ];
+        $match_params->{name} = $names;
```
dylanwh: if `$names` is empty, we can also `return ()` here.

dylanwh:
Finally fixed the tests \o/

extensions/TrackingFlags/Extension.pm
```diff
@@ -301,12 +301,16 @@ sub install_filesystem {
 
 sub active_custom_fields {
   my ($self, $args) = @_;
+  my $wants     = $args->{'wants'};
```
dklawren: This is not passed in by Bugzilla->active_custom_fields so never used.

PR #1066, opened 2019-01-25
no bug - Update README and VS Code tasks for morbo dev server

purelogiq:
Updates the readme to let new users know that they have to run
`start_morbo` before they can visit bmo-web.vm.

Also updates the VS Code tasks to work again.


purelogiq:
@dylanwh Morbo is awesome, it's a much faster dev cycle not having to wait for a provision every change. Especially interested to get @kyoshino's review since he made the original VS Code tasks.

PR #1067, opened 2019-01-25
Bug 1461492 - Add an optional regressed-by field in bugs

kyoshino:
Add the new regression fields, **Regressed by** and **Regresses**, based on the current **Depends on** and **Blocks** implementation. In addition to bug pages, these fields will appear everywhere including the search page and bug lists, and are also available via the API. Views corresponding to dependency trees/graphs are not yet implemented.

This is a built-in feature rather than custom fields provided by an extension. Bugzilla admins can disable the functionality by using the `use_regression_fields` parameter, if needed.

## Bugzilla link

[Bug 1461492 - Add an optional regressed-by field in bugs](https://bugzilla.mozilla.org/show_bug.cgi?id=1461492)


Bugzilla/BugMail.pm
```diff
@@ -133,10 +133,10 @@ sub Send {
     }
   }
 
-  # Add dependencies to referenced bug list on new bugs
+  # Add dependencies and regressions to referenced bug list on new bugs
   if (!$start) {
-    push @referenced_bugs, @{$bug->dependson};
-    push @referenced_bugs, @{$bug->blocked};
+    push(@referenced_bugs,
+      map { @{$bug->{$_}} } qw(dependson blocked regressed_by regresses));
```
dylanwh:
```suggestion
      map { @{$bug->$_} } qw(dependson blocked regressed_by regresses));
```

dylanwh: You can call methods like `$obj->$method` if `$method` contains a method name.

Bugzilla/BugMail.pm
```diff
@@ -595,7 +595,7 @@ sub _get_diffs {
       $diff->{num}       = $comment->count;
       $diff->{isprivate} = $diff->{new};
     }
-    elsif ($diff->{field_name} eq 'dependson' || $diff->{field_name} eq 'blocked') {
+    elsif ($diff->{field_name} =~ /^(dependson|blocked|regress(ed_by|es))$/) {
```
dylanwh:
```suggestion
    elsif ($diff->{field_name} =~ /^(?:dependson|blocked|regress(ed_by|es))$/) {
```
Use a non-capture group since you don't care about the captured pattern

Bugzilla/Field.pm
```diff
@@ -285,6 +285,20 @@ use constant DEFAULT_FIELDS => (
...
   },
+  {
+    name           => 'regressed_by',
+    desc           => 'Regressed by',
+    in_new_bugmail => 1,
+    is_numeric     => 1,
+    buglist        => 1
+  },
+  {
+    name           => 'regresses',
```
dylanwh: This wording feels unnatural. Might just be me.

Bugzilla/Bug.pm
```diff
@@ -677,7 +679,8 @@ sub preload {
   # to the more complex method.
   my @all_dep_ids;
   foreach my $bug (@$bugs) {
-    push(@all_dep_ids, @{$bug->blocked}, @{$bug->dependson});
+    push(@all_dep_ids,
+      map { @{$bug->{$_} || []} } qw(dependson blocked regressed_by regresses));
```
dylanwh:
```suggestion
      map { @{$bug->$_ || []} } qw(dependson blocked regressed_by regresses));
```

Bugzilla/Bug.pm
```diff
@@ -1165,36 +1181,40 @@ sub update {
...
+          undef, $id, $self->id);
+        # Add an activity entry for the other bug.
+        LogActivityEntry($id, $other_field, '', $self->id, $user->id, $delta_ts);
+        # Update delta_ts on the other bug so that we trigger mid-airs.
+        _update_delta_ts($id, $delta_ts);
+      }
 
-    if (scalar(@$removed) || scalar(@$added)) {
-      $changes->{$type} = [join(', ', @$removed), join(', ', @$added)];
+      if (scalar(@$removed) || scalar(@$added)) {
```
dylanwh:
```suggestion
      if (@$removed || @$added) {
```

Might as well make this not needlessly use scalar

Bugzilla/Field.pm
kyoshino: My first iteration used `caused`. Then changed to `causes` but it’s also a noun, could be interpreted as the same meaning as `regressed_by`. Then thought of `introduces` but it would be used for not only for regressions but also features. Then renamed to `regresses` as suggested in the [bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1461492). 

But yes, it may still sound unnatural. I think a field name should be a noun wherever possible, so `regressions` might be the best. In that case the table column name should be different, adding a few lines of tweaks.

kyoshino: Will ask in the bug about the field name.

kyoshino:
So I have renamed the `regresses` field to `regressions` while keeping the table column name. Now I’m not sure if I’m doing the right thing in `Bugzilla/Search.pm` 😼

kyoshino:
I gotta revert the renaming change because it makes the implementation complicated and may introduce potential side effects. It’s okay to use `regresses` internally while using **Regressions** as the UI label. The API field name should also remain `regressions`, which can be mapped like `bug_id`/`id` in 2 places: [Bug.pm](https://github.com/mozilla-bteam/bmo/blob/master/Bugzilla/Bug.pm#L268) and [WebService/Bug.pm](https://github.com/mozilla-bteam/bmo/blob/master/Bugzilla/WebService/Bug.pm#L392).

kyoshino:
Available for testing on https://bugzilla-dev.allizom.org/

kyoshino:
Do not merge yet until we get a green light.

kyoshino:
We gotta ship it later this week.

PR #1068, opened 2019-01-25
Bug 1524174 - Redirect to show_bug.cgi after creating bug or updating a bug

purelogiq:
With this PR post_bug.cgi and process_bug.cgi now redirect to show_bug.cgi after successfully creating/updating a bug successfully. They no longer attempt to render the show_bug page themselves, making things more consistent and less buggy.

These changes work for both the bug-modal UI and the old bug UI.

See #1085 for the update to attachments. This PR should be merged first then that one immediately afterward.

### DIFF
--- a/process_bug.cgi
+++ b/process_bug.cgi
@@ -439,7 +439,7 @@ if (Bugzilla->usage_mode != USAGE_MODE_EMAIL) {
         my $saved_sent_change = {
           id => $sent_change->{params}->{id},
           type => $sent_change->{params}->{type},
-          sent_bugmail => $sent_change->{sent_bugmail},
+          recipient_count => scalar @{$sent_change->{sent_bugmail}->{sent}},
         };
         push @saved_sent_changes, $saved_sent_change;
       }
@@ -462,9 +462,9 @@ if (Bugzilla->usage_mode != USAGE_MODE_EMAIL) {
     foreach my $sent_changes (@all_sent_changes) {
       foreach my $sent_change (@$sent_changes) {
         my $params       = $sent_change->{params};
-        my $sent_bugmail = $sent_change->{sent_bugmail};
+        my $recipient_count = scalar @{$sent_change->{sent_bugmail}->{sent}};
         $vars->{$_} = $params->{$_} foreach keys %$params;
-        $vars->{'sent_bugmail'} = $sent_bugmail;
+        $vars->{'recipient_count'} = $recipient_count;
         $template->process("bug/process/results.html.tmpl", $vars)
           || ThrowTemplateError($template->error());
         $vars->{'header_done'} = 1;
--- a/process_bug.cgi
+++ b/process_bug.cgi
@@ -429,64 +429,53 @@ foreach my $bug (@bug_objects) {
 # Delete the session token used for the mass-change.
 delete_token($token) unless $cgi->param('id');
 
-# BMO: add show_bug_format hook for experimental UI work
-my $format_params = {
-  format => scalar $cgi->param('format'),
-  ctype  => scalar $cgi->param('ctype'),
-};
-Bugzilla::Hook::process('show_bug_format', $format_params);
-if ($format_params->{format} eq 'modal') {
-  my $bug_id = $vars->{bug} ? $vars->{bug}->id : undef;
-  $C->content_security_policy(SHOW_BUG_MODAL_CSP($bug_id));
-}
-my $format = $template->get_format(
-  "bug/show",
-  $format_params->{format},
-  $format_params->{ctype}
-);
-
 if (Bugzilla->usage_mode != USAGE_MODE_EMAIL) {
-  print $cgi->header();
-
-  foreach my $sent_changes (@all_sent_changes) {
-    foreach my $sent_change (@$sent_changes) {
-      my $params       = $sent_change->{params};
-      my $sent_bugmail = $sent_change->{sent_bugmail};
-      $vars->{$_} = $params->{$_} foreach keys %$params;
-      $vars->{'sent_bugmail'} = $sent_bugmail;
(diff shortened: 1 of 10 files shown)


process_bug.cgi
```diff
@@ -27,6 +27,7 @@ use Bugzilla::Status;
 use Bugzilla::Token;
 use Bugzilla::Hook;
 
+use JSON::MaybeXS qw(encode_json);
```
dylanwh: This appears to be unused

dylanwh:
Looks good so far. I bet the tests are failing just for the missing flash messaging

process_bug.cgi
```diff
@@ -429,25 +430,34 @@ foreach my $bug (@bug_objects) {
...
+          id => $sent_change->{params}->{id},
+          type => $sent_change->{params}->{type},
+          sent_bugmail => $sent_change->{sent_bugmail},
+        };
+        push @saved_sent_changes, $saved_sent_change;
+      }
+    }
+    $Bugzilla::App::CGI::C->flash(last_sent_changes => encode_json(\@saved_sent_changes));
+
+    # Redirect thow show_bug.cgi.
```
purelogiq: typo

template/en/default/bug/process/results.html.tmpl
```diff
@@ -29,6 +29,7 @@
 [% PROCESS global/variables.none.tmpl %]
 
 [% UNLESS header_done %]
+  [% PROCESS bug_modal/header.html.tmpl %]
```
purelogiq: Hey guys, I've spent a long time on this:
Unless I manually include the bug_modal header, it doesn't get included, causing everything to render wrong (`bug/process/header.html.tmpl` below should be including it given that `show_bug.cgi` choses to use the modal format). The old `process_bug.cgi` before this PR did things very similar to what I'm doing in show_bug.cgi and the headers were included just fine.

I will keep digging into this but wanted to give an update

dylanwh:
It took a bit to understand this code, will have updates in just a sec

qa/t/test_qa_contact.t
```diff
@@ -195,9 +195,6 @@ $sel->click_ok("commit");
 # The user is no longer the QA contact, and he has no other role
 # with the bug. He can no longer see it.
 
-$sel->wait_for_page_to_load_ok(WAIT_TIME);
```
purelogiq: This test tests that if the user removes themselves in a way that they could no longer see the bug that they would see the 'Changes made to this bug ...' on the resulting Access Denied page. With the redirect flow, it's not as easy to do this since I there is a short circut somewhere beforehand that immediately renders the Access Denied page before show_bug.cgi has a chance to render the bug change results. I think removing this test is OK, we still test that the user can't see the page right below this.

qa/t/test_user_preferences.t
```diff
@@ -249,7 +249,6 @@ $sel->type_ok("comment",
 $sel->click_ok("commit");
 $sel->wait_for_page_to_load(WAIT_TIME);
 $sel->is_text_present_ok("Changes submitted for bug $bug1_id");
-$sel->is_text_present_ok("The next bug in your list is bug $bug2_id");
```
purelogiq: I think this test has been (for a while now) made obsolete by the 'bug/navigate' template partial (which shows the current/next/previous bug in the bug list your browsing). This only appears on the old bug UI as well and it is of no value IMO (it appears on the same page of the next bug, so you're already on the next bug page when you see it...).

qa/t/lib/QA/Util.pm
```diff
@@ -452,11 +452,14 @@ sub check_page_load {
...
 
-  if ($expected_uri->query_param('id')) {
-    if ($expected_uri->query_param('id') eq '__BUG_ID__') {
-      $uri->query_param('id' => '__BUG_ID__');
-    }
+  if ($expected_uri->query_param('id') and $expected_uri->query_param('id') eq '__BUG_ID__') {
+    $uri->query_param('id' => '__BUG_ID__');
+  }
+
+  if ($expected_uri->query_param('list_id') and $expected_uri->query_param('list_id') eq '__LIST_ID__') {
```
purelogiq: Tests were crashing because they expected a hard coded listid (which apparently is generated on every request and not tied to a specific/saved list). I believe it is appropriate to ignore the actual value of the list_id param, hence this function.

purelogiq:
Changes to attachments will be done in https://bugzilla.mozilla.org/show_bug.cgi?id=1524175

dylanwh:
all the tests pass! \o/

dylanwh:
You will be able to remove the hack using the urlbase because I'm going to land #1081 first.

dylanwh:
@kyoshino: Not until the second PR is done. Attachments still also render the showbug page.

purelogiq:
> Now we can remove the `history.replaceState` JavaScript magic from both [modal](https://github.com/mozilla-bteam/bmo/blob/master/extensions/BugModal/web/bug_modal.js#L1448-L1459) and [legacy](https://github.com/mozilla-bteam/bmo/blob/master/template/en/default/bug/show-header.html.tmpl#L55-L66) views 😺

@kyoshino Thanks for pointing those out, I think I will have to remove them in https://bugzilla.mozilla.org/show_bug.cgi?id=1524175 since attachments.cgi still does the old behavior.

purelogiq:
@dylanwh This is ready for review 🎉 

email-sent-receipit

kyoshino:
@purelogiq Can you please rebase this and #1085 to resolve conflicts? The bug still annoys people so we need to fix it soonish 🙂

purelogiq:
@kyoshino Will do, then I will test these changes again.

purelogiq:
@kyoshino This and #1085 are ready to go, tested and fixed selenium tests. Please merge this first before the other one.

PR #1069, opened 2019-01-25
Bug 1523365 - Ensure all requests have the HSTS header (if configured)

psiinon:
The STS header is set on 200's but not on other response codes such as 404s.
The intention is for this to fix that, but I've not been able to test this change.
I've got BMO to run using docker-compose but for some reason that doesnt appear to pick up any of my changes, even when I change other parts of the code that should be much more visible :/
Will happily test if someone can explain to me how I can get my local changes reflected in the docker images.
The conditions were ripped off from https://github.com/mozilla-bteam/bmo/blob/master/Bugzilla/CGI.pm#L593-L602

### DIFF
--- a/Bugzilla/App.pm
+++ b/Bugzilla/App.pm
@@ -127,7 +127,7 @@ sub startup {
   }
   $self->hook(after_dispatch => sub {
     my ($c) = @_;
-    if ($self->req->is_secure
+    if ($c->req->is_secure
       && ! $c->res->headers->strict_transport_security
       && Bugzilla->params->{'strict_transport_security'} ne 'off')
     {
--- a/Bugzilla/App.pm
+++ b/Bugzilla/App.pm
@@ -122,19 +122,22 @@ sub startup {
         else {
           $c->res->headers->cache_control('public, max-age=31536000, immutable');
         }
-        if ($self->https
-          && not $c->res->headers->strict_transport_security
-          && Bugzilla->params->{'strict_transport_security'} ne 'off')
-        {
-          my $sts_opts = 'max-age=' . MAX_STS_AGE;
-          if (Bugzilla->params->{'strict_transport_security'} eq 'include_subdomains') {
-            $sts_opts .= '; includeSubDomains';
-          }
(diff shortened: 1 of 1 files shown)


Bugzilla/App.pm
```diff
@@ -122,6 +122,16 @@ sub startup {
         else {
           $c->res->headers->cache_control('public, max-age=31536000, immutable');
         }
+        if ($self->https
+          && not $c->res->headers->strict_transport_security
```
dylanwh: Here either you should use only the C-style logical operations (`&& || !`) or only the python ones (`and or not`) otherwise the results can be surprising.

dylanwh: Also this entire `if` should be added to a new `after_dispatch` hook, possible outside the `if ($self->mode ne 'development') {` conditional/

```perl
$self->hook(after_dispatch => sub {
my ($c) = @_;
# <insert your code>
});
```

Bugzilla/App.pm
```diff
@@ -122,6 +122,16 @@ sub startup {
         else {
           $c->res->headers->cache_control('public, max-age=31536000, immutable');
         }
+        if ($self->https
```
dylanwh:
```suggestion
        if ($self->req->is_secure
```

psiinon:
Updated based on the feedback.

dylanwh:
This looks good, I think. Tests are failing due to not seeing a 'Bug NNNNN' in the commit history that isn't already on prod.

psiinon:
Ah, what do I need to do to fix that?

Bugzilla/App.pm
```diff
@@ -125,6 +125,19 @@ sub startup {
       }
     );
   }
+  $self->hook(after_dispatch => sub {
+    my ($c) = @_;
+    if ($self->req->is_secure
```
dylanwh:
```suggestion
    if ($c->req->is_secure
```

dylanwh:
build will work with 

```
git commit --amend -m 'Bug 1523365 - Ensure all requests have the HSTS header (if configured)' 
git push -f 
```

PR #1072, opened 2019-01-25
Bug 1523004 - Sort Phabricator revisions by numeric value instead of alphabetically

dklawren:


extensions/PhabBugz/lib/WebService.pm
```diff
@@ -200,7 +200,7 @@ sub bug_revisions {
   }
 
   # sort by revision id
-  @revisions = sort { $a->{id} cmp $b->{id} } @revisions;
+  @revisions = sort { $a->{id} <=> $b->{id} } @revisions;
```
dylanwh: Do these include the `D` prefix or not?

PR #1075, opened 2019-01-28
Bug 1523317 - Exclude Graveyard products from QuickSearch results

kyoshino:
Exclude Graveyard products when no classification, product or component is specified in search terms, just like we are only searching open bugs by default. Given that there are several QuickSearch forms on the site, this will be done server side, in the BMO extension.

## Bugzilla link

[Bug 1523317 - Exclude Graveyard products from QuickSearch results](https://bugzilla.mozilla.org/show_bug.cgi?id=1523317)

### DIFF
--- a/Bugzilla/Search/Quicksearch.pm
+++ b/Bugzilla/Search/Quicksearch.pm
@@ -127,7 +127,7 @@ use constant COMPONENT_EXCEPTIONS => (
 );
 
 # Quicksearch-wide globals for boolean charts.
-our ($chart, $and, $or, $fulltext, $bug_status_set, $ELASTIC);
+our ($chart, $and, $or, $fulltext, $bug_status_set, $bug_product_set, $ELASTIC);
 
 sub quicksearch {
   my ($searchstring) = (@_);
@@ -252,6 +252,16 @@ sub quicksearch {
     unless ($bug_status_set) {
       $cgi->param('bug_status', BUG_STATE_OPEN);
     }
+
+    # Provide a hook to allow modifying the params
+    Bugzilla::Hook::process(
+      'quicksearch_run',
+      {
+        'cgi'             => $cgi,
+        'bug_status_set'  => $bug_status_set,
+        'bug_product_set' => $bug_product_set,
+      }
+    );
 
     # Inform user about any unknown fields
     if (scalar(@unknownFields) || scalar(keys %ambiguous_fields)) {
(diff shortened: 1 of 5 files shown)


Bugzilla/Search/Quicksearch.pm
```diff
@@ -477,6 +487,9 @@ sub _handle_field_names {
         if ($translated eq 'bug_status' || $translated eq 'resolution') {
           $bug_status_set = 1;
         }
+        if ($translated =~ /^(?:classification|product|component)$/) {
+          $bug_product_set = 1;
```
dylanwh: Shouldn't the hook be called *after* this is set?

extensions/BMO/Extension.pm
```diff
@@ -873,6 +873,18 @@ sub quicksearch_map {
   }
 }
 
+sub quicksearch_run {
```
dylanwh: Mention this hook in Bugzilla/Hooks.pm's docs.

Bugzilla/Search/Quicksearch.pm
kyoshino: If I read the code correctly, the hook is called properly with `bug_product_set`. Tested locally and it worked.

kyoshino:
So this PR may have to be updated for the unit tests? Rebasing now.

extensions/BMO/Extension.pm
```diff
@@ -873,6 +873,27 @@ sub quicksearch_map {
...
+    $cgi->param('o1', 'notequals');
+    $cgi->param('v1', 'Graveyard');
+  }
+}
+
+sub quicksearch_test {
+  my ($self, $args) = @_;
+  my $opt = $args->{'opt'};
+
+  $opt{params}->{'f1'} = 'classification';
```
dylanwh:
```suggestion
  $opt->{params}->{'f1'} = 'classification';
```

extensions/BMO/Extension.pm
```diff
@@ -873,6 +873,27 @@ sub quicksearch_map {
...
+  $opt{params}->{'o1'} = 'notequals';
```
dylanwh:
```suggestion
  $opt->{params}->{'o1'} = 'notequals';
```

extensions/BMO/Extension.pm
```diff
@@ -873,6 +873,27 @@ sub quicksearch_map {
...
+  $opt{params}->{'v1'} = 'Graveyard';
```
dylanwh:
```suggestion
  $opt->{params}->{'v1'} = 'Graveyard';
```

t/quicksearch.t
```diff
@@ -122,6 +123,10 @@ sub test_quicksearch {
     }
   }
 
+  # Provide a hook to allow modifying the params. This has to correspond with
+  # the `quicksearch_run` hook
+  Bugzilla::Hook::process('quicksearch_test', { 'opt' => $opt };
```
dylanwh:
```suggestion
  Bugzilla::Hook::process('quicksearch_test', { 'opt' => \%opt };
```

PR #1085, opened 2019-02-01
Bug 1524175 - Redirect to show_bug.cgi after creating or updating an attachment

purelogiq:
- Includes changes from https://github.com/mozilla-bteam/bmo/pull/1068 until that gets merged.
- This PR depends on #1065 being merged.
- As the title suggests this causes BMO to redirect when attachments are created/updated/deleted. This lets us have 1 point (show_bug.cgi) that actually renders the show bug page.
- This works for the bug modal UI and the old UI.
-  ~~TODO: Determine how to handle large recipient lists which won't fit in flash storage.~~ Solved by just listing how many people were emailed.

### DIFF
--- a/attachment.cgi
+++ b/attachment.cgi
@@ -653,31 +653,26 @@ sub insert {
 
   $dbh->bz_commit_transaction;
 
-  # Define the variables and functions that will be passed to the UI template.
-  $vars->{'attachment'} = $attachment;
-
-  # We cannot reuse the $bug object as delta_ts has eventually been updated
-  # since the object was created.
-  $vars->{'bugs'}              = [new Bugzilla::Bug($bugid)];
-  $vars->{'header_done'}       = 1;
-  $vars->{'contenttypemethod'} = $cgi->param('contenttypemethod');
-
+  # Persist details of what changed and redirect to show_bug page.
   my $recipients = {'changer' => $user, 'owner' => $owner};
-  $vars->{'sent_bugmail'} = Bugzilla::BugMail::Send($bugid, $recipients);
+  my $sent_bugmail = Bugzilla::BugMail::Send($bugid, $recipients);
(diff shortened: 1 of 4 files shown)


purelogiq:
Removing the js that rewrites the URLs breaks some selenium tests. I will try to fix those soon. Failures: https://circleci.com/gh/mozilla-bteam/bmo/19542?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

purelogiq:
> Removing the js that rewrites the URLs breaks some selenium tests. I will try to fix those soon. Failures: https://circleci.com/gh/mozilla-bteam/bmo/19542?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

This is still the case, perhaps @kyoshino can take a look in another PR after these are merged? Keeping the JS doesn't hurt for now.

purelogiq:
@dylanwh This is ready for review 🎉

kyoshino:
Looks like sometimes these URL mismatching issues happen due to a race condition, as I saw in #1124. Some of failures in #665 and #778 might be the same.

dylanwh:
> Looks like sometimes these URL mismatching issues happen due to a race condition, as I saw in #1124. Some of failures in #665 and #778 might be the same.

Can you explain a bit more what you mean?

kyoshino:
Since some scripts rewrite the URL dynamically with `history.replaceState()`, race conditions can happen depending on when to check the URL. The tests need a better control, e.g. wait for 1 sec after the page is loaded so the rewritten URL can be checked reliably.

purelogiq:
Looks like master is broken?

kyoshino:
This is from #1351.

kyoshino:
@purelogiq can you force rebase the branch to solve the test failures?

PR #1090, opened 2019-02-04
Bug 1507812 - merge_user.pl should not continue if the old user id has an account in Phabricator

dklawren:


scripts/merge-users.pl
```diff
@@ -11,7 +11,7 @@
 use warnings;
 use lib qw(. lib local/lib/perl5);
 
-=head1 NAME
+=head1 NA
```
dylanwh: pretty sure this is a mistake

PR #1084, opened 2019-02-01
Bug 1522341 - Implement new field for indicating if a bug is a task, enhancement, or defect

kyoshino:
Add new **Type** field to bugs. Just like other bug fields, it will appear everywhere including the search page and bug lists, and are also available via the API. The field values, defaulted to **defect**, **enhancement** and **task** (these are still subject to change), can be edited by Bugzilla admin. The default type per component can also be changed on the Edit Component page. See my [UX spec](https://drive.google.com/file/d/1l596C3yyE1if19SZqnDhYo8Wa2kSGget/view) for details.

The implementation is mostly a copy of the Severity field. While the API field and Quick Search keyword is `type`, the internal field name remains `bug_type` given that `type` is a SQL reserved word.

## Bugzilla link

[Bug 1522341 - Implement new field for indicating if a bug is a task, enhancement, or defect](https://bugzilla.mozilla.org/show_bug.cgi?id=1522341)

PR #1092, opened 2019-02-05
Bug 1461492/1522341 - Schema changes for Regressions & Bug Type fields

kyoshino:
Split from #1067 and #1084. Update the DB schema:

* Add `bug_type` column to `bug` table
* Add `default_bug_type` column to `components` table
* Add new `regressions` table with `regressed_by` and `regresses` columns


kyoshino:
Updated the `set_type` subroutine in #1084 to make sure the field won’t be empty even if the column is nullable.

kyoshino:
I think the last commit [Make bug_type nullable, remove the default value](/mozilla-bteam/bmo/pull/1092/commits/38a7cf085f7829c75973d37e11648501eef81051) was not necessary after all, because the field will be empty until the migration script runs, and the types are already hardcoded in [Bugzilla/DB.pm](https://github.com/mozilla-bteam/bmo/pull/1084/files#diff-5c45e641817f8ebffc9ce08f62734a94). Will send a PR to revert the commit.

PR #1095, opened 2019-02-07
Bug 179115 - changes to Cc: list should not update bug's last changed date

kyoshino:
Stop updating the last modified date = `delta_ts` field when only CC is changed. This makes triagers’ life much easier and helps people focus on real contributions.

## Bugzilla link

[Bug 179115 - changes to Cc: list should not update bug’s last changed date](https://bugzilla.mozilla.org/show_bug.cgi?id=179115)

### DIFF
--- a/Bugzilla/Bug.pm
+++ b/Bugzilla/Bug.pm
@@ -1354,8 +1354,11 @@ sub update {
     }
   );
 
-  # If any change occurred, refresh the timestamp of the bug.
-  if ( scalar(keys %$changes)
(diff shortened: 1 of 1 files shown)


kyoshino:
Do we have any side effect? As far as I can tell from my local test, the change is logged and email notifications are sent afterwards as usual.

dylanwh:
Things like kyle's ETL and external systems could rely on this. We must add a "user-visible" change field or something.

PR #1096, opened 2019-02-07
Bug 1525808 - Remove CC changes from activity stream

kyoshino:
Remove the CC changes, which are hidden by default, and the Show CC Changes option from the modal bug page, because CC changes are not real activities.

## Bugzilla link

[Bug 1525808 - Remove CC changes from activity stream](https://bugzilla.mozilla.org/show_bug.cgi?id=1525808)


kyoshino:
Thanks! 😺

PR #1097, opened 2019-02-07
Bug 1532409 - Introduce Bugzilla::Model (a DBIx::Class::Schema)

dylanwh:


dylanwh:
Perhaps `DBIx::Class::Candy` is too much. It's a dependency of `DBIx::Class::Helpers` which we need at least because it escapes table names that are keywords (like groups). Here it is a little annoying because we can't use `Mojo::Base`. In general I want to use Moo whenever possible, and I fall back to `Mojo::Base` because it's compatible with vanilla perl but also turns on strict, warnings, and so on.

`DBIx::Class::Candy` lets me write `table()` instead of `__PACKAGE__->table` which I like, but it also sets the base. Thoughts?

PR #1102, opened 2019-02-08
remove elasticsearch

dylanwh:


purelogiq:
@dylanwh is there some background story to this PR?

dylanwh:
> @dylanwh is there some background story to this PR?

It's code we're not using and haven't been using, and is likely to not work without additional development.

dylanwh:
I think this was premature. let's explore what this working looks like

PR #1107, opened 2019-02-14
Bug 1402894 - Remove "Restrict this session to this IP" option from login page

kyoshino:
Remove the **Restrict this session to this IP address** option and relevant code since people are moving around more often, 2FA is enabled, less than 20% of users are enabling it, and no one actually wants it.

## Bugzilla link

[Bug 1402894 - Remove "Restrict this session to this IP" option from login page](https://bugzilla.mozilla.org/show_bug.cgi?id=1402894)

### DIFF
--- a/Bugzilla/Install/DB.pm
+++ b/Bugzilla/Install/DB.pm
@@ -780,7 +780,8 @@ sub update_table_definitions {
 
   _add_oauth2_jwt_support();
 
-  _remove_restrict_ipaddr();
+  # Bug 1402894 - kohei.yoshino@gmail.com
+  $dbh->bz_drop_column('logincookies', 'restrict_ipaddr');
 
   ################################################################
   # New --TABLE-- changes should go *** A B O V E *** this point #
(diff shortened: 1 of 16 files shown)


Bugzilla/Install/DB.pm
```diff
@@ -4287,6 +4277,13 @@ sub _add_oauth2_jwt_support {
     {TYPE => 'INT4', NOTNULL => 1});
 }
 
+sub _remove_restrict_ipaddr {
+  my $dbh = Bugzilla->dbh;
+  return unless $dbh->bz_column_info('logincookies', 'restrict_ipaddr');
+
+  $dbh->do('ALTER TABLE logincookies DROP COLUMN restrict_ipaddr');
```
dklawren: $dbh->bz_drop_column('logincookies','restrict_ipaddr');

kyoshino: Ah there is! I was searching `bz_remove_column` instead 😅

kyoshino: And looks like we don’t need a subroutine for that.

PR #1113, opened 2019-02-20
Bug 1391439 - Add ability to capture and attach a screenshot through the Bugzilla UI

kyoshino:
Use the latest [Screen Capture API](https://developer.mozilla.org/en-US/docs/Web/API/Screen_Capture_API/Using_Screen_Capture) currently available in Firefox Nightly/Beta and Chrome to let users take a screenshot directly from the browser. Wow, we don’t need Java nor Flash! 🎉

Note: this API requires HTTPS 🔒 which means the local server must have a TLS certificate to test the functionality. I don’t know how to do it with Vagrant so I’m using the [Apache reverse proxy](https://httpd.apache.org/docs/2.4/howto/reverse_proxy.html) instead.

## Bugzilla link

[Bug 1391439 - Add ability to capture and attach a screenshot through the Bugzilla UI](https://bugzilla.mozilla.org/show_bug.cgi?id=1391439)

### DIFF
--- a/js/attachment.js
+++ b/js/attachment.js
@@ -597,7 +597,7 @@ Bugzilla.AttachmentForm = class AttachmentForm {
     const [date, time] = (new Date().toISOString()).match(/^(.+)T(.+)\./).splice(1);
 
     // Process as a PNG file
-    this.process_file(new File([blob], `Screenshot on ${date} at ${time}`, { type: 'image/png' }));
+    this.process_file(new File([blob], `Screenshot on ${date} at ${time}.png`, { type: 'image/png' }));
     this.update_ispatch(false, true);
 
     // Clean up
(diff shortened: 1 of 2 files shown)


kyoshino:
Sigh. The tests are failing probably because `async` function is not supported. Firefox on CircleCI has to be updated. 😞

kyoshino:
@dylanwh As #1117 demonstrated, [async function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function) is the culprit of the Selenium test failures. It has been supported since Firefox 52, while CircleCI is running Firefox 47. It’s a blocker for front-end dev. [I really want it get updated](https://bugzilla.mozilla.org/show_bug.cgi?id=1480469).

kyoshino:
I’ve solved the test failures by using a traditional `Promise` chain. This can be refactored later using `async` ✌️

kyoshino:
Yet another ES6 stuff here 😺

PR #1106, opened 2019-02-13
Bug 1522348 - Bulk assign open bugs to task, enhancement, defect field

kyoshino:
Happy SQL time! Add a script to do the migration task 1, 2, 3 written in [Bug 1524738](https://bugzilla.mozilla.org/show_bug.cgi?id=1524738#c6). Would like to test this on the dev server.

## Bugzilla link

[Bug 1522348 - Bulk assign open bugs to task, enhancement, defect field](https://bugzilla.mozilla.org/show_bug.cgi?id=1522348)

PR #1118, opened 2019-02-21
Bug 1522341 schema change - Make bug_type non-nullable, re-add default value

kyoshino:
This reverts 38a7cf085f7829c75973d37e11648501eef81051, the last commit in #1092, to make sure the migration goes well in any case.


kyoshino:
So I think we don’t need this after all. We have a migration script in #1106, and upstream users could also have their own migration plan.

PR #1124, opened 2019-02-23
Bug 1496879 - Re-implement bug list navigation with local storage

kyoshino:
Replace the server-side, cookie-based legacy bug list navigation with a lightweight client-side implementation using the local storage. This is required for the new bug list page that will load results client-side, and also [solves a dozen issues](https://bugzilla.mozilla.org/show_bug.cgi?id=1496879#c1) as explained in the bug.

Note to myself: Remove the `profile_search` table in a separate PR after this PR hits production.

## Bugzilla link

[Bug 1496879 - Re-implement bug list navigation with local storage](https://bugzilla.mozilla.org/show_bug.cgi?id=1496879)


Ryuno-Ki:
Hm, I need to dig up the source, but localStorage has some limitations (sync, size).

Maybe you want to use IndexedDB at a later point?
There are some small libraries to simplify its API if needed.

kyoshino:
We can use [KV Storage](https://developers.google.com/web/updates/2019/03/kv-storage) in the future.

kyoshino:
This needs to be updated once #1187 is merged.

kyoshino:
I’m rather thinking about building a single-page app.

PR #1125, opened 2019-02-25
Bug 1472522 - Show image, video, audio, text attachments inline

kyoshino:
Show image, audio, video and text attachments inline on the modal bug page. This uses the [Intersection Observer API](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API) so these attachments will be visible only when it appears on the screen after scrolling the page.

## Bugzilla link

[Bug 1472522 - Show image/text attachments inline](https://bugzilla.mozilla.org/show_bug.cgi?id=1472522)

## Screenshot

inline-attachments


purelogiq:
So far looking good. There wasn't much discussion on the bug but have we though of whether this could be abused and also if we want something like a 'collapse all' button? Sometimes knowing an attachment was created is enough info (since this also does text attachment previews I can imagine that a lot of those would have useless previews if it's like some debug output). What are your thoughts @kyoshino @dylanwh?

kyoshino:
If we find any abusive usage, [tagging the comment](https://wiki.mozilla.org/BMO/comment_tagging) will hide it by default. GitHub also shows images inline, so I’m not too worried about it. Maybe we don’t need the preview of text attachments, but I sometimes feel it’s nice on Slack, and it would be useful for [this kind of text](https://bugzilla.mozilla.org/attachment.cgi?id=9041954&action=edit).

dylanwh:
So it's confirmed that tagging can be done to block any offensive images? @kyoshino

kyoshino:
Not sure who can decide, but I think tags like **abusive** or **offtopic** can be used for attached images as well, given that an attachment and a comment are 1:1.

PR #1127, opened 2019-02-27
Bug 1529368 - Implement REST API for saved searches

kyoshino:
Implement `/rest/saved_searches` so the new search results page can allow user to add/edit/remove a search with one click. I'm not adding a doc for the new API method for now because it's mainly for internal use and we are moving to REST API v2 or GraphQL API in the not-so-distant future.

## Bugzilla link

[Bug 1529368 - Implement REST API for saved searches](https://bugzilla.mozilla.org/show_bug.cgi?id=1529368)

### DIFF
--- a/Bugzilla/Search/Saved.pm
+++ b/Bugzilla/Search/Saved.pm
@@ -296,7 +296,7 @@ sub _whine_subjects {
 
   return $self->{_whine_subjects} if exists $self->{_whine_subjects};
 
-  ($self->{_whine_subjects}) = Bugzilla->dbh->selectcol_arrayref(
+  $self->{_whine_subjects} = Bugzilla->dbh->selectcol_arrayref(
     'SELECT DISTINCT whine_events.subject FROM whine_events
       INNER JOIN whine_queries ON whine_queries.eventid = whine_events.id
       WHERE whine_events.owner_userid = ? AND whine_queries.query_name = ?',
--- a/Bugzilla/Search/Saved.pm
+++ b/Bugzilla/Search/Saved.pm
@@ -128,9 +128,9 @@ sub _check_query {
   my $cgi = new Bugzilla::CGI($query);
   my $query_str = $cgi->param('POSTDATA') || $cgi->param('PUTDATA');
 
-  # The new CGI will inherit the current request method, perhaps of an API call,
-  # so the params will be bogus if sent via POST or PUT. This decodes and fixes
-  # the params before being canonicalizing with `clean_search_url()`.
+  # The new CGI inherits the current request method, perhaps of an API call, so
+  # the params will be bogus if sent via POST or PUT. This decodes and fixes the
+  # params before being canonicalized with `clean_search_url()`.
   if ($query_str) {
     my $uri = URI->new("buglist.cgi?$query_str");
 
--- a/Bugzilla/Search/Saved.pm
+++ b/Bugzilla/Search/Saved.pm
@@ -22,6 +22,8 @@ use Bugzilla::User;
 use Bugzilla::Util;
 
 use Scalar::Util qw(blessed);
+use URI;
+use URI::QueryParam;
 
 #############
 # Constants #
@@ -124,6 +126,21 @@ sub _check_query {
   my ($invocant, $query) = @_;
   $query || ThrowUserError("buglist_parameters_required");
   my $cgi = new Bugzilla::CGI($query);
+  my $query_str = $cgi->param('POSTDATA') || $cgi->param('PUTDATA');
+
+  # The new CGI will inherit the current request method, perhaps of an API call,
+  # so the params will be bogus if sent via POST or PUT. This decodes and fixes
+  # the params before being canonicalizing with `clean_search_url()`.
+  if ($query_str) {
+    my $uri = URI->new("buglist.cgi?$query_str");
+
+    for my $key ($uri->query_param) {
+      $cgi->param($key, $uri->query_param($key));
+    }
+
+    $cgi->delete('POSTDATA', 'PUTDATA');
+  }
+
   $cgi->clean_search_url;
 
   # Don't store the query name as a parameter.
@@ -136,15 +153,20 @@ sub _check_query {
 #########################
 
 sub create {
-  my $class = shift;
-  Bugzilla->login(LOGIN_REQUIRED);
+  my ($class, $params) = @_;
+  my $user  = Bugzilla->login(LOGIN_REQUIRED);
+
+  # Prevent duplicated names from being saved
+  ThrowUserError('saved_search_same_name')
+    if scalar(grep { lc($_->name) eq lc($params->{name}) } @{$user->queries});
+
   my $dbh = Bugzilla->dbh;
-  $class->check_required_create_fields(@_);
+  $class->check_required_create_fields($params);
   $dbh->bz_start_transaction();
-  my $params = $class->run_create_validators(@_);
+  $params = $class->run_create_validators($params);
 
   # Right now you can only create a Saved Search for the current user.
-  $params->{userid} = Bugzilla->user->id;
+  $params->{userid} = $user->id;
 
   my $lif = delete $params->{link_in_footer};
   my $obj = $class->insert_create_data($params);
@@ -219,6 +241,40 @@ sub preload {
     $query->{link_in_footer} = ($links_in_footer{$query->id}) ? 1 : 0;
   }
 }
+
+sub update {
+  my ($self, $params) = @_;
+  my ($name, $url) = @$params{qw(name url)};
+
+  Bugzilla->login(LOGIN_REQUIRED);
+
+  $self->set_name($name) if $name;
+  $self->set_url($url) if $url;
+  $self->SUPER::update(@_);
+}
+
+sub remove {
+  my ($self) = @_;
+  my $dbh = Bugzilla->dbh;
+
+  Bugzilla->login(LOGIN_REQUIRED);
+
+  ThrowUserError('saved_search_used_by_whines', {
+    search_name => $self->name,
+    subjects    => $self->_whine_subjects,
+  }) if $self->used_in_whine;
+
+  $dbh->do('DELETE FROM namedqueries WHERE id = ?', undef, $self->id);
+  $dbh->do('DELETE FROM namedqueries_link_in_footer WHERE namedquery_id = ?',
+    undef, $self->id);
+  $dbh->do('DELETE FROM namedquery_group_map WHERE namedquery_id = ?',
+    undef, $self->id);
+
+  # Now reset the cached queries
+  Bugzilla->memcached->clear({table => 'namedqueries', id => $self->id});
+  Bugzilla->user->flush_queries_cache();
+}
+
 #####################
 # Complex Accessors #
 #####################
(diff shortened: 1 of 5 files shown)


Bugzilla/Search/Saved.pm
```diff
@@ -22,6 +22,8 @@ use Bugzilla::User;
 use Bugzilla::Util;
 
 use Scalar::Util qw(blessed);
+use URI;
+use URI::QueryParam;
```
dylanwh:
```suggestion
use URI::QueryParam;
use List::Util qw( any );
```

Bugzilla/Search/Saved.pm
```diff
@@ -136,15 +153,20 @@ sub _check_query {
...
 
 sub create {
-  my $class = shift;
-  Bugzilla->login(LOGIN_REQUIRED);
+  my ($class, $params) = @_;
+  my $user  = Bugzilla->login(LOGIN_REQUIRED);
+
+  # Prevent duplicated names from being saved
+  ThrowUserError('saved_search_same_name')
+    if scalar(grep { lc($_->name) eq lc($params->{name}) } @{$user->queries});
```
dylanwh:
```suggestion
    if any { lc($_->name) eq lc($params->{name}) } @{$user->queries};
```

Bugzilla/Search/Saved.pm
```diff
@@ -235,16 +291,24 @@ sub edit_link {
...
-  return $self->{used_in_whine} if exists $self->{used_in_whine};
-  ($self->{used_in_whine}) = Bugzilla->dbh->selectrow_array(
-    'SELECT 1 FROM whine_events INNER JOIN whine_queries
-                       ON whine_events.id = whine_queries.eventid
-          WHERE whine_events.owner_userid = ? AND query_name = ?', undef,
-    $self->{userid}, $self->name
-  ) || 0;
-  return $self->{used_in_whine};
+
+  return scalar(@{$self->_whine_subjects}) ? 1 : 0;
```
dylanwh:
```suggestion
  return @{$self->_whine_subjects} ? 1 : 0;
```

`scalar` is only needed when you're inside list. It's a bit like the opposite of the spread operator. Every time you see a `@expr` it's like `...var` in javascript, and `scalar` removes that (implicit) spreading.

Bugzilla/WebService/User.pm
```diff
@@ -504,6 +505,63 @@ sub whoami {
...
+#
+
+sub add_saved_search {
+  my ($self, $params) = @_;
+
+  # The `create()` method fails when a saved search with the same name exists.
+  # The UI should ask in advance if the user wants to override the existing one,
+  # and if the answer is yes, use `update_saved_search()` instead.
+  my $search = Bugzilla::Search::Saved->create({
+    name => $params->{name}, query => $params->{url}, link_in_footer => 1
```
dylanwh: Here is a bit of danger. It might be possible for the `$params->{url}` to be a data structure.
It's not certain this would cause a problem. `Bugzilla::Search::Saved->create($user_input)` would definitely be bad. Still, it's always good to be explicit about what something is.

My other review comments add type checking but I didn't proof them so they may need to be tweaked.

Bugzilla/WebService/User.pm
```diff
@@ -17,6 +17,7 @@ use Bugzilla;
 use Bugzilla::Constants;
 use Bugzilla::Error;
 use Bugzilla::Group;
+use Bugzilla::Search::Saved;
```
dylanwh:
```suggestion
use List::Util qw(first);
use Type::Params qw( compile );
use Type::Utils;
use Types::Standard qw( :types )
```

Bugzilla/WebService/User.pm
```diff
@@ -504,6 +505,63 @@ sub whoami {
   );
 }
 
+#
+# Saved Searches
+#
+
+sub add_saved_search {
+  my ($self, $params) = @_;
```
dylanwh:
```suggestion
  state $check = compile(Object, Dict[name => Str, url => Str]);
  my ($self, $params) = $check->(@_);
```

Bugzilla/WebService/User.pm
```diff
@@ -504,6 +505,63 @@ sub whoami {
...
+
+sub get_saved_searches {
+  my ($self) = @_;
+
+  return [map { $self->_query_to_hash($_) } @{Bugzilla->user->queries}];
+}
+
+sub _get_saved_search_by_id {
+  my ($self, $id) = @_;
+  my ($search) = grep { $_->id == $id } @{Bugzilla->user->queries};
```
dylanwh:
```suggestion
  my $search = first { $_->id == $id } @{Bugzilla->user->queries};
```

Bugzilla/WebService/User.pm
```diff
@@ -504,6 +505,63 @@ sub whoami {
...
+
+sub get_saved_search {
+  my ($self, $params) = @_;
+  my $search = $self->_get_saved_search_by_id($params->{id});
+
+  return $self->_query_to_hash($search);
+}
+
+sub update_saved_search {
+  my ($self, $params) = @_;
```
dylanwh:
```suggestion
  state $check = compile(Object, Dict[name => Str, url => Str]);
  my ($self, $params) = $check->(@_);
```

Bugzilla/WebService/User.pm
```diff
@@ -504,6 +505,63 @@ sub whoami {
...
+  my $search = $self->_get_saved_search_by_id($params->{id});
+
+  $search->update({name => $params->{name}, url => $params->{url}});
+
+  return $self->_query_to_hash($search);
+}
+
+sub remove_saved_search {
+  my ($self, $params) = @_;
```
dylanwh:
```suggestion
  state $check = compile(Object, Dict[id => Int]);
  my ($self, $params) = $check->(@_);
```

Bugzilla/WebService/User.pm
```diff
@@ -504,6 +505,63 @@ sub whoami {
...
+  my ($self, $id) = @_;
+  my ($search) = grep { $_->id == $id } @{Bugzilla->user->queries};
+
+  ThrowUserError('saved_search_not_found') unless $search;
+
+  return $search;
+}
+
+sub get_saved_search {
+  my ($self, $params) = @_;
```
dylanwh:
```suggestion
  state $check = compile(Object, Dict[id => Int]);
  my ($self, $params) = $check->(@_);
```

buglist.cgi
```diff
@@ -219,9 +219,7 @@ sub InsertNamedQuery {
     = grep { lc($_->name) eq lc($query_name) } @{Bugzilla->user->queries};
```
dylanwh:
```suggestion
    = first { lc($_->name) eq lc($query_name) } @{Bugzilla->user->queries};
```

Bugzilla/Search/Saved.pm
```diff
@@ -136,15 +154,20 @@ sub _check_query {
 #########################
 
 sub create {
-  my $class = shift;
-  Bugzilla->login(LOGIN_REQUIRED);
+  my ($class, $params) = @_;
+  my $user  = Bugzilla->login(LOGIN_REQUIRED);
```
dklawren: You do not need to do this in the object modules. The calling code should be the one to handle auth.

my $user = Bugzilla->user;

Bugzilla/Search/Saved.pm
```diff
@@ -219,6 +242,40 @@ sub preload {
     $query->{link_in_footer} = ($links_in_footer{$query->id}) ? 1 : 0;
   }
 }
+
+sub update {
+  my ($self, $params) = @_;
+  my ($name, $url) = @$params{qw(name url)};
+
+  Bugzilla->login(LOGIN_REQUIRED);
```
dklawren: Should not use login in object modules as above.

Bugzilla/Search/Saved.pm
```diff
@@ -219,6 +242,40 @@ sub preload {
...
+  $self->set_name($name) if $name;
+  $self->set_url($url) if $url;
+  $self->SUPER::update(@_);
+}
+
+sub remove {
+  my ($self) = @_;
+  my $dbh = Bugzilla->dbh;
+
+  Bugzilla->login(LOGIN_REQUIRED);
```
dklawren: Same

Bugzilla/WebService/User.pm
```diff
@@ -504,6 +508,67 @@ sub whoami {
...
+  my ($self, $params) = $check->(@_);
+
+  # The `create()` method fails when a saved search with the same name exists.
+  # The UI should ask in advance if the user wants to override the existing one,
+  # and if the answer is yes, use `update_saved_search()` instead.
+  my $search = Bugzilla::Search::Saved->create({
+    name => $params->{name}, query => $params->{url}, link_in_footer => 1
+  });
+
+  return $self->_query_to_hash($search);
```
dklawren: Since we are creating new API methods should always return a hash with a keyword containing the results. We should try to standardize on all result structures where possible.

For example in this case it would be:

```json
{
  "queries": [
      {
         "id": 1,
         "name": "foo",
         "url": "somelongurl",
      }
   ]
}
```

Bugzilla/WebService/User.pm
```diff
@@ -504,6 +508,67 @@ sub whoami {
...
+
+  return $search;
+}
+
+sub get_saved_search {
+  state $check = compile(Object, Dict[id => Int]);
+  my ($self, $params) = $check->(@_);
+  my $search = $self->_get_saved_search_by_id($params->{id});
+
+  return $self->_query_to_hash($search);
```
dklawren:
```json
{
  "queries": [
      {
         "id": 1,
         "name": "foo",
         "url": "somelongurl",
      }
   ]
}
```

Bugzilla/WebService/User.pm
```diff
@@ -504,6 +508,67 @@ sub whoami {
...
+  });
+
+  return $self->_query_to_hash($search);
+}
+
+sub get_saved_searches {
+  my ($self) = @_;
+
+  return [map { $self->_query_to_hash($_) } @{Bugzilla->user->queries}];
+}
```
dklawren:
```json
{
  "queries": [
      {
         "id": 1,
         "name": "foo",
         "url": "somelongurl",
      }
   ]
}
```

Bugzilla/WebService/User.pm
```diff
@@ -504,6 +508,67 @@ sub whoami {
...
+
+sub update_saved_search {
+  state $check = compile(Object, Dict[name => Str, url => Str]);
+  my ($self, $params) = $check->(@_);
+  my $search = $self->_get_saved_search_by_id($params->{id});
+
+  $search->update({name => $params->{name}, url => $params->{url}});
+
+  return $self->_query_to_hash($search);
+}
```
dklawren:
```json
{
  "queries": [
      {
         "id": 1,
         "name": "foo",
         "url": "somelongurl",
      }
   ]
}
```

Bugzilla/WebService/User.pm
```diff
@@ -504,6 +508,67 @@ sub whoami {
...
+sub get_saved_search {
+  state $check = compile(Object, Dict[id => Int]);
+  my ($self, $params) = $check->(@_);
+  my $search = $self->_get_saved_search_by_id($params->{id});
+
+  return $self->_query_to_hash($search);
+}
+
+sub update_saved_search {
+  state $check = compile(Object, Dict[name => Str, url => Str]);
```
dklawren: Missing ID part

state $check = compile(Object, Dict[id => Int, name => Str, url => Str]);

Bugzilla/WebService/User.pm
kyoshino: I think newer methods like this don’t have to follow the legacy, redundant structure, because it’s only available via the REST API and only for internal use (undocumented). The [Component Watching API](https://github.com/mozilla-bteam/bmo/blob/master/extensions/ComponentWatching/lib/WebService.pm) I’ve added recently is also not following the tradition. We are moving to GraphQL sooner or later, so the structure here doesn’t matter 🙂

dklawren: > 
> 
> I think newer methods like this don’t have to follow the legacy, redundant structure, because it’s only available via the REST API and only for internal use (undocumented) for now. The [Component Watching API](https://github.com/mozilla-bteam/bmo/blob/master/extensions/ComponentWatching/lib/WebService.pm) I’ve added recently is also not following the tradition. We are moving to GraphQL sooner or later, so the structure here doesn’t matter 🙂

Not a showstopper for this, but I politely disagree on two points. 1) We should write all APIs as if it could also be used for the general public and not just for internal only reasons. Bugzilla has always lacked in a robust API that could be used to do most actions. Anything we can do to make this better at the same time as fulfilling our own needs the better. And 2) working with the non-obsolete API methods in Phabricator, it is nice to always expect a consistent return data structure and not have to guess at what to expect to be returned. They have been working to deprecate the old API methods that do not adhere to the consistent structure. Also they had to change it to allow for the paging functionality built in. We do not want to put out new methods that people start to use and then have to change them later. Better to do it right now.

I am not aware or knowledgeable about GraphQL yet so you are probably right that the current code may not matter. But even so it would be nice if when we do move over the clients do not have to change very much about how they receive the new data in there code. Would be cool if it all just looked the same, just coming from something different on the backend.

PR #1119, opened 2019-02-21
Bug 1529406 - Record user’s logged-in/out status and track insiders with Google Analytics

kyoshino:
* <del>Start logging actual URLs instead of template names. The query params will also be logged except for sensitive tokens and user-specific bug list IDs</del>
* Start logging users’ logged-in/out status using a [custom dimention](https://support.google.com/analytics/answer/2709828)
* Start logging insiders while prevent security/private bugs being logged
* Remove jQuery dependency

## Bugzilla link

[Bug 1529406 - Record user’s logged-in/out status and track insiders with Google Analytics](https://bugzilla.mozilla.org/show_bug.cgi?id=1529406)

PR #1130, opened 2019-03-01
Bug 1527053 - Can't search for "video" to find all relevant bugs

kyoshino:
Stop treating a word entered in the search bar as a bug alias to avoid unexpected results for users.

## Bugzilla link

[Bug 1527053 - Can't search for "video" to find all relevant bugs](https://bugzilla.mozilla.org/show_bug.cgi?id=1527053)

### DIFF
--- a/Bugzilla/Search/Quicksearch.pm
+++ b/Bugzilla/Search/Quicksearch.pm
@@ -145,8 +145,6 @@ sub quicksearch {
     _bug_numbers_only($searchstring);
   }
   else {
-    _handle_alias($searchstring);
-
     # Retain backslashes and quotes, to know which strings are quoted,
     # and which ones are not.
     my @words = _parse_line('\s+', 1, $searchstring);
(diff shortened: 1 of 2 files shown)


kyoshino:
Since #1119 is not landed, I can’t mention any stats, but some aliases as a quite simple, like [es6](https://bugzilla.mozilla.org/show_bug.cgi?id=es6), [spam](https://bugzilla.mozilla.org/show_bug.cgi?id=spam), [aria](https://bugzilla.mozilla.org/show_bug.cgi?id=aria), [kindle](https://bugzilla.mozilla.org/show_bug.cgi?id=kindle), [camera](https://bugzilla.mozilla.org/show_bug.cgi?id=camera) or [extensions](https://bugzilla.mozilla.org/show_bug.cgi?id=extensions) so the issue may be happening more often than we think.

kyoshino:
This is a confusing behaviour, and I’d instead suggest showing matched bugs on the dropdown menu of the quick search bar (#778).

kyoshino:
Not sure who can decide, but I think tags like **abusive** or **offtopic** can be used for attached images (attachment and comment are 1:1).

purelogiq:
There should be some way to search for the word 'video' via the navigation bar without going to the bug with that alias. Some ideas:

Breaking change: Require some special symbol to search for an alias like `@video` or something.

Non-breaking change: On the resulting bug page with the alias we could have a 'Did you mean to search for all bugs with video?' notification with a link to the regular search.

We could also ask the community what they think.

kyoshino:
> This is a confusing behaviour, and I’d instead suggest showing matched bugs on the dropdown menu of the quick search bar (#778).

I think I want to do this anyway. If a valid bug ID or alias is entered in the search bar, the dropdown should be showing it for faster access.

PR #1131, opened 2019-03-03
Bug 1503483 - Convert redirects to absolute path

kyoshino:
Make sure all the internal redirects are absolute URLs starting with the `basepath` param, which is `/` on BMO. Once this is merged, we can use pretty URLs like `/bug/1503483` or `/user/kohei`.

## Bugzilla link

[Bug 1503483 - Convert redirects to absolute path](https://bugzilla.mozilla.org/show_bug.cgi?id=1503483)


kyoshino:
Solving conflicts...

PR #1132, opened 2019-03-04
Bug 1481106 - Start logging bug_user_last_visit for all visited bugs even if the user is not involved

kyoshino:
[Update the last-visited time for the specified bug and current user](https://bmo.readthedocs.io/en/latest/api/core/v1/bug-user-last-visit.html#update-last-visited) even if the user is not directly involved in. This change allows to enhance the UX, like

* Show the new comments/changes indicator on any bug
* Highlight any bug updated since last visit (like GitHub issue list)
* Search any recently visited bug from the quick search bar (#778)
* Style any visited bug reliably among multiple devices

## Bugzilla link

[Bug 1481106 - Start logging bug_user_last_visit for all visited bugs even if the user is not involved](https://bugzilla.mozilla.org/show_bug.cgi?id=1481106)

### DIFF
--- a/Bugzilla/Bug.pm
+++ b/Bugzilla/Bug.pm
@@ -1363,17 +1363,6 @@ sub update {
     $self->{delta_ts} = $delta_ts;
   }
 
-  # Update last-visited
-  if ($user->can_see_bug($self->id)) {
-    $self->update_user_last_visit($user, $delta_ts);
-  }
-
-  # If a user is no longer involved, remove their last visit entry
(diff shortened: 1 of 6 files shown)


dylanwh:
This is fine, but I suspect `update_user_last_visit` needs to be implemented differently. Right now it has some weird race conditions which we'll see more if more people are using it.

kyoshino:
And the tests doesn’t like me…

kyoshino:
I guess the “weird race condition” is perhaps a UI issue? The blue indicator on bugs will disappear quickly once you (scroll down and) reach new changes. I’ll fix that.

PR #1135, opened 2019-03-05
Bug 1532766 - Make the the application root a static, cache-friendly redirect to /home

dylanwh:
For this patch I also moved around a lot of routes. For review making sure the bzapi and other routes are still the same would be good enough.


template/en/default/global/header.html.tmpl
```diff
@@ -291,7 +295,7 @@
 
 <header id="header" role="banner">
   <div class="inner">
-    <h1 id="header-title" class="title"><a href="[% basepath FILTER none %]" title="Go to home page">[% terms.Bugzilla %]</a></h1>
+    <h1 id="header-title" class="title"><a href="[% c.url_for('/home').to_abs FILTER none %]" title="Go to home page">[% terms.Bugzilla %]</a></h1>
```
dylanwh:
```suggestion
    <h1 id="header-title" class="title"><a href="[% c.url_for('/home').to_abs FILTER html %]" title="Go to home page">[% terms.Bugzilla %]</a></h1>
```

PR #1137, opened 2019-03-08
Bug 1533610 - Search both open and closed bugs by default

kyoshino:


kyoshino:
This needs to be postponed until the new search results page goes live.

PR #1145, opened 2019-03-13
Bug 1283392 - Query for history or comment containing private bug id should return no data for this bug

kyoshino:
Use `Bugzilla::Bug->new()` in the bug comment and history APIs so any private bug won’t be throwing but simply skipped.

I have just noticed that `/rest/bug` supports the experimental, undocumented `permissive` param that returns an error object rather than skipping private bugs. The param is apparently not working for some reason, but this should be the default behaviour in our next generation API. I’m *not* adding the param to the comment and history APIs this time.

## Bugzilla link

[Bug 1283392 - Query for history or comment containing private bug id should return no data for this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1283392)

### DIFF
--- a/Bugzilla/WebService/Bug.pm
+++ b/Bugzilla/WebService/Bug.pm
@@ -318,8 +318,9 @@ sub comments {
       {function => 'Bug.comments', params => ['ids', 'comment_ids']});
   }
 
-  my $bug_ids     = $params->{ids}         || [];
-  my $comment_ids = $params->{comment_ids} || [];
+  my $bug_ids      = $params->{ids}         || [];
+  my $comment_ids  = $params->{comment_ids} || [];
+  my $skip_private = $params->{skip_private} ? 1 : 0;
 
   my $dbh  = Bugzilla->switch_to_shadow_db();
   my $user = Bugzilla->user;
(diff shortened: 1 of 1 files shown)


Bugzilla/WebService/Bug.pm
```diff
@@ -478,10 +486,19 @@ sub history {
...
+  # Cache permissions for bugs. This highly reduces the number of calls to the DB.
+  # visible_bugs() is only able to handle bug IDs, so we have to skip aliases.
+  my @int = grep { $_ =~ /^\d+$/ } @$ids;
+  $user->visible_bugs(\@int);
+
   my @return;
   foreach my $bug_id (@$ids) {
     my %item;
-    my $bug = Bugzilla::Bug->check($bug_id);
+
```
dylanwh: I'd continue to use check and merely add `use Try::Tiny` to the top.

Bugzilla/WebService/Bug.pm
```diff
@@ -478,10 +486,19 @@ sub history {
...
+    # Use new() instead of check() and skip any invisible bug
+    my $bug = Bugzilla::Bug->new({id => $bug_id, cache => 1});
```
dylanwh:
```suggestion
    my $bug;
    try {
      $bug = Bugzilla::Bug->new({id => $bug_id, cache => 1});
    };
```

Bugzilla/WebService/Bug.pm
```diff
@@ -478,10 +486,19 @@ sub history {
...
+    next if $bug->error || !$user->can_see_bug($bug->id);
```
dylanwh:
```suggestion
    next unless $bug;
```

kyoshino:
Given that tests are failing, I’ve introduced the `skip_private` param for backward compatibility. It’s just a temporary workaround so won’t be added to the public API doc.

PR #1148, opened 2019-03-14
Bug 994896 - Add the ability to get comments, attachments, and history using Bug.get

kyoshino:
* Allow to retrieve comments, attachments and history at once with `/rest/bug/[ID]?include_fields=attachments,comments,history`, just like the legacy BzAPI.
* Moved the Comment object’s `raw_text` field from the EditComments extension to core, so `/rest/bug/[ID/comment` and `/rest/bug/[ID]?include_fields=comments` will be consistent. Otherwise it would require an explicit field option like `/rest/bug/[ID]?include_fields=comments,raw_text`.
* Update relevant docs.

## Bugzilla link

[Bug 994896 - Add the ability to get comments, attachments, and history using Bug.get](https://bugzilla.mozilla.org/show_bug.cgi?id=994896)

### DIFF
--- a/Bugzilla/WebService/Bug.pm
+++ b/Bugzilla/WebService/Bug.pm
@@ -110,6 +110,16 @@ use constant ATTACHMENT_MAPPED_RETURNS => {
   mimetype    => 'content_type',
 };
 
+our %api_field_types = (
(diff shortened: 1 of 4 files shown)


kyoshino:
Thanks! 🙌

PR #1156, opened 2019-03-16
Bug 1527841 - Add ability to search for bugs where assignee or needinfo’d people have last seen gt/lt X

kyoshino:
Add new `assignee_last_seen`, `needinfo` and `needinfo_last_seen` fields to let people search bugs based on assignee or needinfo’d user’s last activity. A typical custom search criteria will be:

* Needinfo / is not empty
* Needinfo Last Seen Date / is less than / -2y

## Bugzilla link

[Bug 1527841 - Add ability to search for bugs where assignee or needinfo’d people have last seen gt/lt X](https://bugzilla.mozilla.org/show_bug.cgi?id=1527841)

### DIFF
--- a/Bugzilla/Hook.pm
+++ b/Bugzilla/Hook.pm
@@ -533,6 +533,25 @@ The hash of changed fields. C<< $changes->{field} = [old, new] >>
 
 =back
 
+=head2 buglist_special_parsing
+
+This happens in L<Bugzilla::Search/SPECIAL_PARSING>, which defines special
+parsing methods for user input, such as relative date formats or user pronouns.
+
+Params:
+
+=over
+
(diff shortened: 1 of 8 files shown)


kyoshino:
Need API.

emceeaich:
This will make specifying searches that triage owners need to watch so much easier.

kyoshino:
I’m honestly not sure if this is useful. Active users ignoring needinfos are rather a problem than inactive people never answer questions, eh?

kyoshino:
I was implementing the `needinfo` and `needinfo_detail` fields to `/rest/bug`, but realized that it would overlap with the `flags` and `cc` fields, so just added to `last_activity` to `cc` and other user objects, which will be retuned if the request is authenticated.

PR #1157, opened 2019-03-18
bump version to 20190318.1

kyoshino:
We need to fix #1155 to avoid browser hangs due to certain SVG images. A potential ride-along: #1147 that solves a [bit confusing UI](https://bugzilla.mozilla.org/show_bug.cgi?id=1478901#c3).

### DIFF
--- a/Bugzilla.pm
+++ b/Bugzilla.pm
@@ -13,7 +13,7 @@ use warnings;
 
 use Bugzilla::Logging;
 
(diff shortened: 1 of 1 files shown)


purelogiq:
#1068 and #1085 will be ready in half an hour or so (not including review time) if we want to get those in the next deploy.

kyoshino:
Looks like we don’t need a release today since inline images could be turned off by default.

PR #1159, opened 2019-03-19
Bug 1534084 - Hide Phablicator requests from Attachment list

dklawren:


dklawren:
> but it should say
> 
> > **Attachments** (2 attachments, 4 obsolete attachments)
> 
> instead, given 2 Phab attachments are hidden.

Good catch. will fix.

PR #1165, opened 2019-03-23
Bug 1538115 - Add shortcuts for tracking & status flags

kyoshino:
Allow to use tracking flag pronouns/aliases to search bugs so triage queries don’t have to be updated for each release. The mappings include `cf_tracking_firefox_nightly` -> `cf_tracking_firefox68` and `cf_status_firefox_esr` -> `cf_status_firefox_esr60`.

## Bugzilla link

[Bug 1538115 - Add shortcuts for tracking & status flags](https://bugzilla.mozilla.org/show_bug.cgi?id=1538115)


kyoshino:
Todo:

* Add Thunderbird support
* Use cron to update cache

kyoshino:
@rail just told me the product-details URL will change probably next week. Let’s wait until it happens.

kyoshino:
The URL didn’t change, and the data also remains the same.

rail:
Sorry, I should have provided more details last week. The idea was to move the website to cloudops without changing anything for the consumers. I just wanted to make sure to let people know about the move in case they see something wrong.

kyoshino:
Thanks for the clarification! Tested the PR again and looked fine 🙂

extensions/TrackingFlags/Extension.pm
```diff
@@ -335,6 +367,68 @@ sub active_custom_fields {
   @$$fields = sort { $a->sortkey <=> $b->sortkey } values %field_hash;
 }
 
+sub _get_product_version {
```
dylanwh: Storing this in the request cache is probably fine. But if we want it in memcached it would need its own (to borrow a term from the windows registry) hive to live in.

It appears https://product-details.mozilla.org/1.0/ is behind cloudfront and returns a response in 200s. Do we really need memcaching for this?

dylanwh: Also, the max-age of that response is 300 seconds.

kyoshino: Since the [Crash Stats table](https://github.com/mozilla-bteam/bmo/blob/master/extensions/BMO/web/js/firefox-crash-table.js) is also using `firefox_versions.json`, and the version info is not updated so frequently, we should probably cache it for a certain period of time and [try to use it](https://bugzilla.mozilla.org/show_bug.cgi?id=1543489).

I’ll update this to remove the cron dependency and use `Mojo::UserAgent::Role::Cache`.

kyoshino:
@dylanwh @dklawren Forgot to remove the WIP label but this should be ready. Can you have a look?

sylvestre:
@dylanwh it would be great if you could move that on top of your review list :)

PR #1171, opened 2019-03-27
Bug 1508695 - Incorrect or missing tracking flags on search results in REST API

dylanwh:


dylanwh:
I'm landing this because it's obviously correct as it is what we used to do.

PR #1174, opened 2019-03-28
Bug 1071066 - For bugs with aliases, inline history displays the bug number rather than the bug alias

kyoshino:
Use aliases, if available, for the history of bug list fields like Blocks and Regressions. The same applies to the See Also field that currently shows a bug ID prefixed with “bug”. This PR makes change to both the inline history on the modal bug page and the history page.

## Bugzilla link

[Bug 1071066 - For bugs with aliases, inline history displays the bug number rather than the bug alias](https://bugzilla.mozilla.org/show_bug.cgi?id=1071066)

### DIFF
--- a/Bugzilla/Template.pm
+++ b/Bugzilla/Template.pm
@@ -777,7 +777,7 @@ sub create {
           return sub {
             my $buglist = shift;
             return join(", ",
-              map(get_bug_link($_, $_, $options), split(/ *, */, $buglist)));
+              map { get_bug_link($_, $_, $options) } split(/\s*,\s*/, $buglist));
           };
         },
         1
--- a/Bugzilla/Template.pm
+++ b/Bugzilla/Template.pm
@@ -771,11 +771,17 @@ sub create {
         1
       ],
 
-      bug_list_link => sub {
-        my ($buglist, $options) = @_;
-        return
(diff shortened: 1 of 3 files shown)


Bugzilla/Template.pm
```diff
@@ -771,11 +771,17 @@ sub create {
...
-        return
-          join(", ", map(get_bug_link($_, $_, $options), split(/ *, */, $buglist)));
-      },
+      bug_list_link => [
+        sub {
+          my ($context, $options) = @_;
+          return sub {
+            my $buglist = shift;
+            return join(", ",
+              map(get_bug_link($_, $_, $options), split(/ *, */, $buglist)));
```
dylanwh:
```suggestion
              map { get_bug_link($_, $_, $options) } split(/\s*,\s*/, $buglist));
```

only ever use the block form of `grep` and `map` and friends. The other one is confusing and weird. I mean, more than normal.

Also literal spaces in regex are bad, editors can screw that up and things like `\t` can be hard to find bugs.

kyoshino: I think I copy & pasted from somewhere else. Thanks!

dylanwh: Yeah. it's safe to say a lot of existing code wouldn't pass my code review.

PR #1177, opened 2019-03-28
Bug 1539849 - Multiple uplift request is not working as expected

kyoshino:
#802 didn’t work properly 😞

* The flag IDs set up locally and coded in the uplift template were wrong, because these were copy & pasted from the UI which included **non-breaking hyphens**. So you’ve got a “flag not found” API error due to an ID mismatch. Remove the unnecessary `no_break` filter and fix the template.
* Remove the unnecessary question “Do you want to request approval of these patches as well?” from the comment body.


kyoshino:
Thanks for testing!

PR #1183, opened 2019-03-30
Bug 1539302 - Add "attachment is patch" field to bug

kyoshino:
* Treat Phablicator, MozReview and GitHub pull requests as patches
* Add “Attachment is patch” column to search results

## Bugzilla link

[Bug 1539302 - Add "attachment is patch" field to bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1539302)


kyoshino:
Yeah, that probably makes sense. Will file a separate bug and see if it won’t introduce any side effect.

PR #1187, opened 2019-04-02
Bug 1344091 - Reorganize the header and modules on modal bug UI

kyoshino:
* Reorganize bug modal modules based on my [information architecture plan](https://bugzilla.mozilla.org/show_bug.cgi?id=1344091#c3) so similar fields are properly grouped.
* Hide more fields if it’s empty or unspecified so the view mode will be cleaner: Version, Target, Points, Whiteboard, Triage Owner
* Toolbar: Include both search navigation and page action buttons
* Header: Remove redundant info and instead add simple status (Open/Closed), reported date and updated/resolved date; add menus that are currently on the page footer
* User names on modules: now displayed as Gravatar + nick
* Attachments module: now always displayed with the Attach New File button-like link
* Misc fixes: simplified URL labels, better module header subtitles, etc.

## Bugzilla link

[Bug 1344091 - Move all the related bug fields into the same module](https://bugzilla.mozilla.org/show_bug.cgi?id=1344091)

### DIFF
--- a/extensions/BugModal/template/en/default/bug_modal/attachments.html.tmpl
+++ b/extensions/BugModal/template/en/default/bug_modal/attachments.html.tmpl
@@ -12,6 +12,7 @@
   # obsolete_attachments: array of obsolete attachment objects
   #%]
 
+[% IF active_attachments || obsolete_attachments %]
 <table role="table" class="layout-table" id="attachments">
   [% FOREACH attachment IN bug.attachments %]
     [%
@@ -94,12 +95,19 @@
     </tr>
   [% END %]
 </table>
+[% END %]
 
-<div id="attachments-actions">
-  [% IF obsolete_attachments %]
-    <button type="button" id="attachments-obsolete-btn" class="secondary">Show Obsolete Attachments</button>
-  [% END %]
-</div>
-
-[%# BMO - attachment related warnings %]
-[% Hook.process("warnings") %]
+<footer id="attachments-footer">
+  <div id="attachments-actions">
+    [% IF user.id %]
+      <a href="[% basepath FILTER none %]attachment.cgi?bugid=[% bug.id FILTER uri %]&amp;action=enter" id="attachments-add-link">
+        <button type="button" id="attachments-add-btn" class="secondary">Attach New File</button>
+      </a>
+    [% END %]
+    [% IF obsolete_attachments %]
+      <button type="button" id="attachments-obsolete-btn" class="secondary">Show Obsolete</button>
+    [% END %]
+  </div>
+  [%# BMO - attachment related warnings %]
+  [% Hook.process("warnings") %]
+</footer>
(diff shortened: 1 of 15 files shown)


kyoshino:
Other things need to go first.

kyoshino:
This is now ready for review.

kyoshino:
Fixed a couple of small things. Should be ready.

kyoshino:
@dylanwh @dklawren This is apparently blocking my work removing any table layout from the modal UI and the following mobile support. Would like to see the PR merged.

kyoshino:
I think that’s enough. As this bug blocks various other bugs, it would be great to have this merged soonish 🙂 @dklawren

kyoshino:
Fixed the narrowed `<input>` size that may have regressed at some point. Ready for review 🙆‍♂

dklawren:
Mark as assigned should be one line

image

kyoshino:
Fixed the Mark as Assigned button. Note that the text alignment is still not great overall, and that will be solved in #1366.

dklawren:
A small issue I noticed now that it is on bugzilla-dev:

image

dklawren:
This has now been deployed to bugzilla-dev.allizom.org for feedback.

dklawren:
> 
> 
> A small issue I noticed now that it is on bugzilla-dev:
> 
> image

Nevermind. This looks better once I did shift-reload.

extensions/BugModal/template/en/default/bug_modal/attachments.html.tmpl
```diff
@@ -94,12 +95,19 @@
...
-  [% END %]
-</div>
-
-[%# BMO - attachment related warnings %]
-[% Hook.process("warnings") %]
+<footer id="attachments-footer">
+  <div id="attachments-actions">
+    [% IF user.id %]
+      <a href="[% basepath FILTER none %]attachment.cgi?bugid=[% bug.id FILTER uri %]&amp;action=enter" id="attachments-add-link">
+        <button type="button" id="attachments-add-btn" class="secondary">Attach New File</button>
```
globau: In [bug 1217890](https://bugzilla.mozilla.org/show_bug.cgi?id=1217890) this was changed from a `<button>` to a link to allow for opening in a new tab/window/etc.

It would be better if this were a link styled to look like a button rather than an actual `<button>`.

kyoshino: Ah, this one should be okay. It’s a `<button>` wrapped in `<a>` so the link opens in a new tab with Accel+click. For a better a11y, I’ll just add `role="none"` to the `<button>` to make sure it’s just for presentation.

The Vote button has the problem, but I’ll keep it as is because it’s an existing bug and the voting functionality will be removed in the near future.

kyoshino: You can find a similar `<a>`-wrapped `<button>` on the User Profile page, btw.

kyoshino:
Rebased against the current master, and fixed the `<button>` issue. Let’s ship later this week or next 😺

extensions/BugModal/template/en/default/bug_modal/attachments.html.tmpl
kyoshino: Well, the Edit Profile & Preferences button on the User Profile page doesn’t open in a new tab, probably because it’s not `<button type="button">` but `<input type="button">`. I’ll fix that in a separate bug.

kyoshino:
Okay, I’ll go ahead and merge this. Happy to follow up any remaining issues.

PR #1189, opened 2019-04-02
Bug 1531481 - Add bug type labels to My Dashboard

kyoshino:
Add the Type column to the “big query table” on My Dashboard.

Screenshot_2019-04-02 My Dashboard

## Bugzilla link

[Bug 1531481 - Add bug type labels to My Dashboard](https://bugzilla.mozilla.org/show_bug.cgi?id=1531481)

### DIFF
--- a/extensions/MyDashboard/lib/Queries.pm
+++ b/extensions/MyDashboard/lib/Queries.pm
@@ -39,6 +39,7 @@ use constant QUERY_ORDER => ("changeddate desc", "bug_id");
 # Share with buglist.cgi?
 use constant SELECT_COLUMNS => qw(
   bug_id
+  bug_type
   bug_status
   short_desc
   changeddate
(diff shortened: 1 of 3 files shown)


dylanwh:
fwiw, I second dkl's thoughts. Maybe the type could be a prefix on the bug id? I wonder if that would look ok.

dklawren:
> 
> 
> fwiw, I second dkl's thoughts. Maybe the type could be a prefix on the bug id? I wonder if that would look ok.

That's a good idea. Similar to how we do the lock icon on private bugs in buglist.cgi. Just combine the type icon and bug id link in the ID column.

kyoshino:
I’d like to make the Type column sortable rather than prefixing the ID. The same idea applies to the private/lock icon on `buglsit.cgi` which will also have a separate column as part of the upcoming search results page overhaul.

We’ll eventually get rid of the YUI table, so I think it’s fine to use “T” as a label in the meantime.

kyoshino:
Also: the current `buglist.cgi` shows the Type column at the right hand side of ID, not left, but it’s a known limitation of the template hardcoding the ID column. I’ll change that too as part of the overhaul.

PR #1190, opened 2019-04-02
Bug 1541204 - Bug type field is reset to default when selecting/changing a component, breaks "remember values as bookmarkable template"

kyoshino:
Keep the Type field regardless of the component default value once the user changes it manually.

## Bugzilla link

[Bug 1541204 - Bug type field is reset to default when selecting/changing a component, breaks "remember values as bookmarkable template"](https://bugzilla.mozilla.org/show_bug.cgi?id=1541204)


kyoshino:
Gah, I probably need a bit different fix for a new radio-button-like UI to be introduced in [Bug 1541133](https://bugzilla.mozilla.org/show_bug.cgi?id=1541133). Sorry, but closing this.

PR #1196, opened 2019-04-03
Bug 1541555 - Add facility for requiring an API Key to always come from the same IP address

dylanwh:
It's been requested that an API key for a service owned by @calixteman be accessible from only a single source. This is presumably to hedge against the risk of the key leaking.

We already store the IP that last used an API key, so it seems easiest to add a "sticky" bit that makes the key not work when the IP doesn't match the previous one.

This PR adds the "sticky" column, and if the `last_used_ip` is set (e.g. the API key has been used) it can't be used from another IP. The sticky bit cannot be unset.

# TODO
- [x] file bug
- [x] split out schema change
- [x] docs

### DIFF
--- a/docs/en/rst/using/preferences.rst
+++ b/docs/en/rst/using/preferences.rst
@@ -132,9 +132,14 @@ change your password everywhere.
 You can create more than one API key if required. Each API key has an optional
 description which can help you record what it is used for.
 
-On this page, you can unrevoke, revoke and change the description of existing
+On this page, you can unrevoke, revoke, make sticky, and change the description of existing
 API keys for your login. A revoked key means that it cannot be used. The
 description is optional and purely for your information.
+
+Sticky API keys may only be used from one IP address, which reduces the risk
+of the key being leaked. The IP address is the one the key was last used
+from. The expected workflow is that the sticky bit will be set once your application
+(or script) is setup. The sticky attribute may only be set, it can't ever be unset.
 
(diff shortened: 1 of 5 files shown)


userprefs.cgi
```diff
@@ -872,6 +878,7 @@ sub SaveApiKey {
         }
       }
     }
+
```
dylanwh:
```suggestion
```

dylanwh:
@dklawren re-requesting for documentation review.

docs/en/rst/using/preferences.rst
```diff
@@ -132,10 +132,15 @@ change your password everywhere.
...
-On this page, you can unrevoke, revoke and change the description of existing
+On this page, you can unrevoke, revoke, make sticky, and change the description of existing
 API keys for your login. A revoked key means that it cannot be used. The
 description is optional and purely for your information.
 
+Sticky API keys may only be used from one IP address, which reduces the risk
+of the key being leaked. The IP address is the one the key was last used
+from. The expected workflow is that the sticky bit will be set once your application
+(or script) is setup. The sticky attribute may only be set, it can't ever be unset.
+
```
dklawren: This looks good but I think we should at least have one sentence on the userprefs.cgi page as well. 

Note: Keys marked as sticky may only be used from a single IP address.

dylanwh:
@dklawren is this ok now?

PR #1197, opened 2019-04-03
Bug 1541582 - Set qe-verify to + when manual QA requested upon uplift

mnoorenberghe:
'?' is used ask the dev to asses whether to uplift but that's what the form is already doing.


kyoshino:
Asking in [Bug 1528106 - Update definition of `qe-verify` values?](https://bugzilla.mozilla.org/show_bug.cgi?id=1528106) what should we do.

PR #1198, opened 2019-04-03
Bug 1541618 - Add triage owner to edit components view columns.

kyoshino:
On the Edit Components page, replace the Watch User column with Triage Owner, and reorder the Active column.

## Bugzilla link

[Bug 1541618 - Add triage owner to edit components view columns.](https://bugzilla.mozilla.org/show_bug.cgi?id=1541618)


template/en/default/admin/components/list.html.tmpl
```diff
@@ -105,6 +110,11 @@
        content => my_component.default_assignee.login
      }
   %]
+  [% overrides.triage_owner.name.${my_component.name} = {
+       override_content => 1
+       content => my_component.default_assignee.login
```
dklawren: my_component.triage_owner.login

kyoshino: Ouch. Thanks, fixing now :sweat_smile:

kyoshino:
Somehow the Watch User column is not going away even though I have removed the hook and run `checksetup.pl`. Is the script not removing complied templates properly?

dylanwh:
we shouldn't compile templates at all when MODE==development. Nuke template_cache/ for now.

kyoshino:
Um, the column added by the hook still exists after ` rm -r template_cache/` 🤔

kyoshino:
Figured out why: `vagrant rsync web` doesn’t remove deleted files because [`RSYNC_ARGS`](https://github.com/mozilla-bteam/bmo/blob/master/Vagrantfile#L19-L30) doesn’t have `--delete` option that can be found in the [default](https://www.vagrantup.com/docs/synced-folders/rsync.html#rsync__args). Can we add it?

PR #1209, opened 2019-04-05
Bug 1527459 - Display dependency bugs by type, using the same colour/icon for each

kyoshino:
Show Depends on, Blocks and other bug IDs fields by bug type. For example, [Bug 1451850](https://bugzilla.mozilla.org/show_bug.cgi?id=1451850) will look like this:

Screenshot_2019-04-05 1451850 - (fission)  meta  Project Fission

## Bugzilla link

[Bug 1527459 - Display dependency bugs by type, using the same colour/icon for each](https://bugzilla.mozilla.org/show_bug.cgi?id=1527459)


kyoshino:
The icons are changed in #1210 and it now looks like this:

Screenshot_2019-04-05 1451850 - (fission)  meta  Project Fission(1)

PR #1222, opened 2019-04-10
no bug - Fix missing object error when submitting new bug

kyoshino:
A fallout from #1218. Fix the following error raised when you submit a new bug:

> Can't locate object method "content_security_policy" via package "Bugzilla::CGI" at post_bug.cgi line 307.

**But there's still another problem**: the new bug's title will be undefined, and the browser console also says:

> TypeError: BUGZILLA.user.settings is undefined

### DIFF
--- a/Bugzilla.pm
+++ b/Bugzilla.pm
@@ -13,7 +13,7 @@ use warnings;
 
 use Bugzilla::Logging;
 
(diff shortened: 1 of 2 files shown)


dylanwh:
it's great because I just opened the github UI to file the same PR.

PR #1226, opened 2019-04-11
Bug 1543741 - Blocklist requests getting filed as 'defect' instead of 'task' because of custom form

kyoshino:
Properly check for the bug type defined at not only the instance level but also the component and product level when a new bug is filed without a user-defined type.

## Bugzilla link

[Bug 1543741 - Blocklist requests getting filed as 'defect' instead of 'task' because of custom form](https://bugzilla.mozilla.org/show_bug.cgi?id=1543741)


kyoshino:
The commented-out product level default bug type will be implemented in #1211.

PR #1232, opened 2019-04-13
Bug 1541898 - Allow to search on product/component pair

kyoshino:
Allow to use product/component pairs like “Core::General, IPC” for the Component field in Custom Search. This also works with simple queries like `buglist.cgi?component=Core%3A%3AGeneral&component=IPC`.

## Bugzilla link

[Bug 1541898 - Allow to search on product/component pair](https://bugzilla.mozilla.org/show_bug.cgi?id=1541898)

### DIFF
--- a/Bugzilla/Search.pm
+++ b/Bugzilla/Search.pm
@@ -2871,7 +2871,6 @@ sub _assignee_last_login {
 sub _component_nonchanged {
   my ($self, $args) = @_;
   my $dbh = Bugzilla->dbh;
-  my (@products, @components);
   my $product;
 
   # Allow to search product/component pairs like "Core::General" with a simple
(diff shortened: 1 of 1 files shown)


Bugzilla/Search.pm
```diff
@@ -2843,13 +2843,32 @@ sub _assignee_last_login {
 
 sub _component_nonchanged {
   my ($self, $args) = @_;
+  my $dbh = Bugzilla->dbh;
+  my (@products, @components);
+
+  # Allow to search product/component pairs like "Core::General, IPC"
+  foreach my $word (split(/[\s,]+/, $args->{value})) {
```
dylanwh: I'm concerned that splitting on spaces here is not a good idea because many components include spaces. Some components consider commas too, but I'd be willing to deprecate that.
Alternatively, `Text::ParseWords` can be used.

kyoshino: Ah true. Removing `\s` 😆

kyoshino:
And looks like this is not implemented properly. `Core :: General, bugzilla.mozilla.org :: Email Notifications` returns bugs in 4 components:

* `Core :: DOM: Push Notifications`
* `Core :: General`
* `bugzilla.mozilla.org :: Email Notifications`
* `bugzilla.mozilla.org :: General`

kyoshino:
Supporting “contains any of the words” (`anywords`)  and similar operators is probably difficult because of the spaces in component names. I’ll try to solve the “is equal to” (`equals`) use case.

kyoshino:
Debugged this and realized the args for `build_subselect()` were wrong. The fix is in.

PR #1235, opened 2019-04-17
Bug 1545269 - Request for Bug Dependency Graphs return a 404

dylanwh:
This should expose just the new graphs.

### DIFF
--- a/Bugzilla/App/Static.pm
+++ b/Bugzilla/App/Static.pm
@@ -11,7 +11,7 @@ use Bugzilla::Constants qw(bz_locations);
 
 my $LEGACY_RE = qr{
     ^ (?:static/v(?<version>[0-9]+\.[0-9]+)/) ?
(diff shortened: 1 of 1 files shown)


Bugzilla/App/Static.pm
```diff
@@ -11,7 +11,7 @@ use Bugzilla::Constants qw(bz_locations);
 
 my $LEGACY_RE = qr{
     ^ (?:static/v(?<version>[0-9]+\.[0-9]+)/) ?
-    (?<file>(?:extensions/[^/]+/web|(?:image|skin|j|graph)s)/.+)
+    (?<file>(?:extensions/[^/]+/web|(?:image|skin|j|graph)s|data/webdot).+)
```
purelogiq: Removing the `/` at the end is intentional?

dylanwh:
```suggestion
    (?<file>(?:extensions/[^/]+/web|(?:image|skin|j|graph)s|data/webdot)/.+)
```

PR #1236, opened 2019-04-18
Bug 1545295 - socorro lens chart for crash statistics blocked by CSP (Blocked by Content Security Policy)

dylanwh:


kyoshino:
Actually I tried this a bit last night but it seems to require more work. `metricsgraphics/socorro-lens.html` is not rendered properly. It shows `[% script_nonce FILTER none %]` as is so the script nonce is not outputted.

dylanwh:
yeah, I started testing after PR. It's static -- it can't use [% foo %] directives. So I'll move out the inline js.

dklawren:
> 
> 
> yeah, I started testing after PR. It's static -- it can't use [% foo %] directives. So I'll move out the inline js.

Are you going to do that as part of this PR or a new one?

dylanwh:
It should be all working now. I see this in the firefox console:

```
Content Security Policy: The page’s settings blocked the loading of a resource at inline (“script-src”).
```

But this doesn't happen in not-firefox. The line it refers to is line 3 of the socorro lens file.

PR #1245, opened 2019-04-23
Update of the NDA form. If there is a CASA ticket, use the ticket ins…

sylvestre:
…tead.


sylvestre:
Trying https://github.com/mozilla-bteam/bmo/pull/1246

PR #1246, opened 2019-04-23
no bug - Update of the NDA form

sylvestre:
If there is a CASA ticket, use the ticket instead.


sylvestre:
@dylanwh do you know why circleci is failing ?

dylanwh:
either:

1. this needs to be rebased to `master`, or
2. the commit must reference a bug. Otherwise we can't calculate the build info step.

```
no new revisions.  make sure you run this script before production is updated.
Exited with code 255
```

PR #1249, opened 2019-04-24
Bug 1546539 - [SEO] Many legacy and printable bug pages are cached by Google

kyoshino:
* Drop the `format=default` support from `show_bug.cgi` and redirect visitors to the modal UI.
* Drop `format=multiple` as well if only 1 bug ID is provided.
* Logged-in users who have explicitly disabled the modal UI through the user preference are not affected, will continue to see the legacy UI for now.
* Remove the legacy UI link from the modal UI.
* Remove the For Printing link from the both legacy and modal UI.

## Bugzilla link

[Bug 1546539 - [SEO] Many legacy and printable bug pages are cached by Google](https://bugzilla.mozilla.org/show_bug.cgi?id=1546539)


kyoshino:
Hmm I guess `format=multiple` can also be ditched now. It’s actually not suitable for printing due to the fixed header, but no one has complained about it, which means no one is using it.

kyoshino:
Fixing one more test failure...

PR #1252, opened 2019-04-24
Bug 1546788 - Deselect bug type by default unless component-specific type is defined

kyoshino:
Let users select bug type when they file a new bug unless the selected component doesn’t have the default bug type. Aside from this code change, an [admin task](https://bugzilla.mozilla.org/show_bug.cgi?id=1546815) is required to complete the update on production.

## Bugzilla link

[Bug 1546788 - Deselect bug type by default unless component-specific type is defined](https://bugzilla.mozilla.org/show_bug.cgi?id=1546788)


sylvestre:
@dylanwh Would you mind reviewing it?
We need to see that live (we see a lot of bugs incorrectly classified)
thanks

kyoshino:
Thanks, will do.

kyoshino:
Currently the default value is `defect` code-wise (not scheme-wise) so the population actually won’t happen but I’ve made the column non-null like other fields.

kyoshino:
@dylanwh Should I send a scheme-only PR?

PR #1255, opened 2019-04-29
Bug 1547714 - Attachment with application/octet-stream MIME type should not be previewed even if it’s actually text file

kyoshino:
Prevent attachments from being previewed when it’s marked as binary, even if the content is plaintext. This should have been included in #1168.

## Bugzilla link

[Bug 1547714 - Attachment with application/octet-stream MIME type should not be previewed even if it’s actually text file](https://bugzilla.mozilla.org/show_bug.cgi?id=1547714)


kyoshino:
No, SVG is still forbidden. The trick is `/^image\/(?!vnd|svg).+$/`

PR #1243, opened 2019-04-23
Bug 1540715 - Fix offset of historical dates in secbugs report

purelogiq:
`FILTER time` from Bugzilla's template helpers was inconsistent on
how it rendered the dates of the columns (due to different system
timezones that change based on where the code is being run).

This patch sets manually formats the time (since FILTER time always wants to convert to the local or user's timezone).

PR #1257, opened 2019-04-29
Bug 1546444 - Parameterize time that the security bugs report uses

purelogiq:
Prior to this commit the report could specify a date to base the report
of off, but, it didn't accept a time (hour, min, second).

With this commit the report can be based on the exact day and time it
was run instead of just 0000 UTC of the specified date.

E.g. secbugsreport.pl $(date +'%Y %m %d %H %M %S %z').


purelogiq:
Note this that this change is independent of https://github.com/mozilla-bteam/bmo/pull/1243, which in itself is a valid bug.

purelogiq:
Blocked on getting confirmation from dveditz that he wants to go ahead with this (since it will impact the history again). With confirmation, we will want to have ops update the command that gets run.

PR #1261, opened 2019-05-03
Bug 1377977 - Implement initial version of post-Sandstone theme including Dark Mode

kyoshino:
**This must be shipped by May 16** just in time for the Firefox 67 GA on May 21.

* Introduce a new modern theme featuring both Light & **Dark Mode**, which will be automatically switched according to the system’s appearance setting.
* Demonstrate the new [`prefers-color-scheme`](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme) media query currently supported by Firefox 67+ and Safari 12.1+.
* Typography: Replaced Fira Sans Regular with FiraGO Book for better readability.
* Miscellaneous design improvements.

post-sandstone-dark
post-sandstone-light

## Bugzilla link

[Bug 1377977 - Implement initial version of post-Sandstone theme including Dark Mode](https://bugzilla.mozilla.org/show_bug.cgi?id=1377977)


kyoshino:
@dylanwh I have moved Socorro Lens from static files under `/public/` to a page template under the BMO extension so it can be styled easily. Do you expect any side effect?

dylanwh:
> @dylanwh I have moved Socorro Lens from static files under `/public/` to a page template under the BMO extension so it can be styled better & easily. Do you expect any side effect?
> 
> Screenshot:
> 
> image

Slightly worse performance, but probably not a show-stopper.

kyoshino:
I have removed unnecessary jQuery, Bootstrap, Font Awesome, Open Sans and PT Serif files, so the frontend gets a performance gain 😄

Bugzilla/Constants.pm
```diff
@@ -740,7 +740,7 @@ sub DEFAULT_CSP {
       'https://crash-stop-addon.herokuapp.com',
     ],
     worker_src  => ['none',],
-    img_src     => ['self', 'blob:', 'https://secure.gravatar.com'],
+    img_src     => ['self', 'data:', 'blob:', 'https://secure.gravatar.com'],
```
dylanwh: is data: safe for an img-src?

dylanwh:
The github button looks out of place now

Screen Shot 2019-05-09 at 23 41 47

Bugzilla/Constants.pm
kyoshino: I believe so... [SVG used as image](https://developer.mozilla.org/en-US/docs/Web/SVG/SVG_as_an_Image) cannot execute JavaScript, and MIME types other than `image/*` are simply ignored.

* https://security.stackexchange.com/questions/177412/restrict-js-in-svgs
* https://security.stackexchange.com/questions/94993/is-including-the-data-scheme-in-your-content-security-policy-safe

kyoshino:
This is the current log-in form. The GitHub button has been has been there.

PR #1269, opened 2019-05-08
Bug 1549873 - Create user role tables for product, component, bug and migrate existing tables

kyoshino:
Create the following tables to manage user roles on each level. This allows to make Triage Owner, Assignee and other fields to have multiple users, and to create new roles like Product Owners/Members like GitHub repositories.

* `product_user_map ` — merge `product_reviewers`
* `component_user_map ` — merge `component_cc` and `component_reviewers`
* `bug_user_map` — merge `cc` and `bug_mentors`

## Bugzilla link

[Bug 1549873 - Create user role tables for product, component, bug and migrate existing tables](https://bugzilla.mozilla.org/show_bug.cgi?id=1549873)


kyoshino:
This migrates the existing `cc` table, so requires a downtime of, say, 5 minutes. Can we do it along with the [Triage Owner migration](https://bugzilla.mozilla.org/show_bug.cgi?id=1545327) that depends on it?

kyoshino:
If we cannot avoid a downtime for this, I want to do more things together: 

1. [History migration](https://bugzilla.mozilla.org/show_bug.cgi?id=1509332): replace hardcoded stuff with IDs as much as possible.
2. Assignees and QA Contacts migration: Make them multiple in the DB but keep the UI and most of the code by adding a “shim” to `Bug.pm` like this:

```perl
sub assigned_to {
  return $self->assignees[0];
}
sub assignees {
  # return multiple assignees
}
```

kyoshino:
Okay, let’s migrate Assignees and QA Contacts later. This itself should work now.

kyoshino:
Looks like I should be migrating Assignee anyway so it can be empty or multiple.

kyoshino:
Well, multiple triage owner is the priority, so this can go without the Assignee & QA Contact change.

PR #1274, opened 2019-05-14
Bug 1225902 - Show only flags with requestee in the "Flags You Have Requested" section

kyoshino:
Hide flags without a requestee from My Dashboard, including `in-testsuite?` and `qe-verify?`, so only important info will be displayed.

## Bugzilla link

[Bug 1225902 - Show only flags with requestee in the "Flags You Have Requested" section](https://bugzilla.mozilla.org/show_bug.cgi?id=1225902)

### DIFF
--- a/extensions/MyDashboard/lib/Queries.pm
+++ b/extensions/MyDashboard/lib/Queries.pm
@@ -14,6 +14,7 @@ use warnings;
 use Bugzilla;
 use Bugzilla::Bug;
 use Bugzilla::CGI;
+use Bugzilla::Constants;
 use Bugzilla::Search;
 use Bugzilla::Flag;
 use Bugzilla::Status qw(is_open_state);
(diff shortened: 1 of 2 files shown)


extensions/MyDashboard/lib/Queries.pm
```diff
@@ -313,6 +313,9 @@ sub query_flags {
     # Skip this flag if the bug is not visible to the user
     next if !$visible_bugs{$flag->{'bug_id'}};
 
+    # Skip this flag unless it's specifically requestable and has a requestee
+    next unless $flag->{'requestee'};
+
```
dklawren: I would instead filter out the flags without requestee in the DB instead of looping over 700+ flags here. In lines 256-261:

  if ($type eq 'requestee') {
    $match_params->{'requestee_id'} = $user->id;
  }
  else {
    $match_params->{'setter_id'} = $user->id;
    $match_params->{'requestee_id'} = NOT_NULL;
  }

kyoshino: Thanks! Didn’t know `NOT_NULL` was available. Fixing now.

PR #1275, opened 2019-05-14
Bug 1549637 - Homepage link on the global header is wrong

dylanwh:


template/en/default/search/form.html.tmpl
```diff
@@ -43,10 +43,10 @@ var tms = new Array();
 [%# Create an array of products, indexed by the classification #%]
 
 [% nclass = 0 %]
-[% FOREACH c = classification %]
+[% FOREACH classification = classification %]
```
kyoshino: Hooray TT. Looks like this change introduces another issue 😅 The Classification list will be empty on the Advanced Search page. `cls` or whatever should work.

dylanwh: damnit TT

kyoshino: _/me awaiting an update..._

dklawren:
Seems like it would have been easier just to change the mojo variable to something else?

 $params{self} = $params{controller} = $c;

dylanwh:
> Seems like it would have been easier just to change the mojo variable to something else?
> 
> $params{self} = $params{controller} = $c;

I wanted to be consistent with existing mojo codebases.

PR #1281, opened 2019-05-17
Bug 1547414 - Create Last Major Update field that excludes minor, bulk and automated changes

kyoshino:
* Add the new `major_change_ts` field to a bug that will be updated only when a comment or attachment is added, or the status is changed. 
* Changes by silent users including bots will never update the timestamp.
* The new field is searchable and exposed to the API.

## Bugzilla link

[Bug 1547414 - Create Last Major Update field that excludes minor, bulk and automated changes](https://bugzilla.mozilla.org/show_bug.cgi?id=1547414)

### DIFF
--- a/Bugzilla/Bug.pm
+++ b/Bugzilla/Bug.pm
@@ -93,6 +93,7 @@ sub DB_COLUMNS {
       everconfirmed
       filed_via
       lastdiffed
+      major_change_ts
       op_sys
       priority
       product_id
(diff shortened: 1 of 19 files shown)


kyoshino:
This is still working, ready for review 🙆‍♂

Note that this PR exposes the new field only in the search results page (when you change columns) and the API (via `/rest/bug/######?include_fields=last_change_time`). Other UI changes are coming later.

PR #1286, opened 2019-05-22
Bug 1552885 - Fix issues in the post-Sandstone skin including low contrast, visited links and small font size

kyoshino:
Fix various issues regressed by the new skin.

## Bugzilla link

[Bug 1552885 - Fix issues in the post-Sandstone skin including low contrast, visited links and small font size](https://bugzilla.mozilla.org/show_bug.cgi?id=1552885)


kyoshino:
Removed the background colour for private bugs in favour of a prominent banner like this:

image
image

kyoshino:
Wanted to make it customizable but realized it won’t work in the crash table iframe. I’ll revisit later how to solve it.

kyoshino:
I think most regressions are now covered. Would like to see this deployed soon.

extensions/BugModal/web/bug_modal.css
```diff
@@ -192,14 +192,11 @@ a.activity-ref {
   width: 100%;
 }
 
-.module .field .value input {
+.module .field .value input[type="text"],
+.module .field .value input:not([type]) {
   width: 100%;
```
dklawren: This has caused the dependency and regression fields to look bad. They are now all the way to the edge of the module boundary. This will need to be fixed.

image

dklawren: Other right side text fields such as URL look fine to me.

image

kyoshino: Looks like the problem is `.module .field .value.edit { width: 100%; }`. I'll remove it.

kyoshino:
I haven’t changed the link colour at the top of the modal bug page. It has always been like that, but the previous visited colour was so dark so it was just not so prominent. It’s technically correct, because you just visited the bug.

Honestly I don’t know why the link is there. I’d remove the link rather than overriding the colour. Will also try to make the visited link colour darker while maintaining the minimum contrast.

kyoshino:
A bit purple-ish now rather than pink.

image

kyoshino:
Eh, no, the contrast is not enough. Updating again.

kyoshino:
Checked again, it seems okay. The inspector says 5.20 AA ✅

kyoshino:
Thanks!

PR #1288, opened 2019-05-26
Bug 1545443 - Remove Splinter

kyoshino:
It has been 3 months since review flags have been disabled for Firefox and other products. It’s time to remove the deprecated Splinter extension. I didn’t add Dark Mode to Splinter, so it also [looks bad](https://bugzilla.mozilla.org/show_bug.cgi?id=1554521) now.

## Bugzilla link

[Bug 1545443 - Remove Splinter](https://bugzilla.mozilla.org/show_bug.cgi?id=1545443)


dklawren:
I think there was some discussion in the past that we would not remove until we had some sort of read-only style view for patches/diffs in place. I think Dylan was working with Red Hat maybe on this. Let me see what he says.

kyoshino:
The [legacy diff view](https://bugzilla.mozilla.org/attachment.cgi?id=8997608&action=diff) is still available, and it does support Dark Mode. We should probably add a redirect from Splinter to the corresponding diff page.

dklawren:
> 
> 
> The [legacy diff view](https://bugzilla.mozilla.org/attachment.cgi?id=8997608&action=diff) is still available, and it does support Dark Mode. We should probably add a redirect from Splinter to the corresponding diff page.

But what of the projects who are not yet (or do not plan to) use Phabricator and/or Github for code review? How much work would it really be to fix Splinter to look decent in dark mode? I am not advocating for keeping Splinter, I just think we may need to give notice and some time for people to migrate to something else for commenting on patches.

kyoshino:
Hmm, looks like Thunderbird team is still using Splinter? Then it cannot be removed now. I’ll update the stylesheet.

kyoshino:
Closing this for now.

PR #1291, opened 2019-05-27
Bug 1376609 - Make needinfo field more visible when filing a new bug

kyoshino:
Make it easier to request info on the Enter Bug page by adding a dedicated field. Currently it requires you to set the `needinfo?` flag hidden in the advanced fields, which is quite unintuitive.

image

## Bugzilla link

[Bug 1376609 - Make needinfo field more visible when filing a new bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1376609)

### DIFF
--- a/extensions/Needinfo/Extension.pm
+++ b/extensions/Needinfo/Extension.pm
@@ -69,6 +69,20 @@ sub install_before_final_checks {
   });
 }
 
+sub bug_end_of_create {
+  my ($self, $args) = @_;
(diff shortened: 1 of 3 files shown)


kyoshino:
I’d rather like to change the “Need more information from” label. Maybe “Request info from”?

dklawren:
> 
> 
> I’d rather like to change the “Need more information from” label. Maybe “Request info from”?

I would be fine with "Request more information from" and change it in both places.

PR #1302, opened 2019-05-31
Bug 1555847 - Bugs in core-security-release and pocket-security-sensitive groups have Confidential header, should be Security

kyoshino:
Show the security issue banner on bugs restricted to the `core-security-release` and `pocket-security-sensitive` groups.

## Bugzilla link

[Bug 1555847 - Bugs in core-security-release and pocket-security-sensitive groups have Confidential header, should be Security](https://bugzilla.mozilla.org/show_bug.cgi?id=1555847)

### DIFF
--- a/template/en/default/global/header.html.tmpl
+++ b/template/en/default/global/header.html.tmpl
@@ -417,7 +417,7 @@
   IF bug && bug.groups_in.size;
     is_security_bug = 0;
     FOREACH group IN bug.groups_in;
-      IF group.name.match('^(.*security|\w+sec)$'); is_security_bug = 1; END;
+      IF group.name.match('^(.*security.*|\w+sec)$'); is_security_bug = 1; END;
     END;
 %]
   <div id="private-bug-banner" class="[% IF is_security_bug %]security[% ELSE %]confidential[% END %]">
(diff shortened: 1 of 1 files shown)


dylanwh:
`group.secure_mail`

kyoshino:
But that’s specific to Mozilla and anyone else with the SecureMail extension enabled, right? The banner is in the global header template, and it should work with upstream users as well.

dylanwh:
let's consider the securemail thing to be core. I would rather make that core than depend on regex matches of strings.

kyoshino:
Hmm, okay, but is there any chance of Secure Mail being used for other groups? Maybe not at BMO at this time but I’m thinking about other instances too, given that it will be shipped with Bugzilla 6 either as an extension or as part of the core. Ideally, any confidential stuff should not be sent via plaintext email, right? A simple regex might work better here.

dylanwh:
Confidential vs. security is pretty much the same as encrypted vs. not encrypted in email. It is actually something I'd like to be more visible.

dklawren:
I vote we trigger off of secure_mail for now as that is a good use case for this. If/when we do find that we need groups that need to be designated secure but not encrypted we can revisit this. But I agree that the new private bug banners used to let people know that emails from the particular bug will be encrypted since right now users do not really know.

kyoshino:
Okay, let’s use `group.secure_mail` for now.

kyoshino:
Forgot to say: tested the updated PR locally and it looked good to me.

kyoshino:
I saw a [test failure](https://circleci.com/gh/mozilla-bteam/bmo/24195) but running again solved the issue. Was probably a race condition.

PR #1310, opened 2019-06-05
Bug 1555409 - D32774 not moved out of Draft state

dklawren:
The new change will not update status if the revision is a draft and the status is 'changes-planned'. This happens when someone submits a review using 'moz-phab submit --wip'. Otherwise it should still do the old behavior of moving to 'needs-review'.

Bugzilla Link:
[Bug 1555409 - D32774 not moved out of Draft state](https://bugzilla.mozilla.org/show_bug.cgi?id=1555409)

### DIFF
--- a/extensions/PhabBugz/lib/Feed.pm
+++ b/extensions/PhabBugz/lib/Feed.pm
@@ -508,8 +508,12 @@ sub process_revision_change {
     $attachment->update($timestamp);
   }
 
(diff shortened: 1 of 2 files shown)


dklawren:
> 
> 
> I haven't tested this, but it sounds like you did. Is that ok?

Yes. I tested it locally. If we get this change on bugzilla-dev, I can test there as well.

PR #1315, opened 2019-06-06
Bug 1557360 - Crash Signature textbox is too narrow

kyoshino:
Fix the size of the Crash Signature `<textarea>` plus the position of the Update status flags button in the Crash Data section.

image

## Bugzilla link

[Bug 1557360 - Crash Signature textbox is too narrow](https://bugzilla.mozilla.org/show_bug.cgi?id=1557360)


kyoshino:
Actually this seems too tall. Will fix.

kyoshino:
image

PR #1317, opened 2019-06-06
Bug 1533587 - Rename CC to Subscribers

kyoshino:
Solving a long-standing UX debt: Change the label of the `cc` field from CC (List) to Subscribers. Verbs are also changed accordingly: 

* Add me to CC list / Follow -> Subscribe
* Remove me from the CC list / Unfollow -> Unsubscribe

This PR only changes UI labels and code comments, so internal code and API field continue using `cc`.

## Bugzilla link

[Bug 1533587 - Rename CC to Subscribers](https://bugzilla.mozilla.org/show_bug.cgi?id=1533587)


dklawren:
Actually I accidentally hit the approve button. After talking this over at Allhands, we decided that this does not need to be done at this time and we should hold off on this change.

kyoshino:
I think waiting is not the answer. Sure we are still heavily relying on email notifications, CC is a [jargon](https://bugzilla.mozilla.org/show_bug.cgi?id=611933) and also technically wrong. There’s no reason to make this change now.

dklawren:
This is a change which I feel is not needed and we have other outstanding issues that should be dealt with first. If this is not coming from management or other high level stake holders, we can put this one off. I have not heard of anyone complaining about what Cc or Follow means or claim that it has caused confusion. Closing.

kyoshino:
This is an UX issue that must be solved. Long time users don’t complain because they are familiar with the name from the beginning. New users, however, will be confused because they don’t understand what it means. Why asking management?

dklawren:
> 
> 
> This is an UX issue that must be solved. Long time users don’t complain because they are familiar with the name from the beginning. New users, however, will be confused because they don’t understand what it means. Why asking management?

Is there any user research or data to support the need for this change to be made? Are you pushing for this because other sites use different terminology? We discussed this at Whistler and decided that it wasn't enough impact for this much of a change to be made. Bugzilla has always used the term Cc and I do not recall anyone every making a complaint either in mailing lists or in bug reports. 

I was not meaning to ask management anything really. I was just curious if management had been the one originally requesting this change to be made.

PR #1321, opened 2019-06-07
No bug: update search documentation

emceeaich:
* where saved searches go
* remove self-tagging references


kyoshino:
And this PR needs to be force-rebased so the build error will be gone.

PR #1320, opened 2019-06-07
Bug 1557186 Document relative dates in searches

emceeaich:
Adds documentation for relative searches

PR #1325, opened 2019-06-12
Bug 1557799 - Search By Change History: date range is ignored when field is not specified

kyoshino:
* <del>Select **[Bug creation]** by default because people sometimes forget to select a field name and unexpected results will be returned. It will be simply ignored unless a date range is specified</del>
* Mark `<select name="chfield">` required when a date range is specified
* Remove inline script from the section (and eventually from the entire Advanced Search page)
* Add a link to the relative dates doc being added in #1320

## Bugzilla link

[Bug 1557799 - Search By Change History: date range is ignored when field is not specified](https://bugzilla.mozilla.org/show_bug.cgi?id=1557799)

### DIFF
--- a/js/advanced-search.js
+++ b/js/advanced-search.js
@@ -27,8 +27,8 @@ Bugzilla.AdvancedSearch.HistoryFilter = class HistoryFilter {
     this.$chfieldto = document.querySelector('#chfieldto');
     this.$chfieldto_button = document.querySelector('#chfieldto + button');
 
-    this.$chfieldfrom.addEventListener('input', event => this.on_date_change(event.target));
-    this.$chfieldto.addEventListener('input', event => this.on_date_change(event.target));
+    this.$chfieldfrom.addEventListener('input', event => this.on_date_change(event));
+    this.$chfieldto.addEventListener('input', event => this.on_date_change(event));
 
     // Use on-event handler because `field.js` will update it
     this.$chfieldfrom_button.onclick = () => showCalendar('chfieldfrom');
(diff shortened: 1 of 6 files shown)


kyoshino:
`extensions/BMO/web/js/advanced-search.js` shouldn’t have had the `Bugzilla` namespace, because it’s part of an extension and might conflict with a script in the core. The file actually conflicts here, so I’ve changed the namespace. It can be `Bugzilla.extensions.BMO` but `BMO` should be okay.

kyoshino:
Moved things under the `Bugzilla.BMO` namespace, just like `Bugzilla.BugModal` and `Bugzilla.Review`.

js/advanced-search.js
```diff
@@ -11,6 +11,48 @@
...
+   */
+  constructor() {
+    this.$chfield = document.querySelector('#chfield');
+    this.$chfieldfrom = document.querySelector('#chfieldfrom');
+    this.$chfieldfrom_button = document.querySelector('#chfieldfrom + button');
+    this.$chfieldto = document.querySelector('#chfieldto');
+    this.$chfieldto_button = document.querySelector('#chfieldto + button');
+
+    this.$chfieldfrom.addEventListener('input', event => this.on_date_change(event.target));
+    this.$chfieldto.addEventListener('input', event => this.on_date_change(event.target));
```
dklawren: s/event.target/event/

kyoshino: Aha, why I didn’t notice this… Fixing now.

PR #1331, opened 2019-06-13
Bug 1559248 - Integrate iprepd with BMO

dylanwh:
# NEW localconfig / environment configuration

- BMO_iprepd_url -- the url to the iprepd instance
- BMO_iprepd_key -- the api key

# NEW admin / runtime configuration

These settings are on the admin panel, and the bmo admins can tweak them at runtime without a deployment.

- iprepd_min_reputation -- the min level of reputation needed for an IP to not be blocked.
- iprepd_active -- a boolean to turn off iprepd if it is misbehaving.
- iprepd_email -- an email address, displayed on the block page. This could be the only communication channel open to the person that was blocked, and ideally it would go to both bmo ops and iprepd ops

### DIFF
--- a/Bugzilla/App/Plugin/BlockIP.pm
+++ b/Bugzilla/App/Plugin/BlockIP.pm
@@ -12,15 +12,16 @@ use Type::Library -base, -declare => qw( ResponseType);
 use Type::Utils -all;
 use Types::Standard -all;
 
-declare ResponseType, as Dict [
+declare ResponseType,
+  as Dict [
   object      => Str,
   type        => Str,
   reputation  => Int,
   reviewed    => JSONBool,
   lastupdated => Str,
-  decayafter  => Optional[Str],
+  decayafter  => Optional [Str],
   slurpy Any,
-];
+  ];
 
 use constant BLOCK_TIMEOUT => 60 * 60;
 
@@ -34,7 +35,7 @@ sub register {
   $app->hook(before_routes => \&_before_routes);
   $app->helper(block_ip   => \&_block_ip);
   $app->helper(unblock_ip => \&_unblock_ip);
-  $app->helper(check_ip => \&_check_ip);
+  $app->helper(check_ip   => \&_check_ip);
 
   $app->hook(
     before_server_start => sub {
@@ -42,8 +43,7 @@ sub register {
       $template->process('global/ip-blocked.html.tmpl',
         {block_timeout => BLOCK_TIMEOUT, block_type => 'internal'},
         \$BLOCKED_INTERNAL_HTML);
-      $template->process('global/ip-blocked.html.tmpl',
-        {block_type => 'external'},
+      $template->process('global/ip-blocked.html.tmpl', {block_type => 'external'},
         \$BLOCKED_EXTERNAL_HTML);
       undef $template;
       utf8::encode($BLOCKED_INTERNAL_HTML);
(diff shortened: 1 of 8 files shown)


ckolos:
I'm a little confused about this code; I'm guessing this is in response to our request to have BMO in a state where iprepd could be put in-front of BMO, however, we've not done that yet. Have there been discussions with the iprepd-powers-that-be that A) I'm unaware of and B) guarantee that this code works as expected?

Also, will BMO continue to work as it does today with these changes in place?

dylanwh:
> I'm a little confused about this code; I'm guessing this is in response to our request to have BMO in a state where iprepd could be put in-front of BMO, however, we've not done that yet. Have there been discussions with the iprepd-powers-that-be that A) I'm unaware of and B) guarantee that this code works as expected?
> 
> Also, will BMO continue to work as it does today with these changes in place?

This was the plan discussed previously, at length, and recently with alm both in slack and on the bug that this was cloned from.

In short, everything that works now continues to work.
The following constraints are held:

1) The bugzilla admins (which will be: glob, dkl, emma, kmoir, and tweir upon my departure) can disable the integration entirely
2) The admins can change the threshold at which someone is banned
3) The admins can change the email address contact given in the blocking message.
4) This will never block an IP in bugzilla's internal whitelist.

dylanwh:
Note that only ops can configure (and see) the API key (and the address of the iprepd API), so if both of those are not set, the admin-level settings will do nothing. This is in keeping with standard security pattern where security tokens are configured by the environment and outside of the application config (same as, say, db passwords or memcached)

ckolos:
> This was the plan discussed previously, at length, and recently with alm both in slack and on the bug that this was cloned from.

That's correct, however ops was unaware that this change was planned for "now" and I (representing ops) was unaware that those discussions had taken place with alm in slack. Notably missing from the participants of that discussion is anyone from our team.

> 
> In short, everything that works now continues to work.
> The following constraints are held:
> 
>     1. The bugzilla admins (which will be: glob, dkl, emma, kmoir, and tweir upon my departure) can disable the integration entirely
> 
>     2. The admins can change the threshold at which someone is banned
> 
>     3. The admins can change the email address contact given in the blocking message.
> 
>     4. This will never block an IP in bugzilla's internal whitelist.

Cool. Works for us.

dylanwh:
> > This was the plan discussed previously, at length, and recently with alm both in slack and on the bug that this was cloned from.
> 
> That's correct, however ops was unaware that this change was planned for "now" and I (representing ops) was unaware that those discussions had taken place with alm in slack. Notably missing from the participants of that discussion is anyone from our team.

There is no immediate rush on our part. The plan is same as from https://bugzilla.mozilla.org/show_bug.cgi?id=1521784#c3


> Cool. Works for us.

Cool. I tagged glob and you in the related history of the ops meeting and y'all can chat at the all hands.

Bugzilla/App/Plugin/BlockIP.pm
```diff
@@ -38,18 +67,79 @@ sub _unblock_ip {
...
+
+    return '' unless $params->{iprepd_active};
+
+    my $filter = Bugzilla::Bloomfilter->lookup("rate_limit_whitelist");
+    if ($filter->test($ip)) {
+      INFO("$ip is in the rate limit whitelist, ignoring iprepd.");
+      return '';
+    }
+
+    my $tx   = $c->ua->get($url => {Authorization => "APIKey $iprepd_key"});
```
ameihm0912: I'm not sure if it is possible with the library being used, but one feature that might be useful is a configuration option specifying a maximum end-to-end request timeout for iprepd queries; if the request exceeds this it would just pass. This would prevent any sort of service impact associated with excessive latency.

dylanwh: The mojo framework has defaults for this that are reasonable. Timeouts will behave as pseudo-http responses and be logged by as errors.

dylanwh: (There are two separate limits, connection timeout and inactivity timeout)

Additionally the API is non-blocking so only the current request is impacted and the total number of concurrent requests is unaffected)

PR #1332, opened 2019-06-13
Bug 1317224 - Option 'Ignore Bug Mail' is hidden by default on modal UI

kyoshino:
Show the “Never email me about this bug” checkbox on the modal UI not only in the edit mode but also in the view mode. It was visible by default on the legacy UI, so it’s considered as a regression.

## Bugzilla link

[Bug 1317224 - Option 'Ignore Bug Mail' is hidden by default on modal UI](https://bugzilla.mozilla.org/show_bug.cgi?id=1317224)

### DIFF
--- a/extensions/BugModal/template/en/default/bug_modal/new_comment.html.tmpl
+++ b/extensions/BugModal/template/en/default/bug_modal/new_comment.html.tmpl
@@ -103,7 +103,7 @@
       </tr>
     </table>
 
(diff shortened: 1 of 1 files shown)


dylanwh:
Does this work though? Can I check the checkbox but still have to click edit to submit the change?

kyoshino:
The Save Changes button below the checkbox is always available, so it works without having to enter the edit mode 🙂 

image

PR #1333, opened 2019-06-14
Bug 1559039 - Suggest that a user uses an attachment when a comment is very long

kyoshino:
When large text data like `about:support` or network log is pasted into the comment box, provide an option to upload it as an attachment.

## Bugzilla link

[Bug 1559039 - Suggest that a user uses an attachment when a comment is very long](https://bugzilla.mozilla.org/show_bug.cgi?id=1559039)

### DIFF
--- a/extensions/BugModal/web/bug_modal.js
+++ b/extensions/BugModal/web/bug_modal.js
@@ -1394,7 +1394,7 @@ $(function() {
       const text = event.clipboardData.getData('text');
       const lines = text.split(/(?:\r\n|\r|\n)/).length;
 
-      if (lines < 20) {
+      if (lines < 50) {
         return;
       }
 
--- a/extensions/BugModal/web/bug_modal.js
+++ b/extensions/BugModal/web/bug_modal.js
@@ -1399,8 +1399,10 @@ $(function() {
       }
 
       const extract = text.replace(/(?:\r\n|\r|\n)/, ' ').trim().substr(0, 20);
-      const summary = (window.prompt(`You’re pasting ${lines} lines of text starting with “${extract}”. Enter the ` +
-        'summary below and click OK to upload it as an attachment. Click Cancel to paste it normally.') || '').trim();
+      const summary = (window.prompt(
+        `You’re pasting ${lines} lines of text starting with “${extract}”. ` +
+        'Enter the summary below and click OK to upload it as an attachment. Click Cancel to paste it normally.'
+      ) || '').trim();
 
       if (!summary) {
         return;
--- a/extensions/BugModal/web/bug_modal.js
+++ b/extensions/BugModal/web/bug_modal.js
@@ -1389,6 +1389,44 @@ $(function() {
             saveBugComment(event.target.value);
         });
 
+    // Allow to attach pasted text directly
+    document.querySelector('#comment').addEventListener('paste', event => {
+      const text = event.clipboardData.getData('text');
+      const lines = text.split(/(?:\r\n|\r|\n)/).length;
+
+      if (lines < 20) {
+        return;
+      }
+
+      const extract = text.replace(/(?:\r\n|\r|\n)/, ' ').trim().substr(0, 20);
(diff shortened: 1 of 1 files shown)


extensions/BugModal/web/bug_modal.js
```diff
@@ -1389,6 +1389,44 @@ $(function() {
             saveBugComment(event.target.value);
         });
 
+    // Allow to attach pasted text directly
+    document.querySelector('#comment').addEventListener('paste', event => {
+      const text = event.clipboardData.getData('text');
+      const lines = text.split(/(?:\r\n|\r|\n)/).length;
+
+      if (lines < 20) {
```
mnoorenberghe: Thanks a lot for working on this so quickly :) I do wonder if we should start more conservative since it's not uncommon for comment quotes to end up longer than 20 lines and I often cut and paste reply messages to re-order them if I am replying to multiple comments in one comment of my own. e.g. 
1. Click Reply on Comment 20 and maybe start typing a reply
2. Realize that you want to quote/reply to Comment 19 too
3. Select All in the textarea
4. Click Reply to comment 19
5. Paste the draft text you cut after the comment 19 quote that was inserted

As a data point, pasting the reply from comment 0 of bug 1556953 is already 18 lines with your method of counting and that bug didn't have a long STR. I think the number should be at least 50, maybe 75 to start OR ideally we should treat clipboard contents containing BMO replies specially e.g. looking for "In reply to" in the first line of text being pasted but I don't know if BMO supports multiple languages making that potentially more complicated (though uses the user's own translation for that string would handle when they cut + paste from their own textarea).

mnoorenberghe: Btw. one of the comments that caused me to file this bug was https://bugzilla.mozilla.org/show_bug.cgi?id=1555317#c12 which is 386 lines with your approach.

kyoshino: Okay, let’s bump up the number to 50 🙂

kyoshino: The pasted text won’t be uploaded unless the user explicitly enters the summary (blank + Enter/OK is the same as Cancel) so we can prevent unwanted text from being uploaded as an attachment.

extensions/BugModal/web/bug_modal.js
```diff
@@ -1389,6 +1389,44 @@ $(function() {
...
+      const text = event.clipboardData.getData('text');
+      const lines = text.split(/(?:\r\n|\r|\n)/).length;
+
+      if (lines < 50) {
+        return;
+      }
+
+      const extract = text.replace(/(?:\r\n|\r|\n)/, ' ').trim().substr(0, 20);
+      const summary = (window.prompt(`You’re pasting ${lines} lines of text starting with “${extract}”. Enter the ` +
+        'summary below and click OK to upload it as an attachment. Click Cancel to paste it normally.') || '').trim();
```
dklawren: Nit: insert blank space above "Enter the summary below..." It is less confusing when it is more obvious what it is expected of the user to do with the empty text field.

kyoshino: Fixed!

extensions/BugModal/web/bug_modal.js
```diff
@@ -1399,8 +1399,10 @@ $(function() {
       }
 
       const extract = text.replace(/(?:\r\n|\r|\n)/, ' ').trim().substr(0, 20);
-      const summary = (window.prompt(`You’re pasting ${lines} lines of text starting with “${extract}”. Enter the ` +
-        'summary below and click OK to upload it as an attachment. Click Cancel to paste it normally.') || '').trim();
+      const summary = (window.prompt(
+        `You’re pasting ${lines} lines of text starting with “${extract}”. ` +
```
dklawren: Sorry forgot to mention earlier. HTML encode ${extract} and this should be good to go. r=dkl

kyoshino: This is a prompt message, not HTML, so we don't need `htmlEncode()` here 🙂

PR #1337, opened 2019-06-18
Bug 1512497 - Comment editing should be listed on the bug history page and API

kyoshino:
Add comment revisions to the bug history page and API. Also add comment tag changes to the bug history API.

## Bugzilla link

[Bug 1512497 - Comment editing should be listed on the bug history page and API](https://bugzilla.mozilla.org/show_bug.cgi?id=1512497)

### DIFF
--- a/extensions/EditComments/Extension.pm
+++ b/extensions/EditComments/Extension.pm
@@ -63,6 +63,13 @@ sub install_update_db {
   # because existing admin-edited revisions may contain sensitive info
   $dbh->bz_add_column('longdescs_activity', 'is_hidden',
     {TYPE => 'BOOLEAN', NOTNULL => 1, DEFAULT => 1});
+
+  unless (Bugzilla::Field->new({name => 'comment_revision'})) {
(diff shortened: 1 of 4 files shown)


extensions/EditComments/Extension.pm
```diff
@@ -63,6 +63,13 @@ sub install_update_db {
   # because existing admin-edited revisions may contain sensitive info
   $dbh->bz_add_column('longdescs_activity', 'is_hidden',
     {TYPE => 'BOOLEAN', NOTNULL => 1, DEFAULT => 1});
+
```
dylanwh: Does this impact search? I can't remember what happens for fielddefs that aren't also defined in the search code.

kyoshino: Just checked... Comment Revisions appears in the Custom Search dropdown, which isn’t necessary. I’ll remove this field definition. It still works; it will just say “Comment X” instead of “Comment X Revisions”.

PR #1340, opened 2019-06-19
Bug 1477931 follow-up - Make sure Gravatar and requests are loaded via API

kyoshino:
Let’s try this again once #1212 is merged. This will finish #684 by showing the user’s avatar as well as # of requests or blocking status.

## Bugzilla link

[Bug 1477931 - Show number of review/feedback/needinfo in user autocomplete and prevent person from being added if requests are blocked](https://bugzilla.mozilla.org/show_bug.cgi?id=1477931)


dylanwh:
Tested with the other patch. It's still slower but I think with (well within) the margin of error)

PR #1342, opened 2019-06-20
Bug 1560245 - Specify missing bug field types

kyoshino:
Add the type property to the core fields so the UI can be enriched.

## Bugzilla link

[Bug 1560245 - Specify missing bug field types](https://bugzilla.mozilla.org/show_bug.cgi?id=1560245)


kyoshino:
Closing in favour of #1345.

PR #1343, opened 2019-06-20
Bug 1482447 - Add nofollow to all external links

dylanwh:


dklawren:
test case still not full resolved. taking a look.

PR #1349, opened 2019-06-25
Bug 1472316 - Decommission new-bug page

kyoshino:
Remove the unmaintained `new-bug` page and redirect to `enter_bug.cgi`. The API methods implemented in the BugModal extension are kept for possible later use.

## Bugzilla link

[Bug 1472316 - Decommission new-bug page](https://bugzilla.mozilla.org/show_bug.cgi?id=1472316)


Bugzilla/App/CGI.pm
```diff
@@ -43,7 +43,12 @@ sub setup_routes {
   $r->get('/robots.txt')->to('CGI#robots_cgi');
   $r->any('/login')->to('CGI#index_cgi' => {'GoAheadAndLogIn' => '1'});
   $r->any('/logout')->to('CGI#index_cgi' => {'logout' => '1'});
-  $r->any('/:new_bug' => [new_bug => qr{new[-_]bug}])->to('CGI#new_bug_cgi');
+
+  $r->any('/:new_bug' => [new_bug => qr{new[-_]bug}] => sub {
+    my $c = shift;
+    $c->res->code(301);
+    $c->redirect_to('enter_bugcgi');
+  });
```
dklawren: How about instead:

  $r->any('/:REWRITE_new_bug' => [REWRITE_new_bug => qr{new[-_]bug}])
    ->to('CGI#enter_bug_cgi');

For some reason without the change I was unable to get /new-bug to work new_bug did work.

kyoshino: But that won’t redirect the `new-bug`/`new_bug` to `enter_bug.cgi` 🤔

kyoshino: `redirect_to('CGI#enter_bug_cgi')` didn’t work for me, btw.

dklawren: > 
> 
> But that won’t redirect `new-bug`/`new_bug` to `enter_bug.cgi` 🤔

My suggestion would not do a redirect at all and just execute the CGI#enter_bug_cgi code directly.

For example, see https://github.com/mozilla-bteam/bmo/blob/master/extensions/BMO/Extension.pm#L2820 

But thinking, we should probably leave a redirect in place so the client will know that the new_bug path is no longer valid and is moved permanently. So let's go with:

redirect_to('CGI#enter_bug.cgi')

kyoshino: `enter_bugcgi` works because it’s a route name, but I understand it may be confusing. I’d go ahead with `Bugzilla->localconfig->basepath . 'enter_bug.cgi'`.

PR #1351, opened 2019-06-27
no bug - Do not create qa data if not running qa tests

dklawren:
Needed to not run script to create legacy QA data when running from conduit-suite or other modes. It should only run when running the webservices and selenium tests in CircleCI.


kyoshino:
Hmm why this is failing? The fix in #1322 isn’t working maybe? 🤔

kyoshino:
@dklawren tests are still failing.

PR #1372, opened 2019-07-17
Bug 1565636 - Component description for "Firefox for Android :: Build Config & IDE Support" is out of date

dklawren:
Special one-off script that moves a component while reassigning version and milestones based on a hard coded mapping. If an entry on the map doesn't exist, it just leaves the version and milestone alone as they already exist in the destination product.


dklawren:
> 
> 
> Do we need to do anything with security groups on this, or do the products share the same groups?

Only a handful of bugs are secure out of ~750 so I can manually enabled the missing groups on the firefox build system product.

dklawren:
working on fixes for the sanity test failures

PR #1377, opened 2019-07-19
Bug 1565403 - Log how a bug was filed (via standard, guided, custom bug form or API)

kyoshino:
Add a new field called `filed_via` to each bug that records the bug’s file method: `standard_form`, `guided_form `, `custom_form ` or `api`. All the existing bugs will use `unknown`. It won’t be displayed on the bug, but can be retrieved via the API (`/rest/bug/####?include_fields=filed_via`) as well as bug lists.

## Bugzilla link

[Bug 1565403 - Log how a bug was filed (via standard, guided, custom bug form or API)](https://bugzilla.mozilla.org/show_bug.cgi?id=1565403)

### DIFF
--- a/Bugzilla/Bug.pm
+++ b/Bugzilla/Bug.pm
@@ -91,6 +91,7 @@ sub DB_COLUMNS {
       delta_ts
       estimated_time
       everconfirmed
+      filed_via
       lastdiffed
       op_sys
       priority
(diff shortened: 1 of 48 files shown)


kyoshino:
I’ve changed `advanced_form` to `standard_form` because custom forms call it the standard form. The guided form does call it the advanced form, but it’s inconsistent and confusing because the form includes advanced fields. So it’s better to call it the standard form everywhere.

PR #1379, opened 2019-07-20
Bug 1562364 - Enforce setting bug type when creating bug through API

kyoshino:
Add an admin param for requiring the bug type field. Return an error if it’s required but not specified when the user files a new bug through the API (and UI, though it already makes the field required with HTML/JavaScript).

## Bugzilla link

[Bug 1562364 - Enforce setting bug type when creating bug through API](https://bugzilla.mozilla.org/show_bug.cgi?id=1562364).


dklawren:
Note: We will need to manually set the mandatory bit off on the bug_type field once this goes to production.

PR #1390, opened 2019-07-26
Bug 1569194 - My Dashboard shows Unexpected Error dialog when request is aborted while loading

kyoshino:
Show a simple error message inline instead of using `alert()`. Note: `fetch()` throws a generic `TypeError` instead of an `AbortError` when the user navigates manually, so it’s difficult to know if the request is aborted.

## Bugzilla link

[Bug 1569194 - My Dashboard shows Unexpected Error dialog when request is aborted while loading](https://bugzilla.mozilla.org/show_bug.cgi?id=1569194)

### DIFF
--- a/extensions/MyDashboard/web/js/flags.js
+++ b/extensions/MyDashboard/web/js/flags.js
@@ -43,7 +43,7 @@ $(function () {
                 Y.one(`#${type}_loading`).addClass('bz_default_hidden');
                 Y.one(`#${type}_count_refresh`).removeClass('bz_default_hidden');
 
-                alert(`Failed to load requests:\n\n${message}`);
+                dataTable[type].showMessage(`Failed to load requests.`);
             }
         };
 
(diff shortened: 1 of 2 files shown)


extensions/MyDashboard/web/js/flags.js
```diff
@@ -43,7 +43,7 @@ $(function () {
                 Y.one(`#${type}_loading`).addClass('bz_default_hidden');
                 Y.one(`#${type}_count_refresh`).removeClass('bz_default_hidden');
 
-                alert(`Failed to load requests:\n\n${message}`);
+                dataTable[type].showMessage(`Failed to load requests.`);
```
dklawren: Should we log the actual error message to console?

kyoshino: I already have [`console.log()`](https://github.com/mozilla-bteam/bmo/blob/master/js/util.js#L642) 🙂

PR #1394, opened 2019-07-31
Bug 1512484 - Allow to specify default version per product

kyoshino:
Add `default_version` to products, and make it editable by admin. The implementation is almost the same as Default Milestone. Also make the field optional when filing a bug (especially via the API.)

## Bugzilla link

[Bug 1512484 - Allow to specify default version per product](https://bugzilla.mozilla.org/show_bug.cgi?id=1512484)

### DIFF
--- a/Bugzilla/WebService/Constants.pm
+++ b/Bugzilla/WebService/Constants.pm
@@ -209,8 +209,10 @@ use constant WS_ERROR_CODE => {
   product_name_already_in_use          => 702,
   product_name_diff_in_case            => 702,
   product_must_have_description        => 703,
-  product_must_have_version            => 704,
+  # Removed in favour of default value 'unspecified':
+  # product_must_have_version          => 704,
   product_must_define_defaultmilestone => 705,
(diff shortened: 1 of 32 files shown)


kyoshino:
Note to myself: this requires a separate schema-only PR.

Bugzilla/WebService/Constants.pm
```diff
@@ -209,8 +209,10 @@ use constant WS_ERROR_CODE => {
   product_name_already_in_use          => 702,
   product_name_diff_in_case            => 702,
   product_must_have_description        => 703,
-  product_must_have_version            => 704,
+  # Removed in favour of default value 'unspecified':
```
dklawren: Just remove the two lines

kyoshino:
Thanks @dklawren 😺 Will solve the conflict and you feedback, and send a schema-only PR. Once this is shipped, the default version for the BMO product can be Production instead of Stage, which actually annoys us a bit.

kyoshino:
Solved the conflict in `DB.pm`.

PR #1400, opened 2019-08-01
no bug - Remove jQuery visibility plugin

kyoshino:
It can be replaced with vanilla JS. I’m adding the perf label because we are not using HTTP/2 nor concatenation so less files = better.


emceeaich:
We need a "are we vanilia js" dashboard as you remove all the jquery and yui cruft.

kyoshino:
Yeah, the meta bug is [1493992](https://bugzilla.mozilla.org/show_bug.cgi?id=1493992) but we have a long way to go.

PR #1411, opened 2019-08-12
no bug - Add bug_type field to custom Shield Studies form

kyoshino:
I’ve checked custom forms if they have bug type specified. Other than #1410, only this Shield Studies form doesn’t have it.


kyoshino:
Since @dklawren is on PTO, can @purelogiq take a look? We have to fix this and #1410 quickly, because the bug type is now a required field, and these bugs prevent some people from filing a bug through the custom forms.

PR #1415, opened 2019-08-19
Bug 1574792 - Bugzilla has suffered an internal error:  Bugzilla cannot log you into an external site via GitHub.

dklawren:
- Remove request_uri from Bugzilla/CGI.pm as GitHub login was only thing using.
- Change GitHub target_uri to always redirect back to index.cgi
- Remove login_uri method from GitHub extension as no code was using it.
- Internal error occurring due to no [% urlbase %] in the login templates. github.cgi looks for the urlbase before sending off to GitHub for auth.

### DIFF
--- a/Bugzilla/CGI.pm
+++ b/Bugzilla/CGI.pm
@@ -123,21 +123,6 @@ sub new {
   return $self;
 }
 
-sub target_uri {
-  my ($self) = @_;
-
-  my $base = Bugzilla->localconfig->urlbase;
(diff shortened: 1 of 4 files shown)


purelogiq:
Taking a look now

dklawren:
> 
> 
> This will no longer redirect the user back to the page they were looking at after they login, instead opting to take them to index.cgi. If this is intended then LGTM

Yes this is intended for now.

PR #1427, opened 2019-08-28
Bug 1575881 - Bug open/updated/closed date, reporter and triage owner should be visible in Edit mode

kyoshino:
Show the Opened, Updated and Closed dates on the modal UI header as well as the read-only Reporter and Triage Owner fields even in the Edit mode, because some people are using “Always Enable Edit Mode” to skip the View mode.

## Bugzilla link

[Bug 1575881 - Bug open/updated/closed date, reporter and triage owner should be visible in Edit mode](https://bugzilla.mozilla.org/show_bug.cgi?id=1575881)


kyoshino:
Just rebased against master to solve a conflict with #1426.

PR #1432, opened 2019-08-29
Bug 1576667 - Add creation timestamp to the user_api_keys table to show when an api key was first generated (Schema Only)

dklawren:


purelogiq:
@dklawren red build. also can you update the PR title to specify what's changing in the schema?

dklawren:
> 
> 
> @dklawren red build. also can you update the PR title to specify what's changing in the schema?

Done

PR #1447, opened 2019-10-01
Bug 1583849 - WIP Phabricator revisions not moved out of Draft state

dklawren:


extensions/PhabBugz/lib/Feed.pm
```diff
@@ -519,13 +519,9 @@ sub process_revision_change {
...
-  # Set status to request-review if revision is new and
-  # in draft state and not changes-planned
-  if ($is_new
-      && $revision->status ne 'changes-planned'
-      && ($revision->is_draft && !$revision->hold_as_draft))
-  {
-    INFO("Moving from draft to needs-review");
+  # Removing draft status
+  if ($is_new && $revision->status eq 'draft') {
+    INFO("Removing draft status");
```
globau: if the revision is currently `draft`+`changes-planned`, this will change it to `request-review`, which isn't what we want.

globau (changes requested):

extensions/PhabBugz/lib/Feed.pm
dklawren: Ok after more investigation I still feel like this change is the right way to go. 

1. When a revision is created the status is 'draft'.
2. When moz-phab creates a WIP revision, the status is set to 'changes-planned'.
3. When a revision is created with 'changes-planned' without ever being in the 'request-review' state, it an additional property on the revision, separate from the revision status, called '{"draft.broadcast":false}'.  This means that no email notifications will be send as long as that is false.
4. This is why the 'Draft' label is still displayed on a revision that has a status of 'changes-planned' for D1355 for example.
5. The way to get set the 'draft.broadcast' value to true, is to request review on the revision (set status to 'request-review').
6. Normally moz-phab will do this automatically with --wip is not provided.
7. The BMO change on this PR will therefore only update the status if the current status is 'draft'.  If moz-phab sets the status to 'changes-planned' using --wip, BMO feed daemon will just leave it as is and it will be up to the user or moz-phab to transistion it to 'request-review' status.

globau: The problem with this change is it doesn't address the bug in question.

As we've changed the meaning of `Draft` from Phabricator's default of work-in-progress to not-yet-processed-by-bugzilla we need to ensure the draft badge is never displayed once phabbugz has processed a revision.

Changing the revision's status to `request-review` it isn't a reflection of the user's intent that a WIP revision should be in a `changes-planned` state.  I understand that right now that's the only way to clear the `draft.broadcast` property, but it's bound to be more confusing than leaving the draft badge visible.

I suspect the right thing to do is to remove the code in`DifferentialRevisionViewController.php` which displays the draft badge/tag when shouldBroadcast is false.

globau (changes requested):

PR #1453, opened 2019-10-10
Bug 1587849 - Cannot create markdown comments via the REST API

dklawren:


dklawren:
> 
> 
> As far as I read the [bug comment](https://bugzilla.mozilla.org/show_bug.cgi?id=1587849#c0) and [upstream doc](https://bugzilla.readthedocs.io/en/latest/api/core/v1/comment.html#create-comments), looks like the current behaviour (disable Markdown by default) should remain unchanged. `comment.rst` says
> 
> > All comments created via the API will be considered Markdown (specifically GitHub Flavored Markdown).
> 
> but this sentence is not true and has to be removed.

Ah missed that small detail. Thanks. I will update the patch to default to off but still allow it to be specified which the client wanted.

PR #1466, opened 2019-11-11
Bug 1594844 - Automatically set the bug's assignee when they upload a revision to an unassigned bug

globau:


extensions/PhabBugz/lib/Util.pm
```diff
@@ -115,6 +117,10 @@ sub get_bug_role_phids {
   return [map { $_->phid } @{$phab_users}];
 }
 
+sub is_bug_assigned {
+  return $_->assigned_to->email ne 'nobody@mozilla.org';
```
dklawren: return $_[0]->assigned_to->email ne 'nobody@mozilla.org';

Once I added the [0], everything worked fine. r=dkl

PR #1470, opened 2019-11-18
Bug 1597235 - Setting the mime type of a pasted attachment doesn't work.

kyoshino:
Use a user-defined MIME type for attached text instead of assuming `text/plain`.

## Bugzilla link

[Bug 1597235 - Setting the mime type of a pasted attachment doesn't work.](https://bugzilla.mozilla.org/show_bug.cgi?id=1597235)


emilio:
Thanks @kyoshino!

PR #1471, opened 2019-11-20
Bug 1588221 - Update current s3 storage controller to only upload attachments over a certain size

dklawren:
* Major refactor of the storage classes to make them more extensible for the changes I need to add in the future. Such as storing the S3 JWT data in attachment_storage_class.extra_data.
* Added a check in the S3 storage support for reverting to Database if the attachment is below a set size for performance.
* The use of attachment_storage_class table will allow mixed storages in the same system. Some can be in Database, FileSystem, or S3 and Bugzilla will know where to load from.
* Updated scripts/migrate-attachments.pl to add --migrate option if you want to permanently move data from one storage type to another.


scripts/migrate-attachments.pl
```diff
@@ -25,10 +25,19 @@
 unless ($options{mirror} || $options{copy} || $options{delete}) {
   die <<EOF;
 Syntax:
+    migrate-attachments.pl --migrate source destination
```
globau: `migrate` hasn't been added to the `GetOptions()` call

scripts/migrate-attachments.pl
```diff
@@ -59,33 +68,68 @@
 
 my $dbh = Bugzilla->dbh;
 
+if ($options{migrate}) {
+  if ($options{mirror}->[0] eq $options{mirror}->[1]) {
```
globau: s/mirror/migrate/ for this block

scripts/migrate-attachments.pl
```diff
@@ -59,33 +68,68 @@
 
 my $dbh = Bugzilla->dbh;
 
+if ($options{migrate}) {
```
purelogiq: Do we want an option to only migrate attachments above a certain size?

dklawren: This should technically happen already when migrating from DB to S3. If the 'attachment_s3_minsize' parameter is set, the S3 module will default to storing that data in the DB if less that the minsize value.

When migrating from S3 to DB, all data will be migrated regardless of size which I expect people will be fine with from the most part.

And DB to filesystem and filesystem to DB are basically no performance change from one to the other.

purelogiq: Aha, you're right we're calling the S3 module any which does that. Awesome!

dklawren:
> I don't see any changes since my last review

Sorry i must have fumbled my prior push somehow. I did it correctly this time and the changes are there.

https://github.com/mozilla-bteam/bmo/pull/1471/commits/2a98369f5ce2a0d81bf9379ac70cc54bda660b69

PR #1479, opened 2019-12-11
Bug 1597800 - Send a sec bug report for moderate and low levels

purelogiq:
This augments the existing secbugsreport.pl script to also send a report
on sec moderate and sec low bugs in a separate email according to the
sec teams requests. Will run in the same script so no need for an extra
cron job.


purelogiq:
emailssec

PR #1482, opened 2019-12-13
Bug 1588221 - Update current s3 storage controller to only upload attachments over a certain size

dklawren:
This is take 2 after being reverted before the last BMO production deployment.

Migration code is the only thing changed. Reworked to better allow for timing issues between the time the schema is updated and when the actual support code is deployed.

Also learned from deploying this to bugzilla-dev, we will need a separate schema-only PR that I will create one this has passed review.


purelogiq:
Trying this out by end of day - I have to recreate my big db since I rejiggered somethings when testing hyperv.

dklawren:
> 
> 
> Didn't run into the same migration problems this time around (though the docker changes seemed to introduce some checksetup failures on my end using vagrant, but that is separate from this).
> 
> Correct me if I'm wrong but will we have to merge the schema changes first before this PR?

Yes. I will be doing a separate schema only pr before merging this one. Thanks for the review.

PR #1483, opened 2019-12-18
no bug - Update Vagrant docs and config to improve Windows Hyper-V support

dklawren:


dklawren:
Merging this so we can close it out. We are moving from Vagrant to Docker as the only officially supported dev environment so no further Vagrant changes will be needed.

PR #1485, opened 2020-01-02
Bug 1588221 - Update current s3 storage controller to only upload attachments over a certain size (schema only)

dklawren:
Schema changes only for bug 1588221

### DIFF
--- a/Bugzilla/DB/Schema.pm
+++ b/Bugzilla/DB/Schema.pm
@@ -538,6 +538,7 @@ use constant ABSTRACT_SCHEMA => {
       attachments_ispatch_idx           => ['ispatch'],
     ],
   },
+
   attach_data => {
     FIELDS => [
       id => {
@@ -548,6 +549,20 @@ use constant ABSTRACT_SCHEMA => {
           {TABLE => 'attachments', COLUMN => 'attach_id', DELETE => 'CASCADE'}
       },
       thedata => {TYPE => 'LONGBLOB', NOTNULL => 1},
+    ],
+  },
+
+  attachment_storage_class => {
+    FIELDS => [
+      id => {
+        TYPE       => 'INT3',
+        NOTNULL    => 1,
+        PRIMARYKEY => 1,
+        REFERENCES =>
+          {TABLE => 'attachments', COLUMN => 'attach_id', DELETE => 'CASCADE'}
+      },
+      storage_class => {TYPE => 'varchar(64)', NOTNULL => 1},
+      extra_data    => {TYPE => 'MEDIUMTEXT'}
     ],
   },
 
--- a/Bugzilla/DB/Schema.pm
+++ b/Bugzilla/DB/Schema.pm
@@ -337,7 +337,7 @@ use constant ABSTRACT_SCHEMA => {
         REFERENCES => {TABLE => 'bugs', COLUMN => 'bug_id', DELETE => 'CASCADE'}
       },
       attach_id => {
-        TYPE => 'INT3',
+        TYPE => 'INT5',
         REFERENCES =>
           {TABLE => 'attachments', COLUMN => 'attach_id', DELETE => 'CASCADE'}
       },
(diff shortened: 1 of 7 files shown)


purelogiq:
Might have spoke too soon in #1482. I get this error when running this migration for the first time on a db dump `DBD::mysql::db do failed: Duplicate entry '8388607' for key 'PRIMARY' [for Statement "INSERT INTO attachment_storage_class (id, storage_class) VALUES (?, 'database')"]`.

It was working through a lot of them at a good pace then failed on that bug so perhaps its just one anomaly in my db. Just wanted to get your opinion @dklawren.

dklawren:
> 
> 
> Might have spoke too soon in #1482. I get this error when running this migration for the first time on a db dump `DBD::mysql::db do failed: Duplicate entry '8388607' for key 'PRIMARY' [for Statement "INSERT INTO attachment_storage_class (id, storage_class) VALUES (?, 'database')"]`.
> 
> It was working through a lot of them at a good pace then failed on that bug so perhaps its just one anomaly in my db. Just wanted to get your opinion @dklawren.

It could be an anomaly. Or my logic code is not sane for:

      if (!$dbh->selectrow_array(
        'SELECT id FROM attachment_storage_class WHERE id = ?',
        undef, $attach_id
      ))

That should never try to insert a row that already exists in attachment_storage_class. So I will double check that is right. Also it is strange since your attachments table should not have two rows with the same attach_id value so you may should check that out on your end. and removing the extra row.

purelogiq:
Ok here's my todo:
- [ ] write a query to check if there are dup attach_ids and check my local db dump
- [ ] run said query on dev/prod to see if this will be problem there as well

purelogiq:
Modified block with print statements
```    foreach my $attach_id (@$attach_ids) {
      print "Checking attach_id = " . $attach_id . "\n";
      if (!$dbh->selectrow_array(
        'SELECT id FROM attachment_storage_class WHERE id = ?',
        undef, $attach_id
      ))
      {
        print "Inserting attach_id = " . $attach_id . "\n";
        $dbh->do(
          "INSERT INTO attachment_storage_class (id, storage_class) VALUES (?, 'database')",
          undef, $attach_id
        );
      }
    }```
Result (it checks and skips all the ones before these):
```Checking attach_id = 8388600
Checking attach_id = 8388601
Checking attach_id = 8388602
Checking attach_id = 8388603
Checking attach_id = 8388604
Checking attach_id = 8388605
Checking attach_id = 8388607
Checking attach_id = 8388609
Inserting attach_id = 8388609
Took 379.593497037888 seconds.

Runtime error: DBD::mysql::db do failed: Duplicate entry '8388607' for key 'PRIMARY' [for Statement "INSERT INTO attachment_storage_class (id, storage_class) VALUES (?, 'database')"]```
Looks like somehow the dbh->do is using the attach_id of the last iteration. Even thought the print statement in the same block uses the current iteration id.

**Comment by purelogiq** [comment]
@dklawren Can't believe I'm seeing this but apparently something is happening. Possibly db corruption?

```
Populating attachments_storage_class table...
DBD::mysql::db do failed: Duplicate entry '8388607' for key 'PRIMARY' [for Statement "
       INSERT INTO attachment_storage_class (id, storage_class)
       SELECT attachments.attach_id, 'database' FROM attachments
       WHERE NOT EXISTS (SELECT 1 FROM attachment_storage_class WHERE id = attachments.attach_id)
    "]
```

FWIW I'm using my vagrant setup without perl 5.30 incase that affects anything (will be switching to docker once email bounce changes are in).

dklawren:
Ok this is messed up. Attachments.attach_id and attachment_storage_class.id are the same data type (INT3 and MEDIUMSERIAL) in Bugzilla/DB/Schema.pm. So at some point in the past we changed the data type to bigint(20) in the production database but did not update the code itself. Also bz_schema has MEDIUMSERIAL which is bz_schema is supposed to match the real DB. Will need to figure out how to clean up the mess.

PR #1486, opened 2020-01-08
Bug 1592410 - Don't count bugs twice when using `count_only`.  r=dkl

globau:


globau:
Untested due to https://bugzilla.mozilla.org/show_bug.cgi?id=1607655

PR #1489, opened 2020-01-10
Bug 1608491 - Improve docker environment for developers wanting to work on bmo code

dklawren:
You can sync new code changes locally into the running container using the following:

$ docker-compose exec bmo-web.vm rsync -avP --exclude .git --exclude local /mnt/sync/ /app/

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -50,6 +50,44 @@ services:
       - bmo-db.vm
       - memcached
 
+  bmo-feed.vm:
+    build: *bmo_build
+    command:
+      - perl
+      - extensions/PhabBugz/bin/phabbugz_feed.pl
+      - start
+      - '-d'
+      - '-f'
+    volumes:
+      - bmo-data-dir:/app/data
+    tmpfs:
+      - /tmp
+      - /run
+    environment: *bmo_env
+    restart: always
+    depends_on:
+      - bmo-db.vm
+      - memcached
+
+  bmo-pushd.vm:
+    build: *bmo_build
+    command:
+      - perl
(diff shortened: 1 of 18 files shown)


docker-compose.yml
```diff
@@ -64,10 +67,24 @@ services:
...
+      context: docker/tinyproxy
+      dockerfile: ./Dockerfile
     ports:
-      - "1091:1090"
+      - "1080:1080"
 
 volumes:
   bmo-mysql-db:
   bmo-data-dir:
+  bmo-sync-dir:
```
mitchhentges: You might not need to separately define these volumes out here: I had to do it in `suite` because `volumes_from` was deprecated, and it was the only way to share a host-mount volume between two containers.
However, if the mount is only needed for one container, you should still be able to use the shorthand in the service definition:
```
  bmo-web.vm:
    build: &bmo_build
      context: .
      dockerfile: Dockerfile
    command: dev_httpd
    volumes:
      - bmo-data-dir:/app/data
      - .:/mnt/sync
```

dklawren: Thanks. Now I realize when you need to do that.

dylanwh:
Excellent that the docker env is using morbo!

mitchhentges:
We should probably copy these changes into the other `docker-compose` configs for BMO:
* The one [in `phabricator-extensions`](https://github.com/mozilla-services/phabricator-extensions/blob/master/docker-compose.bmo.yml)
* The one [in `suite`](https://github.com/mozilla-conduit/suite/blob/master/docker-compose.bmo.yml)

PR #1497, opened 2020-02-04
Bug 1613179 - Hide Phabricator related attachments from the attachment list and rely on phabricator table instead

dklawren:


kyoshino:
Is this the same as #1159? It was backed out in #1175 due to [uplift difficulties](https://bugzilla.mozilla.org/show_bug.cgi?id=1534084#c4).

dklawren:
> 
> 
> Is this the same as #1159? It was backed out in #1175 due to [uplift difficulties](https://bugzilla.mozilla.org/show_bug.cgi?id=1534084#c4).

Ah yeah. It was long enough ago I forgot about the old one we backed out. This change came up again at allhands and of course it seemed like an easy fix. But as you recall, it messed up some things. 

What if I were to add a link in the new phabricator table instead to the details page for now?

kyoshino:
Yup, we just need to figure out how to merge the Details links with the Phab section :-)

kyoshino:
Was checking this but it seems I can’t test Phab stuff locally without actual bug data. I should probably request a BMO DB snapshot again 😆

globau:
> What if I were to add a link in the new phabricator table instead to the details page for now?

Not necessary - uplifts are moving to Phabricator; let's wait for that to happen before doing this.

PR #1501, opened 2020-02-06
Bug 1611494 - Bugzilla custom email headers are getting mashed together

dklawren:
Remove custom header encoding function from Mailer.pm and instead use Email::MIME::header_str_set which will handle special encoding for headers for us.


globau:
Looks like this landed without review https://github.com/mozilla-bteam/bmo/commit/07a2f1e5c44fc3e2c7a24b0f71671c196c22ba87

I'm concerned this change will break things by encoding all headers, especially since the old code went to great lengths to avoid encoding.

PR #1502, opened 2020-02-07
Bug 1611281 - Double-escaping of '<' in code areas

dklawren:


S-B99:
Looks good to me imo

Bugzilla/Markdown.pm
```diff
@@ -65,13 +65,18 @@ sub render_html {
     return $html;
   }
 
-  $markdown =~ s{<(?!https?://)}{&lt;}gs;
+  # This is a bit faster since it doesn't engage the regex engine.
```
globau: this line of the comment is a bit out of place as the second line does indeed use a regex.  suggest deleting this line.

PR #1510, opened 2020-02-20
Bug 1612290 - Provide self-service UI for users to reactivate their account after being disabled due to bouncing

dklawren:


template/en/default/account/email/bounced-emails.html.tmpl
```diff
@@ -0,0 +1,54 @@
...
+  # This Source Code Form is "Incompatible With Secondary Licenses", as
+  # defined by the Mozilla Public License, v. 2.0.
+  #%]
+
+[% PROCESS global/header.html.tmpl
+  title = "Bounced Emails"
+%]
+
+<h2>[% title FILTER html %]</h2>
+
```
globau: This could do with a bit more explanatory text.

I'd like to see the current bounce reason called out here, an explanation of what that means and what actions we expect users to take to resolve it prior to reactivating delivery (eg. you may have to contact your email provider to ensure there aren't any issues with your mail account).  We should also state here that if your email bounces more than `BOUNCE_REACTIVATION_MAX` times a Bugzilla administrator will need to reactivate your account after performing more checks.

Bugzilla/Constants.pm
```diff
@@ -666,6 +668,10 @@ use constant EMAIL_LIMIT_EXCEPTION => "email_limit_exceeded\n";
 # (view_job_queue.cgi).
 use constant JOB_QUEUE_VIEW_MAX_JOBS => 2500;
 
+# Maximum number of times an email can bounce for an account
+# before the account is completely disabled.
+use constant BOUNCE_COUNT_MAX => 5;
```
globau: We also need a `BOUNCE_REACTIVATION_MAX` count, and track per-user how many times they have reactivated delivery after bounce-triggered disabling.

Currently if someone's email is completely non-deliverable there's nothing stopping them from immediately reactivating bugmail each and every time they hit 5 bounces.  This will increase our SES bounce rates and might lead to Amazon disabling our account.

Once they hit this cap the `/bounced_emails/` page should direct the user to contact the BMO admins.
editusers will need to be updated to allow admins to reset the per-user reactivation counter.

Bugzilla/App/SES.pm
```diff
@@ -199,16 +199,32 @@ sub _process_bounce {
...
-        my $disable_text;
+        my $bounce_message;
         $template->process('admin/users/bounce-disabled.txt.tmpl',
-          $vars, \$disable_text)
+          $vars, \$bounce_message)
           || die $template->error();
 
-        $user->set_disabledtext($disable_text);
         $user->set_disable_mail(1);
+        $user->set_bounce_count($user->bounce_count + 1);
```
globau: This should be _recent_ bounces, rather than a count since over the entire lifetime of an account.

As implemented if an account has 4 bounces early on, then 5 years later there's a single bounce, bugmail will be disabled.

I'm not sure what the actual timeframe should be (maybe @ckolos can provide some guidance), but something like "5 bounces within the last 4 weeks" feels about right.

Bugzilla/Install/DB.pm
```diff
@@ -805,6 +805,9 @@ sub update_table_definitions {
   # Bug 1576667 - dkl@mozilla.com
   _populate_api_keys_creation_ts();
 
+  # Bug XXX - dkl@mozilla.com
```
globau: `XXX`

Bugzilla/App/BouncedEmails.pm
```diff
@@ -0,0 +1,56 @@
...
+  if ( $self->param('enable_email')
+    && $user->id == $other_user->id
+    && $other_user->email_disabled)
+  {
+    my $token = $self->param('token');
+    check_token_data($token, 'bounced_emails');
+
+    $other_user->set_email_enabled(1);
+    $other_user->update();
+
```
globau: this should have a `audit` entry to mirror the disabled message.

Bugzilla/App/SES.pm
dklawren: > 
> 
> This should be _recent_ bounces, rather than a count since over the entire lifetime of an account.
> 
> As implemented if an account has 4 bounces early on, then 5 years later there's a single bounce, bugmail will be disabled.
> 
> I'm not sure what the actual timeframe should be (maybe @ckolos can provide some guidance), but something like "5 bounces within the last 4 weeks" feels about right.

Actually I failed to mention here that there is also a new checkbox on the editusers.cgi user view that is for resetting the bounce count to zero. So when a user is deactivated BOUNCE_COUNT_MAX times and the admin has to enable their account manually, the admin will also need to check the box to reset the count. There is a link from the editusers.cgi view to the age where all of the bounce messages are listed similar to what the user sees.

That being said, I did toy with the idea in B/App/SES.pm of when a bounce occurs, and the user has not had a bounce in in less than 30 days, then set the bouce count to 1 instead of whatever the value currently is. I think that will work out fine for now.

Bugzilla/Constants.pm
dklawren: > 
> 
> We also need a `BOUNCE_REACTIVATION_MAX` count, and track per-user how many times they have reactivated delivery after bounce-triggered disabling.
> 
> Currently if someone's email is completely non-deliverable there's nothing stopping them from immediately reactivating bugmail each and every time they hit 5 bounces. This will increase our SES bounce rates and might lead to Amazon disabling our account.
> 
> Once they hit this cap the `/bounced_emails/` page should direct the user to contact the BMO admins.
> editusers will need to be updated to allow admins to reset the per-user reactivation counter.

This should not happen as when we set the disabledtext in SES.pm at 5 bounces, the user is automatically logged out the next time they access any page and when they try to login they get the message stating why they are disabled. So they will no longer be able to get to /bounced_emails/ after 5 bounces.

Bugzilla/App/BouncedEmails.pm
dklawren: We already store the disabled messages in the audit log in B/A/SES.pm. They never go away.

Bugzilla/App/SES.pm
globau: I'm still not sure having a perm bounce count is ideal, however I'm happy to leave it as-is and see how it works out.

Bugzilla/Constants.pm
globau: Ah right, I misread some code.  `BOUNCE_COUNT_MAX` does exactly what we need :)

template/en/default/account/email/bounced-emails.html.tmpl
```diff
@@ -0,0 +1,61 @@
...
+[% IF user.id == other_user.id AND other_user.bounce_count AND other_user.email_disabled %]
+<p>
+  Due to issues delivering email to your account, we have temporarily disabled email notifications
+  being sent to it. If you feel like the issue has been resolved, you may reactivate email delivery
+  below. After a maximum of [% bounce_max FILTER html %] bounces, we will disable logging in to your
+  account and you will need to contact an administrator to reactivate it.</p>
```
globau: nice!

PR #1513, opened 2020-02-25
Bug 1237874 - File size unit always plural: "1 bytes"

stellarbeam:
modified:   Bugzilla/Template.pm to account for 1 byte file size

### DIFF
--- a/Bugzilla/Template.pm
+++ b/Bugzilla/Template.pm
@@ -807,12 +807,7 @@ sub create {
         my %units = ('KB' => 1024, 'MB' => 1024 * 1024, 'GB' => 1024 * 1024 * 1024,);
 
         if ($data < 1024) {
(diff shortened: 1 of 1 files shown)


Bugzilla/Template.pm
```diff
@@ -807,7 +807,12 @@ sub create {
         my %units = ('KB' => 1024, 'MB' => 1024 * 1024, 'GB' => 1024 * 1024 * 1024,);
 
         if ($data < 1024) {
-          return "$data bytes";
+          if($data == 1) {
+            return "1 byte";
+          }
+          else {
+            return "$data bytes";
+          }
```
globau: This looks good, but it would be more idiomatic to use a ternary operator:
`return "$data byte" . ($data == 1 ? '' : 's');`

PR #1516, opened 2020-02-26
Bug 1612290 - Provide self-service UI for users to reactivate their account after being disabled due to bouncing (schema only)

dklawren:
Sorry for not catching this sooner. When running checksetup.pl on bugzilla-dev, checksetup.pl tries to access profiles before the schema is updated and fails with the error that profiles.bounce_count does not exist. This will make the schema changes needed first. A second deployment will be done that contains the rest of the functionality.


dklawren:
Not sure why the extra commits on my fork are listed in the commits but the Files Changed looks correct.

PR #1523, opened 2020-03-06
Bug 1517996 - Add a link back to the user's profile from the Activity Report page

KWierso:
…

### DIFF
--- a/extensions/BMO/lib/Reports/UserActivity.pm
+++ b/extensions/BMO/lib/Reports/UserActivity.pm
@@ -322,6 +322,8 @@ sub report {
     $vars->{'incomplete_data'} = $incomplete_data;
     $vars->{'operations'}      = \@operations;
 
+    $vars->{user} = Bugzilla::User->new({ name => $input->{who}, cached => 1 });
+
     my @bug_ids = sort { $a <=> $b } keys %bug_ids;
     $vars->{'bug_ids'} = \@bug_ids;
   }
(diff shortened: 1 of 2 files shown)


extensions/BMO/lib/Reports/UserActivity.pm
```diff
@@ -322,6 +322,8 @@ sub report {
     $vars->{'incomplete_data'} = $incomplete_data;
     $vars->{'operations'}      = \@operations;
 
+    $vars->{user} = Bugzilla::User->new({ name => $input->{who}, cached => 1 });
```
dklawren: This is inside the for loop that stores each activity item. So depending on how many activity rows there are, it will be set each time. Also user_activity.html allows for displaying more than one person grouped together. Normally from user_profile.html, the link to user_activity.html would only be a single person so wouldn't be an issue with that case. But we need to be able to handle both cases.

So out side of the for loop you need to create $var->{users}:

```perl
$var->{users}
  = [ map { Bugzilla::User->new({ name => $_, cache => 1}) } @who ];
```

extensions/BMO/template/en/default/pages/user_activity.html.tmpl
```diff
@@ -104,6 +104,10 @@
   </p>
 [% END %]
 
+<p>
+  <a href="[% basepath FILTER none %]user_profile?user_id=[% user.id FILTER uri %]"> User Profile</a>
+</p>
```
dklawren: We need to deal with the possibility that multiple users were searched on instead of always coming from user_profile.html so we may need a loop here:

```
<p>
  User Profiles:
  [% FOREACH u = users %]
    <a href="[% basepath FILTER none %]user_profile?user_id=[% u.id FILTER uri %]">[% user.name FILTER html %]</a>[% ", " IF NOT loop.last() %]
  [% END %]
</p>

PR #1528, opened 2020-03-11
Bug 1621278 - Make bug IDs on search results proper bug links

kyoshino:
Linkify all bug IDs on the search results page, including the fixed Bug ID column. Somehow the Duplicates field’s column type is `FIELD_TYPE_BUG_ID` so I’ll keep it to avoid any potential breakage and make the linkification work for both single bug ID and multiple bug IDs.

## Bugzilla link

https://bugzilla.mozilla.org/show_bug.cgi?id=1621278

### DIFF
--- a/template/en/default/list/table.html.tmpl
+++ b/template/en/default/list/table.html.tmpl
@@ -209,12 +209,13 @@
     </td>
     [% END %]
     <td class="first-child bz_id_column">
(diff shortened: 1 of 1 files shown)


kyoshino:
@dklawren hmm I forgot how to fix the build error after a “bump version” commit 🤔

PR #1533, opened 2020-03-18
Bug 1622867 - Add an equivalent to `/latest/configuration` to the REST api

dklawren:
Took a different approach to this. This is a native Mojo route and will take the place of

/rest/configuration
/latest/configuration
/bzapi/configuration

I have tried to make sure it is the same JSON structure as /bzapi/configuration.

Another thing I did was introduce API versioning support. Eventually all of the API code from Bugzilla/WebServices/*.pm will be moved under Bugzilla/API/V1/*.pm and given Mojo style routes and params. Should make the move easier now with the changes in this PR. Then the entire Bugzilla/WebServices* dir structure can just simply be removed effectively decommissioning XMLRPC/JSONRPC.


Bugzilla/API/V1/Configuration.pm
```diff
@@ -0,0 +1,273 @@
...
+use Bugzilla::Product;
+use Bugzilla::Status;
+
+use List::MoreUtils qw(any);
+use Mojo::Base qw( Mojolicious::Controller );
+use Mojo::JSON qw( true false );
+
+sub setup_routes {
+  my ($class, $r) = @_;
+  $r->get('/latest/configuration')->to('V1::Configuration#configuration');
```
globau, line 25: If we're providing backwards compat we should also hook up routes for `/rest/configuration` and `/bzapi/configuration`, even if they just redirect to the canonical `/latest/configuration` endpoint.

edit: ah, I see you're defining `/rest/configuration` in CGI.pm and `/bzapi/configuration` in the BzAPI extension.  Both of those should be moved here.

Bugzilla/API/V1/Configuration.pm
```diff
@@ -0,0 +1,273 @@
...
+  my $user = Bugzilla->user;
+
+  my $can_cache = !$self->param('product') && !$self->param('flags');
+
+# Using config.* will clear this data in memcache when Bugzilla changes are made to products, components, etc.
+  my $cache_key = 'config.configuration';
+  if ($can_cache) {
+    my $result = Bugzilla->memcached->get_config({key => $cache_key});
+    return $self->render(json => $result) if defined $result;
+  }
```
globau, line 39: `Bugzilla->memcached->set_config()` is never called.

Bugzilla/API/V1/User.pm
```diff
@@ -0,0 +1,41 @@
...
+
+package Bugzilla::API::V1::User;
+
+use 5.10.1;
+use Mojo::Base qw( Mojolicious::Controller );
+use Mojo::JSON qw( true false );
+
+sub setup_routes {
+  my ($class, $r) = @_;
+  $r->get('/api/user/profile')->to('V1::User#user_profile');
```
globau, line 16: I know this is an existing route, but why is the `/api/` not `/rest/`?

extensions/BzAPI/Extension.pm
```diff
@@ -291,6 +288,9 @@ sub app_startup {
 
   Bugzilla::App::CGI->load_one('bzapi_cgi', 'extensions/BzAPI/bin/rest.cgi');
   $r->any('/extensions/BzAPI/bin/rest.cgi/*PATH_INFO')->to('CGI#bzapi_cgi');
+  # Use Bugzilla::API::V1::Configuration
+  $r->any('/latest/configuration')->to('V1::Configuration#configuration');
+  $r->any('/bzapi/configuration')->to('V1::Configuration#configuration');
```
globau, line 293: As per above, move into `Bugzilla::API::V1::Configuration`

Bugzilla/App/CGI.pm
```diff
@@ -35,6 +35,9 @@ sub setup_routes {
   $r->any('/bug/<id:num>')->to('CGI#show_bug_cgi');
   $r->any('/<id:num>')->to('CGI#show_bug_cgi');
 
+  # Route to Bugzilla::API::V1::Configuration for now
+  $r->get('/rest/configuration')->to('V1::Configuration#configuration');
+
```
globau, line 40: As per above, move into `Bugzilla::API::V1::Configuration`

Bugzilla/API/V1/Configuration.pm
globau, line 25: It's weird that the versioning is _internal_ only; the main point of versioning an API is for _external_ consumers – eg `/rest/v1/bug/12345` vs `/rest/v2/bug/12345`; that way consumers are isolated from breaking changes.

As it stands I'm unclear on the benefit on this versioning work.

globau (changes requested):

Bugzilla/API/V1/Configuration.pm
dklawren: /rest/bug is the same as /rest/v1/bug as the original API paths are always considered v1. For v2 and beyond, people would then need to start including versions in the path.

Bugzilla/API/V1/User.pm
dklawren: Because /api/user/profile is what the OAuth2 code in Phabricator currently uses. This was something Dylan set up to create a new path for Mojo implemented API calls. I like /api better than /rest but we will need to leave it as /rest for the foreseeable future. I will add another /rest/user/profile path below so that both work the same.

globau (approved):

PR #1540, opened 2020-03-25
Bug 1183759 :  Keyword suggestions should not show keywords which have been already selected 

manyaagarwal:
Made changes to the edit bug and add bug workflow.

### DIFF
--- a/extensions/BugModal/web/bug_modal.js
+++ b/extensions/BugModal/web/bug_modal.js
@@ -638,14 +638,19 @@ $(function() {
                         forceFixPosition: true,
                         lookup: function(query, done) {
                             query = query.toLowerCase();
+                            var that = $('#keywords');
+                            var activeValues = that.val().split(',');
+                            activeValues.forEach((o,i,a) => a[i] = a[i].trim());
                             var matchStart =
                                 $.grep(keywords, function(keyword) {
-                                    return keyword.toLowerCase().substr(0, query.length) === query;
+                                    if(!(activeValues.includes(keyword)))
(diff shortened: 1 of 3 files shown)


extensions/BugModal/web/bug_modal.js
```diff
@@ -638,14 +638,19 @@ $(function() {
                         forceFixPosition: true,
                         lookup: function(query, done) {
                             query = query.toLowerCase();
+                            var that = $('#keywords');
```
emceeaich: Could you use `let` instead of creating `that` to keep `#keywords` in scope?

emceeaich: Also, using `document.querySelect()` instead of `$` will make it easier to migrate away from jQuery in the future.

manyaagarwal: Sure, thanks!  Added the changes.

js/field.js
```diff
@@ -716,14 +716,18 @@ $(function() {
                 lookup: function(query, done) {
                     var values = BUGZILLA.autocomplete_values[that.data('values')];
                     query = query.toLowerCase();
+                    var activeValues = that.val().split(',');
```
dklawren: This line needs to be changed as currently it will only read in the first time and not get any updates to the field if someone removes or adds a keyword. Change to:

var activeValues = $('#keywords').val().split(',');

This will cause it to re-read each time the current set of field values.

manyaagarwal: Should I use document.querySelector instead of '$' to make migration easier as Emma suggested?

PR #1541, opened 2020-03-26
upgrade to latest mojolicious and fix logging glue

dylanwh:
It looks like MojoX::Log::Log4perl::Tiny is not maintained, and it won't work with newer mojolicious. 

So here I upgrade it and switch out to MojoX::Log::Log4perl which was last updated in October 2019. Not sure if this works but hoping it will


dylanwh:
Bah, ignore this, I'm trying this out in bugzilla/harmony first. Sorry for the noise.

PR #1546, opened 2020-04-07
Bug 1628467 - Button to stop auto-refresh in My dashboard

LissetCuevas:
Switch button added to stop or start auto-refresh in My dashboard. 
@dklawren @emceeaich

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@
 # 2. https://circleci.com/docs/2.0/building-docker-images/
 #
 
-version: 2
+version: 2.1
 
 defaults:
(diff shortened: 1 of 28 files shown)


LissetCuevas:
> Looks good to me. Something we can do later as an enhancement is to store the current setting in a cookie so that it is either on all of the time or off even if the page is reloaded. We can do that as a separate bug.

I thought that but I didn't know how to store it but I'll check the cookies and I'll do it. Is there some part of the project that already uses cookies? To use it as a guide.

And also if we store the setting maybe could be useful if the user can select how often wants the auto-refresh. Do you think it could be useful or not?

PR #1559, opened 2020-04-28
Bug 1633846 - Enable retries for uploading attachments to S3 when an error occurs

dklawren:


Makefile.PL
```diff
@@ -49,7 +49,7 @@ my %requires = (
   'DBI'                                 => '1.614',
   'DBIx::Connector'                     => 0,
   'DBIx::Class'                         => 0,
-  'DBIx::Class::Helpers'                => 0,
+  'DBIx::Class::Helpers'                => '== 2.034002',
```
globau, line 52: this seems unrelated; why does this need pinning now, and if it's required why aren't the other dbix classes pinned?

dklawren: It is related in that any version above 2.034002 causes an error in the Model code Dylan wrote for sending data to STMO. The newer version came was released after the last time we generated a bmo-slim image. The Helpers module seems to be independent to the DBIx::Class module so the Class module does not need to be pinned currently.

PR #1568, opened 2020-05-14
Bug 1638022 - Phabbugz should not try to set needs-review when the revision is closed or abandoned

dklawren:


dklawren:
Failing because of the no new revisions issue and not because of the code itself.

extensions/PhabBugz/lib/Feed.pm
```diff
@@ -523,9 +523,11 @@ sub process_revision_change {
   # in draft state and not changes-planned
```
globau, line 523: this comment needs to be updated too

PR #1569, opened 2020-05-17
no bug - clear request cache in Glue plugin, before setting usage mode

dylanwh:
This patch fixes the behavior of `Bugzilla->usage_mode` so it is always consistent, and also cleans things up by keeping this in one place.

The trick is to replace `before_dispatch` with `around_dispatch`, and use a localized package variable. As `around_dispatch` wraps the entire request, the localized package variable (`$cleanup_guard`) will not be cleared until dispatch is totally done.

Prior to this change, even if you were to set the `usage_mode` after clearing the request cache, it would be cleared at an *unpredictable* time as Mojolicious does not clear the stash when it used to. Granted, we should have never relied on that but meh.

Anyway, this fixes a number of problems and potential problems.

### DIFF
--- a/Bugzilla/App/CGI.pm
+++ b/Bugzilla/App/CGI.pm
@@ -73,7 +73,7 @@ sub load_one {
     tie *STDOUT, 'Bugzilla::App::Stdout', controller => $c;   ## no critic (tie)
 
     # the finally block calls cleanup.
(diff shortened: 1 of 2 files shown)


dylanwh:
Even this won't fix it; I'll have a better fix done shortly.

PR #1584, opened 2020-07-01
Bug 1645768 - Please add 'See Also' support for GitLab

dklawren:


Bugzilla/BugUrl/GitLab.pm
```diff
@@ -20,9 +20,10 @@ use base qw(Bugzilla::BugUrl);
 sub should_handle {
   my ($class, $uri) = @_;
 
-# GitLab issue URLs can have the form:
+# GitLab issue and merge request URLs can have the form:
 # https://gitlab.com/projectA/subprojectB/subprojectC/../issues/53
```
zalun, line 24: This link is for issues only.
Should we change this to `../[issues|merge_requests]/53` ?

PR #1590, opened 2020-07-09
Bug 1651591 - remove preloading of fonts and ga; r=dkl

globau:
Font preloading has been broken for more than a year, and doesn't appear
to work correctly even when provided URLs which exist; removed.

google-analytics likely does little to improve page load time
performance against all the other assets, is loaded async, and exists
outside of the GoogleAnalytics making it easy to miss when updating that
extension (eg. the preloading doesn't honour DNT); also removed.

### DIFF
--- a/Bugzilla/CGI.pm
+++ b/Bugzilla/CGI.pm
@@ -410,26 +410,6 @@ sub header {
 
   $self->{_header_done} = 1;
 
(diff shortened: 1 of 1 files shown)


emceeaich:
Google Analytics is managed through the admin screens. So we should remove the configuration code (they are in a plugin.)

Also should investigate GA alternatives.

PR #1593, opened 2020-07-21
Bug 1654370 - Remove remaining code that references Firefox OS from BMO code base

dklawren:


extensions/BMO/Extension.pm
```diff
@@ -2647,19 +2632,6 @@ sub _split_crash_signature {
       extract_multiple($crash_signature, [sub { extract_bracketed($_[0], '[]') }])];
 }
 
-sub enter_bug_entrydefaultvars {
```
zalun, line 2650: I'm not that familiar with the flow of the code.
You haven't removed any calls to that function.

Is that OK?

dklawren: It is a hook subroutine in an extension. So when a callout is made in the core code it looks for any extensions that have that subroutine defined and runs it. So removing it from the extensions/BMO/Extension.pm file will not cause any problems with other extension that may be using that hook.

PR #1595, opened 2020-07-22
Bug 1652863 - setting the needinfo flag when filing a new bug in Core or Toolkit does not cause the textbox for user information to pop up

dklawren:


dklawren:
This fix was confirmed on bugzilla-dev.allizom.org so I am merging for a quick release.

PR #1606, opened 2020-08-13
Bug 1658846 - Allow users to enable and disable their webhooks

LissetCuevas:
@dklawren I made some little changes that I saw, and the UI is like this:
image

### DIFF
--- a/extensions/Webhooks/Extension.pm
+++ b/extensions/Webhooks/Extension.pm
@@ -21,6 +21,7 @@ use Bugzilla::User;
 use Bugzilla::Logging;
 use Bugzilla::Extension::Webhooks::Webhook;
 use Bugzilla::Extension::Push::Util;
+use Bugzilla::Util;
 use Try::Tiny;
 
 #
@@ -77,6 +78,7 @@ sub user_preferences {
 
   my $input = Bugzilla->input_params;
   my $user  = Bugzilla->user;
+  my $push  = Bugzilla->push_ext;
   my $vars  = $args->{vars};
 
   if ($args->{'save_changes'}) {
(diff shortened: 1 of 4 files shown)


extensions/Webhooks/Extension.pm
```diff
@@ -143,10 +146,26 @@ sub user_preferences {
...
 
+      my $dbh      = Bugzilla->dbh;
+      my $webhooks = Bugzilla::Extension::Webhooks::Webhook->match(
+        {user_id => $user->id});
+
+      $dbh->bz_start_transaction();
+      foreach my $webhook (@$webhooks) {
+        my $connector = $push->connectors->by_name('Webhook_' . $webhook->id);
+        my $config = $connector->config;
+        $config->{enabled} = trim($input->{$connector->name . ".enabled"});
```
dklawren, line 164: We may need to validate the input here to make sure that the value is either 'Enabled' or 'Disabled'. Sneaky users can modify values before POST to circumvent checks to do bad things.

LissetCuevas: I hadn't thought about sneaky users before :open_mouth:

extensions/Webhooks/template/en/default/hook/global/user-error-errors.html.tmpl
```diff
@@ -18,4 +18,8 @@
   [% title = "Define a URL" %]
   You didn't define a URL for your webhook. Define one.
 
+[% ELSIF error == "invalid_option" %]
```
dklawren, line 21: `webhook_invalid_option`

extensions/Webhooks/Extension.pm
```diff
@@ -161,8 +161,13 @@ sub user_preferences {
...
         my $config = $connector->config;
-        $config->{enabled} = trim($input->{$connector->name . ".enabled"});
-        $config->update();
+        my $status = trim($input->{$connector->name . ".enabled"});
+        if ( $status eq 'Enabled' || $status eq 'Disabled' ){
+          $config->{enabled} = $status;
+          $config->update();
+        }else{
+          ThrowUserError('invalid_option');
+        }
```
dklawren, line 170: good

PR #1610, opened 2020-08-19
Bug 1658845 - Allow users to see their own queue for their webhooks

LissetCuevas:
The queue count is added as a column in the webhooks table and if exist a queue, is created a link to the webhooks_queues page.

### DIFF
--- a/extensions/Webhooks/template/en/default/account/prefs/webhooks.html.tmpl
+++ b/extensions/Webhooks/template/en/default/account/prefs/webhooks.html.tmpl
@@ -124,7 +124,7 @@ window.onload = function() {
       <th>Product</th>
       <th>Component</th>
       <th>Status</th>
-      <th>Queue</th>
+      <th>Errors</th>
     </tr>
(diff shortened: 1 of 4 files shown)


extensions/Webhooks/template/en/default/account/prefs/webhooks.html.tmpl
```diff
@@ -156,15 +146,23 @@ window.onload = function() {
...
       </td>
+      <td>
+        [% IF connector.backlog.count%]
+          <a href="page.cgi?id=webhooks_queues.html&amp;[% ~%]
+                  webhook=[% webhook.id FILTER uri %]">
+            [% connector.backlog.count FILTER html %]
+          </a>
+        [% ELSE %]
+          [% connector.backlog.count FILTER html %]
+        [% END %]
```
dklawren, line 164: If we do rename to **Errors** as mentioned above, I would just not display anything in the column row instead of 0.

PR #1611, opened 2020-08-19
Bug 1658841 - Webhooks should include comments added to a bug

LissetCuevas:
@dklawren I'm trying to use the `HTML::Strip` module but it throws me an error to remove html, I think I need to install the module on the project but I'm not sure.
Here I took the message when a comment is created and modify it in order it seems as current change event message from webhooks, and also are sent as part of change event, as you suggest.
And the description is sent as a parameter that changes but not as part of the bug body, because it is sent as a comment too, and for that reason is sent in another message when a bug is created not in the same:
```
"event": {
      "action": "change",
      "time": "2020-07-24T20:11:22",
      "user": {
          "id": 1,
            "login": "admin@bmo.test",
            "real_name": "Vagrant User"
      },
      "changes": [
          {
            "field": "description",
            "removed": "",
            "added": "This is a my description"
            }
        ],
}
```
If we want to send it in all webhooks messages as part of the bug body maybe I can modify the push message payload and add it.

### DIFF
--- a/extensions/Push/lib/Connector/Webhook.pm
+++ b/extensions/Push/lib/Connector/Webhook.pm
@@ -74,13 +74,11 @@ sub should_send {
   if ($product eq $bug->product
       && ($component eq $bug->component || $component eq 'any'))
   {
-    if ($event =~ /create/ && $message->routing_key eq 'bug.create') {
-      return 1;
-    }elsif ($event =~ /change/ && $message->routing_key =~ /^bug\.modify/) {
-      return 1;
-    }elsif ($event =~ /comment/ && $message->routing_key eq 'comment.create') {
-      return 1;
-    }elsif ($event =~ /attachment/ &&  $message->routing_key eq 'attachment.create') {
+    if (($event =~ /create/ && $message->routing_key eq 'bug.create')
+      || ($event =~ /change/ && $message->routing_key =~ /^bug\.modify/)
+      || ($event =~ /comment/ && $message->routing_key eq 'comment.create')
+      || ($event =~ /attachment/ &&  $message->routing_key eq 'attachment.create'))
+    {
       return 1;
     }
   }
--- a/extensions/Push/lib/Connector/Webhook.pm
+++ b/extensions/Push/lib/Connector/Webhook.pm
@@ -26,7 +26,6 @@ use JSON qw(decode_json encode_json);
 use LWP::UserAgent;
 use List::MoreUtils qw(any);
 use Try::Tiny;
-#use HTML::Strip;
 
 sub new {
   my ($class,$webhook_id) = @_;
@@ -77,7 +76,11 @@ sub should_send {
   {
     if ($event =~ /create/ && $message->routing_key eq 'bug.create') {
       return 1;
-    }elsif ($event =~ /change/ && ($message->routing_key =~ /\Qbug.modify\E/ || $message->routing_key eq 'comment.create')) {
+    }elsif ($event =~ /change/ && $message->routing_key =~ /^bug\.modify/) {
+      return 1;
+    }elsif ($event =~ /comment/ && $message->routing_key eq 'comment.create') {
+      return 1;
+    }elsif ($event =~ /attachment/ &&  $message->routing_key eq 'attachment.create') {
       return 1;
     }
   }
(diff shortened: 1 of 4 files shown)


extensions/Push/lib/Connector/Webhook.pm
```diff
@@ -76,7 +77,7 @@ sub should_send {
   {
     if ($event =~ /create/ && $message->routing_key eq 'bug.create') {
       return 1;
-    }elsif ($event =~ /change/ && $message->routing_key =~ /\Qbug.modify\E/) {
+    }elsif ($event =~ /change/ && ($message->routing_key =~ /\Qbug.modify\E/ || $message->routing_key eq 'comment.create')) {
```
dklawren, line 80: \Q\E are not needed since we are looking for an exact string so let's go ahead and  make the line:

```perl
}elsif ($event =~ /change/ && ($message->routing_key =~ /^bug\.modify/ || $message->routing_key eq 'comment.create')) {
```

extensions/Push/lib/Connector/Webhook.pm
```diff
@@ -96,12 +97,23 @@ sub send {
 
     my $bug_data   = $self->_get_bug_data($payload);
     my $is_private = $bug_data->{is_private};
+
+    if ($message->routing_key eq 'comment.create'){
+      my $comment->{field} = $payload->{comment}->{number} == 0 ? 'description' : 'comment';
+      #$comment->{added}    = HTML::Strip->new()->parse($payload->{comment}->{body});
+      $comment->{added}    = $payload->{comment}->{body};
```
dklawren, line 104: There is no need do strip HTML from the comment as we are sending the raw text and not the fully formatted text you see in the web UI. quoteUrls() in the code is what changes URLs into clickable links, etc in the web UI and we do not need that for JSON data.

Also if we can somehow load the Bugzilla/Comment.pm object for the comment that is added, we can call `$comment->body_full` which will give us extra text such as `Attachment NNNN created` or `*** This bug has been marked as a duplicate of NNNN ***`.

extensions/Webhooks/template/en/default/account/prefs/webhooks.html.tmpl
```diff
@@ -129,7 +135,6 @@ window.onload = function() {
   <thead>
     <tr>
       <th>Remove</th>
-      <th>ID</th>
```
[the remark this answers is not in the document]
dklawren: Or if we leave it out, then the data columns need to be fixed to be the right width and align properly.

extensions/Push/lib/Connector/Webhook.pm
```diff
@@ -94,17 +96,35 @@ sub send {
...
+    if ($bug_is_private){
+      if ($target eq 'bug'){
+        delete @{$payload}{bug};
+        $payload->{bug}->{id}         = $bug_data->{id};
+        $payload->{bug}->{is_private} = $bug_is_private;
+      }
+      else{
+        delete @{$payload->{$target}}{bug};
+        $payload->{$target}->{bug}->{id}         = $bug_data->{id};
+        $payload->{$target}->{bug}->{is_private} = $bug_is_private;
```
dklawren, line 121: Same here. _int($bug_data->{id}) and _boolean($bug_is_private). Basically anywhere that data is being added to the JSON that is to be sent as part of the payload, needs to use the proper methods from `extensions/lib/Push/lib/Serialize.pm`.

extensions/Push/lib/Connector/Webhook.pm
```diff
@@ -94,17 +96,35 @@ sub send {
...
-      delete @{$payload}{bug};
-      if($payload->{event}->{action} eq 'modify'){
+    my $bug_data       = $self->_get_bug_data($payload);
+    my $bug_is_private = $bug_data->{is_private};
+    my $target         = $payload->{event}->{target};
+
+    my $target_is_private = $payload->{$target}->{is_private};
+    if ($target_is_private && ($target eq 'attachment' || $target eq 'comment')){
+      my $target_id = $payload->{$target}->{id};
+      delete @{$payload}{$target};
```
dklawren, line 106: For comments, if the bug is private or the comment is private, we need to remove the content for the comment. For example if we add a comment to a private bug, we need to make sure to remove `$payload->{comment}->{body}`.

dklawren (changes requested):
Looking almost done. Just one or two more fixes. Thanks!

extensions/Push/lib/Connector/Webhook.pm
```diff
@@ -20,6 +20,7 @@ use Bugzilla::Attachment;
 use Bugzilla::Extension::Webhooks::Webhook;
 use Bugzilla::Extension::Push::Constants;
 use Bugzilla::Extension::Push::Util;
+use Bugzilla::Extension::Push::Serialize qw(_integer _boolean _bug);
```
dklawren, line 23: Sorry for the delay with this review. The company gave us all a few days off for the holiday in the US. 

This will not work unfortunately the way you have it as Serialize.pm does not export those functions using Exporter.pm. If I make a change to a private but I get a function not found error when the webhook is processed.

An example of using Exporter.pm can be seen in https://github.com/mozilla-bteam/bmo/blob/master/extensions/Push/lib/Util.pm#L23-L35

The easiest fix would be to copy the _int() and _boolean() functions from Serialize.pm and put them in the Webhook.pm module since they are very small.

As for the call to _bug(), this is not necessary as the data from `$message->payload_decoded` should be ok to use directly.

dklawren (changes requested):

extensions/Push/lib/Connector/Webhook.pm
LissetCuevas: Oh yes, sorry about that.

dklawren (changes requested):
Looks good. All done except for a bug I found when testing. The JSON data structure for a created comment has the wrong value for ID . It is returning the bug ID instead of the comment ID. 

https://github.com/mozilla-bteam/bmo/blob/master/extensions/Push/lib/Serialize.pm#L269

Needs to be changed from `$comment->bug_id` to `$comment->id`. After that it will be done.

dklawren (approved):
Cool! r=dkl

LissetCuevas:
Thanks for the reviews :)

PR #1620, opened 2020-09-15
Message for disabled accounts should reflect Mozilla's environment

jasp00:
Message for disabled accounts should reflect Mozilla's environment better. Just in case, I clarify that I care about Mozilla and their users, which is proved by my past contributions.


jasp00:
Thank you. I am glad to contribute.

PR #1626, opened 2020-09-23
Bug 1612286 - Single sign-on through Mozilla IAM

dklawren:


oauth2.cgi
```diff
@@ -0,0 +1,39 @@
...
+
+# Get access token from OAuth2 provider;
+my $resp = $c->oauth2->get_token('oauth2');
+
+# Store access token for use by OAuth2 login info getter
+$cache->{oauth2_client_userinfo} = $c->oauth2->userinfo($resp->{access_token});
+
+my $user = Bugzilla->login(LOGIN_REQUIRED);
+
+if ($user && $user->in_group('no-sso-auth')) {
```
globau, line 34: Can you give examples of why we'd want to exclude users that are "too powerful" – that seems to be the opposite to what we should be doing.

oauth2.cgi
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/env perl
```
globau, line 1: shouldn't this be a mojo route rather than a cgi?

template/en/default/account/auth/login.html.tmpl
```diff
@@ -84,6 +84,12 @@
   </form>
 </div>
 
+[% IF Param('oauth2_client_enabled') %]
+  [% INCLUDE account/auth/oauth2_login_button.html.tmpl
+     target => target
+  %]
+[% END %]
```
globau, line 91: Can you please include a screenshot of how the two login buttons are presented

globau (changes requested):
I see this is WIP so some quick comments.

oauth2.cgi
dklawren: I was duplicating what we do currently with Github auth where we do not allow people with heightened security permissions to login via Github. But thinking it about it more now, it is OK to allow users of Mozilla IAM to be able to login regardless of permissions. And the fact that we could one day make all people use SSO for auth. Will remove.

oauth2.cgi
dklawren: That was my original plan. But after trying various hacks to make it work I gave in and made it a small cgi. Reason is that the way our current Auth stack works relies heavily on running inside the CGI compatibility wrapper that we use with Mojolicious. I would either need to do quite a bit of refactoring or duplicating a lot of code to make it work. 

My future plan is to start moving most of the supporting code (login, utilities, etc.) into native Mojo and then doing these things will be much less painful.

template/en/default/account/auth/login.html.tmpl
dklawren: image

image

globau: It'll need to be styled to match GitHub's button (or vice versa, although having the service's icon is super useful).

dklawren: Probably should use something more standard such as 

https://lipis.github.io/bootstrap-social/

This is what Lando has currently

PR #1629, opened 2020-09-25
Bug 1667321 - remove hardcoded date ranges from Release Tracking Report

jcristau:
These haven't been updated in 4+ years, so just keep the custom ranges.

This also removes the "Anytime" range. I'm not sure if that's being used
in practice (it was at the bottom of the long hardcoded list so I
personally never even noticed it before looking at the code), so that
may or may not be OK.


dklawren (approved):
lgtm. thanks! r=dkl

PR #1631, opened 2020-09-30
Bug 1668308 - Unauthenticated searches can generate massive cookies

dklawren:
Removed all use of BUGLIST cookie for non-authenticated users. Logged in users still use the DB for storing last buglist.cgi queries.


globau (approved):
lgtm

PR #1632, opened 2020-10-01
Bug 1668439 - fix lightbox image height when page is scrolled

globau:


dklawren (approved):
Works correctly. r=dkl

PR #1634, opened 2020-10-05
Bug 1665646 - Make it clear when a bug's description (comment 0) has multiple authors

globau:
- change `$comment->get_activity` so the cached value is independent from the `activity_sort_order` argument
- add `$comment->update_text` to remove duplication of comment text updating code
- add `editors` and `edited` to bug_modal's activity stream
- teach bug_modal how to render a comment with multiple authors
- add hook to bug_modal's activity stream to allow other extensions to alter the stream before rendering
- update edit_comments extension to alter bug_model's activity stream when comment 0 has been edited by more than one visible user
- add a "multiple user" icon to gravator


extensions/EditComments/Extension.pm
```diff
@@ -131,72 +133,112 @@ sub _comment_is_editable_by {
...
 
-  if ($activity_sort_order eq "oldest_to_newest") {
-    @activity = reverse @activity;
-  }
+  # Update comment object
+  $self->{thetext} = $new_text;
 
-  $self->{'activity'} = \@activity;
+  # Clear memcached entry of comment 0 authors
+  print STDERR "CLEAR CACHE c0:$comment_id\n";
```
dklawren, line 237: Remove debugging line.

dklawren:
Overall the code looks fine. Just one thing that is probably not an issue. When a reporter edits their own comment twice, it show "Multiple Authors" and it shows the same gravatar twice. 

And related if, if a different user (not the author) edits the comment, and then the author edits it again, you see "Multple Authors" correctly and then the gravatar of the author, the gravatar of the other user and then another gravatar of the author again. Could we a) Separate in some way the gravatar of the original author from the ones that are editors, and b) If the original author edits their comment in the future, maybe skip displaying their avatar again?

Otherwise no issues and seems to all work correctly.

dklawren (approved):
Works as expected. r=dkl

PR #1635, opened 2020-10-05
Bug 1669330 - Should not be possible to override bug type requirement when submitting a bug using API

dklawren:
We are currently blocking from creating/updating a bug via API/UI but not if the bug_type is '--' which in the DB is a legal value. What we mean when we set the 'require_bug_type' data/params setting, we do not want '--' to be allowed. Now enforcing this properly.


t/report-ping-simple.t
```diff
@@ -1,3 +1,4 @@
+
```
globau, line 1: accidental blank line

globau (approved):

PR #1638, opened 2020-10-08
Bug 1670092 - Update CircleCI configuration for BMO to follow new guidelines for downloading Docker Hub images

dklawren:
Only build image once and then store the cached layers for use by each job since they cannot be shared otherwise.
Login to docker hub for each job before pulling any images (currently on works for commits to master).


.circleci/config.yml
```diff
@@ -104,24 +117,37 @@ jobs:
...
       - checkout
-      - *build_image
+      - *docker_login
       - attach_workspace:
           at: /app/build_info
       - deploy:
           command: |
             [[ -n "$DOCKERHUB_REPO" && -n "$DOCKER_USER" && -n "$DOCKER_PASS" ]] || exit 0
-            docker login -u "$DOCKER_USER" -p "$DOCKER_PASS"
+            echo "${DOCKER_PASS}" | docker login -u="${DOCKER_USER}" --password-stdin
```
globau, line 150: Shouldn't we already be logged in at this point due to the `docker_login` step?

globau (approved):

PR #1640, opened 2020-10-12
Bug 1670574 - The lightbox close button should be visible.

globau:
- reposition the lightbox's close button out from under the header


dklawren:
Works as expected. r=dkl

dklawren (approved):

PR #1646, opened 2020-10-21
Bug 1670319 - Allow use of OAuth2 bearer token as authentication for API calls same as API keys

dklawren:
- Added new api:modify scope that a user must agree to in order to use an access token as an API key.
- Updated the mojo-oauth2.t test to properly test the different scenarios.
- Also fixed a bug in the original implementation of OAuth2 which wasn't harmful as we only had one possible scope til now.


Bugzilla/Auth/Login/APIKey.pm
```diff
@@ -39,36 +38,57 @@ sub app_id {
...
-  my $api_key   = Bugzilla::User::APIKey->new({name => $api_key_text});
-  my $remote_ip = remote_ip();
+  # First check for an API key in the header or passed as query params
+  my $api_key_text = trim($cgi->http('X_BUGZILLA_API_KEY'))
+    || trim($params->{'Bugzilla_api_key'});
 
-  if (!$api_key or $api_key->api_key ne $api_key_text) {
+  # Also allow use of OAuth2 bearer tokens to access the API
+  # substr() removes the "Bearer " portion of the header
+  my $oauth_token = trim($cgi->http('Authorization')) || '';
```
globau, line 51: `//` would be clearer here to indicate that it's to avoid `undef`.

Bugzilla/Auth/Login/APIKey.pm
```diff
@@ -39,36 +38,57 @@ sub app_id {
...
+  $oauth_token = substr($oauth_token, 7) if $oauth_token;
```
globau, line 52: I'm not sure why this is happening as all the code does is check if this header was present; the contents of `$oauth_token` is ignored here.

If you want to retain the substringing, the following would be clearer and remove the need to document what the substr call is achieving:

`$oauth_token =~ s/^Bearer //i if $oauth_token;`

Bugzilla/Auth/Login/APIKey.pm
```diff
@@ -39,36 +38,57 @@ sub app_id {
...
+
+    $api_key->update_last_used($remote_ip);
+    $self->set_app_id($api_key->app_id);
+
+    return {user_id => $api_key->user_id};
   }
 
-  $api_key->update_last_used($remote_ip);
-  $self->set_app_id($api_key->app_id);
+  if ($oauth_token) {
```
globau, line 83: At first glance there doesn't appear to be any validation of `$oauth_token`.

Please add a comment pointing to `Bugzilla::App::Plugin::OAuth2` where the validation occurs.

t/mojo-oauth2.t
```diff
@@ -47,129 +54,151 @@ $t->ua->max_redirects(1);
...
-$t->post_ok(
-  '/oauth/access_token' => {Referer => $referer} => form => {
-    client_id     => $oauth_client->{client_id},
-    client_secret => $oauth_client->{secret},
-    code          => $auth_code,
-    grant_type    => 'authorization_code',
-    redirect_uri  => '/oauth/redirect',
-  }
-)->status_is(400)->json_is('/error' => 'invalid_grant');
+    {Authorization => 'Bearer ' . $access_data->{access_token}})->status_is(401);
```
globau, line 102: I'd like to see a test where you pass in garbage as an access-token to ensure that fails correctly.

globau (changes requested):

extensions/Gravatar/Extension.pm
```diff
@@ -27,7 +27,9 @@ BEGIN {
 
 sub _user_gravatar {
   my ($self, $size) = @_;
-  if ($self->setting('show_my_gravatar') eq 'Off') {
+  if ( $self->setting('show_my_gravatar')
+    && $self->setting('show_my_gravatar') eq 'Off')
+  {
```
globau, line 32: Remove unrelated change.

extensions/Review/Extension.pm
```diff
@@ -125,9 +125,12 @@ sub _user_review_count {
 }
 
 sub _user_reviews_blocked {
-  return $_[0]->settings->{block_reviews}->{value} eq 'on';
+  return
+    exists $_[0]->settings->{block_reviews}
+    && $_[0]->settings->{block_reviews}->{value} eq 'on';
```
globau, line 130: Remove unrelated change.

globau (approved):

PR #1647, opened 2020-10-25
Bug 1673064 - BMO phabbugz server errors when a revision has a package owner set as a reviewer

dklawren:


extensions/PhabBugz/lib/Revision.pm
```diff
@@ -293,6 +293,7 @@ sub _build_reviews {
   my %by_phid = map { $_->{reviewerPHID} => $_ } @{$self->reviewers_raw};
   my @users;
   foreach my $phid (keys %by_phid) {
+    next if $phid !~ /^PHID-PROJ/ && $phid !~ /^PHID-USER/;
```
globau, line 296: It would be clearer to remove this and just change the unconditional `else` to check for `/^PHID-USER/`;

globau (changes requested):

globau (approved):

PR #1653, opened 2020-11-01
Bug 1671262 - Forbid reporters to edit the priority, severity and type fields unless they have editbugs privs

dklawren:


extensions/BMO/Extension.pm
```diff
@@ -687,6 +687,11 @@ sub bug_check_can_change_field {
       }
     }
   }
+  elsif (($field eq 'bug_severity' || $field eq 'priority')
+    && !$user->in_group('editbugs'))
+  {
+    push(@$priv_results, PRIVILEGES_REQUIRED_EMPOWERED);
```
globau, line 693: This change doesn't prevent the `type` field from changing.

Personally I agree that the type field should be editable by the reporter; however, the bug reporter and commit description says otherwise.  Check with Jon to confirm that the type field should be here, and either update the code or the commit description.

This branch will be skipped if the current user is in `canconfirm` (likewise for the reopening of old bugs check).  Perhaps these `elsif`s should be just `if` blocks to avoid that footgun?

Also, there should be a comment making it clear what's going on here and for parity with the other checks.

globau (changes requested):

extensions/BMO/Extension.pm
dklawren: > 
> 
> This change doesn't prevent the `type` field from changing.
> 
> Personally I agree that the type field should be editable by the reporter; however, the bug reporter and commit description says otherwise. Check with Jon to confirm that the type field should be here, and either update the code or the commit description.
> 

Yes I also thought that and left that part of the request out. I was going to update the bug to change the specs but didn't yet. Some products we require the reporter to choose for themselves instead of setting a default value.

> This branch will be skipped if the current user is in `canconfirm` (likewise for the reopening of old bugs check). Perhaps these `elsif`s should be just `if` blocks to avoid that footgun?
> 
> Also, there should be a comment making it clear what's going on here and for parity with the other checks.

I will look after these suggested changes

globau: Looks like this has stalled.

Let's proceed with as per the bug and prevent the reporter from _editing_ the type, priority, and severity fields.

> Some products we require the reporter to choose [the type] for themselves instead of setting a default value.

Reporters should still be able to set those fields at bug creation time; the goal is prevent new users from modifying them after the bug has been created.

dklawren:
Closing due to substantial bit-rot. The PR will need to be redone to support the new way of implementing Mozilla rules.

PR #1655, opened 2020-11-10
Bug 1675625 - Refactor error handing in Mojolicious to allow for native mojo code to something other than Bugzilla::Error

dklawren:


Bugzilla/App/API.pm
```diff
@@ -24,13 +27,27 @@ sub setup_routes {
   push @$namespaces, 'Bugzilla::API';
   $r->namespaces($namespaces);
 
+  # Backwards compat with /api/user/profile which Phabricator
```
globau, line 30: incomplete comment

Bugzilla/API/V1/User.pm
```diff
@@ -13,8 +13,7 @@ use Mojo::JSON qw( true false );
 
 sub setup_routes {
   my ($class, $r) = @_;
-  $r->get('/api/user/profile')->to('V1::User#user_profile');
-  $r->get('/rest/user/profile')->to('V1::User#user_profile');
+  $r->get('/profile')->to('V1::User#user_profile');
```
globau, line 16: Won't this end up being `/rest/profile` instead of `/rest/user/profile`?

globau (changes requested):
You'll also need to update the routes in `Bugzilla/API/V1/Teams.pm`.

Bugzilla/API/V1/User.pm
dklawren: Yeah this is tricky. Having this as /rest/user/profile conflicts with Bugzilla::WebService::User::get in that the older code thinks 'profile' is a user name to return data for and instead errors that 'profile' cannot be found. With #1657 I will be able to properly fix this but for now I have renamed the endpoint to '/rest/user_profile'.

globau (approved):

PR #1656, opened 2020-11-11
Bug 1673948 - Add "Team Name" as a search column

globau:
Will not work without configuring the new `report_component_teams` parameter.

### DIFF
--- a/Bugzilla/API/V1/Teams.pm
+++ b/Bugzilla/API/V1/Teams.pm
@@ -7,6 +7,7 @@
 
 package Bugzilla::API::V1::Teams;
 use 5.10.1;
+use Bugzilla::Constants;
 use Mojo::Base qw( Mojolicious::Controller );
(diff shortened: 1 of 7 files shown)


Bugzilla/API/V1/Teams.pm
```diff
@@ -0,0 +1,33 @@
...
+  $r->get('/rest/config/component_security_teams')->to('V1::Teams#component_security_teams');
+}
+
+sub component_teams {
+  my ($self) = @_;
+  $self->render(
+    json => decode_json(Bugzilla->params->{report_component_teams})
+  );
+}
+
```
dklawren, line 25: Probably would want these to be authentication required API calls.

dklawren (approved):
Otherwise looks good. r=dkl

dklawren (approved):
r=dkl

PR #1661, opened 2020-11-17
Bug 1677419 - Token cancellation and verification should take an extra step to confirm

dklawren:


globau (changes requested):
This looks good, but only covers the cancellation case.

dklawren:
Ah right. It was in my head that people were mostly complaining about Outlook, etc. visiting the cancellation link when they wanted to accept the change instead. I will update to add a confirmation page for accepting the change as well.

dklawren:
> This looks good, but only covers the cancellation case.

Actually looking into this more, you do not need the confirmation page for 'confirming the email change' as the user already gets a form where they have to enter the old email address before the change is actually made. So we should be good with the current code.

globau (approved):

PR #1663, opened 2020-11-18
Bug 1678128 - Additional CORS headers are missing in the new versioned MOJO based REST API

dklawren:
Missed adding CORS headers for new API endpoints. Added _insert_rest_headers() function. This will not be needed in as many places once everything resides under a single /rest path.

### DIFF
--- a/qa/t/rest_bugzilla.t
+++ b/qa/t/rest_bugzilla.t
@@ -29,6 +29,11 @@ $t->ua->max_redirects(1);
 # Check for version returned
 $t->get_ok($url . 'rest/version')->status_is(200)->json_has('/version');
 
+# Change for proper CORS headers
+my $headers = $t->tx->res->headers;
+ok($headers->header('Access-Control-Allow-Origin') eq '*');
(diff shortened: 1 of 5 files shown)


qa/t/rest_bugzilla.t
```diff
@@ -29,6 +29,11 @@ $t->ua->max_redirects(1);
 # Check for version returned
 $t->get_ok($url . 'rest/version')->status_is(200)->json_has('/version');
 
+# Change for proper CORS headers
```
globau, line 32: s/Change/Check/

qa/t/rest_bugzilla.t
```diff
@@ -29,6 +29,11 @@ $t->ua->max_redirects(1);
...
+my $headers = $t->tx->res->headers;
+ok($headers->header('Access-Control-Allow-Origin') eq '*');
+ok($headers->header('Access-control-allow-headers') =~ /x-bugzilla-api-key/);
```
globau, line 35: nit: inconsistent capitalisation of header name

globau (approved):

PR #1674, opened 2020-12-07
Bug 1677416 - Use inputmode attribute for MFA.

makotokato:
Actually, BMO on mobile uses `<input type="number">` to input verify code by MFA. This is incorrect to input it. `type=number` won't input "0" as head character.

Best practice for 2FA is `type="tel"` (https://web.dev/sign-in-form-best-practices/) or use `inputmode` attribute. All modern mobile browsers (inc. Fenix) support `inputmode` attribute, so we should use `inputmode` instead of `type=tel` or `type=number`.


dklawren (approved):
Changes look good. Thanks for the contribution!

PR #1678, opened 2020-12-17
Bug 1682221 - Change Team tracking from the JSON params to a component field

dklawren:


template/en/default/admin/components/edit-common.html.tmpl
```diff
@@ -94,6 +94,11 @@
        %]
     </td>
 </tr>
+<tr>
+  <th class="field_label"><label for="team_name">Team Name:</label></th>
+  <td><input size="64" maxlength="64" name="team_name" id="team_name"
+             value="[%- comp.team_name || 'Mozilla' FILTER html %]"></td>
```
globau, line 100: Please make this an autocomplete field pulling from the list of current team names (we already use `devbridgeAutocomplete` in a few places for this).

That'll make it easier to avoid typos when using the same name in multiple components.

globau (changes requested):
The following items from the bug don't appear to have been addressed:
* add an API call to return a list of all team names
* add an API call to return a list of all components for a given team name
* ensure the team name is visible on editcomponents.cgi lists (ie. `admin/components/list.html.tmpl`)
* change the security bug report's configuration to just a list of team names

Bugzilla/API/V1/Teams.pm
```diff
@@ -7,29 +7,26 @@
...
-    || return $self->user_error('invalid_username');
-  $self->render(json => decode_json(Bugzilla->params->{report_component_teams}));
-}
-
-sub component_security_teams {
-  my ($self) = @_;
-  $self->bugzilla->login(LOGIN_REQUIRED)
-    || return $self->user_error('invalid_username');
-  $self->render(json => decode_json(Bugzilla->params->{report_secbugs_teams}));
+  $self->bugzilla->login();
```
globau, line 21: There's nothing confidential in team names (note `/rest/configuration` doesn't require authentication); let's make this public.

Bugzilla/Teams.pm
```diff
@@ -86,6 +91,45 @@ sub check_value {
   return $json_structure->check($value) ? '' : 'Invalid structure';
 }
 
+sub team_names {
+  return Bugzilla->dbh->selectcol_arrayref(
+    "SELECT DISTINCT team_name FROM components WHERE team_name != 'Mozilla' ORDER BY team_name"
```
globau, line 96: `Mozilla` is a valid team name and shouldn't be excluded.

Bugzilla/Teams.pm
```diff
@@ -86,6 +91,45 @@ sub check_value {
...
+  my $query = "
+    SELECT products.name AS product,
+           components.name AS component,
+           components.team_name AS team
+      FROM components INNER JOIN products ON components.product_id = products.id";
+  if (@team_names) {
+    $query .= ' WHERE components.team_name IN (' . join(',', ('?') x @team_names) . ')';
+  }
+  else {
+    $query .= " WHERE components.name != 'Mozilla'";
```
globau, line 113: `Mozilla` is a valid team name and shouldn't be excluded.

scripts/migrate-component-teams.pl
```diff
@@ -0,0 +1,70 @@
...
+          Bugzilla->input_params->{watch_user} = $component_obj->watch_user->login;
+          $component_obj->set_team_name($team_name);
+          $component_obj->update();
+        }
+      }
+    }
+  }
+}
+
+$dbh->do("UPDATE components SET team_name = 'Mozilla' WHERE team_name IS NULL");
```
globau, line 68: This isn't required.

scripts/secbugsreport.pl
```diff
@@ -34,6 +34,25 @@
...
+  'Mobile'
+  'Networking',
+  'Other',
+  'Platform',
+  'Pocket and User Journey',
+  'Privacy and Security',
+  'Services',
+  'Web Extensions',
+];
+
```
globau, line 55: This should be a parameter, not a constant.

template/en/default/admin/components/edit-common.html.tmpl
globau: As you noted this still needs autocomplete.

globau: Might be better to default to an empty value to force thought around selecting the correct team name.

Bugzilla/Teams.pm
```diff
@@ -24,6 +26,7 @@ use Try::Tiny qw(try catch);
 use Type::Utils;
 use Types::Standard qw(:types);
 
+# DEPRECATED: Will be removed in future release
```
globau, line 29: As far as I can tell these methods won't be used with this change; should be safe to delete them now rather than at a later date.

Bugzilla/Teams.pm
```diff
@@ -86,6 +91,45 @@ sub check_value {
...
+    my $product   = $row->{product};
+    my $component = $row->{component};
+    my $team      = $row->{team};
+    next if !Bugzilla->user->can_see_product($product);
+    $teams->{$team} ||= {};
+    $teams->{$team}->{$product} ||= [];
+    push @{$teams->{$team}->{$product}}, $component;
+  }
+
+  return $teams;
```
globau, line 130: Lets sort the result, either in SQL or Perl.

globau (changes requested):

dklawren:
> There's nothing confidential in team names (note `/rest/configuration` doesn't require authentication); let's make this public. 

This is an optional login so we filter in or out private products in Bugzilla::Team::get_team_info(). If someone passes a valid API key they may want to see all product components that they have access to.

dklawren:
> As far as I can tell these methods won't be used with this change; should be safe to delete them now rather than at a later date. 

I left them in because the migrate-component-teams.pl needs one or the other to take the old JSON and update the component table values. Then we could remove methods and one of the config parameters in a later PR as it will then no longer be needed. Otherwise I will need to copy the JSON into the migration script __DATA__ and use it from there. As long as it doesn't change from now til then.

dklawren:
> > As far as I can tell these methods won't be used with this change; should be safe to delete them now rather than at a later date.
> 
> I left them in because the migrate-component-teams.pl needs one or the other to take the old JSON and update the component table values. Then we could remove methods and one of the config parameters in a later PR as it will then no longer be needed. Otherwise I will need to copy the JSON into the migration script **DATA** and use it from there. As long as it doesn't change from now til then.

I will remove them after all with the next revision. We can leave the old JSON data in report_secbugs_teams, run the migration script, and then update the param value to the new single team per line format. This way a second PR to remove code will not be needed.

globau (approved):

globau (approved):

PR #1680, opened 2021-01-04
Bug 1612286 - Single sign-on through Mozilla IAM

dklawren:


Bugzilla/Util.pm
```diff
@@ -971,6 +971,20 @@ sub extract_nicks {
   return grep { defined $_ } @nicks;
 }
 
+sub mojo_user_agent {
+  my $ua = Mojo::UserAgent->new(
+    request_timeout     => 30,
+    connect_timeout     => 5,
+    inactivity_timesout => 30
```
globau, line 978: set the user-agent here so we identify as bugzilla

globau (approved):
rubberstamp for deployment to dev

PR #1685, opened 2021-01-26
Bug 1682600 - Webhook Event Change Uses Alternate naming from REST API

dklawren:


extensions/Push/lib/Serialize.pm
```diff
@@ -11,13 +11,17 @@ use 5.10.1;
...
 
+use Bugzilla::Bug;
 use Bugzilla::Constants;
 use Bugzilla::Extension::Push::Util;
 use Bugzilla::Version;
 
 use Scalar::Util 'blessed';
 use JSON ();
 
+my %translated_field_names = reverse %{Bugzilla::Bug::FIELD_MAP()};
```
globau, line 22: `reverse` isn't necessary here

globau (approved):

extensions/Push/lib/Serialize.pm
dklawren: It is necessary for outgoing data. For incoming data you would not need to reverse which is true.

PR #1693, opened 2021-02-11
Bug 1692222 - Show bug ids in reference fields when user has editbugs even if the user cannot see the actual bugs

dklawren:
* Also removed the update_token test from a couple of the webservice tests as the update_token is (has never been?) not used anyway.

### DIFF
--- a/Bugzilla/Bug.pm
+++ b/Bugzilla/Bug.pm
@@ -4536,8 +4536,9 @@ sub bug_alias_to_id {
 sub list_relationship {
   my ($table, $my_field, $target_field, $bug_id, $exclude_resolved) = @_;
   my $cache = Bugzilla->request_cache->{"bug_$table"} ||= {};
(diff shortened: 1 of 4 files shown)


globau:
As per our 1:1, this is just a discussion point, not an actual request.

Lafi0 (approved):

Lafi0 (approved):

Lafi0 (approved):

Lafi0 (approved):

Lafi0 (approved):

Lafi0 (approved):

Lafi0 (approved):

dveditz:
@dklawren  Since this PR was written you fixed https://bugzilla.mozilla.org/show_bug.cgi?id=1628759 which also hides inaccessible "See Also" references. I think we need a new patch that has code to expose those, too?

PR #1698, opened 2021-02-23
Bug 1693860 - [WEBHOOKS] The Webhook push connector needs to return more details when an error occurs instead of just the code

dklawren:
- Also update line in Extension.pm sub user_preferences to always set $push->set_config_last_modified() when changes are made to user webhooks as some cases were being missed.

### DIFF
--- a/extensions/Push/lib/Connector/Webhook.pm
+++ b/extensions/Push/lib/Connector/Webhook.pm
@@ -157,7 +157,7 @@ sub send {
 # Private methods
 
 sub _user_agent {
-  my $ua = Mojo::UserAgent->new(request_timeout => 30);
+  my $ua = Mojo::UserAgent->new(request_timeout => 30, connect_timeout => 5);
   if (my $proxy = Bugzilla->params->{proxy_url}) {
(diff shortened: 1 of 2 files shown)


extensions/Push/lib/Connector/Webhook.pm
```diff
@@ -155,19 +157,13 @@ sub send {
...
 sub _user_agent {
-  my ($self) = @_;
-
-  my $ua = LWP::UserAgent->new(agent => 'Bugzilla');
-  $ua->timeout(10);
-  $ua->protocols_allowed(['http', 'https']);
-
-  if (my $proxy_url = Bugzilla->params->{proxy_url}) {
-    $ua->proxy(['http', 'https'], $proxy_url);
+  my $ua = Mojo::UserAgent->new(request_timeout => 30);
```
globau, line 160: Lets also set `connect_timeout` to a low value here (eg. 5 seconds) to avoid waiting up to 30 seconds for the initial connection to a broken server/route.

globau (changes requested):

globau (approved):

PR #1704, opened 2021-03-10
Bug 1697642 - Record login method in login cookies table to see how users are authenticating

dklawren:


Bugzilla/Install/DB.pm
```diff
@@ -822,6 +822,9 @@ sub update_table_definitions {
   $dbh->bz_add_column('components', 'team_name',
     {TYPE => 'varchar(255)', NOTNULL => 1, DEFAULT => "'Mozilla'"});
 
+  # Bug XXXX - dkl@mozilla.com
```
globau, line 825: missing bug number

Bugzilla/Auth/Persist/Cookie.pm
```diff
@@ -39,8 +39,11 @@ sub persist_login {
 
   my $ip_addr = remote_ip();
 
-  $dbh->do('INSERT INTO logincookies (cookie, userid, ipaddr, lastused)
-    VALUES (?, ?, ?, NOW())', undef, $login_cookie, $user->id, $ip_addr);
+  # Record the auth method the user used to get the new cookie
+  my $auth_type = ref $user->authorizer->{_info_getter}->{successful} || 'Unknown';
```
globau, line 43: You should be calling `$user->authorizer->successful_info_getter` instead of reaching into a private attribute.

Better yet to avoid repetition with `Bugzilla::Auth::Persist::Cookie`, add an `authentication_method` sub to `Bugzilla::User` which does the same.  Note `Bugzilla::Auth::Persist::Cookie` wraps its logic in `eval`, so it's worth doing that here too.

globau: Looking at the code it appears that `->{successful}` contains an object; it would be better to add a short name (eg. max 8 chars) to each `Bugzilla::Auth::Login` package and persist that.

Bugzilla/Install/DB.pm
```diff
@@ -822,6 +822,9 @@ sub update_table_definitions {
...
+  $dbh->bz_add_column('logincookies', 'auth_type',{TYPE => 'varchar(256)'});
```
globau, line 826: `256` is very large for our needs; after switching to storing the login provider's name update this column's size accordingly (here and in the schema def).

globau (changes requested):

globau (approved):

PR #1711, opened 2021-04-26
Bug 1705743 - Make frequently reported bugs more obvious for enter bug form and duplicates.cgi more intuitive

dklawren:


template/en/default/reports/duplicates-table.html.tmpl
```diff
@@ -44,7 +44,7 @@
     { name => "id", description => "$terms.Bug #" },
     { name => "count", description => "Dupe<br>Count" },
     { name => "delta",
-      description => "Change in last<br>$changedsince day(s)" },
+      description => "Duped in last<br>$changedsince day(s)" },
```
globau, line 47: The changes from `Change` to `Duped`, here and in the form, aren't accurate.

The field we're showing is the last change date.  With the default query now restricted to open bugs only, the table will incorrectly indicate that the bugs have been duplicated.  If a bug was updated (eg. with a comment) after it was marked as a duplicate then the date will also be incorrect.

Revert back to the original description of this column.

globau (changes requested):

template/en/default/reports/duplicates-table.html.tmpl
dklawren: > 
> 
> The changes from `Change` to `Duped`, here and in the form, aren't accurate.
> 
> The field we're showing is the last change date. With the default query now restricted to open bugs only, the table will incorrectly indicate that the bugs have been duplicated. If a bug was updated (eg. with a comment) after it was marked as a duplicate then the date will also be incorrect.
> 
> Revert back to the original description of this column.

Unfortunately this is not the case. The changed since value is for any bugs that have had their resolution changed to DUPLICATE in the span of time, not *any* change to the bug.

https://github.com/mozilla-bteam/bmo/blob/master/duplicates.cgi#L172-L184

This is why I changed the description so that it better explained what kind of change it was talking about.

globau: Ah - updating the description makes sense then, thank you.

globau (approved):

PR #1713, opened 2021-04-27
Bug 1708052 - Create edittriageowners group allowing specific users to edit triage owners of components but nothing else

dklawren:
Details:

- Create new edittriageowners system group and wrap the triage owner related fields with the new permission.
- Fixed work around for input_params and the ComponentWatching extension.
  - Originally ComponentWatching relied on input_params for setting watch_user from editcomponents.cgi. Since a person who only has edittriageowners cannot edit watch_user, an error would be generated on update. 
- Also fixed some issues related to new bug description templates
  - The way bug description templates were implemented, if a product has template defined and a component also had a template defined, the product template would always win out which was not as designed.

### DIFF
--- a/editcomponents.cgi
+++ b/editcomponents.cgi
@@ -21,6 +21,8 @@ use Bugzilla::Component;
 use Bugzilla::Teams qw(team_names);
 use Bugzilla::Token;
 
+use List::Util qw(any);
+
 my $cgi      = Bugzilla->cgi;
 my $template = Bugzilla->template;
 my $vars     = {};
@@ -29,27 +31,30 @@ my $vars     = {};
 # so all actions point to the same page.
 $vars->{'doc_section'} = 'components.html';
 
-#
-# Preliminary checks:
-#
-
-my $user = Bugzilla->login(LOGIN_REQUIRED);
-
 print $cgi->header();
 
-$user->in_group('editcomponents')
-  || scalar(@{$user->get_products_by_permission('editcomponents')})
-  || ThrowUserError("auth_failure",
-  {group => "editcomponents", action => "edit", object => "components"});
+#
+# Often used variables
+#
+my $user              = Bugzilla->login(LOGIN_REQUIRED);
+my $product_name      = trim($cgi->param('product')   || '');
+my $comp_name         = trim($cgi->param('component') || '');
+my $action            = trim($cgi->param('action')    || '');
+my $showbugcounts     = (defined $cgi->param('showbugcounts'));
+my $token             = $cgi->param('token');
+my $editcomponents    = $user->in_group('editcomponents');
+my $edittriageowners  = $editcomponents || $user->in_group('edittriageowners');
+my $editable_products = $user->get_products_by_permission('editcomponents');
 
 #
-# often used variables
+# Preliminary permission checks:
 #
-my $product_name = trim($cgi->param('product')   || '');
-my $comp_name    = trim($cgi->param('component') || '');
-my $action       = trim($cgi->param('action')    || '');
-my $showbugcounts = (defined $cgi->param('showbugcounts'));
-my $token         = $cgi->param('token');
+
+$editcomponents
+  || scalar @{$editable_products}
+  || $edittriageowners
+  || ThrowUserError("auth_failure",
+  {group => "editcomponents", action => "edit", object => "components"});
 
 #
 # product = '' -> Show nice list of products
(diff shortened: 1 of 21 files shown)


editcomponents.cgi
```diff
@@ -29,28 +31,31 @@ my $vars     = {};
 # so all actions point to the same page.
 $vars->{'doc_section'} = 'components.html';
 
+print $cgi->header();
+
 #
-# Preliminary checks:
+# Often used variables
 #
+my $user              = Bugzilla->login(LOGIN_REQUIRED);
```
globau, line 39: All other code calls `Bugzilla->login` before `$cgi->header`.

To avoid surprises please move this before the header is printed.

editcomponents.cgi
```diff
@@ -229,30 +244,42 @@ if ($action eq 'update') {
...
+    $component->set_default_assignee($default_assignee);
+    $component->set_default_qa_contact($default_qa_contact);
+    $component->set_team_name($team_name);
+    $component->set_cc_list(\@initial_cc);
+    $component->set_is_active($isactive);
+    $component->set_bug_description_template($bug_desc_template);
+  }
+
+  if ($edittriageowners) {
+    # This should really be fixed one day
```
globau, line 274: This comment isn't great, especially since it doesn't indicate what's wrong.

Since this is new code, why isn't today the day to fix it?  🙂

editcomponents.cgi
```diff
@@ -29,28 +31,31 @@ my $vars     = {};
...
+# Often used variables
 #
+my $user              = Bugzilla->login(LOGIN_REQUIRED);
+my $product_name      = trim($cgi->param('product')   || '');
+my $comp_name         = trim($cgi->param('component') || '');
+my $action            = trim($cgi->param('action')    || '');
+my $showbugcounts     = (defined $cgi->param('showbugcounts'));
+my $token             = $cgi->param('token');
+my $editcomponents    = $user->in_group('editcomponents');
+my $edittriageowners  = $editcomponents || $user->in_group('edittriageowners');
```
globau, line 46: Setting `edittriageowners` to true if `editcomponents` is true makes it harder to reason about relevant permissions, and is a footgun (as shown by this causing an issue later in the code).

Keep `edittriageowners` separate from `editcomponents` and explicitly check both if required.

editcomponents.cgi
```diff
@@ -229,30 +244,42 @@ if ($action eq 'update') {
...
+    $component->set_default_qa_contact($default_qa_contact);
+    $component->set_team_name($team_name);
+    $component->set_cc_list(\@initial_cc);
+    $component->set_is_active($isactive);
+    $component->set_bug_description_template($bug_desc_template);
+  }
+
+  if ($edittriageowners) {
+    # This should really be fixed one day
+    if (!$editcomponents) {
```
globau, line 275: Here's the `edittriageowners` footgun - right now it's impossible for `edittriageowners` to be true and `editcomponents` to be false.

extensions/ComponentWatching/Extension.pm
```diff
@@ -183,16 +185,22 @@ sub object_end_of_update {
   my $object     = $args->{object};
   my $old_object = $args->{old_object};
   my $changes    = $args->{changes};
+  my $user       = Bugzilla->user;
+
   return unless $object->isa('Bugzilla::Component');
+  return
+    unless $user->in_group('editcomponents')
+    || any { $_->id == $object->product->id }
+  @{$user->get_products_by_permission('editcomponents')};
```
globau, line 194: This doesn't appear to be related to the bug in question.

template/en/default/admin/components/edit.html.tmpl
```diff
@@ -42,36 +42,39 @@
...
   </table>
 
    <input type="hidden" name="action" value="update">
    <input type="hidden" name="componentold" value="[% comp.name FILTER html %]">
    <input type="hidden" name="product" value="[% product.name FILTER html %]">
    <input type="hidden" name="token" value="[% token FILTER html %]">
-   <input type="submit" value="Save Changes" id="update"> or <a
-        href="[% basepath FILTER none %]editcomponents.cgi?action=del&amp;product=
+   <input type="submit" value="Save Changes" id="update">
+   [% IF editcomponentsup %]
```
globau, line 73: `editcomponentsup` typo?

globau (changes requested):

globau (approved):

PR #1714, opened 2021-05-04
Bug 1690201 - Require a comment when changing the severity field

dklawren:
Not quite ready for review but wanted to ask for some guidance.

Mostly everything works except for the blocking of submission of the bug_modal form. With the event handlers added the .save-btn submit buttons, my 'submit' handler is never called some reason and I cannot figure out why. If I comment out  
https://github.com/mozilla-bteam/bmo/blob/master/extensions/BugModal/web/bug_modal.js#L689 then my submit handler is executed and the alert is displayed. But the submit still happens with either a blank comment of the comment template and that is not desired behavior. My handler returns false as I thought would work. 

I can't see any other 'submit' handlers that may be conflicting. I would assume that the click handler for the save-btn's which then call this.form.submit() would trigger my 'submit' handler and then fully stop the form submissions (#changeform).

Ideas?

### DIFF
--- a/duplicates.cgi
+++ b/duplicates.cgi
@@ -34,6 +34,7 @@ use constant DEFAULTS => {
   fully_exclude_status  => ['CLOSED'],
   partly_exclude_status => ['VERIFIED'],
   except_resolution     => ['INVALID', 'WONTFIX'],
+  openonly              => 'yes',
   changedsince          => 7,
   maxrows               => 20,
   sortby                => 'count',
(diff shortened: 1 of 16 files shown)


globau:
Please clear the review request until it's ready for review so it doesn't sit on my dashboard 🙂

I don't see anything that immediately jumps out to me, but it's been more than 5 years since I worked on this sort of code.  It would be best to ask an engineer on the team for help.

dklawren:
> 
> 
> Please clear the review request until it's ready for review so it doesn't sit on my dashboard 🙂

Since github doesnt have a feedback flag and I wasnt sure you would see the question if I just simple created the WIP PR, I did the review request. 

> I don't see anything that immediately jumps out to me, but it's been more than 5 years since I worked on this sort of code. It would be best to ask an engineer on the team for help.

K. I will see who I can get some help with debugging as I need to see why my event handler is failing to execute in the event chain. If I inspect the 'changeform' form DOM in the inspection tool, it shows that the event is properly attached to it. But putting in console.log() calls in the code, the handler is never executed and the form is always submitted.

PR #1715, opened 2021-05-13
Bug 1690203 - Automatically set priority to P1 when severity is set to S1

dklawren:


extensions/MozChangeField/template/en/default/hook/bug_modal/header-end.html.tmpl
```diff
@@ -0,0 +1,9 @@
+[%# This Source Code Form is subject to the terms of the Mozilla Public
+  # License, v. 2.0. If a copy of the MPL was not distributed with this
+  # file, You can obtain one at http://mozilla.org/MPL/2.0/.
+  #
+  # This Source Code Form is "Incompatible With Secondary Licenses", as
+  # defined by the Mozilla Public License, v. 2.0.
+  #%]
+
+[% javascript_urls.push("extensions/MozChangeField/web/js/severitys1priorityp1.js") %]
```
globau, line 9: That filename looks like soup, `severity-s1-priority-p1.js` would be much easier to read.

extensions/MozChangeField/lib/Post/SeverityS1PriorityP1.pm
```diff
@@ -0,0 +1,44 @@
...
+
+use Bugzilla::Field;
+
+sub evaluate_create {
+  my ($self, $args) = @_;
+  my $bug       = $args->{bug};
+  my $timestamp = $args->{timestamp};
+
+  if ($bug->bug_severity eq 'S1' && $bug->priority ne 'P1' && grep { $_ eq 'P1' }
+    @{get_legal_field_values('priority')})
```
globau, line 21: `grep` in a boolean context is inefficient; use `any` (see `PerlCritic/ProhibitBooleanGrep`)

globau (changes requested):
The follow item from the spec has not been implemented:
- set a tooltip explaining that the field value is locked due to the severity

extensions/MozChangeField/web/js/severitys1priorityp1.js
```diff
@@ -0,0 +1,59 @@
...
+   * Initialize a new SeverityS1PriorityP1 instance.
+   */
+  constructor() {
+    this.priority = document.querySelector("#priority");
+    this.severity = document.querySelector("#bug_severity");
+
+    if (this.severity && this.priority) {
+      this.severity_current_value = this.severity.value;
+      this.priority_current_value = this.priority.value;
+      this.severity.addEventListener("change", () => this.severity_onselect());
```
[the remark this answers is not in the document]
dklawren: Tried that initially and did not work as 'this' is not bound properly. Using => style functions explicitly binds 'this' based on context and worked properly.

globau (approved):

PR #1718, opened 2021-05-24
Bug 1712619 - [GITHUBAUTH] Use authorization header instead of access_token query parameter when retrieving email addresses

dklawren:


extensions/GitHubAuth/lib/Client.pm
```diff
@@ -95,14 +95,17 @@ sub get_access_token {
...
 sub get_user_emails {
   my ($self, $access_token) = @_;
-  my $uri = URI->new(GH_USER_EMAILS_URI);
-  $uri->query_form(access_token => $access_token);
 
-  my $response = $self->user_agent->get($uri, Accept => 'application/json');
+  $self->user_agent->default_header(
+    'Accept'        => 'application/json',
+    'Authorization' => "token $access_token"
+  );
```
globau, line 102: These headers should only be sent on this request, not on this request and all future calls.
Pass the headers to the UA's `get` call instead.

globau (changes requested):

PR #1725, opened 2021-06-16
Bug 1690204 - Automatically set `tracking-?` for the current nightly/beta releases for all S1 bugs

dklawren:


extensions/MozChangeField/lib/Post/SetTrackingSeverityS1.pm
```diff
@@ -0,0 +1,149 @@
...
+      $changes->{$flag_name} = [$old_value, '?'];
+
+      # Update the name/value pair in the bug object
+      $old_bug->{$flag_name} = '?';
+      $bug->{$flag_name} = '?';
+    }
+  }
+}
+
+sub _fetch_product_version_file {
```
globau, line 121: This is exactly the same code that exists in the BMO extension; move it into the Bugzilla namespace and call it from the BMO ext and here.

extensions/MozChangeField/lib/Post/SetTrackingSeverityS1.pm
```diff
@@ -0,0 +1,149 @@
...
+    Bugzilla->memcached->set_data({
+      key   => $key,
+      value => $versions,
+
+      # Cache for 30 minutes if the data is available, otherwise retry in 5 min
+      expires_in => $response->is_success ? 1800 : 300,
+    });
+  }
+
+  my ($nightly) = split /\./, $versions->{FIREFOX_NIGHTLY};
```
globau, line 144: This will throw errors if the values weren't available from product-details

extensions/MozChangeField/lib/Post/SetTrackingSeverityS1.pm
```diff
@@ -0,0 +1,149 @@
...
+  my $versions = Bugzilla->request_cache->{$key}
+    || Bugzilla->memcached->get_data({key => $key});
+
+  unless ($versions) {
+    my $ua = Mojo::UserAgent->new;
+    if (my $proxy_url = Bugzilla->params->{'proxy_url'}) {
+      $ua->proxy->http($proxy_url);
+    }
+
+    my $response = $ua->get(PD_ENDPOINT . $key . '.json')->result;
```
globau, line 132: Since this is a blocking call (something I don't like at all but I understand this is existing code) it should have a short timeout.

extensions/MozChangeField/lib/Post/SetTrackingSeverityS1.pm
```diff
@@ -0,0 +1,149 @@
...
+    $versions = Bugzilla->request_cache->{$key}
+      = $response->is_success ? decode_json($response->body) : {};
+    Bugzilla->memcached->set_data({
+      key   => $key,
+      value => $versions,
+
+      # Cache for 30 minutes if the data is available, otherwise retry in 5 min
+      expires_in => $response->is_success ? 1800 : 300,
```
globau, line 140: This data changes every 4 weeks, checking every 30 minutes seems excessive.

extensions/MozChangeField/web/js/set-tracking-severity-s1.js
```diff
@@ -0,0 +1,112 @@
...
+        )[0];
+        request_opt.disabled = false;
+        request_opt.selected = true;
+        flag.title = this.flag_curr_titles[flag.name];
+      }
+    });
+  }
+
+  // Below borrowed from extensions/BMO/web/js/firefox-crash-table.js
+  async fetchProductDetails() {
```
globau, line 85: we already have this information cached server-side; pass it into the page rather than re-fetching on every page load.

extensions/MozChangeField/web/js/set-tracking-severity-s1.js
```diff
@@ -0,0 +1,112 @@
...
+   * Initialize a new SetTrackingSeverityS1 instance.
+   */
+  constructor() {
+    this.priority = document.querySelector("#priority");
+    this.severity = document.querySelector("#bug_severity");
+    this.flags = document.querySelector("div.edit-show table.tracking-flags");
+
+    if (this.severity && this.priority && this.flags) {
+      this.sev_curr_value = this.severity.value;
+      //this.severity.addEventListener("change", () => this.severity_onselect());
```
globau, line 28: remove commented out code

globau (changes requested):

extensions/MozChangeField/web/js/set-tracking-severity-s1.js
dklawren: Yep. I first check for the existence of the meta[name="firefox-versions"] element which has the JSON already populated for show_bug.cgi. If that is not present, we then try to load it ourselves.

globau: We _must not_ duplicate the logic here.  If the data isn't available then fallback to sane defaults rather than attempting to fetch here.

globau (changes requested):

extensions/MozChangeField/web/js/set-tracking-severity-s1.js
dklawren: Sorry for the confusion here. I can definitely move this code from here and firefox-crash-table.js to the util.js file to remove duplication. But what would be sane defaults be in this case other than just returning an empty JSON object? I could not just return some arbitrary version numbers.

globau: This code needs to be removed; the data should _always_ come from the server.  If the perl code is unable to fetch the product details then attempting to do it client-side will only DDoS the server.

If product-details is down the best course of action here is to do nothing – not set tracking-? flags.

extensions/BMO/Extension.pm
```diff
@@ -2514,37 +2512,9 @@ sub _split_crash_signature {
...
-
-  return $versions;
-}
-
 sub _get_product_version {
   my ($product, $channel, $detail) = @_;
-  my $versions = _fetch_product_version_file($product);
+  my $versions = fetch_product_version_file($product);
   my $version = $versions->{PRODUCT_CHANNELS->{$product}->{$channel}->{json_key}};
   return $version if $detail;
```
globau, line 2519: The `return 0 unless $versions;` line should happen before this return.
Also, because `$versions` is a hashref, it'll always be true in this context, that line needs to be `return 0 unless %$versions;`

extensions/MozChangeField/lib/Post/SetTrackingSeverityS1.pm
```diff
@@ -0,0 +1,127 @@
...
+      $old_bug->{$flag_name} = '?';
+      $bug->{$flag_name}     = '?';
+    }
+  }
+}
+
+sub _fetch_nightly_beta_versions {
+  my $versions  = fetch_product_version_file('firefox');
+  my ($nightly) = split /\./, $versions->{FIREFOX_NIGHTLY};
+  my ($beta)    = split /\./, $versions->{LATEST_FIREFOX_RELEASED_DEVEL_VERSION};
```
globau, line 121: Both of these splits will generate `Use of uninitialized value` warnings if `$versions` is empty.

extensions/MozChangeField/web/js/set-tracking-severity-s1.js
```diff
@@ -0,0 +1,102 @@
...
+  /**
+   * Initialize a new SetTrackingSeverityS1 instance.
+   */
+  constructor() {
+    this.priority = document.querySelector("#priority");
+    this.severity = document.querySelector("#bug_severity");
+    this.flags = document.querySelector("div.edit-show table.tracking-flags");
+    this.firefox_versions = document.querySelector(
+      'meta[name="firefox-versions"]'
+    );
```
globau, line 27: Can you check that `[% IF firefox_versions %]` in Template Toolkit returns false if `firefox_versions` is an empty hashref (Perl returns true in this case)?

Bugzilla/Util.pm
```diff
@@ -971,6 +972,32 @@ sub extract_nicks {
   return grep { defined $_ } @nicks;
 }
 
+sub fetch_product_version_file {
```
globau, line 975: This name confuses me – what `file` is this returning?

globau (changes requested):

globau (approved):

PR #1727, opened 2021-06-17
Bug 1690201 - Require a comment when changing the severity field

dklawren:


extensions/MozChangeField/lib/Post/CommentOnSeverity.pm
```diff
@@ -0,0 +1,51 @@
...
+use 5.10.1;
+use Moo;
+
+use Bugzilla::Error;
+use Bugzilla::Util qw(trim);
+
+use constant DEFAULT_COMMENT =>
+  'Changing severity to S? because of <rationale>.';
+
+sub evaluate_create {
```
globau, line 19: A comment is only required when _changing_ the severity field, not during bug creation.

Remove all code relating to bug creation.

extensions/MozChangeField/web/js/comment-on-severity.js
```diff
@@ -0,0 +1,63 @@
...
+  }
+
+  /**
+   * Called when severity select is changed. Insert or remove the comment text
+   */
+  severity_onselect() {
+    // Set comment required warning
+    if (this.severity.value != this.curr_severity) {
+      this.severity_comment_required.innerHTML =
+        ' <a href="https://firefox-source-docs.mozilla.org/bug-mgmt/guides/severity.html" target="_blank">Comment Required</a>';
```
globau, line 42: This should be a transient top of page notification with the text from the bug.

ie. use `#floating-message`
- it might make sense to position that as `fixed` instead of `absolute` to ensure it's always on screen
- add a click handler that opens the severity guide link in a new window, removing it once the floating-message is no longer visible

globau (changes requested):

globau (changes requested):

extensions/BugModal/web/bug_modal.css
```diff
@@ -1072,6 +1072,10 @@ h3.change-name a {
   margin: 4px;
 }
 
+.floating-message-fixed {
+  position: fixed;
+}
+
```
globau, line 1078: I don't see where this is used.

globau (approved):

PR #194, opened 2017-08-03
Bug 1386336 - Generate robots.txt at checksetup.pl from a template

umohm1:

PR #1734, opened 2021-07-13
Bug 1720283 - bugzilla-dev.allizom.org pages found by Google search, should get excluded by robots.txt

dklawren:


extensions/SiteMapIndex/Extension.pm
```diff
@@ -88,8 +88,6 @@ sub install_before_final_checks {
     print STDERR get_text('sitemap_requirelogin'), "\n";
     return;
   }
-
-  return if (Bugzilla->localconfig->urlbase ne 'https://bugzilla.mozilla.org/');
```
globau, line 92: This whole sub is a no-op and should be deleted.

Prior to [Bug 1386336](https://github.com/mozilla-bteam/bmo/pull/194/files#diff-8dbb7ef93c864e1fe9c0cca78918404459e5b85d93bf4dfb221dc345d6852f8c) there was a call to `$self->_fix_robots_txt();` at the end of the sub.

template/en/default/robots.txt.tmpl
```diff
@@ -2,7 +2,8 @@ User-agent: *
 Disallow: [% basepath FILTER none %]
 Crawl-delay: 30
 
-[% IF NOT Bugzilla.localconfig.urlbase.matches("bugzilla-dev") %]
+[% USE Bugzilla %]
+[% IF NOT Bugzilla.localconfig.urlbase.match("bugzilla-dev") %]
```
globau, line 6: Lets make is so we only generate a permissive robots.txt on production, rather than just excluding bugzilla-dev.  That'll make sure that any transient/test instances we spin up in future won't be indexed.

globau (changes requested):

template/en/default/robots.txt.tmpl
```diff
@@ -2,7 +2,8 @@ User-agent: *
 Disallow: [% basepath FILTER none %]
 Crawl-delay: 30
 
-[% IF NOT Bugzilla.localconfig.urlbase.matches("bugzilla-dev") %]
+[% USE Bugzilla %]
+[% IF Bugzilla.localconfig.urlbase.match("bugzilla.mozilla.org") %]
```
globau, line 6: This should be an exact string match on production's `urlbase` rather than a substring regex match.

globau (changes requested):

globau (approved):

PR #1735, opened 2021-07-15
Bug 1612286 - Updated perl dependencies for IAM

dklawren:
* Added modules need by upcoming IAM support
* Updated Perl dependencies in general


Makefile.PL
```diff
@@ -80,6 +82,7 @@ my %requires = (
   'Mojo::JWT'                           => '0.07',
   'MojoX::Log::Log4perl'                => '0.04',
   'Mojolicious'                         => '8.42',
```
globau, line 84: This should be updated to reflect the major version bump.

Also need to update `MIN_PERL_VERSION` to at least match Mojo's (5.16.0 as of 8.50), although it might make sense to make it match what we do in docker.

globau (changes requested):

globau (approved):

PR #1743, opened 2021-07-29
Bug 1722985 - Update to latest Perl dependencies to support IAM

dklawren:
This is the second pass at this update as the previous had to be reverted due to causing some emails to be double encoded. The issue was the upstream Email::MIME module changed the format of the 'Content-type' header which our Mailer.pm code did not expect.

https://github.com/rjbs/Email-MIME/commit/e80d8a435dfe167a7883a6350f8c2dc4a8b5039f

Old header format:
`Content-Type: text/plain; charset="UTF-8"`

New header format:
`Content-Type: text/plain; charset=UTF-8`

So basically this section of code was always executing:
https://github.com/mozilla-bteam/bmo/blob/master/Bugzilla/Mailer.pm#L203-L218

I am including the simple email fix in this bug as well as the files related to the update Perl dependencies.


Bugzilla/Mailer.pm
```diff
@@ -198,12 +198,14 @@ sub MessageToMTA {
     my ($part) = @_;
     return if $part->parts > 1;    # Top-level
     my $content_type = $part->content_type || '';
-    $content_type =~ /charset=['"](.+)['"]/;
+    $content_type =~ /charset=(.+)/;
+    my $charset = $1;
+    $charset =~ s/['"]//g;
```
globau, line 203: This will throw an `uninitialized value` warning if the charset is not present.

Bugzilla/Mailer.pm
```diff
@@ -198,12 +198,14 @@ sub MessageToMTA {
     my ($part) = @_;
     return if $part->parts > 1;    # Top-level
     my $content_type = $part->content_type || '';
-    $content_type =~ /charset=['"](.+)['"]/;
+    $content_type =~ /charset=(.+)/;
```
globau, line 201: This is still fragile — there's no guarantee that `charset` will be the last part of the header.

Instead of reinventing the wheel, use [Email::MIME::ContentType](https://metacpan.org/pod/Email::MIME::ContentType) to parse the content-type.

It's currently pulled in by Email::MIME, but be sure to add it as an explicit dependency in case Email::MIME changes.

globau (changes requested):

Bugzilla/Template.pm
```diff
@@ -911,6 +911,16 @@ sub create {
...
+
+      # Fixes "Wide character in substitution (s///) at [..]/Template/Filters.pm line 62."
+        collapse => sub {
+          my $content = $_[0];
+          no warnings 'utf8';
+          $content =~ s/^\s+//;
+          $content =~ s/\s+$//;
+          $content =~ s/\s+/ /g;
+          return $content;
+        },
```
globau, line 923: Please file a new bug for this unrelated change (or morph bug 1639906).

globau (changes requested):

globau (approved):

PR #1744, opened 2021-08-03
Bug 1639906 - Fix Wide character (U+2192) in substitution error

dklawren:


Bugzilla/Template.pm
```diff
@@ -911,6 +911,16 @@ sub create {
...
       none => sub { return $_[0]; },
+
+      # Fixes "Wide character in substitution (s///) at [..]/Template/Filters.pm line 62."
+      collapse => sub {
+        my $content = $_[0];
+        no warnings 'utf8';
+        $content =~ s/^\s+//;
+        $content =~ s/\s+$//;
+        $content =~ s/\s+/ /g;
+        return $content;
```
globau, line 922: There's a reasonable perf hit in creating a variable here instead of working on `$_[0]` directly; given how frequently `collapse` is called we should avoid that.

This also has slightly different logic than `Template::Filters`, it would be best to mirror exactly what that does just with the `no warnings` addition:

```
sub {
    no warnings 'utf8';
    for ($_[0]) { s/^\s+//; s/\s+$//; s/\s+/ /g };
    $_[0]
}

globau (changes requested):

globau (approved):

PR #1745, opened 2021-08-05
Bug 1722612 - Add support for sending IP addresses of authentication failures and abusive behavior to the iprepd API for rate limiting

dklawren:


Bugzilla.pm
```diff
@@ -787,15 +787,36 @@ sub check_rate_limit {
...
-        "[rate_limit] action=$action, ip=$ip, limit=$limit, name=$name");
+        "[rate_limit] action=$action, ip=$ip, limit=$full_limit, name=$name");
       if ($action eq 'block') {
         request_cache->{mojo_controller}->block_ip($ip);
         $throw_error->();
       }
     }
   }
+
+  # Send information about this event to the iprepd API if active
```
globau, line 800: As discussed on slack informing iprepd about failed authentication attempts should not be part of the rate limiting logic.

Decouple this completely from rate limiting; add a method to report the violation to iprepd, and call that directly from the points in the code where authentication failed.

globau (changes requested):

Bugzilla.pm
```diff
@@ -787,15 +788,46 @@ sub check_rate_limit {
...
+      $params->{iprepd_base_url} . '/violations/type/ip/' . $ip,
+      {'Authorization' => 'Bearer ' . $params->{iprepd_client_secret}} => json =>
+        $payload
+    );
+    if ($tx->res->code != 200) {
+      die 'Expected HTTP 200, got '
+        . $tx->res->code . ' ('
+        . $tx->error->{message} . ') '
+        . $tx->res->body;
+    }
```
globau, line 824: It's now better to use `$tx->result` than digging into `$tx->res` directly.

Something like the following should work (untested):
```
my $res = $tx->result;
die $res->message . ' ' . $res->body unless $res->is_success;
```

globau (changes requested):
There's a few other places we should consider adding:
- Bugzilla::Auth::Login::APIKey when the wrong IP is provided on a sticky api-key, and when a revoked api-key is used
- Bugzilla::Auth::Login::Cookie where auth_invalid_token and invalid_cookies_or_token are thrown
- Bugzilla::Auth::Login::CGI where auth_untrusted_request is thrown

PR #1747, opened 2021-08-23
Bug 1727027 - bug is not a valid parameter for the Bugzilla::Comment::create function

dklawren:
When a comment is added it is not good idea to run $bug->update multiple times are it will try to add the same comment over again giving the duplicate key error. This change update Bugzilla::Bug::update to be able to be ran again and not try to re-add the comments.

### DIFF
--- a/Bugzilla/Bug.pm
+++ b/Bugzilla/Bug.pm
@@ -1281,20 +1281,22 @@ sub update {
   }
 
   # Comments and comment tags
-  foreach my $comment (@{$self->{added_comments} || []}) {
+  $self->{added_comments} ||= [];
+  while (my $comment = shift @{$self->{pending_comments} || []}) {
 
-    # Override the Comment's timestamp to be identical to the update
-    # timestamp.
+    # Override the Comment's timestamp to be identical to the update timestamp.
     $comment->{bug_when} = $delta_ts;
     $comment = Bugzilla::Comment->insert_create_data($comment);
     if ($comment->work_time) {
       LogActivityEntry($self->id, "work_time", "", $comment->work_time, $user->id,
         $delta_ts);
     }
-    foreach my $tag (@{$self->{added_comment_tags} || []}) {
+    foreach my $tag (@{$self->{pending_comment_tags} || []}) {
       $comment->add_tag($tag) if defined $tag;
     }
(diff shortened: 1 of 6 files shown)


Bugzilla/Bug.pm
```diff
@@ -1784,7 +1786,7 @@ sub _check_bug_status {
     @valid_statuses = @{$invocant->statuses_available};
     $product        = $invocant->product_obj;
     $old_status     = $invocant->status;
-    my $comments = $invocant->{added_comments} || [];
+    my $comments = $invocant->{pending_comments} || [];
```
globau, line 1789: If the `_check*` methods are called after the first `$bug->update()` then they will incorrectly report that no comments have been added as the comments would have been moved from `pending_comments` to `added_comments`.

globau (changes requested):
This approach is likely to be problematic as each call to `$bug->update()` happens in its own db transaction, and there's a few other places that assume that `added_comments` always contains the full set of comments.

It would be better to _prevent_ `$bug->update()` from being called multiple times, and instead ensure that all changes are performed within the single `update()` call and database transaction.  That'll avoid situations like what was mentioned in the reporting bug where a bug was partially updated.

globau (changes requested):
There are other `MozChangeField` handlers that also call `$bug->update`, these should be fixed too.

Also please add code to _prevent_ `$bug->update` from being called multiple times.

globau (changes requested):
`CrashKeywordSetSeverity.pm` is still calling `$bug->update()` – `evaluate_create()` is triggered within a bug's `create()` DB transation, I'm surprised that calling `$bug->update()` at this point works!  Best to change this to a direct DB update too.

The Voting ext calls `$bug->update()` from within the `bug_end_of_update` hook.

extensions/MozChangeField/lib/Post/SeverityS1PriorityP1.pm
```diff
@@ -24,8 +24,10 @@ sub evaluate_create {
     @{get_legal_field_values('priority')}
     )
   {
-    $bug->set_priority('P1');
-    $bug->update($timestamp);
+# Should call $bug->update here so set directly
```
globau, line 27: nit: bad indent

globau (approved):

PR #1752, opened 2021-09-07
Bug 1729495 - Update showdependencygraph.cgi to not create a png file in data/webdot and instead include as base64 data

dklawren:


template/en/default/bug/dependency-graph.html.tmpl
```diff
@@ -53,7 +53,7 @@
 </p>
 
 [% IF image_map %]
-  <img src="[% image_url %]" alt="Dependency graph" usemap="#imagemap">
+  <img src="data:image/png;base64,[% image_data FILTER html %]" alt="Dependency graph" usemap="#imagemap">
```
globau, line 56: There's no need to `FILTER html` here as we know the content is base64; let's do `FILTER none` to avoid the overhead.

showdependencygraph.cgi
```diff
@@ -267,31 +268,15 @@ if ($webdotbase =~ /^https?:/) {
 }
 else {
   # Local dot installation
+  my $image_data;
+  open my $pipe, '-|', "\"$webdotbase\" -Tpng $filename";
```
globau, line 272: Lets take this opportunity to switch to the open's safer pipe list form and add error handling:
```
open(my $pipe, '-|', $webdotbase, '-Tpng', $filename)
  || ThrowCodeError(...);
```

globau (changes requested):

globau (approved):

PR #1764, opened 2021-10-12
Bug 1696781 - Disable OAuth2 integration in createaccount.cgi, and enable it in token.cgi after creating account.

arai-a:
for https://bugzilla.mozilla.org/show_bug.cgi?id=1696781

This fixes 2 issues:

1) In the current code, the redirect for OAuth2 is disabled only for `token.cgi`.
That means, if the user arrives bugzilla without an account, they cannot create account in `createaccount.cgi`.

2) In the current code, the redirect for OAuth2 is disabled for `token.cgi`, even after creating account.
That means, the user cannot continue OAuth2 process until they click some link in the page.

This patch fixes the above by:
  1. disble OAuth2 redirect for `createaccount.cgi`
  2. enable Oauth2 redirect for `token.cgi` if the account is created

I used `$cgi->param('token_account_created')` to pass that information from `token.cgi` to `Bugzilla->login`.
If there's better way that won't break any other code, let me know.

### DIFF
--- a/Bugzilla.pm
+++ b/Bugzilla.pm
@@ -284,6 +284,14 @@ sub login {
   my $do_logout     = $cgi->param('logout');
   my $on_token_page = $script_name eq '/token.cgi';
 
+  my $is_creatingaccount = 0;
+  if ($script_name eq '/createaccount.cgi'
+    || ($on_token_page && !$cgi->param('token_account_created')))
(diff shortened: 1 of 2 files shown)


dklawren:
Thanks for your contribution! Merging.

PR #1765, opened 2021-10-13

dklawren:
* Migration from data/params to database without losing custom parameters
* Maybe create a Bugzilla::Config::Param module for doing operations on the params table instead of INSERT/SELECT/UPDATE directly in Bugzilla::Config


mitchhentges:
Can do, I'll take a look later today. :+1:

PR #1767, opened 2021-10-13
Bug 1659599 - [WEBHOOKS] Implement authentication token in Webhooks (schema only)

dklawren:


dklawren:
Merged without review as code was already reviewed as per pull request 1755

PR #1781, opened 2021-11-23
Bug 1742458 - cf_crash_signature doesn't always exist in some bugs

dklawren:


mitchhentges (approved):
The quick fix makes sense to me, though it might be worth adding some of the details from [this comment](https://bugzilla.mozilla.org/show_bug.cgi?id=1742458#c1) into the commit message itself.

PR #1782, opened 2021-11-23
Bug 1628759 - Hide bugs in See Also fields from users without access

dklawren:


dklawren:
Not yet ready for review. Want to do some more testing.

dklawren:
Ok ready for review. Verified that someone with out permission to see the bug referred to cannot remove the reference.

qa/t/webservice_bug_get_bugs.t
```diff
@@ -121,6 +120,14 @@ sub post_success {
...
   }
 
+  # See also to a private bug should not display for the public bug
+  if (!$is_private_bug && !$is_private_user) {
+    ok(
+      !exists $bug->{see_also} || !@{$bug->{see_also}},
+      'See also to a private bug should not display for the public bug and normal user'
+    );
+  }
+
```
globau, line 130: I'd like to see a few more tests here to cover the whole matrix: public/public, public/private, private/public, and private/private

qa/t/webservice_bug_get.t
```diff
@@ -124,6 +123,14 @@ sub post_success {
...
   }
 
+  # See also to a private bug should not display for the public bug
+  if (!$is_private_bug && !$is_private_user) {
+    ok(
+      !exists $bug->{see_also} || !@{$bug->{see_also}},
+      'See also to a private bug should not display for the public bug and normal user'
+    );
+  }
+
```
globau, line 133: I'd like to see a few more tests here to cover the whole matrix: public/public, public/private, private/public, and private/private

globau (changes requested):
Users should be prevented from adding or removing bugs they cannot see to/from see-also, both via the API and at bug creation time.

globau (approved):

PR #1785, opened 2021-12-08
Bug 1744995 - Client bug bounty form should set type defect instead of task

dklawren:


mitchhentges (approved):
Looks good!
It looks like there's 5 commits in this stack, even though the meat is in the last one? Unsure what's going on there, might be worth taking a look :smile:

zzzeid (approved):
lgtm - probably worth rebasing.

dklawren:
I merged my fork with the latest upstream commits and published those. Then I created a new worktree against the latest local master and this what I got when I pushed the branch to GH for review. It has happened before and have never quite understood why it happens. But the files changed and the diff always reflect the actual changes I made and want to be reviewed.

mitchhentges:
Let me know if you want to do a lil Zoom call in the future to sort out your local git repo, sounds like there's shenanigans afoot :)

PR #1791, opened 2022-01-14
Bug 1749479 - Add "regression" keyword and set "Has Regression Range" to "yes" on all bugs with a non-empty "Regressed By" field

dklawren:


scripts/bug1749479.pl
```diff
@@ -0,0 +1,106 @@
...
+$auto_user->{bless_groups} = [Bugzilla::Group->get_all];
+Bugzilla->set_user($auto_user);
+
+my $query = {
+  o1 => 'isnotempty',
+  f1 => 'regressed_by',
+  f2 => 'keywords',
+  o2 => 'notequals',
+  v2 => 'regression',
+};
```
globau, line 38: This won't set `cf_has_regression_range` if a bug has the `regression` keyword, but not `cf_has_regression_range`.  eg [Bug 1750444](https://bugzilla.mozilla.org/show_bug.cgi?id=1750444)

globau: This would be much easier to read in field-operator-value order.

scripts/bug1749479.pl
```diff
@@ -0,0 +1,106 @@
...
+my ($data) = $search->data;
+
+my $bug_count = @$data;
+if ($bug_count == 0) {
+  warn "There are no bugs to update.\n";
+  exit 1;
+}
+
+# if running from command line
+if (-t STDIN) {
```
globau, line 50: Isn't this a command line only script?

scripts/bug1749479.pl
```diff
@@ -0,0 +1,106 @@
...
+  # the current last modified timestamp.
+  if (!$regressed_by_ts) {
+    $regressed_by_ts = $bug->delta_ts;
+  }
+
+  my $field = Bugzilla::Field->new({name => 'cf_has_regression_range'});
+  $field || die 'Custom field cf_has_regression_range field does not exist';
+  $bug->set_custom_field($field, 'yes');
+
+  my $keyword = Bugzilla::Keyword->new({name => 'regression'});
```
globau, line 88: Do we have to create new `Bugzilla::Field` and `Bugzilla::Keyword` objects for each bug?
I'd expect the script to run much faster if we can reuse these objects.

scripts/bug1749479.pl
```diff
@@ -0,0 +1,106 @@
...
+  );
+  # If no timestamp was found in the activity table, it was added
+  # at the time the bug was created. In this case we will just use
+  # the current last modified timestamp.
+  if (!$regressed_by_ts) {
+    $regressed_by_ts = $bug->delta_ts;
+  }
+
+  my $field = Bugzilla::Field->new({name => 'cf_has_regression_range'});
+  $field || die 'Custom field cf_has_regression_range field does not exist';
```
globau, line 85: `Bugzilla::Field->check(…)` will avoid the need to explicitly check the result.

scripts/bug1749479.pl
globau: `->check(…)` here too

globau (changes requested):

globau (approved):

PR #1792, opened 2022-01-15
Bug 1749827 - Phabbugz should also disable users in Phabricator when disabled in BMO using Conduit API user.edit

dklawren:


Adams214565 (approved):

extensions/PhabBugz/Extension.pm
```diff
@@ -31,14 +35,14 @@ sub template_before_process {
   return unless $file =~ /bug_modal\/(header|edit).html.tmpl$/;
 
   if (my $bug = exists $vars->{'bugs'} ? $vars->{'bugs'}[0] : $vars->{'bug'}) {
-    my $has_revisions = 0;
+    my $has_revisions         = 0;
```
mitchhentges, line 38: FWIW, This feels like an unrelated change.
Also, I wish we had an autoformatter to avoid unexpected spacing changes like this.

extensions/PhabBugz/Extension.pm
```diff
@@ -54,6 +58,34 @@ sub webservice {
   $args->{dispatch}->{PhabBugz} = "Bugzilla::Extension::PhabBugz::WebService";
 }
 
+sub object_end_of_update {
+  my ($self,   $args)    = @_;
+  my ($object, $changes) = @$args{qw(object changes)};
```
mitchhentges, line 63: To make sure I understand this right: `$object` is a BMO user model, right?
Can we name it `$user` instead?

Or, alternatively, call it `$object`, but after the `!$object->isa('Bugzilla::User')` guard, then do `my $user = $object`? Might make this easier to follow 🤔

extensions/PhabBugz/Extension.pm
```diff
@@ -54,6 +58,34 @@ sub webservice {
...
+    if ($object->is_enabled) {
+      $phab_user->enable_user;
+    }
+    else {
+      $phab_user->disable_user;
+    }
+  }
+  catch {
+    WARN('ERROR: Error updating Phabricator user status for Bugzilla user '
+        . $object->login);
```
mitchhentges, line 85: I suppose if this ever fails, we'll probably miss the warning, but we'll still once-in-a-while ensure that Phabricator's "disabled users" are up-to-date with Bugzilla, is that right?

Since this (guessed 😄) situation is still strictly better than today's situation, that sounds good to me 👍

extensions/PhabBugz/lib/User.pm
```diff
@@ -199,5 +200,27 @@ sub get_phab_bugzilla_ids {
...
+}
+
+sub disable_user {
+  my ($self) = @_;
+  my $params = {
+    objectIdentifier => $self->phid,
+    transactions     => [{type => 'disabled', value => true,}]
+  };
+  request('user.edit', $params);
+}
```
mitchhentges, line 223: Nit: I might join these into a `sub set_user_enabled_status($enabled)`, then smooth out the call site as well to collapse the `if/else`.

mitchhentges (approved):
Makes sense to the me 👍

extensions/PhabBugz/Extension.pm
dklawren: Yeah. It is not optimal in that it is best effort and doesn't retry on failure. And we would need to know the error was occurring by either watch the logs or noticing a user still is enabled in Phab in future. Not sure if there is a security risk in having a user not be disabled in Phab and disabled in BMO. The user would remain logged into Phab for a long time even if not logged in BMO. We could instead send an email to bmo-admins instead of just logging a warning. Then we could address it manually. I weighed this against just out right failing when someone disables the user in BMO and did not want to hinder that process since #servicedesk handles that,

extensions/PhabBugz/lib/User.pm
dklawren: Will do that.

extensions/PhabBugz/Extension.pm
mitchhentges: If it's not too tricky, then sending an email to `bmo-admins` (or otherwise triggering an alert notification in our monitoring infrastructure) sounds great!
I'm happy to land that in a separate patch if you'd prefer to land this one sooner, especially considering how this is a drastic improvement on the status quo.

extensions/PhabBugz/Extension.pm
```diff
@@ -54,6 +59,48 @@ sub webservice {
...
+  my $user = $object;
+
+  my $orig_error_mode = Bugzilla->error_mode;
+  Bugzilla->error_mode(ERROR_MODE_DIE);
+
+  try {
+    my $phab_user
+      = Bugzilla::Extension::PhabBugz::User->new_from_query({ids => [$user->id]});
+    return if !$phab_user;
+    $phab_user->set_user_enabled_status(($user->is_enabled ? 1 : 0));
```
mitchhentges, line 82: Just curious: why are you passing in `0`/`1` instead of `$user->is_enabled` directly?

extensions/PhabBugz/lib/User.pm
```diff
@@ -199,5 +200,17 @@ sub get_phab_bugzilla_ids {
...
-1;
+#################
+#   Helpers     #
+#################
 
+sub set_user_enabled_status {
+  my ($self, $enabled) = @_;
+  my $params = {
+    objectIdentifier => $self->phid,
+    transactions     => [{type => 'disabled', value => ($enabled ? false : true),}]
```
mitchhentges, line 211: Can you do `!$enabled` here?

extensions/PhabBugz/template/en/default/email/phab_enabled_status.txt.tmpl
```diff
@@ -0,0 +1,22 @@
...
+
+From: [% Param('mailfrom') %]
+To: [% Param('maintainer') %]
+Subject: [[% terms.Bugzilla %]] Unable to set Phabricator enabled status for [% user.login FILTER none %]
+X-Bugzilla-Type: admin
+
+Attempting to update enabled status automatically for [% user.login FILTER none %]
+in Phabricator failed with following error:
+
+[% error FILTER none %]
```
mitchhentges, line 19: Thanks for putting the `error` in here 👍

mitchhentges (approved):
Nice, thanks for sending that notification email 👍

extensions/PhabBugz/Extension.pm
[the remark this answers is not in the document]
dklawren: Bugzilla::Extension::PhabBugz::User->new_from_query) also makes an API call to Phabricator so it should be included.

extensions/PhabBugz/Extension.pm
```diff
@@ -54,6 +59,48 @@ sub webservice {
...
+  next
+    if !$params->{phabricator_enabled}
+    || !$object->isa('Bugzilla::User')
+    || !$changes
+    || !$changes->{disabledtext};
+
+  my $user = $object;
+
+  my $orig_error_mode = Bugzilla->error_mode;
+  Bugzilla->error_mode(ERROR_MODE_DIE);
```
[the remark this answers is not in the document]
dklawren: This changes the formatting of any exceptions to be that of a command line script. The default error mode, if I do not change it, is set to display a web page display which is then included with the normal output when a user is disabled in BMO. By changing it to command line it is suppressed from the UI and we just send an email instead including the actual error message from Phabricator.

extensions/PhabBugz/Extension.pm
dklawren: Good point. Just figured it would be more clear the to the viewer but it is technically redundant. I will change that.

extensions/PhabBugz/lib/User.pm
dklawren: Not sure how that would translate JSON wise. The Phabricator API requires an explicit true or false JSON value and not a 1 or 0. Perl does support using !$enabled. I will try it out.

extensions/PhabBugz/Extension.pm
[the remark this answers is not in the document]
mitchhentges: Nice, thanks for the context, that makes sense :)

extensions/PhabBugz/lib/User.pm
dklawren: So it has to be set to JSON true or false which is two function provided by the Mojo::JSON library. So I will leave this as is. Anything else I get "Exception when processing transaction of type "disabled": Error while reading "value": Expected boolean (true or false), got something else." errors from Phabricator.

mitchhentges: Oh, this `true` and `false` was from a _third party library_?
Ok, carry on :)

PR #1798, opened 2022-01-26
Bug 1751601 - Bugzilla::User->visible_bugs causes fatal error if one of the bug ids passed is an alias and not a bug id

dklawren:


Bugzilla/Bug.pm
```diff
@@ -4635,15 +4635,21 @@ sub GetBugActivity {
         my @filtered_added;
         foreach my $value (split /, /, $removed) {
           if (substr($value, 0, length $url_base) eq $url_base) {
-            my ($bug_id) = $value =~ /id=(\d+)$/;
-            next if !Bugzilla->user->can_see_bug($bug_id);
+            my ($bug_id) = $value =~ /id=(\w+)$/;
+            if ($bug_id !~ /^\d+$/) {
+              $bug_id = bug_alias_to_id($bug_id);
```
mitchhentges, line 4640: What is a "bug alias"?

Bugzilla/BugUrl/Bugzilla/Local.pm
```diff
@@ -73,7 +74,17 @@ sub _check_value {
     $uri = $class->SUPER::_check_value($uri);
   }
 
-  my $ref_bug_id  = $uri->query_param('id');
+  # If bug ID is an alias, we want to store the value in the DB
+  # as the actual ID instead of the alias for visibility checking
+  # and other parts of the code.
```
mitchhentges, line 79: Are there already cases of aliases (instead of IDs) in the database that we should clean up with a migration?

mitchhentges (approved):

Bugzilla/Bug.pm
dklawren: A bug alias is just a unique text identifier for a bug instead of using the ID. It has to be unique and can be used in places that accept a bug ID. 

Example: https://bugzilla.mozilla.org/show_bug.cgi?id=MediaSession

Bugzilla/BugUrl/Bugzilla/Local.pm
dklawren: Surprisingly not many.

bug_id | added | removed
-- | -- | --
1736952 | https://bugzilla.mozilla.org/show_bug.cgi?id=proton |  
1702365 | https://bugzilla.mozilla.org/show_bug.cgi?id=hdr |  
960875 | https://bugzilla.mozilla.org/show_bug.cgi?id=letterboxing |  
1644950 | https://bugzilla.mozilla.org/show_bug.cgi?id=avif |  
1512337 | https://bugzilla.mozilla.org/show_bug.cgi?id=puppeteer |  
1480562 | https://bugzilla.mozilla.org/show_bug.cgi?id=antitracking |  
812437 | https://bugzilla.mozilla.org/show_bug.cgi?id=appstartup |  
1138101 | https://bugzilla.mozilla.org/show_bug.cgi?id=POODLEBITE |  

No fixup migration is necessary as the changes in Bugzilla/Bug.pm will do the separate lookup on the fly if it gets a URL that doesn't have an integer bug id.

mitchhentges: Hmm, but then other consumers that access the database may be "suprised" if they run into "alias" data instead of raw "id" data.
IMO, cleaning this data in the database (especially since there's so few) is worth it because:
* It simplifies code now (you don't need the separate lookup on-the-fly)
* It reduces complexity and maintenance effort in the future (you won't need to worry about aliases getting changed/removed, causing broken links, you won't need to add new on-the-fly lookups for new queries that access this data, etc)

dklawren: Cool. I will manually clean up the current ones in the DB and update the PR to not include the extra steps to workaround the old entries.

mitchhentges (approved):
Noice!

PR #1799, opened 2022-01-26
Bug 1751597 - DBD::mysql::db do failed: Column 'method' cannot be null [for Statement "INSERT INTO user_request_log

dklawren:


Bugzilla.pm
```diff
@@ -588,7 +588,7 @@ sub log_user_request {
   my $cgi         = $class->cgi;
   my $user_id     = $class->user->id;
   my $request_url = $cgi->request_uri // '';
-  my $method      = $cgi->request_method;
+  my $method      = $cgi->request_method // 'GET';
```
mitchhentges, line 591: Is `$cgi` from a 3rd party? If not, perhaps we should default `request_method` to `'GET'` earlier in the request lifecycle.

mitchhentges (approved):

Bugzilla.pm
dklawren: Yeah it comes from upstream CGI Perl module
https://metacpan.org/release/LEEJO/CGI-4.53/source/lib/CGI.pm#L2902

So in our case some requests, the environment $ENV{'REQUEST_METHOD'} is not getting set properly and is undefined.

mitchhentges: > So in our case some requests, the environment $ENV{'REQUEST_METHOD'} is not getting set properly and is undefined.

This might be worth investigating further to see why it's not getting set properly.

PR #1800, opened 2022-01-26
Bug 1751596 - Use of uninitialized value $component in string ne at /app/extensions/BMO/Extension.pm line 2266.

dklawren:
This just sets the component value to '' if not provided to eliminate the Perl warning from going to the logs.

### DIFF
--- a/extensions/BMO/Extension.pm
+++ b/extensions/BMO/Extension.pm
@@ -2253,6 +2253,7 @@ sub forced_format {
   # note: this is also called from the guided bug entry extension
   my ($product, $component) = @_;
   return undef unless defined $product;
+  $component ||= '';
 
   # always work on the correct product name
   $product = Bugzilla::Product->new({name => $product, cache => 1})


extensions/BMO/Extension.pm
```diff
@@ -2253,6 +2253,7 @@ sub forced_format {
   # note: this is also called from the guided bug entry extension
   my ($product, $component) = @_;
   return undef unless defined $product;
+  $component ||= '';
```
mitchhentges, line 2256: This seems weird, but if this is the conventional way for the default value of a Perl argument to be equivalent to `None`, then :+1:

mitchhentges (approved):

extensions/BMO/Extension.pm
dklawren: Yeah. For older versions of Perl that is the customary way to give default values. Python is much nicer in this regard. Newer Perl has support for experimental features such as 'signatures' which gives Perl more Python like syntax.

dklawren: https://www.effectiveperlprogramming.com/2015/04/use-v5-20-subroutine-signatures/

PR #1801, opened 2022-01-28
Bug 1751524 - [MYSQL8] The groups table in Bugzilla conflicts with reserved keyword in MySQL 8.x

dklawren:


cpanfile.snapshot
```diff
@@ -22,83 +22,83 @@ DISTRIBUTIONS
       Algorithm::Diff::_impl 1.201
     requirements:
       ExtUtils::MakeMaker 0
-  Alien-Build-2.41
```
mitchhentges: Is this file equivalent to a `cargo.lock`/`package-lock.json`/`requirements.txt`?
If so, would it be possible to split changing dependency versions to a different patch?

Bugzilla/Bug.pm
```diff
@@ -4180,7 +4180,7 @@ sub groups {
       . " THEN 1 ELSE 0 END,"
       . " CASE WHEN groups.id IN($grouplist) THEN 1 ELSE 0 END,"
       . " isactive, membercontrol, othercontrol"
-      . " FROM groups"
+      . " FROM " . $dbh->quote_identifier('groups')
```
mitchhentges, line 4183: IMO, if we're already pretty MYSQL-specific, it might be simpler and easier to read to manually quote this eh?
```
...
. ' FROM "groups"'
...
```

I'm guessing that `dbh->quote_identifier()` is still a good idea if the table name comes from a variable, though.

What do you think?

Bugzilla/DB/Mysql.pm
```diff
@@ -330,15 +330,6 @@ sub bz_setup_database {
...
 
   if ($self->utf8_charset eq 'utf8mb4') {
-    my %global = map {@$_}
-      @{$self->selectall_arrayref(q(SHOW GLOBAL VARIABLES LIKE 'innodb_%'))};
-    my $utf8mb4_supported
-      = $global{innodb_file_format} eq 'Barracuda'
-      && $global{innodb_file_per_table} eq 'ON'
-      && $global{innodb_large_prefix} eq 'ON';
-
-    die install_string('mysql_innodb_settings') unless $utf8mb4_supported;
```
mitchhentges, line 340: In the bug/commit message/PR description, can you elaborate why this is no longer needed? Perhaps link to the specific part of the release notes that obsolete this?

mitchhentges, line 340: In the bug/commit message/PR description, can you elaborate why this is no longer needed? Perhaps link to the specific part of the release notes that obsolete this?

Bugzilla/DB/Schema.pm
```diff
@@ -2480,7 +2490,7 @@ sub _get_create_index_ddl {
   my $sql = "CREATE ";
   $sql .= "$index_type " if ($index_type && $index_type eq 'UNIQUE');
   $sql
-    .= "INDEX $index_name ON $table_name \(" . join(", ", @$index_fields) . "\)";
+    .= "INDEX \`$index_name\` ON \`$table_name\` \(" . join(", ", @$index_fields) . "\)";
```
mitchhentges, line 2493: Why do the backticks need to be `\escaped` here?

Bugzilla/Report/Ping/Simple.pm
```diff
@@ -49,8 +49,7 @@ sub _build_validator {
     creation_ts      => joi->string->required,
   });
 
-  return JSON::Validator->new(
-    schema => Mojo::JSON::Pointer->new($schema->compile));
+  return JSON::Validator->new(schema => $schema);
```
mitchhentges, line 52: This is an unrelated change (probably due to the `cpan.snapshot` update) and should be in another patch IMO

extensions/Voting/Extension.pm
```diff
@@ -899,7 +899,8 @@ sub _remove_votes {
 sub _confirm_if_vote_confirmed {
   my ($id, $product, $timestamp) = @_;
   my $bug = ref $id ? $id : new Bugzilla::Bug({id => $id, cache => 1});
-  $product ||= $bug->product_obj;
+  $product   ||= $bug->product_obj;
+  $timestamp ||= Bugzilla->dbh->selectrow_array('SELECT LOCALTIMESTAMP(0)');
```
mitchhentges, line 903: I can't find where `$timestamp` is used 🤔

mitchhentges (changes requested):
Something went funky with my github review comments, sorry. I think github choked 😨 

Anyways:
* Hopefully we can always escape table names with "`", that would be a lot more convenient for readability and review
* I'm not going to check over syntax and string-join correctness here and leave that to your hand-testing, because that will be a more effective way of avoiding regressions IMHO
* Nice work!

extensions/Voting/Extension.pm
dklawren: https://github.com/mozilla-bteam/bmo/blob/master/extensions/Voting/Extension.pm#L911

It happens when a bug is updated and triggers a voting change, we want the voting related comment to have the same timestamp as the other bug change(s).

Bugzilla/Report/Ping/Simple.pm
dklawren: Yes this is due to the newer version of JSON::Validator due to the new bmo-slim-perl image built with the latest CPAN dependencies. I can pin the version of JSON::Validator to an older version and build a new image. The reason for the new image in the first place was due to needing new version of the mysql perl bindings to support the move the 8.x.

Bugzilla/Bug.pm
dklawren: PostgreSQL uses double quotes instead and I do not want to knowingly break Pg. Using ->quote_identifier allows for any other changes needed in the future.

cpanfile.snapshot
dklawren: It is like a requirements.txt. The snapshot file is the list of the actual versions that were installed when all of the latest packages were pulled down. I will see if I can minimize the amount of changes.

dklawren:
Sorry about the extra commits. I git rebased from main into this branch. Also I know why the tests are failing right now as I need to push the new image to docker hub and restart them.

mitchhentges:
No worries, just ping me when this is ready for re-review :)

dklawren:
Ugh it also added all of the files from the rebased commits to this PR. Hmm what did i do

dklawren:
Ok. cleaned up with force push (sorry) and added a couple more places that needed to be updated.

Bugzilla/Report/Ping/Simple.pm
mitchhentges: Ah, that's fair.
Would you mind splitting the bumping of the new `bmo-slim-perl` image (plus fixing deprecated usages, such as `JSON::Validator` here) from the MSYQL8 changes?

That way, if there's risk of breakage, it may be easier to track down which change caused it (changed system + dependencies versus the signficant MySQL change).

Bugzilla/DB/Mysql.pm
```diff
@@ -332,10 +332,17 @@ sub bz_setup_database {
   if ($self->utf8_charset eq 'utf8mb4') {
     my %global = map {@$_}
       @{$self->selectall_arrayref(q(SHOW GLOBAL VARIABLES LIKE 'innodb_%'))};
+
+    # Newer versions of MySQL the default value for innodb_file_format is Barracuda
```
mitchhentges, line 336:
```suggestion
    # In newer versions of MySQL the default value for innodb_file_format is Barracuda
```

Also, if you could specify a rough range (is "in MySQL 8+" correct?) that would help for readers of this comment a few years from now, when "newer version of MySQL" means something different :)

Bugzilla/Install/DB.pm
```diff
@@ -1933,12 +1931,11 @@ sub _convert_groups_system_from_groupset {
 
     # Identify admin group.
     my ($admin_gid)
-      = $dbh->selectrow_array("SELECT id FROM groups WHERE name = 'admin'");
+      = $dbh->selectrow_array("SELECT id FROM `groups` WHERE name = 'admin'");
```
mitchhentges, line 1934: Let's do `$dbh->quote_identifier()` here

Bugzilla/Search.pm
```diff
@@ -1435,8 +1435,9 @@ sub _translate_join {
   if ($extra_condition) {
     $extra_condition = " AND $extra_condition";
   }
-
-  my @join_sql = "$join JOIN $table AS $name"
+  my $sql_table
+    = $table eq 'groups' ? Bugzilla->dbh->quote_identifier($table) : $table;
```
mitchhentges, line 1439: We could simplify this by unconditionally quote-ing it, though I'm happy either way.

userprefs.cgi
```diff
@@ -533,17 +533,18 @@ sub DoPermissions {
...
-      . ") ORDER BY name");
+      . ') ORDER BY name');
   foreach my $group (@$groups) {
     my ($nam, $desc) = @$group;
     push(@has_bits, {"desc" => $desc, "name" => $nam});
   }
   $groups = $dbh->selectall_arrayref(
     'SELECT DISTINCT id, name, description
-                                          FROM groups
+                                          FROM `groups`
```
mitchhentges, line 547: For consistency, let's `$dbh->quote_identifier()` here

mitchhentges (approved):
Looks good, nice work on the migration.
There's a couple missed `$dbh->quote_identifier()` calls that should probably be addressed, and from a commit-structure POV I think it'd be better to split the perl package (and docker) changes from the MySQL changes (though this part is more of a nit).

Send it 🚀

Bugzilla/Search.pm
dklawren: I tried this in some cases the $table value is actually another string of SQL such as "foo JOIN ON (SELECT blah FROM blar)" So I can only quote if table is groups.

Bugzilla/Install/DB.pm
dklawren: Missed that one.

PR #1802, opened 2022-02-01
Bug 1752918 - Remove duplicate column names in buglist.cgi

dklawren:


buglist.cgi
```diff
@@ -512,6 +514,9 @@ if (defined $params->param('columnlist')) {
   }
   else {
     @displaycolumns = split(/[ ,]+/, $params->param('columnlist'));
+    # Weed out duplicate column names
+    @displaycolumns = uniq @displaycolumns;
+    $params->param('columnlist', join ',', @displaycolumns);
```
mitchhentges, line 519: Oh, do we both peel out `$params->param` into `@displaycolumns`, then continue to use `$params->param` elsewhere? That's a bit of a bummer :(

mitchhentges (approved):
Makes sense to me if I've understood the context correctly :)

buglist.cgi
dklawren: Yeah unfortunately.

https://github.com/mozilla-bteam/bmo/blob/master/buglist.cgi#L1049-L1051

Bugzilla never one any code design awards :)

PR #1804, opened 2022-02-03
Bug 1753429 - add ability to search for disabled/enabled users

globau:
Adds two new advanced search operators:
- is disabled
- is not disabled

These operators only work with user fields.


dklawren (approved):
r=dkl. Works as expected.

PR #1805, opened 2022-02-03
Bug 1666921 - Remove all code related to auth delegation from BMO as OAuth2 provider support replaces it

dklawren:
Also removed couple small places where MozReview code still existed.


Bugzilla/Token.pm.orig
```diff
@@ -0,0 +1,813 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
```
mitchhentges, line 1: I don't think you want this `.orig` file in your patch :)

mitchhentges (approved):
Red lines are good lines, thanks for cleaning this up :)

PR #1806, opened 2022-02-03

dklawren:
Second attempt on this feature


Bugzilla/Config.pm
```diff
@@ -249,72 +276,72 @@ sub update_params {
...
-  $params{urlbase} = Bugzilla->localconfig->urlbase;
-  __PACKAGE__->_write_file(Data::Dumper->Dump([\%params], ['*param']));
+  my ($params) = @_;
+  $params ||= Bugzilla->params;
+
+  unless (Bugzilla->request_cache->{no_database}) {
+    try {
+      foreach my $key (keys %{$params}) {
+        if (my $param = Bugzilla::Config::Param->new({name => $key})) {
+          my $value = $params->{$key} || ($param->is_numeric ? 0 : '');
```
mitchhentges, line 286: Oh, I thought that after [this commit](https://github.com/mozilla-bteam/bmo/pull/1765/commits/57bd5ff7d5f35b20e2ebeb4b8020d902cbf56d63), we were dropping the `is_numeric` part of parameters. Can you clarify why it's been added back?

t/sqlite-memory.t
```diff
@@ -12,17 +12,18 @@ use Test::More;
...
   $ENV{BMO_db_driver}   = 'sqlite';
   $ENV{BMO_db_name}     = ':memory:';
 }
+
+use Bugzilla::Test::MockParams;
+
 use Bugzilla;
 BEGIN { Bugzilla->extensions }
 
-
```
mitchhentges, line 25: We can simplify the diff here by reverting these unnecessary changes.

mitchhentges (changes requested):
Can you provide details on why the previous implementation (#1765) was reverted, and what's different with this one?

Bugzilla/Config.pm
```diff
@@ -95,7 +100,29 @@ sub update_params {
...
+    my %file_params = %{$s->varglob('param')};
+
+    WARN('Migrating old parameters from data/params to database');
+    foreach my $key (keys %file_params) {
+      $param->{$key} = $file_params{$key};
+    }
+
+    WARN('Backing up old params file');
+    rename("$datadir/params", "$datadir/params.old")
+      or die "Rename params file failed: $!";
```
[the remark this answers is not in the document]
dklawren: When running in GCP, we will no longer use a shared file system for $datadir so this will be a non-issue.

Bugzilla/Config.pm
dklawren: New changes do drop is_numeric and just overwrite current value always without needed to do the comparison.

dklawren:
Please try not to hate for me this but after much thought I decided to refactor this one last time. Just further proving the old saying "Perfection is the enemy of good".  I made the Config.pm file a proper class/object instead of just being a file of utility functions. So this did away with Param.pm and Config.pm just loads all params from the database table when Bugzilla::Config->new is called. If there are no params in the table yet (i.e. new install or migrating from old file based params), it loads all default values instead, Then if a param is to be updated code will just do $object->set_param($param_name, $new_value). And then it is written back to the DB with $object->update. We do similar in most other class modules in Bugzilla. To migrate old params, checksetup.pl will call Bugzilla::Config->new->migrate_params(). I will do that manually when we bring up the new cluster in AWS but before we move it over to replace the current running cluster. I also made a few optimizations in the test scripts to work with the new way of doing things. Thanks!

Bugzilla/Config.pm
```diff
@@ -11,375 +11,366 @@ use 5.10.1;
...
+      else {
+        $dbh->do('INSERT INTO params (name, value) VALUES (?, ?)',
+          undef, $key, $new_value);
+      }
+    }
+
+    # And now we have to reset the params cache
+    Bugzilla->memcached->set_params($params);
+    delete Bugzilla->request_cache->{params};
+    delete Bugzilla->request_cache->{params_obj};
```
mitchhentges, line 115: Do you know why we have two caches - one in `memcache`, and one here in-memory (if I understand correctly)?
Perhaps we need only one local cache: I'm guessing the in-memory `Bugzilla->request_cache` one.

Bugzilla/Config.pm
```diff
@@ -11,375 +11,366 @@ use 5.10.1;
...
     }
   }
 
   # Change from a boolean for quips to multi-state
-  if (exists $param->{'usequip'} && !exists $param->{'enablequips'}) {
-    $new_params{'enablequips'} = $param->{'usequip'} ? 'on' : 'off';
+  if (exists $params->{'usequip'}
+    && !exists $params->{'enablequips'})
+  {
+    $new_params{'enablequips'} = $params->{'usequip'} ? 'on' : 'off';
```
mitchhentges, line 170: Oof, that `$param` -> `$params` rename hurts a little in making this diff a little more aggressive.
If you don't mind splitting such impactful changes into separate commits, that could make reviewing a little bit easier.

Bugzilla/Status.pm
```diff
@@ -139,8 +141,11 @@ sub is_open_state {
 }
 
 sub closed_bug_statuses {
-  my @bug_statuses = Bugzilla::Status->get_all;
-  @bug_statuses = grep { !$_->is_open } @bug_statuses;
+  my @bug_statuses;
+  try {
+    @bug_statuses = Bugzilla::Status->get_all;
+    @bug_statuses = grep { !$_->is_open } @bug_statuses;
+  };
```
mitchhentges, line 148: Was this change needed as part of this patch?

mitchhentges (changes requested):
Would you mind splitting this up into more easy-to-parse commits? I'm completely OK with a force-push to clean up the current history if it makes that easier.

There's a lot of `perl` knowledge that I need to pick up on (`shift`, `__PACKAGE__`, how you're invoking these functions and managing the cache), and it's a lot trickier with all the changes going on.

Some things I'd recommend being standalone commits:
1. The baseline change of using the database rather than the file and having a local cache
2. Large-scale variable renames (feel free to group these, but something which should definitely be in here is the `$param -> $params` change)
3. I'm :100:% ok with formatting changes in the PR, but if they're in there, then please maintain a clean commit history so that I can factor them out during review.

Bugzilla/Config.pm
```diff
@@ -11,375 +11,366 @@ use 5.10.1;
...
-    }
+  my $cache = Bugzilla->request_cache;
+  if ($cache->{params_obj}) {
+    return $cache->{params_obj};
   }
-}
 
-# END INIT CODE
+  my $self = {params => {}};
+  bless($self, $class);
```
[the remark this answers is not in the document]
dklawren: ref() returns the class that $invocant belongs to. In this case it would probably be `Bugzilla::Config`. Bless creates an pseudo object that Perl uses for OOP. So if I bless() a hash such as $self (normally what people name it, sort of like this in JS), then I can then use pointer notation ($object->do_something()) and any functions defined can be used as methods. The functions just assume the first param will be the $self reference.

Bugzilla/Config.pm
dklawren: memcache is a cached of just the hash of parameter name and values. It can be used instead of reading from the params table of the database.

In memory is the request_cache that is destroyed at the end of each client request.

PR #1808, opened 2022-02-12
Bug 1754481 - Make mozilla-employee-confidential group read only for people who use IAM for authentication

dklawren:


extensions/MozillaIAM/template/en/default/hook/admin/users/userdata-end.html.tmpl
```diff
@@ -0,0 +1,36 @@
...
+      span.setAttribute('title',
+        'Changes made by admins will be overwritten next ' +
+        'time the user\'s People profile is updated.');
+      span.innerHTML = '(IAM Managed)';
+      label.appendChild(span);
+    }
+    document.querySelector('input[name="group_[% moz_emp_group_id FILTER js %]"]').disabled = true;
+    document.querySelector('input[name="bless_[% moz_emp_group_id FILTER js %]"]').disabled = true;
+});
+</script>
```
mitchhentges, line 36: Can we do this we render the user settings page, rather than modifying it afterwards with javascript?
This will have a couple benefits:
1. People who have javascript disabled will have a better experience
2. It will be easier to follow where work happens: a future developer won't be stuck looking at the user settings page template, wondering how some inputs appear to be being disabled.

extensions/MozillaIAM/Extension.pm
```diff
@@ -143,6 +145,30 @@ sub oauth2_client_handle_redirect {
   return;
 }
 
+sub object_end_of_set_all {
```
mitchhentges, line 148: Are there other kinds of ways that the `User` object can be modified that we should guard against?

mitchhentges (changes requested):

extensions/MozillaIAM/template/en/default/hook/admin/users/userdata-end.html.tmpl
dklawren: It comes down to how extensions work in Bugzilla. MozillaIAM is an extension and not part of the core code. The template that generates the user permissions admin page do not have extension hooks in the places that render each row of the permissions table. There is only the hook that is called at the end of the table so that was the reason for using jscript to update the DOM after render. I can do it the other way, fixing it pre-render, but will take adding additional extension hooks. 

The code is supposed to be a mild deterrent really to keep future admins from messing with that particular group setting and is not meant to be bullet-proof really. If someone wants to update the group, they can just blank out the value for IAM Username, save, and then edit the permission. The think is that the next time the user logs in with Mozilla IAM, it will repopulate the username and set the permission anyway.

mitchhentges: Ah, that makes sense, I didn't realize the extension-y model going on here.
This is ok then :+1:  :)

mitchhentges (approved):

extensions/MozillaIAM/Extension.pm
```diff
@@ -192,6 +192,28 @@ sub userprefs_can_change_email_password {
   }
 }
 
+sub admin_editusers_action {
```
mitchhentges, line 195: Why can't this change be part of `object_end_of_set_all`?
In which workflows is `object_end_of_set_all()` called, and in which ones is this called instead?

dklawren: Unfortunately in editusers.cgi, legacy code still updates the group mapping tables using direct SQL and does not use the object methods properly like $user->set_all() etc. The REST API and other places do use the proper object methods for updating group permissions which is why we needed to update both places. I will look into fixing editusers.cgi properly in the future.

mitchhentges (approved):

PR #1810, opened 2022-02-17
Bug 1755884 - https://bugzilla.mozilla.org/ does not show logged in user as logged in

dklawren:
As part of the BMO IAM commit, this change snuck in and should not have. It was part of some debugging work that I was doing to figure out how the quick redirect from / to /home that we have in places was causing problems for other functionality. This patch adds it back so that when people go to / it will show them as logged in properly.


mitchhentges (approved):

dklawren:
Thanks

PR #1816, opened 2022-02-22
Bug 1753889 - "remember these as my default search list options" is not saving "classification" choices

dklawren:


extensions/BMO/web/js/advanced-search.js
```diff
@@ -42,6 +42,10 @@ Bugzilla.BMO.AdvancedSearch = class AdvancedSearch {
 
     // Select non-Graveyard classifications
     for (const $option of this.$classifications.options) {
+      // Return if already selected (part of the default search params for query page)
+      if ($option.selected) {
+        return;
+      }
```
mitchhentges, line 48: Can you elaborate in the commit message how this fixes the issue?

mitchhentges (changes requested):

extensions/BMO/web/js/advanced-search.js
mitchhentges: Oh, I think I understand: if any `$option` was already selected, then we stop the automatic-setting of `$option.selected` to the default.

However, isn't this order-dependent? Won't this behave incorrectly in the following two scenarios?
1. (Unlikely, I suppose): `Graveyard` is moved to be higher in the list
2. The user _unselects_ `Client Software`, then chooses to remember the search options. This code will start by setting `Client Software` to the default value (enabled), then get to `Developer Infrastructure`, realize that it's been selected due to "search options being remembered", then stop at that point. The end result being: `Client Software` is selected even though it was supposed to be remembered as "unselected".

dklawren:
The way the default search is saved is that the current form choices are saved in the database as url query string. When a user loads the form page with nothing selected from the address and query string, it will look for the query string that is stored in the database as their default query. It will then create a new cgi object in query.cgi with the string loaded from the DB and then auto-fill the form choices from that. The javascript that I am fixing on looks at query string values from the origin query string and address used to load the page and doesn't know about the new choices loaded from the database later.

One thing that I could do that might be a simpler fix, is to pass a positive value to the templates if the  query being used was loaded from the database and not do execute the javascript if the value is true.

Another thing would be to do just remove the javascript completely and update the system wide default query values to preselect all classifications other than the graveyard one which does the same as the javascript. The only downside would be if we add new classifications, we have to remember to update the default system wide query as well. I assume the javascript was originally added to avoid remembering to do that. But to be honest the classifications change very seldomly so really isn't that big of deal.

Thoughts?

dklawren:
Decided to go with the last option. This removes extra unneeded code to maintain and we can accomplish the same updating the default query string and just adding classifications as needed later.

PR #1817, opened 2022-02-23
Bug 1671262 - Require EDITBUGS to edit the priority, severity and type fields

dklawren:


dklawren:
Will fix Selenium tests before asking for review.

qa/t/test_bug_edit.t
```diff
@@ -67,9 +67,6 @@ $sel->is_text_present_ok('has been added to the database',
 go_to_bug($sel, $bug1_id);
 $sel->select_ok("rep_platform", "label=Other");
 $sel->select_ok("op_sys",       "label=Other");
-$sel->select_ok("priority",     "label=Highest");
-$sel->check_ok('//input[@name="bug_type" and @value="defect"]');
-$sel->select_ok("bug_severity", "label=blocker");
```
globau, line 72: Rather than removing checks for the input field, we should check to ensure that it isn't present.

globau (changes requested):

globau (approved):

PR #1819, opened 2022-03-03
Bug 1757970 - For private bugs, add placeholder to release announcement text instead of failing

dklawren:
This allows us to commit a secure bug to github, deploy live and then verify that the fix works. All before the bug is made publicly available. This was a request of the security team.

### DIFF
--- a/scripts/build-bmo-push-data.pl
+++ b/scripts/build-bmo-push-data.pl
@@ -173,7 +173,18 @@ sub fetch_bug {
   die 'missing id' unless $bug_id;
 
   my $response = _get('bug/' . $bug_id, {include_fields => BUG_FIELDS,});
-  return $response->{bugs}->[0];
+
+  if (exists $response->{bugs}) {
+    return $response->{bugs}->[0];
+  }
+  else {
+    return {
+      bug_id     => $bug_id,
+      status     => 'RESOLVED',
+      resolution => 'FIXED',
+      summary    => '*** Place holder for summary ***'
+    };
+  }
 }
 
 sub _get {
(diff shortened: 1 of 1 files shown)


scripts/build-bmo-push-data.pl
```diff
@@ -199,7 +210,13 @@ sub _get {
   }
 
   my $response = $ua->request($request);
-  if ($response->code !~ /^2/) {
+  if ($response->code =~ /^2/) {
+    return decode_json($response->decoded_content);
+  }
+  elsif ($response->code =~ /^4/) {
+    return {};
+  }
```
mitchhentges, line 218: Is this just used for Github integration, or will this caught case have unwanted behaviour changes elsewhere?

scripts/build-bmo-push-data.pl
```diff
@@ -173,7 +173,18 @@ sub fetch_bug {
...
+    return $response->{bugs}->[0];
+  }
+  else {
+    return {
+      bug_id     => $bug_id,
+      status     => 'RESOLVED',
+      resolution => 'FIXED',
+      summary    => '*** Place holder for summary ***'
+    };
+  }
```
mitchhentges, line 187: Is there a way that we can specifically provide this information if the bug is indeed private, rather than for every case where we get a `4xx` error when accessing the bug?

mitchhentges (approved):
I trust your judgement here, but I'm a little nervous about unexpected extraneous behaviour changes.

scripts/build-bmo-push-data.pl
dklawren: This only affects creating text file artifacts in CircleCI that help me with posting announcements to various locations such as the wiki, blog and firefox-tools discourse. It doesn't affect anything outside of that.

scripts/build-bmo-push-data.pl
dklawren: I can look for the specific 4xx code that happens when a bug is private instead. Will update.

mitchhentges (approved):
Send it :rocket:

PR #1823, opened 2022-03-11
Bug 1191821 - The Add button when trying to add something to the See Also field is pointless

dklawren:
I agree that the Add button is a nuisance and we should just show single line text fields by default when clicking Edit. This is what happens in this patch for the see also field.


mitchhentges (approved):
Makes sense to me 👍

PR #1824, opened 2022-03-25

dklawren:
Decided easier to just create a new PR. Went through all of the changes and made small purposed commits so that the flow of changes are easier to understand. Please let me know how I can improve the process in the future as well.


dklawren:
> Please continue to review when you can. I will address the failing test cases (forgot to run those locally ☹️ before submitting)

Tests should all be passing now. Please review when can.

Bugzilla/Config.pm
```diff
@@ -11,91 +11,123 @@ use 5.10.1;
...
+###############################
 
-  my $entry = $params{$name};
+sub set_param {
+  my ($self, $name, $value) = @_;
 
-  # sanity check the value
+  die "Unknown param $name" unless (exists $self->{params}->{$name});
 
+  # Sanity check the value
```
mitchhentges, line 82: Language nit: perhaps replace the terminology here with "validate the value"? I was recently asked to not use the terms "sanity" or "sane" to describe correct values.

Bugzilla/Config/Common.pm
```diff
@@ -151,21 +153,25 @@ sub check_opsys {
...
-    return "Must be a valid closed status: one of "
-      . join(', ', @closed_bug_statuses);
-  }
+  try {
+    my @closed_bug_statuses = map { $_->name } closed_bug_statuses();
+    if (!grep($_ eq $bug_status, @closed_bug_statuses)) {
+      return "Must be a valid closed status: one of "
+        . join(', ', @closed_bug_statuses);
+    }
+  };
```
mitchhentges, line 162: Should we `catch` for a specific exception type? [It sounds like `try` without `catch` is supposed to be for exceptional scenarios](https://metacpan.org/pod/Try::Tiny):

> You can also use it like a standalone eval to catch and ignore any error conditions. Obviously, this is an extreme measure not to be undertaken lightly:

Bugzilla/Config.pm
```diff
@@ -133,6 +133,27 @@ sub migrate_params {
...
+    my $s = Safe->new;
+    $s->rdo("$datadir/params");
+    die "Error reading $datadir/params: $!"    if $!;
+    die "Error evaluating $datadir/params: $@" if $@;
+    my %file_params = %{$s->varglob('param')};
+
+    WARN('Migrating old parameters from data/params to database');
+    foreach my $key (keys %file_params) {
+      $param->{$key} = $file_params{$key};
+    }
```
mitchhentges, line 145: Where does the write to the database occur here? I don't see a call to `update` 🤔

mitchhentges (approved):
Looks good, thanks for splitting this up into separate commits. 👍

Bugzilla/Config.pm
dklawren: Will do.

Bugzilla/Config.pm
dklawren: The write happens at the bottom of the routine after all other checks and modifications have been performed.

mitchhentges (approved):

PR #1826, opened 2022-04-01
Bug 1761594 - Bugzilla should reject redirect_uri tampering when being used for authentication with Phabricator

dklawren:
This pull request adds another value to the client data stored in Bugzilla. The new value is a hostname that the client will have Bugzilla redirect back to once the user has approved the login request. This will allow Bugzilla to verify that the redirect hostname provided in the URL has not been changed to a different location which could steal the users token. If the redirect URL and its hostname has been altered, then the user is redirect back to the client site as if access request was not granted at all.

We can also put this change on bugzilla-dev for extra testing in addition to code review here.

### DIFF
--- a/Bugzilla/App/OAuth2/Provider/Clients.pm
+++ b/Bugzilla/App/OAuth2/Provider/Clients.pm
@@ -68,19 +68,29 @@ sub create {
   my $description = $self->param('description');
   my $id          = $self->param('id');
   my $secret      = $self->param('secret');
+  my $hostname    = $self->param('hostname');
   my $scopes      = $self->every_param('scopes');
   $description
-    or return $self->code_error('param_required', {param => 'description'});
-  $id     or return $self->code_error('param_required', {param => 'id'});
-  $secret or return $self->code_error('param_required', {param => 'secret'});
+    or return $self->code_error('param_required',
+    {function => 'admin/oauth/provider/create', param => 'description'});
+  $id
+    or return $self->code_error('param_required',
+    {function => 'admin/oauth/provider/create', param => 'id'});
+  $secret
+    or return $self->code_error('param_required',
+    {function => 'admin/oauth/provider/create', param => 'secret'});
+  $hostname
+    or return $self->code_error('param_required',
+    {function => 'admin/oauth/provider/create', param => 'hostname'});
   any { $_ > 0 } @{$scopes}
-    or return $self->code_error('param_required', {param => 'scopes'});
+    or return $self->code_error('param_required',
+    {function => 'admin/oauth/provider/create', param => 'scopes'});
   my $token = $self->param('token');
   check_token_data($token, 'create_oauth_client');
 
   $dbh->do(
-    'INSERT INTO oauth2_client (client_id, description, secret) VALUES (?, ?, ?)',
-    undef, $id, $description, $secret);
+    'INSERT INTO oauth2_client (client_id, description, secret, hostname) VALUES (?, ?, ?, ?)',
+    undef, $id, $description, $secret, $hostname);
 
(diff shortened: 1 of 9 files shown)


Bugzilla/App/Plugin/OAuth2/Provider.pm
```diff
@@ -111,11 +113,15 @@ sub _resource_owner_confirm_scopes {
     return undef;
   }
 
+  # Deny access if hostname of redirect_uri doesn't match
+  # the hostname assigned to the client id
```
cgsheeh, line 117: Can you add this comment to the function definition?

Bugzilla/App/OAuth2/Provider/Clients.pm
```diff
@@ -68,19 +68,29 @@ sub create {
...
+    {function => 'admin/oauth/provider/create', param => 'description'});
+  $id
+    or return $self->code_error('param_required',
+    {function => 'admin/oauth/provider/create', param => 'id'});
+  $secret
+    or return $self->code_error('param_required',
+    {function => 'admin/oauth/provider/create', param => 'secret'});
+  $hostname
+    or return $self->code_error('param_required',
+    {function => 'admin/oauth/provider/create', param => 'hostname'});
```
cgsheeh, line 84: What does the extra `function` parameter do here?

cgsheeh (changes requested):
LGTM, just want a clarification on the `function` parameter and a suggestion to move the comment into the function.

For future reviews could you please add a commit message outlining what you're doing and why? That will make reviews much easier, given I don't know Perl or the BMO codebase. ;)

Bugzilla/App/OAuth2/Provider/Clients.pm
dklawren: It allows the error message to be show a more exact location in which the error occurred. So in this example, the user will know that the endpoint was missing the required parameter and value.

The error message `param_required` is a small portion of a larger template at `templates/en/default/global/code-error.html.tmpl` and is used in many places. So the function variable in the template could also display a method or code function instead of an endpoint. For example `Bugzilla::Bug->create()', etc.

Bugzilla/App/Plugin/OAuth2/Provider.pm
dklawren: Will do.

dklawren:
> LGTM, just want a clarification on the `function` parameter and a suggestion to move the comment into the function.
> 
> For future reviews could you please add a commit message outlining what you're doing and why? That will make reviews much easier, given I don't know Perl or the BMO codebase. ;)

Yeah I will do better at that and give more detail explanation of what the changes are doing. I originally thought the bug would give enough context and background info but that is not always the case I admit.

fkiriakos07:
@dklawren  I have a quick question, is the hostname stored in the database when the oauth application is created? if so, would phabricator oauth app need to be recreated for the validation to work?

cgsheeh (approved):
Sorry for the delay, I forgot to update the status after you commented about the function parameter.

Longkali2021:
Hi I am the sender of this problem report (irwan)
Hey problem has been months, when will the fix start and when will it end so I can retest this issue.

Best Regards...
Irwan

dklawren:
> Hi I am the sender of this problem report (irwan) Hey problem has been months, when will the fix start and when will it end so I can retest this issue.

Since this was proven not to be a full account take over level issue, we are working it as a normal security improvement that will go out with the normal deployment coming up this week. We were unable to deploy it last week due to myself being out of office. Hopefully you will be able to retest in the next day or so.
Thanks!

PR #1827, opened 2022-04-01
Bug 1759306 - Add a field for Number of See Also

dklawren:


Bugzilla/Search.pm
```diff
@@ -623,6 +625,7 @@ sub COLUMNS {
 
     'attachments.count'  => 'COUNT(DISTINCT map_attachments_count.attach_id)',
```
cgsheeh, line 626: Any idea why it's `XXX.count` sometimes but `XXX_count` other times?

cgsheeh (approved):

Bugzilla/Search.pm
dklawren: Inconsistency is the best guess. I prefer attachments_count instead of attachments.count since the latter looks like it is referring to a column in the attachments table instead of an identifier of a special search term. But we won't go back and change all of them to match unless we need to for now. Technically they both work the same under the hood.

PR #1828, opened 2022-04-02
Bug 1762134 - Update BMO user guide to list "Project Flags" too in addition to "Tracking Flags"

dklawren:


cgsheeh (approved):
Makes sense to me. What templating engine is this?

dklawren:
> Makes sense to me. What templating engine is this?

Template toolkit (http://template-toolkit.org/) Been around for a long time. Probably faster alternatives out there by now but would be a big effort to migrate to something else.

PR #1830, opened 2022-04-05
Bug 1763003 - Write script to fix dependency corruption in database for a large number of older bugs

dklawren:


dklawren:
I have uploaded and ran this script on bugzilla-dev with success. I just wanted a quick look over to make me feel better before running on production.

globau (approved):

PR #1837, opened 2022-04-24
Bug 1430732 - Comments from a logged in user should be stylable differently

branisha:
CSS class "yourself" is added to "change-set" div comment if logged in user is commenter.

This is my first PR :tada:
If I made any mistakes or if I missed anything, please let me know and I'll fix it.


dklawren:
You will also need to add a style to the BugModal css file at /home/dkl/devel/github/mozilla/bmo/master/extensions/BugModal/web/bug_modal.css for the 'yourself' class. We do not do changes that rely on someone elses userstyle customization so you will need the style to be part of Bugzilla itself.

dklawren (changes requested):
Add style definition to /home/dkl/devel/github/mozilla/bmo/master/extensions/BugModal/web/bug_modal.css for this new class.

PR #1840, opened 2022-04-28
Bug 1765917 - Open Redirectionwhen using OAuth2 authentication

dklawren:
1. Bugzilla/App.pm sets the usage mode and stores away the universal controller for use by native Mojo code and routes.
2. Provider.pm now throws a user exception when  the redirect_uri hostname does not match the one configured for the OAutht2 client instead of just returning a false value as before. Before, when a false is returned, BMO just redirected to whatever redirect_uri had set and gave a request denied error to the client which was the incorrect solution in this case.
3. Error.pm removes the message value from stash as it was interfering with a different message value that is used sometimes in the templates and various BMO pages. In this case, message was unused and not needed.

### DIFF
--- a/Bugzilla/App.pm
+++ b/Bugzilla/App.pm
@@ -17,7 +17,7 @@ use FileHandle;    # this is for compat back to 5.10
 use Bugzilla          ();
 use Bugzilla::BugMail ();
 use Bugzilla::CGI     ();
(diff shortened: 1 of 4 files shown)


cgsheeh (approved):
LGTM.

PR #1841, opened 2022-04-29
Bug 1752191 - [WEBHOOKS] Allow Webhook Setup with "Any" Product

dklawren:
1. Update table schema to allow product_id column to be NULL which means any product.
2. Update product_name() and component_name() in Webhooks.pm to return 'Any' if product_id and/or component_id are NULL values in the DB.
3. Update the Push connector for Webhooks to allow all messages if product equals 'Any'.
4. Create a new permission group configuration option to limit amount of users who can select 'Any' as a product choice. Defaults to 'admin'.
5. Extended current test_webhooks.t test script to test for the 'Any' product support.
6. Fixed a current bug in the edit webhooks form where the component field name was incorrectly set to 'api_key'.

### DIFF
--- a/extensions/Webhooks/lib/Webhook.pm
+++ b/extensions/Webhooks/lib/Webhook.pm
@@ -79,24 +79,33 @@ sub component_id {
 
 sub product {
   my ($self) = @_;
-  return $self->{product} ||=
-    Bugzilla::Product->new({id => $self->{product_id}, cache => 1});
+  return $self->{product} if exists $self->{product};
+  $self->{product}
+    = $self->{product_id}
+    ? Bugzilla::Product->new({id => $self->{product_id}, cache => 1})
+    : undef;
+  return $self->{product};
 }
 
 sub product_name {
   my ($self) = @_;
-  return $self->{product_name} ||= $self->{product_id} ? $self->product->name : '';
+  return $self->{product_name} ||= $self->product ? $self->product->name : 'Any';
 }
 
 sub component {
   my ($self) = @_;
-  return $self->{component} ||= $self->{component_id}
-    ? Bugzilla::Component->new({id => $self->{component_id}, cache => 1}) : undef;
+  return $self->{component} if exists $self->{component};
+  $self->{component}
+    = $self->{component_id}
+    ? Bugzilla::Component->new({id => $self->{component_id}, cache => 1})
+    : undef;
+  return $self->{component};
 }
 
 sub component_name {
   my ($self) = @_;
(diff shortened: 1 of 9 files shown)


extensions/Webhooks/lib/Webhook.pm
```diff
@@ -79,24 +79,33 @@ sub component_id {
 
 sub product {
   my ($self) = @_;
-  return $self->{product} ||=
-    Bugzilla::Product->new({id => $self->{product_id}, cache => 1});
+  return $self->{product} if exists $self->{product};
```
cgsheeh, line 82: Does the second return actually have an effect? My non-Perl brain sees this as returning `null` or `$self->{product}`, with the second return being unreachable. Maybe Perl just skips the line if the condition is false?

I'd lean towards making this an `if` with the return inside for readability, like:

```
if exists $self->{product}
    return $self->{product}
endif
```

If this is a standard Perl thing that I'm just not familiar with it's fine to keep. :+1:

cgsheeh (approved):
Question: could we use the value `Any` instead of `null` in the database? Is there a precedent of using `null` to be `Any`/`All` etc in BMO? I think it's fine either way, but it might make things slightly simpler if we don't need to update DB schema and convert `null` to `Any` in the codebase - just check if the value is `Any` in the DB itself.

dklawren:
> Question: could we use the value `Any` instead of `null` in the database? Is there a precedent of using `null` to be `Any`/`All` etc in BMO? I think it's fine either way, but it might make things slightly simpler if we don't need to update DB schema and convert `null` to `Any` in the codebase - just check if the value is `Any` in the DB itself.

I would need to create an actual 'Any' product in the products table and then hide it somehow in the web UI for different pages. This is because the webhooks.product_id column has a foreign key relationship to the products.id column in the products table. By allowing it to be set to NULL, it will still work without creating an actual product.

extensions/Webhooks/lib/Webhook.pm
dklawren: Yeah Perl allows for you to shorten an 'if' statement by writing it backwards. So you example you wrote is exactly the same thing. $self is essentially a hash (dictionary) and $self->{product} is a key of the hash. If the key exists, we have ran product() at least once before and can return the value early without looking up again. 

If I did `return $self->{product} if $self->{product};` then it would return early if it was a true value such as an object. But since in our case $self->{product} can also be an undefined (false) value and still be valid, then we just return it if it exists at all.

PR #1844, opened 2022-05-04
Bug 1767563 - Fix the response table for the /rest/user endpoint

suhaibmujahid:
The Docker development setup does not support my machine (Mac M1), so this PR is not tested yet. I will check the preview produced by `docs/readthedocs.org:bmo` to verify the fix.


suhaibmujahid:
Based on the preview in https://bmo--1844.org.readthedocs.build/en/1844/api/core/v1/user.html#get-user, the bug has been fixed.

PR #1847, opened 2022-05-06
Bug 1767526 - Add auditing to the Voting extension to record changes to votes for bug and user

dklawren:
* Create a new Vote.pm class that stores the current vote data for a given user and bug id. By subclassing Object.pm we get the create() and remove_from_db() methods.
* Since there is not currently a table to track vote changes per user/bug over time, I decided to use the audit_log table to record the changes since they do not need to be visible in the UI. The audit_log table was not exactly compatible so I overrode the audit_log() method to use the bug_id as the identifier instead of the object id (normally the primary key of the table).
* Extension.pm changes are mostly to remove the direct SQL calls where appropriate and use the object methods for storing instead.

### DIFF
--- a/extensions/Voting/lib/Vote.pm
+++ b/extensions/Voting/lib/Vote.pm
@@ -44,9 +44,10 @@ use constant VALIDATORS => {vote_count => \&_check_vote_count};
 ################################
 
 # We override the parent audit_log so we can store the bug id
-# value instead of the vote table id. Also is we are creating a
+# value instead of the vote table id. Also when we are creating a
 # new vote table entry, then we record it as a vote_count change
-# from 0 to the new count or vice versa.
+# from 0 to the new count and the opposite when removing a vote
+# table entry.
 sub audit_log {
   my ($self, $changes) = @_;
   my $new_changes = $changes;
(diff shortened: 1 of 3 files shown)


extensions/Voting/lib/Vote.pm
```diff
@@ -0,0 +1,92 @@
...
+
+  $self->SUPER::audit_log($new_changes);
+}
+
+################################
+####     Validators         ####
+################################
+
+sub _check_vote_count {
+  my ($invocant, $count) = @_;
```
cgsheeh, line 70: What is `$invocant` here?

extensions/Voting/lib/Vote.pm
```diff
@@ -0,0 +1,92 @@
...
+);
+
+use constant VALIDATORS => {vote_count => \&_check_vote_count};
+
+################################
+####     Overrides          ####
+################################
+
+# We override the parent audit_log so we can store the bug id
+# value instead of the vote table id. Also is we are creating a
```
cgsheeh, line 47: "Also is" is a typo I think?

cgsheeh (approved):

extensions/Voting/lib/Vote.pm
dklawren: https://perldoc.perl.org/perlootut#Methods

"When we make a method call, Perl arranges for the method's invocant to be passed as the first argument. Invocant is a fancy name for the thing on the left side of the arrow. The invocant can either be a class name or an object."

PR #1851, opened 2022-05-13
Bug 1769246 - Make the header always above main content, in order to show request menu above other notifications.

arai-a:
`.new-changes-link` has `z-index: 10`, so header should have higher z-index, so that `.dropdown-content` inside it is shown above `.new-changes-link`


dklawren (approved):
LGTM Thanks! r=dkl

PR #1852, opened 2022-05-13
Bug 1769284 - Show phabricator links in user_profile

arai-a:
This adds 2 links in the `user_profile` page in new "Phabricator" section, under "Review Queue" section.
One link for user's profile page that shows recent activity: [example](https://phabricator.services.mozilla.com/p/arai/)
Other link for user's open revisions: [example](https://phabricator.services.mozilla.com/differential/?responsiblePHIDs[0]=PHID-USER-ovlr37yfp3loy53g6wgo&statuses[0]=open()&order=newest&bucket=action)

~~The data is retrieved during showing the user_profile page, with 2 requests, 5 seconds timeout for each.~~
~~If any error happens (including the user not found on phabricator), or if phabricator is not configured, nothing is shown.~~

### DIFF
--- a/extensions/PhabBugz/Extension.pm
+++ b/extensions/PhabBugz/Extension.pm
@@ -33,6 +33,9 @@ sub template_before_process {
   return unless Bugzilla->user->id;
   return unless Bugzilla->params->{phabricator_enabled};
   return unless Bugzilla->params->{phabricator_base_uri};
+
+  $vars->{phabricator_available} = 1;
+
   return unless $file =~ /bug_modal\/(header|edit).html.tmpl$/;
 
   if (my $bug = exists $vars->{'bugs'} ? $vars->{'bugs'}[0] : $vars->{'bug'}) {
--- a/extensions/PhabBugz/lib/WebService.pm
+++ b/extensions/PhabBugz/lib/WebService.pm
@@ -211,6 +211,8 @@ sub user {
 
   $self->_check_phabricator();
 
+  Bugzilla->login(LOGIN_REQUIRED);
+
   my $bmo_user_id = $params->{user_id};
 
   my $response = request(
--- a/extensions/PhabBugz/lib/WebService.pm
+++ b/extensions/PhabBugz/lib/WebService.pm
@@ -205,6 +205,55 @@ sub bug_revisions {
   return {revisions => \@revisions};
 }
 
+sub user {
+  state $check = compile(Object, Dict [user_id => Int]);
+  my ($self, $params) = $check->(@_);
+
+  $self->_check_phabricator();
+
+  my $bmo_user_id = $params->{user_id};
+
+  my $response = request(
+    'bmoexternalaccount.search',
+    {
+      accountids => [$bmo_user_id],
+    }
+  );
+  if (scalar(@{$response->{result}}) == 0) {
+    return {};
+  }
+
+  my $phid = $response->{result}[0]{phid};
+
+  $response = request(
+    'user.query',
+    {
+      phids => [$phid],
+    }
+  );
+  if (scalar(@{$response->{result}}) == 0) {
+    return {};
+  }
+
+  my $user_name = $response->{result}[0]{userName};
+  my $real_name = $response->{result}[0]{realName};
+
+  my $base_url = Bugzilla->params->{phabricator_base_uri};
+  $base_url =~ s{/$}{};
+  my $user_url = "$base_url/p/$user_name/";
+  my $revisions_url = "$base_url/differential/?responsiblePHIDs%5B0%5D=$phid&statuses%5B0%5D=open()&order=newest&bucket=action";
+
+  return {
+    user => {
+      phid => $phid,
+      userName => $user_name,
+      realName => $real_name,
+      userURL => $user_url,
+      revisionsURL => $revisions_url,
+    },
+  };
+}
+
 sub rest_resources {
   return [
     # Bug permission checks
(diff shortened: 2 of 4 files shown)


Bugzilla/User.pm
```diff
@@ -2754,6 +2755,127 @@ sub set_bounce_count {
...
+# Returns error happened during ensure_phab if any.
+sub phab_error {
+  my $self = shift;
+  return $self->{phab_error};
+}
+
+sub phab_base_url {
+  (my $phab_url = Bugzilla->params->{phabricator_base_uri}) =~ s{/$}{};
+  return $phab_url;
+}
```
arai-a, line 2877: I'm not sure if this is the right place.
let me know if there's better place, or if this info is already available in the template.

arai-a:
Applied some cleanup

arai-a:
This takes the option (a) in [the bug comment](https://bugzilla.mozilla.org/show_bug.cgi?id=1769284#c0)

If increasing the time taken by `user_profile` page isn't acceptable, I can switch to the option (b), with redirection page added.
in that case:
  * the link will be shown regardless of whether the user has phabricator account
  * the username isn't known, so the link text will be just "Account" or something (not much problem I think)

arai-a:
Here's screenshot of this patch:
bmo-phab-profile

If the user has review request, it's shown under "Review Queue" section

arai-a:
Currently, the request to the phabricator API is done for each request.

The cache is created on the `User` instance.
So,  `target.phab_phid()` in the template requests the info and create cache, and  `target.phab_user_name()` and `target.phab_real_name()` uses the cached data,
but reloading the page triggers the API request again.

Other option is to store it in the BMO database on the initial request,
but this results in outdated cache once the the user name or the real name is modified.
To be clear, real name is only for displaying text, so not much problem (we can omit it),
but user name is used in the URL, so outdated cache is a problem.

To my understanding, the user name cannot be changed by user, but administrator can change it in case of trouble.
So, if we cache it in the BMO database, there needs to be a cache invalidation process.

arai-a:
Just realized this is very different than what the phabricator revisions view in the bug page works (and how `phabbugz` works)
I'll look into refactoring this to be dynamically generated on the client side, integrating the user/review query to `phabbugz` and communicating it after the page load.

arai-a:
Updated to utilize `PhabBugz` extension.

This patch adds `/rest/phabbugz/user/{bmo_user_id}` endpoint, that returns the following data:
```
{
  user: {
    "phid": "<User PHID>",
    "userName": "<Phabricator username>",
    "realName": "<Phabricator real name>",
    "userURL": "user page URL",
    "revisionsURL": "search page URL for the user's open revisions",
  }
}
```

If phabricator is set up on the server side, the "Phabricator" section is inserted in the `user_profile` page's reponse, with the "User" row and "Revisions" row with empty.

The `phab_user.js` script starts querying to the above endpoint, with setting "User" and "Revisions" field with "Loading..." text,
and once the response is received, those fields are replaced with links:
  * "User" : link to user's profile page that shows recent activity: [example](https://phabricator.services.mozilla.com/p/arai/)
  * "Revisions": link to user's open revisions: [example](https://phabricator.services.mozilla.com/differential/?responsiblePHIDs%5B0%5D=PHID-USER-ovlr37yfp3loy53g6wgo&statuses%5B0%5D=open()&order=newest&bucket=action)

When error happens during the request, the error is shown in the "User" row, and "Revisions" row is set to empty.

arai-a:
possible improvement would be to query `differential.revision.search` on the server side and return the length of review queue, and show it.
But so far I'm not sure how to exclude "Waiting on Other Reviewers" from the result...

arai-a:
> But so far I'm not sure how to exclude "Waiting on Other Reviewers" from the result...

I found the solution.
querying with `{"constraints": {"reviewerPHIDs": ["<phid>"], "statuses": ["needs-review"]}, "attachments": {"reviewers": true}}` returns necessary information (actually, that's `extensions/PhabBugz/lib/WebService.pm` is already doing...).
So, if that also helps, and adding more server-side request is acceptable, we can add it.

dklawren (changes requested):
Overall everything looks great and no functionality or style issues were found. My only request is that phab_user.js be updated to use Vanilla JS instead of rely on jQuery. I realize we have quite a lot of jQuery code already in BMO but we made a choice a while back that all new code would be done using vanilla js and that we would b phasing out the use of jQuery. As time allows we will go back and refactor some of the legacy jQuery code. Let me know if this is a problem or if you need any assistance. Thanks!

arai-a:
Yeah, I just followed the other code.  I'm happy to switch to Vanilla JS :)
Will update shortly.

dklawren:
> Yeah, I just followed the other code. I'm happy to switch to Vanilla JS :) Will update shortly.

Understood. We have way more jQuery that I am comfortable with. We have some pages still using YUI2 (Yahoo's old JS framework) which is completely unsupported upstream which I have been trying to get around to porting over to plain JS.

dklawren (changes requested):

dklawren:
> I found the solution. querying with `{"constraints": {"reviewerPHIDs": ["<phid>"], "statuses": ["needs-review"]}, "attachments": {"reviewers": true}}` returns necessary information (actually, that was `bug_revisions` in `extensions/PhabBugz/lib/WebService.pm` is already doing...). So, if that also helps, and adding more server-side request is acceptable, we can add it.

I am not sure how much value it would be to add the additional call to print the number of revisions. Users can just click through to the Phabricator page and get the information if they need it. If there is a call for it then we can add the extra as a separate bug.

dklawren (approved):
All good! r=dkl

PR #1854, opened 2022-05-17
Bug 1769541 - Update group permission needed to view the groups a user is a member of using the REST API

dklawren:
If a user is a member of `mozilla-employee-confidential` then they are by default able to see the groups a user is a member of. This should be ok for employees to see but not the general public. The old group was `editusers` which was too narrow and a group we cannot just add people to for no good reason. Another part that used the `editusers` group was viewing whether a user was disabled or not and if they had email turned off. I changed that to be `disabledusers` instead which is the permission group allowed to edit those fields for any user in BMO anyway.

### DIFF
--- a/Bugzilla/WebService/User.pm
+++ b/Bugzilla/WebService/User.pm
@@ -309,7 +309,7 @@ sub get {
       last_activity_time => $self->type('dateTime', $user->last_activity_time),
       };
 
-    if (Bugzilla->user->in_group('editusers')) {
+    if (Bugzilla->user->in_group('disableusers')) {
       $user_info->{email_enabled}     = $self->type('boolean', $user->email_enabled);
       $user_info->{login_denied_text} = $self->type('string',  $user->disabledtext);
     }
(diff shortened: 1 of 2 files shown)


suhaibmujahid:
@dklawren thank you for doing this.

> Another part that used the `editusers` group was viewing whether a user was disabled or not and if they had email turned off. I changed that to be `disabledusers` instead which is the permission group allowed to edit those fields for any user in BMO anyway.

I guess you mean "the login denial text" not "whether a user was disabled or not". I think viewing if the user is disabled (`can_login`) already doesn't require `editusers` permissions:

https://github.com/mozilla-bteam/bmo/blob/b261658ecfe7843ad254d843ec464e157dd80c89/Bugzilla/WebService/User.pm#L306

suhaibmujahid:
It will be great to have the docs in sync:


https://github.com/mozilla-bteam/bmo/blob/4cbf51d2e5f1c387ed162c7724682dda85d9643d/docs/en/rst/api/core/v1/user.rst#L369-L372


https://github.com/mozilla-bteam/bmo/blob/4cbf51d2e5f1c387ed162c7724682dda85d9643d/docs/en/rst/api/core/v1/user.rst#L405-L408

cgsheeh (approved):

dklawren:
> It will be great to have the docs in sync:

Ah thanks for the heads up. I will update that as well but I will not say the actual group name.

dklawren:
> I guess you mean "the login denial text" not "whether a user was disabled or not". I think viewing if the user is disabled (`can_login`) already doesn't require `editusers` permissions:

I think it is OK to leave the can_login outside of the permission check. We mainly don't want general users to see the actual disabled text itself in case it has any sensitive information.

PR #1856, opened 2022-05-20
Bug 1770442 - Fix t/critic.t to not just look in scripts directory for violations

dklawren:
So this PR basically just resets the perlcritichistory file to what we have currently while looking at all pl, pm, and cgi files instead of the scripts dir only. This will at the least help keep new violations from occurring with new code. 

After this, I think a good plan of attack for the current violations is to take the severity level down to 5 (gentle) and then file a bug for each violation policy group fixing one at a time until all clear. Then raise the severity to 4, rinse and repeat. Realistically I don't think we would ever get to a clean severity level 1 (brutal) since we just do not have the time and resources. And probably not the best use of time anyway. But at lease we could know out some of the low hanging fruit.

Thoughts?


Administration21 (approved):

t/critic.t
```diff
@@ -8,10 +8,14 @@
 use 5.10.1;
 use strict;
 use warnings;
-use lib qw(. lib local/lib/perl5);
+use lib qw(. lib local/lib/perl5 t);
+
+use Support::Files;
 use Test::More;
 
+my @testitems = (@Support::Files::testitems, @Support::Files::test_files);
```
globau, line 16: This now excludes files in the scripts directory, which should be covered by linting.

Perhaps `@Support::Files::testitems` should be extended to include all files under scripts with a first line ending in ` perl`?

t/.perlcritic-history
```diff
@@ -1,141 +1,144 @@
...
-            'Perl::Critic::Policy::ValuesAndExpressions::ProhibitQuotesAsQuotelikeOperatorDelimiters' => 0,
-            'Perl::Critic::Policy::Variables::RequireNegativeIndices' => 0,
-            'Perl::Critic::Policy::InputOutput::RequireBriefOpen' => 5,
+            'Perl::Critic::Policy::ValuesAndExpressions::RequireUpperCaseHeredocTerminator' => 0,
+            'Perl::Critic::Policy::Variables::ProhibitReusedNames' => 144,
+            'Perl::Critic::Policy::Freenode::Wantarray' => 0,
+            'Perl::Critic::Policy::ValuesAndExpressions::ProhibitLongChainsOfMethodCalls' => 16,
+            'Perl::Critic::Policy::Modules::RequireEndWithOne' => 214,
+            'Perl::Critic::Policy::BuiltinFunctions::ProhibitBooleanGrep' => 229,
+            'Perl::Critic::Policy::Miscellanea::ProhibitUnrestrictedNoCritic' => 0
```
globau, line 142: 📈

globau (changes requested):

globau (changes requested):
> `scripts/start_morbo → scripts/start_morbo.pl` 100755 → 100644

This will result in `Permission denied` when DaemonControl execs the script.

globau (approved):

t/.perlcritic-history
Administration21: > 📈
[](
# Clone my vim settings to a .vim directory
git clone git@github.com:lucasr/vim-settings.git .vim

# Link standard vim settings file to the one .vim directory
ln -s .vim/vimrc ~/.vimrc

# Fetch all bundle submodules
cd .vim
- git submodules update --init)

PR #1860, opened 2022-05-27
Bug 1678376 - Make t/001compile.t Great Again

dklawren:
Mainly this added some code to create a temp test DB using SQLite to allow the compiling of the Bugzilla/Model/* modules which require a DB. Also some critic warnings fixed in Archive.pm and 001compile.t

This does not *yet* test extension modules as there is added complexity to get them to load without lots of redefined warnings. I will address extensions as a separate bug.

### DIFF
--- a/Bugzilla/Attachment/Archive.pm
+++ b/Bugzilla/Attachment/Archive.pm
@@ -80,7 +80,7 @@ sub _build_input_fh {
     croak "I will not read and write a file at the same time";
   }
   my $file = $self->file;
-  return IO::File->new($self->file, '<:bytes') or die "cannot read $file: $!";
+  return IO::File->new($self->file, '<:bytes') || die "cannot read $file: $!";
 }
 
 sub _build_output_fh {
--- a/t/001compile.t
+++ b/t/001compile.t
@@ -20,16 +20,20 @@ use Support::Files;
 use Test::More;
 
 BEGIN {
-  if ($ENV{CI}) {
-    plan skip_all => 'Not running compile tests in CI.';
-    exit;
-  }
   plan tests => @Support::Files::testitems + @Support::Files::test_files;
 
   use_ok('Bugzilla::Constants');
   use_ok('Bugzilla::Install::Requirements');
   use_ok('Bugzilla');
+
+  if (-f 'data/db/model_test') {
+    unlink 'data/db/model_test' || die $!;
+  }
+  $ENV{test_db_name} = 'model_test';
 }
+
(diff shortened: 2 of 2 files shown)


albertoEnstein (approved):

t/001compile.t
```diff
@@ -20,16 +20,20 @@ use Support::Files;
...
-    exit;
-  }
   plan tests => @Support::Files::testitems + @Support::Files::test_files;
 
   use_ok('Bugzilla::Constants');
   use_ok('Bugzilla::Install::Requirements');
   use_ok('Bugzilla');
+
+  if (-f 'data/db/model_test') {
+    unlink 'data/db/model_test' || die $!;
```
globau, line 30: This needs to be either `unlink('data/db/model_test') || die $!` or `unlink 'data/db/model_test' or die $!`

t/001compile.t
```diff
@@ -45,28 +49,28 @@ sub compile_file {
...
 
-  if ($file =~ s/\.pm$//) {
+  if ($file =~ s/[.]pm$//) {
     $file =~ s{/}{::}g;
     use_ok($file);
     return;
   }
 
-  open(my $fh, $file);
+  open my $fh, '<', $file || die $!;
```
globau, line 58: Same `||` vs `or` bug

t/001compile.t
```diff
@@ -45,28 +49,28 @@ sub compile_file {
...
   my $bang = <$fh>;
-  close $fh;
+  close $fh || die $!;
```
globau, line 60: Same `||` vs `or` bug

Bugzilla/Attachment/Archive.pm
```diff
@@ -80,7 +80,7 @@ sub _build_input_fh {
     croak "I will not read and write a file at the same time";
   }
   my $file = $self->file;
-  return IO::File->new($self->file, '<:bytes') or die "cannot read $file: $!";
+  return IO::File->new($self->file, '<:bytes') || die "cannot read $file: $!";
```
globau, line 83: This is unrelated and shouldn't be touched here.

globau (changes requested):

Bugzilla/Attachment/Archive.pm
dklawren: When running t/001compile.t, I get the following error if this left as-is:

```
Possible precedence issue with control flow operator at Bugzilla/Attachment/Archive.pm line 83
```

globau (approved):

PR #1867, opened 2022-06-15
Bug 1771132 - Consider dropping the "Has Regression Range" field

dklawren:
This PR removes the data that contained the products that cf_has_regression_range was visible for. The field is now disabled and so this is no longer needed.


cgsheeh (approved):
WFM, though "Consider" in the title is not really accurate. Can you make that `drop the "Has Regression Range'" field` before landing?

dklawren:
> WFM, though "Consider" in the title is not really accurate. Can you make that `drop the "Has Regression Range" field` before landing?

Ah good point. The bug originally was to gauge consent from other to remove the field and the bug finally got approval. Will change the commit summary as well as the bug summary .

PR #1871, opened 2022-06-21
Bug 1774351 - Allow adding GitHub projects to "See Also"

dklawren:
Adds the ability to add github project links to the see also field. The links take the form of https://github.com/USER_OR_TEAM_OR_ORGANIZATION_NAME/REPOSITORY_NAME/projects/111 which the regex will check for.


cgsheeh (approved):
LGTM - would be better if we had a test for this though.

dklawren:
> LGTM - would be better if we had a test for this though.

Good idea. I made a bug to add the test separately as we will need to test all of the other possible endpoints as well.
https://bugzilla.mozilla.org/show_bug.cgi?id=1775300

PR #1872, opened 2022-06-21
Bug 1775303 - make `Last Updated DESC` work as a buglist order-by value

globau:


dklawren (approved):

PR #1875, opened 2022-06-24
Bug 1776435: clarify "Update Bug" docs to show bug in URL is mandatory

cgsheeh:


zzzeid (approved):

dklawren (approved):
Thanks for the clarification text. r=dkl

PR #1876, opened 2022-06-24
Bug 1776503 - [SELF-SERVICE] New form to add new version and milestone to select products when a new Firefox release comes out

dklawren:
This new form will allow for quickly adding a new Firefox related milestone/version to the products selected. More details are in the bug.

To test:

```bash
$ docker compose down && docker compose build && docker compose up
```

Set your browsers proxy to localhost:1080. Go to http://bmo.test and login as the admin@mozilla.bugs user. Password is in the README.  You can then add a previous milestone and version to the Firefox product of 'Firefox 100' and '100 Branch' respectively. Then access the new release form from the Administrative menu. You should be able to select Firefox from both selects and add 101 to the test fields. When looking at the versions and milestones list, you should see the new 'Firefox 101' and '101 Branch' values create.

Ping me for any questions.

### DIFF
--- a/Bugzilla/App/BMO/NewRelease.pm
+++ b/Bugzilla/App/BMO/NewRelease.pm
@@ -20,8 +20,8 @@ use Bugzilla::Util qw(fetch_product_versions trim);
 use Scalar::Util qw(blessed);
 
 use constant TEMPLATES => {
-  milestones => {'_default' => '%%% Branch'},
-  versions   => {
+  milestone => {'_default' => '%%% Branch'},
+  version   => {
     'Calendar'      => 'Thunderbird %%%',
     'Chat Core'     => 'Thunderbird %%%',
     'MailNews Core' => 'Thunderbird %%%',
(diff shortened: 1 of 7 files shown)


dklawren:
Need a few more changes before ready for review.

cgsheeh (changes requested):

Bugzilla/App/BMO/NewRelease.pm
```diff
@@ -179,4 +166,18 @@ sub _add_value {
   }
 }
 
+# Helper function to return a fully formatted version or milestone
+sub _get_formatted_value {
+  my ($product, $value, $type) = @_;
+  use Bugzilla::Logging;
+  DEBUG("product: $product");
```
cgsheeh, line 173: Development debugging lines? If we do want to log this we should import `Bugzilla::Logging` at the top.

cgsheeh (changes requested):
LGTM except the debug calls.

Bugzilla/App/BMO/NewRelease.pm
dklawren: Ugh. Totally left that there by mistake. Pushing change to remove.

cgsheeh (approved):

PR #1878, opened 2022-06-30
Bug 1767084 - [WEBHOOKS] Disable webhooks after large number of errors and notify webhook owner.

dklawren:
This PR changes the Push and Webhooks extension to automatically disable webhooks when they reach a specific amount of errors. This keeps our system from continuously trying to send data to endpoints that are not working or are no longer valid. The owner will then receive an email about the hook having issues and can go in and reactivate it after they have fixed the issue hopefully. Admins can still disable hooks themselves if needed just like before.

Please get in touch with any questions or comments.


extensions/Webhooks/template/en/default/admin/params/webhooks.html.tmpl
```diff
@@ -14,7 +14,11 @@
 [%
    param_descs = {
      webhooks_enabled => "Enable or disable the webhooks feature.",
-     webhooks_group => "The name of the group of users who can create webhooks."
+     webhooks_group => "The name of the group of users who can create webhooks.",
      webhooks_any_product_group => "The name of the group of users who can create webhooks for any product."
```
cgsheeh, line 18: Do we need a comma here?

cgsheeh (approved):

extensions/Webhooks/template/en/default/admin/params/webhooks.html.tmpl
dklawren: They are not required by the template engine, but I added here for consistency.

PR #1880, opened 2022-07-06
Bug 1764246 - Address issues with triaged keyword

globau:
Fixes:
- `triaged` keyword should only apply to defects
- allow updates to bugs still using old severity values
- allow the use of N/A on defects (after initially being set to S\d)


dklawren (approved):
LGTM. r=dkl

PR #1889, opened 2022-07-15
Bug 1772408 - Not properly forbidding reporters to set the priority, severity and type fields unless they have editbugs privs for new bugs

dklawren:
These changes allow BMO to filter the priority and severity values on the new bug form. Also priority and severity are not cloned from another bug as we will need to triage the new bug just as if it was entered fresh. Also A new post rule is added to catch updating of the fields by non-permitted users using REST API calls.

### DIFF
--- a/extensions/MozChangeField/lib/Post/TypePriSevEditbugs.pm
+++ b/extensions/MozChangeField/lib/Post/TypePriSevEditbugs.pm
@@ -1,0 +1,28 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+#
+# This Source Code Form is "Incompatible With Secondary Licenses", as
+# defined by the Mozilla Public License, v. 2.0.
+
+package Bugzilla::Extension::MozChangeField::Post::TypePriSevEditbugs;
+
+use 5.10.1;
+use Moo;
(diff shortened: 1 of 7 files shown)


extensions/MozChangeField/lib/Post/TypePriSevEditbugs.pm
```diff
@@ -0,0 +1,28 @@
...
+
+sub evaluate_change {
+  my ($self, $args) = @_;
+  my $changes  = $args->{changes};
+  my $editbugs = $args->{editbugs};
+
+  # If changing the bug_type, severity, or priority, editbugs is required.
+  foreach my $field (qw(bug_type bug_severity priority)) {
+    if (exists $changes->{$field} && !$editbugs) {
+      ThrowUserError('mozchangefield_field_needs_editbugs', {field => $field});
```
globau, line 23: This should return the error and reasons as all other checkers instead of raising the error directly.

globau (changes requested):

extensions/MozChangeField/lib/Post/TypePriSevEditbugs.pm
dklawren: Those type of return values only apply to Pre instances (bug_check_can_change_field) and Post instances are for after changes have been submitted. So Post instances only full stop using Throw{Code|User}Error() or they modify the data in some way.

globau (approved):

PR #1893, opened 2022-07-20
Bug 1780308 - Allow a group of users to delete attachments when needed due to sensitive information

dklawren:
Change the required group permissions from admin to new can_delete_attachments when showing the option to delete an attachment contents.

### DIFF
--- a/qa/t/test_bug_edit.t
+++ b/qa/t/test_bug_edit.t
@@ -40,7 +40,7 @@ check_page_load($sel, q{http://HOSTNAME/editgroups.cgi});
 $sel->title_is("Edit Groups");
 $sel->click_ok("link=Master");
 check_page_load($sel,
-  q{http://HOSTNAME/editgroups.cgi?action=changeform&group=29});
+  q{http://HOSTNAME/editgroups.cgi?action=changeform&group=30});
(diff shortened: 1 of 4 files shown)


qa/t/test_bug_edit.t
```diff
@@ -40,7 +40,7 @@ check_page_load($sel, q{http://HOSTNAME/editgroups.cgi});
 $sel->title_is("Edit Groups");
 $sel->click_ok("link=Master");
 check_page_load($sel,
-  q{http://HOSTNAME/editgroups.cgi?action=changeform&group=29});
+  q{http://HOSTNAME/editgroups.cgi?action=changeform&group=30});
```
cgsheeh, line 43: What does this do?

cgsheeh (approved):

qa/t/test_bug_edit.t
dklawren: Tech debt. I need to make the test script smarter in how it gets the group id for the Master group. This step might actually be able to removed. But what it does it assumes that the last group that was added is the id 29 (now 30) and since I added new permission group in this PR, it also increased the number needed in the test script by 1 :(

PR #1896, opened 2022-07-21
Bug 1512484 - Allow to specify default version per product

dklawren:
This was originally created by @kyoshino in #1394, but since the repo was deleted I had to create a new PR with the patch from the previous PR. So this is his code. I have looked it over and so far it all looks good and works but I will do some further testing to make sure all use cases are working.


dklawren:
A schema update only is part of the PR #1899

dklawren:
Overall this looks good as far as reviewing the code from my point of view. Maybe if someone can take a look and make sure I didnt miss anything obvious

Bugzilla/Product.pm
```diff
@@ -452,6 +444,26 @@ sub _check_default_milestone {
...
+sub _check_default_version {
+  my ($invocant, $version) = @_;
+
+  $version = trim($version);
+
+  if (ref $invocant) {
+
+    # The default version must be one of the existing versions.
+    my $mil_obj
+      = new Bugzilla::Version({name => $version, product => $invocant});
```
globau, line 456: This variable should be `$ver_obj`

Bugzilla/Version.pm
```diff
@@ -113,31 +113,78 @@ sub bug_count {
   return $self->{'bug_count'};
 }
 
+# Copied from Milestone::update()
```
globau, line 116: This comment doesn't provide any value; remove

Bugzilla/Version.pm
```diff
@@ -113,31 +113,78 @@ sub bug_count {
...
   if (exists $changes->{value}) {
-    my $dbh = Bugzilla->dbh;
+
+    # The version value is stored in the bugs table instead of its ID.
     $dbh->do(
       'UPDATE bugs SET version = ?
                   WHERE version = ? AND product_id = ?', undef,
-      ($self->name, $old_self->name, $self->product_id)
+      ($self->name, $changes->{value}->[0], $self->product_id)
+    );
```
globau, line 131: I'm not sure it's correct to update bugs when the _default_ version changes.

I can see the reasoning for Milestone, as that's in the future, but the version field captures the version the bug is applicable to and it probably shouldn't be touched once it has been set by the reporter, even if it's the old default value.

Bugzilla/Version.pm
```diff
@@ -113,31 +113,78 @@ sub bug_count {
...
     );
+    Bugzilla->memcached->clear({table => 'products', id => $self->product_id});
   }
+  $dbh->bz_commit_transaction();
+  Bugzilla->memcached->clear_config();
+
   return $changes;
 }
 
+# Copied from Milestone::remove_from_db()
```
globau, line 147: Also not valuable

template/en/default/admin/versions/confirm-delete.html.tmpl
```diff
@@ -65,36 +65,33 @@
...
+  <table border="0" cellpadding="20" width="70%" bgcolor="red">
+  <tr><td>
+    There
+    [% IF version.bug_count > 1 %]
+      are [% version.bug_count FILTER none %] [%+ terms.bugs %]
+    [% ELSE %]
+      is 1 [% terms.bug %]
+    [% END %]
+    entered for this version! When you delete this version,
+    <b><blink>ALL</blink></b> of these [% terms.bugs %] will be retargeted
```
globau, line 78: Don't use `<blink>`

globau (changes requested):

Bugzilla/Version.pm
globau: I was mistaken; this is when the _version_ is updated, not when the _default version_ is updated.

globau (approved):

PR #1897, opened 2022-07-22
Bug 1780693 - Use selected-control-background-color for the hovered phabricator revision row.

arai-a:
Fore [bug 1780693](https://bugzilla.mozilla.org/show_bug.cgi?id=1780693)

hover


dklawren (approved):
r=dkl

PR #1898, opened 2022-07-22
Bug 1780733 - Sort phabricator revisions by stack order.

arai-a:
This patch sorts the revisions by the stack order.

This patch does not show any graph, so if there's branch, that information is not shown, but at least the revisions table shows revisions in the order that reviewer needs to look.

On the server side, use `edge.search` API to get child revisions (in `_get_children_map` function), and add the info into the response,
by adding `"children": ["D1111", "D1112"],` property to the revision data (in `bug_revisions` function).

I've used dummy data during testing the feature, instead of actually calling `request` function.
So there might be some bug in the `request` function call.

On the client side, use the `children` property to sort the revisions in the stack order (in `sortRevisions` function).

The sort algorithm first sorts the revisions as partially ordered.

If there's no branch, then the result is already sorted, and the remaining operations is no-op.

```
D1
|
D2
|
D3
```

If there's branch, then the result at this point looks like the following:
(note that the line is not shown in the revision table)

```
D1
|
D2
|
+--+
|  |
D3 |
|  |
|  D5
|  |
D4 |
   |
   D6
```

Then split branches into contiguous range (in `splitBranches` function):

```
D1
|
D2
|
+--+
|  |
D3 |
|  |
D4 |
   |
   D5
   |
   D6
```

The sort algorithm adds pseudo root node before performing sort (in `addPseudoRoot` function), so, if there are multiple roots (such as, 2 patches submitted separately, and "Child" relation is not yet set), they're treated as a branch on the pseudo root.

### DIFF
--- a/extensions/PhabBugz/lib/WebService.pm
+++ b/extensions/PhabBugz/lib/WebService.pm
@@ -146,6 +146,8 @@ sub bug_revisions {
   my $error = $SearchResult->validate($response);
   ThrowCodeError('phabricator_api_error', {reason => $error}) if defined $error;
 
+  my $children_map = $self->_get_children_map($response);
+
   my $revision_status_map = {
     'abandoned'       => 'Abandoned',
     'accepted'        => 'Accepted',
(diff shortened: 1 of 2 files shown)


dklawren (approved):
r=dkl

PR #1899, opened 2022-07-22
Bug 1512484 - Allow to specify default version per product (schema only)

dklawren:
This PR only does the DB changes needed to support the landing of #1896. The migration code will add the new column to the products table, and then looking at the version for each product, make an assumption of which version will be set to a default. We can then go into the admin UI later and change the default to something else if needed.


Bugzilla/Install/DB.pm
```diff
@@ -829,6 +829,9 @@ sub update_table_definitions {
   $dbh->bz_add_column('oauth2_client', 'hostname',
     {TYPE => 'varchar(255)', NOTNULL => 1, DEFAULT => "''"});
 
+  # Bug 1512484 - kohei.yoshino@gmail.com
```
cgsheeh, line 832: Maybe this should be changed to `dkl@mozilla.com`? Though I don't know what purpose the email address serves here.

cgsheeh (approved):

Bugzilla/Install/DB.pm
dklawren: Yeah in that file it is consistent with other changes that list the bug id and author. And since Kohei wrote that code and the address is still valid, it is best to leave it like that.

PR #1900, opened 2022-07-25
Bug 1781121 - moving both the status and reviewer columns before the summary for phabricator list make it more readable

dklawren:
This is a simple patch that reorders the columns in the Phabricator Revisions table so that the (possibly long) summary is the last column.


cgsheeh (approved):
Love it!

PR #1903, opened 2022-07-29
Bug 1523847 - Order revisions in the phabricator panel from newest to oldest similar to Phabricator and Lando

dklawren:


extensions/PhabBugz/lib/WebService.pm
```diff
@@ -204,8 +204,8 @@ sub bug_revisions {
     push @revisions, $revision_data;
   }
 
-  # sort by revision id
-  @revisions = sort { $a->{sortkey} <=> $b->{sortkey} } @revisions;
+  # sort by newest to oldest revision id
+  @revisions = sort { $b->{sortkey} <=> $a->{sortkey} } @revisions;
```
arai-a, line 208: The order here is used only as initial order before sorting patches in the stack order.

And the `sortRevision` in client side expects the order to be "old to new".

https://github.com/mozilla-bteam/bmo/blob/ad2c7ec6135c322e8a8df16ed363d6c92a0e5d48/extensions/PhabBugz/web/js/phabricator.js#L116

https://github.com/mozilla-bteam/bmo/blob/ad2c7ec6135c322e8a8df16ed363d6c92a0e5d48/extensions/PhabBugz/web/js/phabricator.js#L28-L30

Also `sortRevisions` sorts the stack order in "root to leaf", that's almost same as "old to new".

So, if we show "new to old" order, the list needs to be reversed after the sort:

https://github.com/mozilla-bteam/bmo/blob/ad2c7ec6135c322e8a8df16ed363d6c92a0e5d48/extensions/PhabBugz/web/js/phabricator.js#L156

arai-a (changes requested):
Thanks!
but this is not right place to reverse the order.

extensions/PhabBugz/lib/WebService.pm
dklawren: Thanks for the review. Somehow I didn't think this would be simple :) Will take a look at it today.

dklawren:
Work being done for this in #1906

PR #1904, opened 2022-07-29
Bug 1782257 - /rest/component endpoint - handle cases where a component name has `/`

dklawren:
Updated the routing for the /rest/component API endpoints to allow for component names with / in the name. If the product has a / in the name, then named query parameters must be used.

### DIFF
--- a/docs/en/rst/api/core/v1/component.rst
+++ b/docs/en/rst/api/core/v1/component.rst
@@ -19,6 +19,7 @@ To get information about the General component under the Firefox product:
    GET /rest/component/Firefox/General
 
 To get information about a component where the product name contains a slash (/) character.
+Named parameters must be used instead of path based parameters.
 
 .. code-block:: text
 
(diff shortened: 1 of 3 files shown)


docs/en/rst/api/core/v1/component.rst
```diff
@@ -18,6 +18,12 @@ To get information about the General component under the Firefox product:
 
    GET /rest/component/Firefox/General
 
+To get information about a component where the product name contains a slash (/) character.
+
+.. code-block:: text
+
+   GET /rest/component?product=Firefox%20%2F%20Bugs&component=General
```
cgsheeh, line 25: We could make the help text clearer by explicitly stating "named parameters must be used instead of path based parameters".

cgsheeh (changes requested):
Please clarify the docs but otherwise this is good. :+1:

cgsheeh (approved):

PR #1905, opened 2022-07-29
Bug 1782314 - When creating a new account using OAuth2 or Github, show warning text about emails being visible and other rules

dklawren:


template/en/default/account/auth/verify_account_creation.html.tmpl
```diff
@@ -7,21 +7,57 @@
...
+      and Communications Terms of Use</a> for more information.
+    </li>
+    <li>
+      You understand that your conduct on this site is subject to both <a href="[% basepath FILTER none %]page.cgi?id=etiquette.html">[% terms.Bugzilla %] Etiquette</a>,
+      and the <a href="https://www.mozilla.org/about/governance/policies/participation/">Mozilla Community Participation Guidelines</a>.
+      By creating an account, you agree to abide by them.
+    </li>
+  </ul>
+</div>
+
```
globau, line 62: As this is an exact copy from the content in `extensions/BMO/template/en/default/account/create.html.tmpl` can we shift it into its own file and include it into both locations?

I'm concerned that in the future we'll update one location and not the other.

globau (changes requested):
Looks good aside from the duplication.

globau (approved):

PR #1909, opened 2022-08-02
Bug 1324011 - "See also" elements should be links in emails

dklawren:
This patch creates links instead of plain text values if the field that is changes is the See Also field. bugmail.html.tmpl is the template generates the HTML version of a bug change email. Text versions do not change.

### DIFF
--- a/extensions/BMO/template/en/default/email/bugmail.html.tmpl
+++ b/extensions/BMO/template/en/default/email/bugmail.html.tmpl
@@ -141,6 +141,8 @@
         <td class="c2">
           [% IF change.field_name == "bug_id" %]
             [% new_value FILTER bug_link(bug, full_url => 1) FILTER none %]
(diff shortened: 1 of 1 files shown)


cgsheeh (approved):
So was the old code was displaying the link as text without the `a` tag, or just displaying the bug number? I would imagine this wouldn't work with just the bug number as it would look like `<a href="12345">`.

Assuming you've tested this it's r+. :)

dklawren:
> So was the old code was displaying the link as text without the `a` tag, or just displaying the bug number? I would imagine this wouldn't work with just the bug number as it would look like `<a href="12345">`.
> 
> Assuming you've tested this it's r+. :)

It was displaying the fully qualified link to the local bug but just as text without anchor tags similar to external links. So this fixes both cases.

PR #1910, opened 2022-08-03
Bug 1782806 - move of bug to different component shouldn't drop its keywords

dklawren:
This issue was only occurring when changing component via the REST API and not through the UI, This was due the team name changing and using del_keyword(). del_keyword() had not yet initialized $self->{keyword_objects} to the real list so it assumed it was empty which essentially removed any keywords that were there. I updated del_keyword() to initialize by calling $self->keyword_objects in the for each loop which fixes this issue for API calls.

I also remove two unused variables in the extension hook as well as return early if not using triaged keyword for the product.


globau (approved):
ty!

PR #1912, opened 2022-08-09
Bug 1783044 - Unable to link profile to People due to error page not being rendered properly

dklawren:
When running code under fully native Mojo code instead of the Mojo wrapper around legacy cgi scripts (all files with .cgi) we should not use the cgi method of printing the HTTP headers. The OAuth2 authentication process uses native Mojo code but yet calls a utility function in Token.pm that assumes it is running under a cgi environment. So when it tries to do `print $cgi->header` an error is thrown about the header() call masking the actual page that is supposed to be displayed explaining the real error. 

So I have update the utility function in Token.pm to check if it is running under Mojo native or cgi and return a data structure instead if running Mojo. Otherwise it prints the warning page as usual.

Provider.pm is updated to receive the result from the utility function and display a similar warning except it renders it itself. There may be other cases where something similar needs to be done but this one change should help this issue in particular.


cgsheeh (approved):
r=sheehan

PR #1917, opened 2022-08-16
Bug 1524850 - Documented password guidelines do not match implementation

dklawren:
* Removes dependency on Data::Password::passwdqc
* Reverts to using simple regex checking of passwords
* Increase min password length to 12 characters
* Update password tests to work with different password requirements


Bugzilla/User.pm
```diff
@@ -2877,6 +2877,36 @@ sub user_id_to_login {
...
+
+  my $complexity_level = Bugzilla->params->{password_complexity};
+  if ($complexity_level eq 'bmo') {
+    my $features = 0;
+
+    $features++ if $password =~ /[a-z]/;
+    $features++ if $password =~ /[A-Z]/;
+    $features++ if $password =~ /[0-9]/;
+    $features++ if $password =~ /[^A-Za-z0-9]/;
+    $features++ if length($password) > 12;
```
globau, line 2904: This is different from the stated requirements of "Password must contain at least one letter, a number and a special character".

A password of `abcdefghijkl1` passes the regex tests but shouldn't.

This also drops support for passphrases; we should continue to allow them.

Bugzilla/Auth/Verify/DB.pm
```diff
@@ -71,27 +71,27 @@ sub check_credentials {
...
-      $user->set_password_change_reason(
-        "You must change your password for the following reason: $reason");
-
-      # Remove disabled text if user was previously disabled due to inactivity
-      if ($old_reason eq 'Inactive Account') {
-        $user->set_disabledtext('');
-        $user->set_disable_mail(0);
-      }
-
-      $user->update();
```
globau, line 91: We shouldn't remove the logic that forces weak passwords to be updated.

Bugzilla/Auth/Verify/DB.pm
```diff
@@ -71,27 +71,27 @@ sub check_credentials {
...
+    my $check = validate_password_check($password);
+    if ($check) {
+      return {
+        failure    => AUTH_ERROR,
+        user_error => $check,
+        details    => {locked_user => $user}
+      };
```
globau, line 80: If I'm reading this right won't this block users from logging in with the correct password if that password doesn't match the current password requirements?

conf/checksetup_answers.txt
```diff
@@ -1,6 +1,6 @@
 $answer{'ADMIN_EMAIL'}          = 'admin@mozilla.bugs';
 $answer{'ADMIN_OK'}             = 'Y';
-$answer{'ADMIN_PASSWORD'}       = 'password01!';
+$answer{'ADMIN_PASSWORD'}       = 'password012!';
```
globau, line 3: You'll need to update the password in the README

Bugzilla/Constants.pm
```diff
@@ -283,8 +283,8 @@ use constant AUTH_NO_SUCH_USER => 5;
 use constant AUTH_LOCKOUT      => 6;
 
 # The minimum and maximum length a password will have.
-# BMO uses 8 characters.
-use constant USER_PASSWORD_MIN_LENGTH => 8;
+# BMO uses 12 characters.
+use constant USER_PASSWORD_MIN_LENGTH => 12;
```
globau, line 287: We shouldn't increase the min password length as part of this change.

globau (changes requested):

dklawren:
Need to fix couple more things.

qa/config/generate_test_data.pl
```diff
@@ -537,7 +537,7 @@ BEGIN
 
       Bugzilla::User->create({
         login_name    => $watch_user,
-        cryptpassword => Bugzilla->passwdqc->generate_password(),
+        cryptpassword => generate_random_password(14), # Increase to pass password complexity
```
globau, line 540: This comment doesn't make sense

qa/t/webservice_user_create.t
```diff
@@ -111,7 +111,7 @@ foreach my $rpc ($jsonrpc, $xmlrpc) {
         password  => PASSWORD_TOO_SHORT
       },
       error =>
-        'The password does not meet our security requirements for the following reason: too short',
+        'The password must be at least 12 characters long.',
```
globau, line 114: Use the const instead of hardcoding the length

Bugzilla/User.pm
```diff
@@ -2877,6 +2877,43 @@ sub user_id_to_login {
   return $login || '';
 }
 
+sub validate_password {
```
globau, line 2880: I loved the clarity that `assert_password_is_secure` and `assert_passwords_match` provides.

What do you feel about having a single function `assert_valid_password` that always throws?  Callers can use `eval` (or ideally `try/catch`) to handle errors if required (we're already doing that in one place).

Bugzilla/Install.pm
```diff
@@ -514,14 +514,9 @@ sub _prompt_for_password {
...
     chomp $pass2;
-    my $pwqc = Bugzilla->passwdqc;
-    my $ok   = $pwqc->validate_password($password);
-    if (!$ok) {
-      print "\n", $pwqc->reason, "\n";
-      undef $password;
-    }
-    elsif ($password ne $pass2) {
-      print "\n", "passwords do not match\n";
+    eval { validate_password($password, $pass2); };
```
globau, line 517: `try/catch` would be clearer

Bugzilla/User.pm
```diff
@@ -2877,6 +2877,43 @@ sub user_id_to_login {
...
+  elsif ((defined $matchpassword) && ($password ne $matchpassword)) {
+    return 'password_mismatch';
+  }
+
+  my $complexity_level = Bugzilla->params->{password_complexity};
+  if ($complexity_level eq 'bmo') {
+    my $features = 0;
+
+    # A password can be 3 (or more) or more 4 (or more) letter words
+    if ($password =~ /\W{4}\s+\W{4}\s+\W{4}/) {
```
globau, line 2901: This regex is wrong; it's matching on *non-word* characters (so `"""" """" """"` passes but `test test test` doesn't), isn't anchored to the start or end of the string, only matches words that are exactly four chars long, and Perl's definition of "word" (`[0-9a-zA-Z_]`) shouldn't be used as it's English only and doesn't include chars such as `'` that exist in words.

We probably want something like `/^\S{4,}(?:\s+\S{4,}){2,}/`

Please add tests for lots of variations of passwords.

globau (changes requested):

qa/config/generate_test_data.pl
dklawren: I will make it more descriptive of what is going on. Basically the way generate_random_password was working, I was getting validation errors since sometimes it would not include enough of the qualities to pass. Making the password 14 characters long instead of the default 12, made it then pass each time when creating a new user account.

Bugzilla/User.pm
dklawren: Thanks for catching that. I meant to use \w instead of \W. You regex definitely looks better so I will try that and also update the automated tests.

Bugzilla/User.pm
dklawren: Sounds good to me. Will do that.

qa/t/webservice_user_create.t
```diff
@@ -13,13 +13,17 @@ use strict;
...
+use Test::More tests => 99;
 my ($config, $xmlrpc, $jsonrpc, $jsonrpc_get) = get_rpc_clients();
 
 use constant NEW_PASSWORD => 'UiX1Shuuchid';
 use constant NEW_FULLNAME => 'WebService Created User';
 
-use constant PASSWORD_TOO_SHORT => 'a';
+use constant PASSWORD_TOO_SHORT     => 'a';
+use constant PASSWORD_TOO_FEW_WORDS => 'bip bop boop';
+use constant PASSWORD_NOT_COMPLEX   => 'abcdefghijk1';
```
globau, line 24: This is missing a test for a valid pass phrase.

There should also be tests that exercise all of the possible valid combinations; eg. there isn't anything here that tests for symbols.

Perhaps this would be easier as a unit test?

globau (changes requested):

globau (changes requested):

dklawren:
Just realized an issue with my new changes so hold off on review.

dklawren:
Now ready for re-review.

globau (approved):

PR #1919, opened 2022-08-18
Bug 1555583 - Remove whine queries owned by disabled users no longer with the company

dklawren:
Current behavior is we ignore whether an user is disabled or not and continue to send their whine events indefinitely. As you can imagine, this not ideal. This patch just skips the current event, resets the timer and does not send out the email. Then when the account is no longer disabled, the next schedule event will execute and send the email.

The other option was to delete the events when a user has been disabled, but that is not ideal as some accounts only get disabled temporarily and we do not want them to have to recreate the whines all over again.

### DIFF
--- a/whine.pl
+++ b/whine.pl
@@ -303,6 +303,16 @@ sub get_next_event {
 while (my $event = get_next_event) {
 
   my $eventid = $event->{'eventid'};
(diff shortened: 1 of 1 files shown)


cgsheeh (approved):
Makes sense. :+1:

PR #1920, opened 2022-08-26
Bug 1783576 - Add field to REST API that returns timestamp of bug last change by an actual person

dklawren:
This PR adds a new field that can be returned if explicitly requested with the get bug API endpoint. It is NOT returned by default. It is called `last_change_time_non_bot` and is the last time the bug was changed by a non-bot Bugzilla account. There is a new config parameter that allows storing a comma or space separated list of email addresses that are considered *bot* accounts. Any account in the list, their change will be ignored when returning the last_change_time_non_bot value. A test cases has also been added that tests to make sure the right timestamps are returned depending on which account updated the bug last.

### DIFF
--- a/Bugzilla/Bug.pm
+++ b/Bugzilla/Bug.pm
@@ -267,28 +267,28 @@ use constant MAX_LINE_LENGTH => 254;
 # of Bugzilla. (These are the field names that the WebService and email_in.pl
 # use.)
 use constant FIELD_MAP => {
-  blocks                => 'blocked',
-  commentprivacy        => 'comment_is_private',
-  creation_time         => 'creation_ts',
-  creator               => 'reporter',
-  description           => 'comment',
-  depends_on            => 'dependson',
-  dupe_of               => 'dup_id',
-  id                    => 'bug_id',
-  is_confirmed          => 'everconfirmed',
-  is_cc_accessible      => 'cclist_accessible',
-  is_creator_accessible => 'reporter_accessible',
-  last_change_time      => 'delta_ts',
+  blocks                   => 'blocked',
+  commentprivacy           => 'comment_is_private',
+  creation_time            => 'creation_ts',
+  creator                  => 'reporter',
+  description              => 'comment',
+  depends_on               => 'dependson',
+  dupe_of                  => 'dup_id',
+  id                       => 'bug_id',
+  is_confirmed             => 'everconfirmed',
+  is_cc_accessible         => 'cclist_accessible',
+  is_creator_accessible    => 'reporter_accessible',
+  last_change_time         => 'delta_ts',
   last_change_time_non_bot => 'delta_ts_non_bot',
-  comment_count         => 'longdescs.count',
-  platform              => 'rep_platform',
-  regressions           => 'regresses',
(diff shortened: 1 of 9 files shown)


Bugzilla/Bug.pm
```diff
@@ -279,6 +279,7 @@ use constant FIELD_MAP => {
   is_cc_accessible      => 'cclist_accessible',
   is_creator_accessible => 'reporter_accessible',
   last_change_time      => 'delta_ts',
+  last_change_time_non_bot => 'delta_ts_non_bot',
```
cgsheeh, line 282: Nit: it seems like the entire `FIELD_MAP` should have it's indentation updated here.

Bugzilla/Bug.pm
```diff
@@ -4411,6 +4412,60 @@ sub bug_alias_to_id {
     undef, $alias);
 }
 
+# Last changed timestamp of a non-bot user
+sub delta_ts_non_bot {
+  my ($self) = @_;
+  return $self->{delta_ts_non_bot} if exists $self->{delat_ts_non_bot};
```
cgsheeh, line 4418: Typo: `$self->{delat_ts_non_bot}` -> `$self->{delta_ts_non_bot}`

I suppose this means this is missing test coverage.

cgsheeh (changes requested):
So I would assume this change means that all Bugzilla automation users now need to effectively "register" with BMO admins to be added to the automation user list. Not sure how many automation users we know of but just calling out the manual/non-self-service nature of that change. 

Have we considered adding something to the user settings such that we can mark accounts as bots? Then developers could create their bot accounts and set them as such without intervention from admins. It could be a setting hidden behind `editfields` or something to avoid allowing anyone to create a known bot account. If there aren't many bot accounts on BMO it might not be worth the effort, just throwing it out there.

Bugzilla/Bug.pm
dklawren: Nah just means that the cached value is never used so technically would still work but be less performant. Sorry for missing the misspelling.

dklawren:
> So I would assume this change means that all Bugzilla automation users now need to effectively "register" with BMO admins to be added to the automation user list. Not sure how many automation users we know of but just calling out the manual/non-self-service nature of that change.
> 
> Have we considered adding something to the user settings such that we can mark accounts as bots? Then developers could create their bot accounts and set them as such without intervention from admins. It could be a setting hidden behind `editfields` or something to avoid allowing anyone to create a known bot account. If there aren't many bot accounts on BMO it might not be worth the effort, just throwing it out there.

Sorry didn't reply to this sooner. This is a fair point and something that has been considered in the past. There is bug for it already I think. But yes having a setting in the profile designating the account as a bot account would be a good thing to have. This is basically done already in Phabricator as well. For now Marco and his group who manage auto-nag will just give us the initial list of bots to add to the config parameter. Then as a separate bug we can look at adding the checkbox to a profile account and manage the list that way instead. I am not sure people will remember or know to do this themselves when they create the account so I think admin intervention will still be needed.

cgsheeh (approved):
Makes sense!

PR #1921, opened 2022-08-26
Bug 1787294 - Allow optional tolerance in the /rest/user endpoint when querying users that are not exist

dklawren:
This PR adds  a new `permissive` parameter which is similar to what we already have for getting a list of bugs. This parameter allows for ignoring errors when passing in a list of names to query for user information. If one or more cause an error such as not found, etc. it will just add the error to a `faults` list and return the ones that were found. This does not change the API in a way that would break any current clients so it should not be an issue.


cgsheeh (approved):
LGTM.

PR #1923, opened 2022-08-31
Bug 1787966 - Refactor duplicate fields for the user's last activity in /rest/user endpoint

dklawren:


cgsheeh (approved):
Yay for less tech debt. :)

PR #1926, opened 2022-09-07
Bug 1789654 - User matching for user related fields in the web UI is broken due to regression caused by bug 1787954

dklawren:
I failed to notice that the /rest/user/suggest API endpoint internally calls the same function that the /rest/user endpoint uses which I changed the data passed in bug 1787954.

So this changes /rest/user/suggest to match the data structure needed by the webservice_user_get hook function and also removed some redundant code.

### DIFF
--- a/Bugzilla/WebService/User.pm
+++ b/Bugzilla/WebService/User.pm
@@ -153,7 +153,7 @@ sub suggest {
   return {users => []} if length($s) < 3;
 
   my $dbh    = Bugzilla->dbh;
-  my @select = ('userid AS id', 'realname AS real_name', 'login_name AS name');
+  my @select = ('userid AS id');
   my $order  = 'last_activity_ts DESC';
   my $where;
   state $have_mysql = $dbh->isa('Bugzilla::DB::Mysql');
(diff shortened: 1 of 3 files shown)


Bugzilla/WebService/User.pm
```diff
@@ -183,23 +183,32 @@ sub suggest {
...
-    . join(', ', @select)
-    . " FROM profiles WHERE $where ORDER BY $order LIMIT 25";
-  my $results = $dbh->selectall_arrayref($sql, {Slice => {}});
+  my $results = $dbh->selectall_arrayref(
+    "SELECT "
+      . join(', ', @select)
+      . " FROM profiles WHERE $where ORDER BY $order LIMIT 25",
+    {Slice => {}}
+  );
+  my $user_objects = Bugzilla::User->new_from_list([map { $_->{id} } @$results]);
```
globau, line 192: When originally authored there was a reasonable performance hit using the User objects for the number of objects we create.  Please ensure you don't regress performance.

cgsheeh (approved):

cgsheeh:
Could we add a test that would catch this and similar regressions?

Bugzilla/WebService/User.pm
dklawren: Interestingly, this does not impact performance from what we have currently.  Reason being is that the extensions hooks for the Review and Gravatar extensions already loaded full user objects for the entire list of user ids in order to add their data to the results. This was before bug 1787954. So the changes are just doing it earlier in the code path. 

That being said, we could definitely make this perform better than it does currently as a different bug and also add some testing for performance regressions which we do not have at the moment for any aspect of BMO afaict.

globau: Thanks - makes sense.

PR #1928, opened 2022-09-13
Bug 1790623 - Extend the Triage Owners report to include an indication if the Triage Owner is a member of the default security group

dklawren:
This PR adds a new column to the existing triage owners reporter. The column shows Yes or No if the triage owner for the component is in the default security group for the product the component belongs to. Triage.pm is the module that generates the data for the template and does the heavy lifting of checking for membership, etc. Only viewers who are in the mozilla-employee-confidential group should be able to see the column in the table. 

Thanks!

### DIFF
--- a/extensions/BMO/template/en/default/pages/triage_owners.html.tmpl
+++ b/extensions/BMO/template/en/default/pages/triage_owners.html.tmpl
@@ -102,6 +102,13 @@
           <tr>
             <th>[% r.product FILTER html %]</th>
             <th>Owner</th>
+            [% IF r.in_prod_security_group.defined %]
+              <th>
+                <span title="Triage owner is a member of the default security group for this product">
+                  Prod Sec Group Member
+                </span>
+              </th>
+            [% END %]
             [% FOREACH count IN r.bug_counts %]
               <th class="count">
                 [% IF count.key == "total"%]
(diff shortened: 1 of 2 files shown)


extensions/BMO/template/en/default/pages/triage_owners.html.tmpl
```diff
@@ -102,6 +102,13 @@
           <tr>
             <th>[% r.product FILTER html %]</th>
             <th>Owner</th>
+            [% IF r.in_prod_security_group.defined %]
```
cgsheeh, line 105: Shouldn't we also check that the value is `'Yes'` here?

cgsheeh (changes requested):

extensions/BMO/template/en/default/pages/triage_owners.html.tmpl
dklawren: > Shouldn't we also check that the value is `'Yes'` here?

No, this just determines if the column, with its header, will be displayed or not if the value is defined.

cgsheeh: Ah, makes sense.

cgsheeh (approved):

PR #1934, opened 2022-10-03
Bug 1726504 - Markdown tables are rendered incorrectly

nandaleite:
## Description 

Update markdown table CSS to render correctly in any width.

Final result:
markdown-tables-desktop


dklawren (approved):
Looks good and thanks for the fix. r=dl;

PR #1938, opened 2022-10-13
Bug 1795160 - OAuth2 client admin UI form field for description is too small and should be size of DB column

dklawren:
- Increase size of field and maxsize for creation and editing
- Cleaned up formatting as well for edit form.


cgsheeh (approved):
LGTM. Before merging, should this be in CSS somewhere?

dklawren:
> LGTM. Before merging, should this be in CSS somewhere?

CSS is for presentation and cannot determine the max characters allowed for a input element that I know of. There may be a way I do not know about. I know you can set it via JS though.

PR #1943, opened 2022-10-20
Bug 1796692 - Update triage owners report to alter permitted status values depending on if the product belongs to core-security

dklawren:
This PR mainly tweaks the way Yes/No is displayed if the products default security group is core-security. Other product/component rows will determine Yes/No the previous way. The core security team is a small group who then triage the bugs to other products/components which use other *-core-security groups. So if the triager owner is in one or more of the *-core-security groups (but maybe not in core-security itself) then treat it like they are in core-security.

### DIFF
--- a/extensions/BMO/lib/Reports/Triage.pm
+++ b/extensions/BMO/lib/Reports/Triage.pm
@@ -14,6 +14,7 @@ use warnings;
 use Bugzilla::Component;
 use Bugzilla::Constants;
 use Bugzilla::Error;
+use Bugzilla::Group;
 use Bugzilla::Product;
 use Bugzilla::User;
 use Bugzilla::Util qw(detaint_natural trim url_quote);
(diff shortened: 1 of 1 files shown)


cgsheeh (approved):
Just to clarify what this does: when the report is shown this will basically make the answer to the question "will this person be able to see core security bugs once they have been triaged into the correct component" be "yes"?

dklawren:
> Just to clarify what this does: when the report is shown this will basically make the answer to the question "will this person be able to see core security bugs once they have been triaged into the correct component" be "yes"?

Basically yes. Since members of the core-security team will see the initial bug reports only, they will triage the new bugs and then assign them to a different *-core-security group as needed which the real triage owner will then be able to see if they are in the group. So we are just stating 'Yes' on the report of the triage owner is in any of the *-core-security groups as a shortcut.

PR #1948, opened 2022-10-25
Bug 1791629: teach `phab-bot` to set `#release-managers!` review when uplift form submitted

cgsheeh:
Teaches `phab-bot` to set `#release-managers!` review when
the uplift request form is submitted. Parses feed story text
for an uplift request change and uses the `stackGraph` field
to request review on each revision in the stack.

### DIFF
--- a/extensions/PhabBugz/lib/Feed.pm
+++ b/extensions/PhabBugz/lib/Feed.pm
@@ -400,7 +400,7 @@ sub process_revision_change {
     );
 
     # Get `#release-managers` review group.
-    my $release_managers_phid = Bugzilla::Extension::PhabBugz::Project
+    my $release_managers_group = Bugzilla::Extension::PhabBugz::Project
       ->new_from_query({name => 'release-managers'});
 
     if (!$release_managers_group) {
--- a/extensions/PhabBugz/lib/Feed.pm
+++ b/extensions/PhabBugz/lib/Feed.pm
@@ -401,8 +401,17 @@ sub process_revision_change {
 
     # Get `#release-managers` review group.
     my $release_managers_phid = Bugzilla::Extension::PhabBugz::Project
-      ->new_from_query({name => 'release-managers'})
-      ->phid;
+      ->new_from_query({name => 'release-managers'});
+
+    if (!$release_managers_group) {
+      WARN(
+        "Uplift request change detected but `#release-managers` was " .
+        "not found on Phabricator."
+      );
+      return;
+    }
+    
+    my $release_managers_phid = $release_managers_group->phid;
 
     # Request `#release-managers` review on each revision.
     foreach my $stack_revision_phid (keys %{$revision->stack_graph_raw}) {
--- a/extensions/PhabBugz/lib/Feed.pm
+++ b/extensions/PhabBugz/lib/Feed.pm
@@ -422,11 +422,6 @@ sub process_revision_change {
 
       # Add `#release-managers!` review and set revision status.
       $stack_revision->add_reviewer("blocking($release_managers_phid)");
-
-      if ($stack_revision->status ne 'needs-review') {
-        $stack_revision->set_status('request-review');
-      }
-
       $stack_revision->update();
 
       INFO("Requested #release-managers review of $stack_revision_phid.");
--- a/extensions/PhabBugz/lib/Feed.pm
+++ b/extensions/PhabBugz/lib/Feed.pm
@@ -351,6 +351,13 @@ sub group_query {
   }
 }
 
+sub _is_uplift_request_form_change {
+  # Return `true` if the story text is an uplift request change.
+  my ($story_text) = @_;
+
+  return $story_text =~ /\s+uplift request field/;
+}
+
 sub process_revision_change {
   state $check = compile($Invocant, Revision, LinkedPhabUser, Str);
   my ($self, $revision, $changer, $story_text) = $check->(@_);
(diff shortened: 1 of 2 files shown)


extensions/PhabBugz/lib/Feed.pm
```diff
@@ -351,6 +351,13 @@ sub group_query {
   }
 }
 
+sub _is_uplift_request_form_change {
+  # Return `true` if the story text is an uplift request change.
+  my ($story_text) = @_;
+
+  return $story_text =~ /\s+uplift request field/;
+}
+
```
dklawren, line 360: Seems overkill to have a subroutine to perform a string regex and nothing else. 

Maybe just have

if ($story_text =~ /\s+uplift request field/) {
  ...
}

It's fine either way.

extensions/PhabBugz/lib/Feed.pm
```diff
@@ -384,6 +391,41 @@ sub process_revision_change {
...
+      "requesting `#release-managers` review."
+    );
+
+    # Get `#release-managers` review group.
+    my $release_managers_phid = Bugzilla::Extension::PhabBugz::Project
+      ->new_from_query({name => 'release-managers'})
+      ->phid;
+
+    # Request `#release-managers` review on each revision.
+    foreach my $stack_revision_phid (keys %{$revision->{stack_graph_raw}}) {
```
dklawren, line 408: Since you added this to Revision.pm as an attribute then you can just access it as a normal method.
It is preferred over accessing the internal data directly.


    foreach my $stack_revision_phid (keys %{$revision->stack_graph_raw}) {

extensions/PhabBugz/lib/Feed.pm
```diff
@@ -384,6 +391,41 @@ sub process_revision_change {
...
+    INFO(
+      "Uplift request form submitted on $revision_phid, " .
+      "requesting `#release-managers` review."
+    );
+
+    # Get `#release-managers` review group.
+    my $release_managers_phid = Bugzilla::Extension::PhabBugz::Project
+      ->new_from_query({name => 'release-managers'})
+      ->phid;
+
```
dklawren, line 406: We may need some code here to make sure the new_from_query() was successful. In case the release-managers project does not yet exist or was removed. 

Should we skip the feed entry? Error out and keep trying until success? As it is written, if the new_from_query() does not find the project, and you try to access ->phid, it will crash.

extensions/PhabBugz/lib/Feed.pm
```diff
@@ -384,6 +391,41 @@ sub process_revision_change {
...
+      # Add `#release-managers!` review and set revision status.
+      $stack_revision->add_reviewer("blocking($release_managers_phid)");
+
+      if ($stack_revision->status ne 'needs-review') {
+        $stack_revision->set_status('request-review');
+      }
+
+      $stack_revision->update();
+
+      INFO("Requested relman review of $stack_revision_phid.");
```
dklawren, line 423: nit-pick: Change relman to release-managers.

extensions/PhabBugz/lib/Feed.pm
dklawren, line 408: Since you added this to Revision.pm as an attribute then you can just access it as a normal method.
It is preferred over accessing the internal data directly.


    foreach my $stack_revision_phid (keys %{$revision->stack_graph_raw}) {

dklawren (changes requested):

extensions/PhabBugz/lib/Feed.pm
cgsheeh: I changed this to just skip the message if `release-managers` doesn't exist.

extensions/PhabBugz/lib/Feed.pm
```diff
@@ -384,6 +391,50 @@ sub process_revision_change {
...
+  # Change is a submission of the uplift request form.
+  if (_is_uplift_request_form_change($story_text)) {
+    my $revision_phid = $revision->phid;
+    INFO(
+      "Uplift request form submitted on $revision_phid, " .
+      "requesting `#release-managers` review."
+    );
+
+    # Get `#release-managers` review group.
+    my $release_managers_phid = Bugzilla::Extension::PhabBugz::Project
```
dklawren, line 403: Need to change $release_managers_phid to $release_managers_group

dklawren (changes requested):

dklawren (changes requested):
OK so I was able to do some functionality testing this afternoon and ran into an error condition that we will need to figure out. When I reviewed and accepted a test revision, I then created an uplift request. Once I created and save the uplift request, phabbugz then tried to update the revision back to needs-review and failed. It is able to add a project but it cannot update the status since the phab-bot user is not the author of the revision.

Error:
```
FATAL | Bugzilla.Extension.PhabBugz.Feed | Code: ERR-CONDUIT-CORE Reason: Validation errors: - You can not request review of this revision because you are not the author of the revision
and it is not currently a draft.  at Bugzilla/Error.pm line 99.
```

Unfortunately we cannot update the revision as the revision author since we don't have an api key for them. And I don't think making phab-bot an admin would work anyway and we don't want to do that. Thoughts?

cgsheeh:
> OK so I was able to do some functionality testing this afternoon and ran into an error condition that we will need to figure out. When I reviewed and accepted a test revision, I then created an uplift request. Once I created and save the uplift request, phabbugz then tried to update the revision back to needs-review and failed. It is able to add a project but it cannot update the status since the phab-bot user is not the author of the revision.
> 
> Error:
> 
> ```
> FATAL | Bugzilla.Extension.PhabBugz.Feed | Code: ERR-CONDUIT-CORE Reason: Validation errors: - You can not request review of this revision because you are not the author of the revision
> and it is not currently a draft.  at Bugzilla/Error.pm line 99.
> ```
> 
> Unfortunately we cannot update the revision as the revision author since we don't have an api key for them. And I don't think making phab-bot an admin would work anyway and we don't want to do that. Thoughts?

I think we can actually get away with removing the bit where we add `needs-review`. In theory the revision will be `needs-review` already, so we should just be able to add `#release-managers!` review and let the developer handle setting `needs-review` again if the status is set to something else.

cgsheeh:
> OK so I was able to do some functionality testing this afternoon and ran into an error condition that we will need to figure out. When I reviewed and accepted a test revision, I then created an uplift request. Once I created and save the uplift request, phabbugz then tried to update the revision back to needs-review and failed. It is able to add a project but it cannot update the status since the phab-bot user is not the author of the revision.
> 
> Error:
> 
> ```
> FATAL | Bugzilla.Extension.PhabBugz.Feed | Code: ERR-CONDUIT-CORE Reason: Validation errors: - You can not request review of this revision because you are not the author of the revision
> and it is not currently a draft.  at Bugzilla/Error.pm line 99.
> ```
> 
> Unfortunately we cannot update the revision as the revision author since we don't have an api key for them. And I don't think making phab-bot an admin would work anyway and we don't want to do that. Thoughts?

Also, thank you for doing functionality test of this PR! Much appreciated. :)

dklawren (approved):
I made a similar change earlier in my testing and this fixed the issue. so r=dkl

PR #1950, opened 2022-10-27
Bug 1196291 - Prevent users without editbugs from commenting on resolved bugs, unless they are already involved in the bug

dklawren:
This PR adds a new module that blocks adding a comment to a CLOSED bug if the user is not in editbugs, or does not have a direct role in the bug such as reporter, assignee, or qa contact. The template will show text in place of the comment field if they are blocked with a reason. This also blocks adding comments via the API if the user is not allowed. Let me know if you have any questions.


cgsheeh (approved):
LGTM.

PR #1951, opened 2022-10-31
Bug 1798382 - Bugzilla queries that involve change history are rejected by MySQL when using date shortcuts and  the calculated timestamp is an invalid date

dklawren:
Unless you can think of a more clever way to fix, this does fix this rare occurring issue without requesting the user to just run it again the next day.

Suppose another way would be to use DataTime to validate the timestamp and catch the error if one occurs. Decrement the day until the error goes away.

### DIFF
--- a/Bugzilla/Search.pm
+++ b/Bugzilla/Search.pm
@@ -22,6 +22,7 @@ use Bugzilla::Error;
 use Bugzilla::Field;
 use Bugzilla::Group;
 use Bugzilla::Keyword;
+use Bugzilla::Logging;
 use Bugzilla::Search::Clause;
 use Bugzilla::Search::ClauseGroup;
 use Bugzilla::Search::Condition qw(condition);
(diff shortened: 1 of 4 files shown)


dklawren:
Ugh. Closing. Tree had other stuff that wasn't supposed to be there.

Bugzilla/Search.pm
```diff
@@ -2371,6 +2374,19 @@ sub SqlifyDate {
         return sprintf("%4d-%02d-01 00:00:00", $year + 1900, $month + 1);
       }
       else {
+        # If day is 31 but month only has 30 days, then set day to 30
```
globau, line 2377: It would be better to use a library that performs correct date math rather than hand rolling, especially when it comes to handling leap years.

I'd also expect dates that overflow months to map to the following month; eg. 2022-11-31 -> 2022-12-01

PR #1952, opened 2022-11-01
Bug 1798382 - Bugzilla queries that involve change history are rejected by MySQL when using date shortcuts and  the calculated timestamp is an invalid date

dklawren:
Removes all manual date math and instead uses DateTime for date math.

### DIFF
--- a/Bugzilla/Search.pm
+++ b/Bugzilla/Search.pm
@@ -32,6 +32,7 @@ use Bugzilla::Util;
 use Data::Dumper;
 use Date::Format;
 use Date::Parse;
+use DateTime;
 use List::MoreUtils qw(all any firstidx part uniq);
 use POSIX qw(INT_MAX);
 use Scalar::Util qw(blessed);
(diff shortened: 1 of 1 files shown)


Bugzilla/Search.pm
```diff
@@ -2318,10 +2319,10 @@ sub _date_translate {
 sub SqlifyDate {
   my ($str) = @_;
   my $fmt = "%Y-%m-%d %H:%M:%S";
+  my $date = DateTime->now();
```
globau, line 2322: Should we be using the database's current time rather than the webheads?

dklawren: All pods should be in sync time wise but there could be some difference. I thought about it and was trying to keep the extra SQL calls to a minimum to get the current timestamp each time SqlifyDate is called. But maybe in would be insignificant? Thoughts?

dklawren: I take that back, the MySQL server itself is a separate instance and may not be time synced.

globau: Fair enough, let's leave it client-side.

globau (approved):

PR #1956, opened 2022-11-03
Bug 1798961 - Due to bug 1798382, a bug was introduced that cause all full date strings to be converted to todays date

dklawren:


cgsheeh (approved):
I guess this is kind of analogous to `str2time($str)` from before de802b26a2ed811d6fc63c3be26b51e7fb81bcad?

This function is a bit tough to follow IMO, it could use a cleanup.

dklawren:
> I guess this is kind of analogous to `str2time($str)` from before [de802b2](https://github.com/mozilla-bteam/bmo/commit/de802b26a2ed811d6fc63c3be26b51e7fb81bcad)?
> 
> This function is a bit tough to follow IMO, it could use a cleanup.

Yeah sorry I didnt add a description initially explaining what was happening. datetime_from() is a function in Bugzilla::Util that is used in numerous places around the code base. It basically does similar to str2time() but it returns a DateTime object instead of a unix epoch time. Also it validates that the time str is correct and formatted properly otherwise returns undefined and we throw an error in that case.  $date->strftime($fmt) will return the timestamp in the default format. I also remove the dependency on Date::Format and Date::Parse since they are no longer needed in Search.pm

PR #1970, opened 2022-11-21
Bug 1801309 - Only create new attachments when a pull request is initially opened so as to not create possible duplicates

dklawren:


cgsheeh (approved):
:+1:

PR #1973, opened 2022-11-24
no bug - Quoting groups in remove_idle_group_members script

dlactin:
`groups` is a reserved word in MySQL 8 and needs to be quoted for this script to support MySQL 8.

### DIFF
--- a/scripts/remove_idle_group_members.pl
+++ b/scripts/remove_idle_group_members.pl
@@ -37,7 +37,7 @@ my $expired = $dbh->selectall_arrayref(
   q{SELECT DISTINCT profiles.userid AS user_id,
             groups.id AS group_id
        FROM profiles JOIN user_group_map ON profiles.userid = user_group_map.user_id
-            JOIN groups ON user_group_map.group_id = groups.id
+            JOIN `groups` ON user_group_map.group_id = groups.id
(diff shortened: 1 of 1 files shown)


scripts/remove_idle_group_members.pl
```diff
@@ -37,7 +37,7 @@
   q{SELECT DISTINCT profiles.userid AS user_id,
             groups.id AS group_id
        FROM profiles JOIN user_group_map ON profiles.userid = user_group_map.user_id
-            JOIN groups ON user_group_map.group_id = groups.id
+            JOIN `groups` ON user_group_map.group_id = groups.id
```
dklawren, line 40: Instead of using ` quotes directly we use a helper function called $dbh->quote_identifier('groups') in case another DB other than MySQL uses different type of quotes for the same thing. So in this case it would be:

```perl
JOIN ' .  $dbh->quote_identifier('groups') . ' ON user_group_map.group_id = groups.id
```

I can fix this if you want. It was missed somehow when I had to do the same change in many other places in the code.

dklawren (changes requested):

scripts/remove_idle_group_members.pl
dlactin: Sure, that would be great!

PR #1975, opened 2022-11-27
Bug 1802203 - Change email delivery functionality in Bugzilla to Email::Sender instead of Email::Send

dklawren:
This pull request changes the dependency that we were using to something better maintained. In the past using Email::Send was fine as we were just piping emails locally to the sendmail command which then relayed the email to AWS SES, etc. For GCP we are now using SMTP directly and since we have never used it before, several bugs were found in using Email::Send and its SMTP transport. With the better maintained, Email::Sender library, I have not encountered the same issues and everything so far seems to work fine. I have installed this patch on bugzilla-dev.allizom.org and test both file based delivery as well as SMTP to SES using the credentials provided by CloudOps.


globau (commented):
I don't have time to do an actual review, but nothing jumped out at me from a quick read through the changes.

dklawren:
> I don't have time to do an actual review, but nothing jumped out at me from a quick read through the changes.

Thanks glob.

Bugzilla/Mailer.pm
```diff
@@ -221,23 +188,31 @@ sub MessageToMTA {
...
-
-    # From - <date> is required to be a valid mbox file.
-    print TESTFILE "\n\nFrom - "
-      . $email->header('Date') . "\n"
-      . $email->as_string;
-    close TESTFILE;
+    my $test_fh = IO::File->new($filename, 'a');
+    $transport = Email::Sender::Transport::Print->new({fh => $test_fh});
   }
-  else {
```
zzzeid, line 234: Since `None` is still a valid choice here for `$method`, should we still be handling `else` (i.e., a default transport?)? Or perhaps `None` should be removed from the [list of choices](https://github.com/mozilla-bteam/bmo/pull/1975/files#diff-f71e4290382723e65f9c98e4d27682ad6dce7dbc1d0895997dd0beda1f66eaa5R51)?

Bugzilla/Mailer.pm
```diff
@@ -120,37 +117,17 @@ sub MessageToMTA {
...
-    $email->header_set('From', $from);
-
-    # Sendmail adds a Date: header also, but others may not.
-    if (!defined $email->header('Date')) {
-      $email->header_set('Date', time2str("%a, %d %b %Y %T %z", time()));
-    }
+  # Sendmail will automatically append our hostname to the From
+  # address, but other mailers won't.
+  my $urlbase = Bugzilla->localconfig->urlbase;
+  $urlbase =~ m|//([^:/]+)[:/]?|;
```
zzzeid, line 123: Is this variable used anywhere?

zzzeid (approved):
Looks good to me, just a couple of minor comments.

Bugzilla/Mailer.pm
dklawren: None is processed earlier up at https://github.com/mozilla-bteam/bmo/blob/master/Bugzilla/Mailer.pm#L46
It just basically exits out early.

Bugzilla/Mailer.pm
dklawren: yes and no. We save `Bugzilla->localconfig->urlbase` value into a variable so we can perform a proper regex on the value. Then in the next line, you will see `$urlbase` with a regex match performed against it. And then next, `$hostname = $1;` is using the matched content from inside the regex parenthesis from the previous regex. So `$urlbase` is used but only briefly.

PR #1976, opened 2022-11-28
Bug 1802894 - Please delist "Data Platform and Tools" from new bug page

dklawren:


extensions/BMO/template/en/default/global/choose-product.html.tmpl
```diff
@@ -123,10 +123,6 @@
       name="Mozilla Localizations"
       icon="localization.png"
     %]
-    [% INCLUDE easyproduct
-      name="Data Platform and Tools"
-      icon="telemetry.png"
```
cgsheeh, line 128: Can we `git rm telemetry.png` also?

cgsheeh (changes requested):

cgsheeh (approved):

PR #1978, opened 2022-11-30
Bug 1783577 - Add ability to associate a URL with a comment keyword

dklawren:
I will push this change to bugzilla-dev where the functionality can be properly tested rather than run this locally. The commit messages describe what changes are occurring for each file.

Update: This is now installed on bugzilla-dev.allizom.org. I have added an entry to the tag urls table for the tag 'mozilla' which will show a link to 'https://www.mozilla.org'. So you can go to any bug and add the tag 'mozilla' to a comment and you will see the more info link next to it.


Bugzilla/DB/Schema.pm
```diff
@@ -467,6 +467,16 @@ use constant ABSTRACT_SCHEMA => {
...
   },
 
+  longdescs_tags_url => {
+    FIELDS => [
+      id  => {TYPE => 'MEDIUMSERIAL', NOTNULL => 1, PRIMARYKEY => 1},
+      tag => {TYPE => 'varchar(64)',  NOTNULL => 1},
+      url => {TYPE => 'varchar(512)', NOTNULL => 1},
+    ],
+    INDEXES =>
+      [longdescs_tags_url_tag_url_idx => {FIELDS => [qw(tag url)], TYPE => 'UNIQUE'},],
```
globau, line 477: We'll need an index on just `tag` given that's the primary lookup

extensions/BugModal/web/comments.js
```diff
@@ -269,48 +269,54 @@ $(function() {
...
                 span.prepend($('<a role="button" aria-label="Remove" class="remove">×</a>').click(deleteTag));
             }
+            if (tag_urls && tag_urls[this]) {
+                var infoLink = $('<a aria-label="More Information" target="_blank"></a>');
+                infoLink.attr('href', tag_urls[this].htmlEncode());
+                var img = $('<img width="16" height="16" title="Click for more information"/>');
+                img.attr('src', BUGZILLA.config.basepath + 'extensions/BugModal/web/images/help.png');
+                infoLink.append(img);
+                span.append(infoLink);
+            }
```
globau, line 316: Rather than duplicating rendering both server and client side, would it be better to extract the comment tags template into its own template, then always request a html fragment from the client?

extensions/BugModal/web/comments.js
```diff
@@ -269,48 +269,54 @@ $(function() {
...
+            var span = $("<span/>")
+                .addClass("comment-tag")
+                .data("tag", this)
+                .text(this);
             if (BUGZILLA.user.can_tag) {
                 span.prepend($('<a role="button" aria-label="Remove" class="remove">×</a>').click(deleteTag));
             }
+            if (tag_urls && tag_urls[this]) {
+                var infoLink = $('<a aria-label="More Information" target="_blank"></a>');
+                infoLink.attr('href', tag_urls[this].htmlEncode());
```
globau, line 311: My jquery's pretty rusty, but I'm pretty sure you don't need to encode the value here.

globau (changes requested):
UI:
- the whole tag should be a hyperlink, not just the icon
- remove the icon; it isn't a good fit size or style-wise
- set the cursor to `help`; this matches all other places on show_bug where we provide more information (eg. field labels)
- don't set a blank target, let the user decide if they want a new tab or not (this also matches links to field labels)
- a tag's linking is dropped if you cancel adding a tag

globau:
Can bugzilla-dev.allizom.org be updated with the most recent changes?

Bugzilla/WebService/Bug.pm
```diff
@@ -429,6 +429,12 @@ sub _translate_comment {
   # Don't load comment tags unless enabled
   if (Bugzilla->params->{'comment_taggers_group'}) {
     $comment_hash->{tags} = [map { $self->type('string', $_) } @{$comment->tags}];
+    $comment_hash->{tag_urls} = {};
+    foreach my $tag (@{$comment->tags}) {
+      my $url = $comment->tag_url($tag);
+      $comment_hash->{tag_urls}->{$tag} = $url if $url;
+    }
+    $comment_hash->{comment_tag_html} = $self->_get_comment_tag_html($comment);
```
globau, line 437: `comment_tag_html` shouldn't be exposed via the public API as it's internal only to the Bugzilla UI.

Instead this should be a `BugModel` endpoint that returns just the html (similar to `bug_model/cc/\d`).

globau (changes requested):

Bugzilla/WebService/Bug.pm
dklawren: > `comment_tag_html` shouldn't be exposed via the public API as it's internal only to the Bugzilla UI.
> 
> Instead this should be a `BugModel` endpoint that returns just the html (similar to `bug_model/cc/\d`).

This would require a second call. I originally added the html option to the core add/remove endpoints to be able to do it all in one call. I will see what kind of performance change this would make.

globau: I thought that `comment_tag_html` is the only field used from that response.

dklawren: > I thought that `comment_tag_html` is the only field used from that response.

the refreshTags(), which just retrieves the current tag list,  does only use comment_tag_html, but  when you add/remove a tag, the endpoint will also return the refreshed comment_tag_html which is then passed to renderTags() to use. If I move the comment html potion to its own endpoint, I will need to first add/remove the tag with the first call, then call refreshTags() which will make a second call to get the rendered html. It might be just fine to do it that way and not noticeably slower. I will give it a try.

globau: Rather than making two calls you could return the html and the other fields in the internal endpoint.

extensions/BugModal/web/comments.js
```diff
@@ -367,11 +360,11 @@ $(function() {
...
         // update Bugzilla
         try {
-            var result = await Bugzilla.API.put(
+            await Bugzilla.API.put(
               `bug/comment/${commentID}/tags`,
-              { add: newTags, comment_tag_html: 1 }
+              { add: newTags }
             );
-            renderTags(commentNo, result.comment_tag_html);
+            refreshTags(commentNo, commentID);
```
globau, line 367: I think it would be better to have the single internal bug-modal call return all the required tag info rather than making two calls, but would like to check the UX with two calls.

Please let me know when this change is on bugzilla-dev.

dklawren:
The latest code has now been installed to bugzilla-dev.allizom.org

cgsheeh (approved):
LGTM.

PR #1991, opened 2022-12-19
Bug 1805681 - github pull request webhook returns 400

dklawren:
This PR changes to code to show 400 failures only if the JSON is invalid or the signature is invalid. All other errors should return 200 even though they are non-fatal errors. I have added a new key/value called `error` that is a boolean. If true, then a `message` is provided with the error message but the status code is 200. This should show green results in the webhook log in Github unless something serious happens.


cgsheeh (approved):
WFM!

PR #1992, opened 2022-12-19
Bug 1806477 - Secbugsreport CronJob failing in latest release

dklawren:
Fixed bad SQL.


cgsheeh (approved):
Looks like we used to do `JOIN fielddefs AS field` before Bug 1804082.

PR #1993, opened 2022-12-21
Bug 1805453: implement Bugzilla-updating logic from Phab uplift extension in Phab Feed reader

cgsheeh:
Re-implements the Bugzilla updating logic in the Phabricator Uplifts extension into BMO PhabBugz feed reader.

First add the uplift request form to `Revision.pm` so it is accessible from the feed. The form is either an empty
array or a `Dict` of `Any` keys, where the keys are the form questions and the values are the answers.
Move uplift request form processing into its own function to avoid a massive `process_revision_change` function.
Add A function `format_uplift_request_as_markdown` to render the form as a markdown comment, and add a
`readable_answer` function to convert `1` and `0` values to `yes` and `no` respectively. Use the markdown rendering
function as the contents of a comment on the bug when the form is submitted. Add a check for the `Needs manual QE test` question to set `qe-verify` to `+` if the users reponse is `true`.

### DIFF
--- a/extensions/PhabBugz/lib/Feed.pm
+++ b/extensions/PhabBugz/lib/Feed.pm
@@ -358,6 +358,79 @@ sub _is_uplift_request_form_change {
   return $story_text =~ /\s+uplift request field/;
 }
 
+sub process_uplift_request_form_change {
+    # Process an uplift request form change for the passed revision object.
+    my ($revision) = @_;
+
+    # Take no action if the form is empty.
+    # TODO test
+    if ($revision->uplift_form->is_empty) {
+        INFO("Uplift request form field cleared, ignoring.");
+        return;
+    }
+
+    my $revision_phid = $revision->phid;
+    INFO(
+      "Uplift request form submitted on $revision_phid, " .
+      "requesting `#release-managers` review."
+    );
+
+    # Get `#release-managers` review group.
+    my $release_managers_group = Bugzilla::Extension::PhabBugz::Project
+      ->new_from_query({name => 'release-managers'});
+
+    if (!$release_managers_group) {
+      WARN(
+        "Uplift request change detected but `#release-managers` was " .
+        "not found on Phabricator."
+      );
+      return;
+    }
+
+    my $release_managers_phid = $release_managers_group->phid;
+
+    # Request `#release-managers` review on each revision.
+    foreach my $stack_revision_phid (keys %{$revision->stack_graph_raw}) {
+      # Query Phabricator for each revision object related to the updated revision.
+      my $stack_revision = Bugzilla::Extension::PhabBugz::Revision->new_from_query(
+        {phids => [$stack_revision_phid]}
+      );
+
+      # Add `#release-managers!` review and set revision status.
+      $stack_revision->add_reviewer("blocking($release_managers_phid)");
+      $stack_revision->update();
+
+      INFO("Requested #release-managers review of $stack_revision_phid.");
+    }
+
+    # Comment the uplift form rendered on Bugzilla
+    # TODO add rendered form to storage on Phabricator side.
+    # TODO test
+    my $comment_data = {
+        'bug_id' => $revision->bug->id,
+        'is_markdown' => 1,
+    };
+    my $comment = Bugzilla::Comment->insert_create_data($comment_data);
+
+    # If manual QE is required, set the Bugzilla flag
+    # TODO test
+    if ($revision->uplift_form->{"Needs manual QE test"}) {
+        # TODO how to set bug flags?
+        # TODO set the `qe-verify` flag to `?` here.
+        my $new_flags = {
+            type_id  => $flagtype->id,
+            status   => '?',
+            setter   => $setter,
+            flagtype => $flagtype,
+        };
+
+        $revision->bug->set_flags($revision->bug->flags, $new_flags);
+        $revision->bug->update();
+    }
+    
+    # TODO commit here?
+}
+
 sub process_revision_change {
   state $check = compile($Invocant, Revision, LinkedPhabUser, Str);
   my ($self, $revision, $changer, $story_text) = $check->(@_);
@@ -393,40 +466,7 @@ sub process_revision_change {
 
   # Change is a submission of the uplift request form.
   if (_is_uplift_request_form_change($story_text)) {
-    my $revision_phid = $revision->phid;
-    INFO(
-      "Uplift request form submitted on $revision_phid, " .
-      "requesting `#release-managers` review."
-    );
-
-    # Get `#release-managers` review group.
-    my $release_managers_group = Bugzilla::Extension::PhabBugz::Project
-      ->new_from_query({name => 'release-managers'});
-
-    if (!$release_managers_group) {
-      WARN(
-        "Uplift request change detected but `#release-managers` was " .
-        "not found on Phabricator."
-      );
-      return;
-    }
-
-    my $release_managers_phid = $release_managers_group->phid;
-
-    # Request `#release-managers` review on each revision.
-    foreach my $stack_revision_phid (keys %{$revision->stack_graph_raw}) {
-      # Query Phabricator for each revision object related to the updated revision.
-      my $stack_revision = Bugzilla::Extension::PhabBugz::Revision->new_from_query(
-        {phids => [$stack_revision_phid]}
-      );
-
-      # Add `#release-managers!` review and set revision status.
-      $stack_revision->add_reviewer("blocking($release_managers_phid)");
-      $stack_revision->update();
-
-      INFO("Requested #release-managers review of $stack_revision_phid.");
-    }
-
+    process_uplift_request_form_change($revision);
     return;
   }
 
--- a/extensions/PhabBugz/lib/Feed.pm
+++ b/extensions/PhabBugz/lib/Feed.pm
@@ -458,14 +458,14 @@ sub process_uplift_request_form_change {
       INFO("Found `qe-verify` flag to update.");
       $qe_verify_flag = $flag;
 
-      if ($qe_verify_flag->status ne '?') {
+      if ($qe_verify_flag->status ne '+') {
         # Set the flag to `?`.
         $bug->set_flags([{
           id     => $qe_verify_flag->id,
-          status => '?',
+          status => '+',
         }], []);
 
-        INFO("Set `qe-verify` flag to `?`.");
+        INFO("Set `qe-verify` flag to `+`.");
       }
 
       last;
--- a/extensions/PhabBugz/lib/Feed.pm
+++ b/extensions/PhabBugz/lib/Feed.pm
@@ -445,31 +445,48 @@ sub process_uplift_request_form_change {
   $bug->add_comment($comment_content, $comment_params);
 
   # If manual QE is required, set the Bugzilla flag.
-  # TODO test
   if ($revision->uplift_request->{"Needs manual QE test"}) {
     INFO("Needs manual QE test is set.");
 
-    # Find the current `qe-verify` flag state.
-    my $qe_verify_flag;
+    my @old_flags;
+    my @new_flags;
+
+    # Find the current `qe-verify` flag state if it exists.
     foreach my $flag (@{$bug->flags}) {
+      my $name = $flag->type->name;
+
       # Ignore for all flags except `qe-verify`.
       next if $flag->type->name ne 'qe-verify';
+      INFO("Found `qe-verify` flag.");
 
-      INFO("Found `qe-verify` flag to update.");
-      $qe_verify_flag = $flag;
+      if ($flag->status ne '+') {
+        INFO("Setting status to `+` for qe-verify.");
 
-      if ($qe_verify_flag->status ne '+') {
         # Set the flag to `?`.
-        $bug->set_flags([{
-          id     => $qe_verify_flag->id,
-          status => '+',
-        }], []);
+        push(@old_flags, {id => $flag->id, status => '+'});
 
         INFO("Set `qe-verify` flag to `+`.");
       }
 
       last;
     }
+
+    # If we didn't find an existing `qe-verify` flag to update, add it now.
+    if (!@old_flags) {
+      FATAL("CREATING NEW QE-VERIFY FLAG");
+      my $qe_flag = Bugzilla::FlagType->new({name => 'qe-verify'});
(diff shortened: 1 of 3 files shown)


extensions/PhabBugz/lib/Feed.pm
```diff
@@ -427,6 +403,70 @@ sub process_revision_change {
...
     }
 
+    # Comment the uplift form rendered on Bugzilla
+    # TODO add rendered form to storage on Phabricator side.
+    # TODO test
+    my $comment_data = {
+        'bug_id' => $revision->bug->id,
+        'is_markdown' => 1,
+    };
+    my $comment = Bugzilla::Comment->insert_create_data($comment_data);
```
cgsheeh, line 413: I need to add a markdown comment to the bug here.

extensions/PhabBugz/lib/Feed.pm
```diff
@@ -427,6 +403,70 @@ sub process_revision_change {
...
+        my $new_flags = {
+            type_id  => $flagtype->id,
+            status   => '?',
+            setter   => $setter,
+            flagtype => $flagtype,
+        };
+
+        $revision->bug->set_flags($revision->bug->flags, $new_flags);
+        $revision->bug->update();
+    }
```
cgsheeh, line 429: I need to check the value of one of the keys in `uplift_form` and set `qe-verify` flag to `?` if it is `true`.

extensions/PhabBugz/lib/Revision.pm
```diff
@@ -52,6 +52,7 @@ has stack_graph_raw  => (
 );
 has subscriber_count => (is => 'ro',   isa => Int);
 has bug              => (is => 'lazy', isa => Object);
+has uplift_request   => (is => 'ro',   isa => Dict [ slurpy Any ]);
```
cgsheeh, line 55: Is this how I would add the `uplift_request` field without having to specify the exact values of the keys?

extensions/PhabBugz/lib/Revision.pm
```diff
@@ -166,7 +168,18 @@ sub BUILDARGS {
...
+#             "User impact if declined": "Some impact.",
+#             "Code covered by automated testing": true,
+#             "Fix verified in Nightly": true,
+#             "Needs manual QE test": false,
+#             "Steps to reproduce for manual QE testing": "none",
+#             "Risk associated with taking this patch": "low",
+#             "Explanation of risk level": "Affects only nondefault configurations.",
+#             "String changes made/needed": "none",
+#             "Is Android affected?": true
+#         }
```
cgsheeh, line 182: Note: I'm going to change this slightly because I would like to include the markdown comment already rendered.

extensions/PhabBugz/lib/Revision.pm
dklawren: Yeah

extensions/PhabBugz/lib/Feed.pm
```diff
@@ -358,6 +358,79 @@ sub _is_uplift_request_form_change {
...
+  }
+
+  # Comment the uplift form rendered on Bugzilla
+  # TODO add rendered form to storage on Phabricator side.
+  # TODO test
+  my $comment_data = {
+    'bug_id' => $revision->bug->id,
+    'is_markdown' => 1,
+  };
+  my $comment = Bugzilla::Comment->insert_create_data($comment_data);
```
dklawren, line 413: You would do `$revision->bug->add_comment();` here

extensions/PhabBugz/lib/Feed.pm
```diff
@@ -358,6 +358,79 @@ sub _is_uplift_request_form_change {
...
+
+  # If manual QE is required, set the Bugzilla flag
+  # TODO test
+  if ($revision->uplift_form->{"Needs manual QE test"}) {
+    # TODO how to set bug flags?
+    # TODO set the `qe-verify` flag to `?` here.
+    my $new_flags = {
```
dklawren, line 420: You will need to first load the flag type object

```
my $qe_flag = Bugzilla::FlagType->new({name => 'qe-verify'});
if ($qe_flag) {
  [...]
}
```

extensions/PhabBugz/lib/Feed.pm
```diff
@@ -358,6 +358,79 @@ sub _is_uplift_request_form_change {
...
+      type_id  => $flagtype->id,
+      status   => '?',
+      setter   => $setter,
```
dklawren, line 423: $setter will need to be the ID of the user setting the flag. Would this be phab-bot?

dklawren (changes requested):

extensions/PhabBugz/lib/Feed.pm
cgsheeh: Yes in this case it would be `phab-bot`. How would I get the ID for that user?

We could also potentially use the `$changer` variable from `process_revision_change`. Can we assume that any user who has access to a revision (ie to submit the uplift form on Phabricator) would also be able to update the bug as needed? If that's the case I'd be fine with making all these bug changes as the user who submitted the change in Phabricator.

dklawren: > Yes in this case it would be `phab-bot`. How would I get the ID for that user?
> 
> We could also potentially use the `$changer` variable from `process_revision_change`. Can we assume that any user who has access to a revision (ie to submit the uplift form on Phabricator) would also be able to update the bug as needed? If that's the case I'd be fine with making all these bug changes as the user who submitted the change in Phabricator.

`my $phab_user = Bugzilla::User->new({name => PHAB_AUTOMATION_USER});`

It might be that the $changer does not have access to the bug and so it is better that we use the phab-bot user making the bug changes. 

In process_revision_change(), we set the current user the phab-bot user already so it is making any changes to bugs. 
It happens at: 

`my $restore_prev_user = set_phab_user();`

extensions/PhabBugz/lib/Feed.pm
```diff
@@ -393,40 +470,7 @@ sub process_revision_change {
...
-
-      # Add `#release-managers!` review and set revision status.
-      $stack_revision->add_reviewer("blocking($release_managers_phid)");
-      $stack_revision->update();
-
-      INFO("Requested #release-managers review of $stack_revision_phid.");
-    }
-
+    process_uplift_request_form_change($revision, $changer);
     return;
```
dklawren, line 474: Just do `return process_uplift_request_form_change($revision, $changer);` here in case a return value is needed by the caller.

extensions/PhabBugz/lib/Feed.pm
```diff
@@ -358,6 +358,83 @@ sub _is_uplift_request_form_change {
...
+
+  # If manual QE is required, set the Bugzilla flag.
+  # TODO test
+  if ($revision->uplift_form->{"Needs manual QE test"}) {
+    my $qe_flag = Bugzilla::FlagType->new({name => 'qe-verify'});
+    # TODO what to do if !$qe_flag ?
+    if ($qe_flag) {
+      my $new_flags = {
+        type_id  => $qe_flag->id,
+        status   => '?',
```
dklawren, line 426: As of today, '?' is disabled for qe-verify and only '+' and '-' are availiable. The proper value now is '+'.

extensions/PhabBugz/lib/Feed.pm
```diff
@@ -358,6 +358,83 @@ sub _is_uplift_request_form_change {
...
+        setter   => $changer->bugzilla_user->id,
```
dklawren, line 427: When this will execute, we have not yet ran set_phab_user() so the comment and flag setting might fail if the bug is private. Since the phabbugz daemon is running locally, no default user is set so we need to do it explicitly. 

We might be ok just moving set_phab_user earlier in process_revision_change() right before the uplift check.

dklawren (changes requested):

extensions/PhabBugz/lib/Feed.pm
cgsheeh: This is what I did and I can leave comments as that user in suite. :+1:

extensions/PhabBugz/lib/Feed.pm
cgsheeh: What should `process_uplift_request_form_change` return?

extensions/PhabBugz/lib/Feed.pm
cgsheeh: I think everything in the current patch state is now working, except for the `qe-verify` flag setting. I'm giving it a try with `+` as the value to set in case that's the problem.

dklawren:
The main critic.t error that needs to be fixed is the use of parenthesis with builtin perl functions such as push, etc. So the changes needed below:

```diff
diff --git a/extensions/PhabBugz/lib/Feed.pm b/extensions/PhabBugz/lib/Feed.pm
index 06d677ced..486246e35 100644
--- a/extensions/PhabBugz/lib/Feed.pm
+++ b/extensions/PhabBugz/lib/Feed.pm
@@ -464,7 +464,7 @@ sub process_uplift_request_form_change {
         INFO("Setting status to `+` for qe-verify.");
 
         # Set the flag to `?`.
-        push(@old_flags, {id => $flag->id, status => '+'});
+        push @old_flags, {id => $flag->id, status => '+'};
 
         INFO("Set `qe-verify` flag to `+`.");
       }
@@ -477,19 +477,19 @@ sub process_uplift_request_form_change {
       FATAL("CREATING NEW QE-VERIFY FLAG");
       my $qe_flag = Bugzilla::FlagType->new({name => 'qe-verify'});
       if ($qe_flag) {
-        push(@new_flags, {
+        push @new_flags, {
           flagtype => $qe_flag,
           setter => $phab_bot_user,
           status => '+',
           type_id => $qe_flag->id,
-        });
+        };
       }
     }
```

Also, non-fatal but should be changed, is the Perl styling dictates use of single quotes (') for string literals and double quotes (") if you are doing variable interpolation. So for example:

```diff
diff --git a/extensions/PhabBugz/lib/Feed.pm b/extensions/PhabBugz/lib/Feed.pm
index 06d677ced..520cc680f 100644
--- a/extensions/PhabBugz/lib/Feed.pm
+++ b/extensions/PhabBugz/lib/Feed.pm
@@ -368,11 +368,11 @@ sub readable_answer {
 
   # Return "yes" for `1`.
   if ($answer) {
-    return "yes";
+    return 'yes';
   }
 
   # Return "no" for `0`.
-  return "no";
+  return 'no';
 }
 
 sub format_uplift_request_as_markdown {
```

I realize Python just uses double quotes most of the time which is fine. We can leave those unless you want to change them to clean up the warnings.

cgsheeh:
> The main critic.t error that needs to be fixed is the use of parenthesis with builtin perl functions such as push, etc. So the changes needed below:
> 
> ```diff
> diff --git a/extensions/PhabBugz/lib/Feed.pm b/extensions/PhabBugz/lib/Feed.pm
> index 06d677ced..486246e35 100644
> --- a/extensions/PhabBugz/lib/Feed.pm
> +++ b/extensions/PhabBugz/lib/Feed.pm
> @@ -464,7 +464,7 @@ sub process_uplift_request_form_change {
>          INFO("Setting status to `+` for qe-verify.");
>  
>          # Set the flag to `?`.
> -        push(@old_flags, {id => $flag->id, status => '+'});
> +        push @old_flags, {id => $flag->id, status => '+'};
>  
>          INFO("Set `qe-verify` flag to `+`.");
>        }
> @@ -477,19 +477,19 @@ sub process_uplift_request_form_change {
>        FATAL("CREATING NEW QE-VERIFY FLAG");
>        my $qe_flag = Bugzilla::FlagType->new({name => 'qe-verify'});
>        if ($qe_flag) {
> -        push(@new_flags, {
> +        push @new_flags, {
>            flagtype => $qe_flag,
>            setter => $phab_bot_user,
>            status => '+',
>            type_id => $qe_flag->id,
> -        });
> +        };
>        }
>      }
> ```
> 
> Also, non-fatal but should be changed, is the Perl styling dictates use of single quotes (') for string literals and double quotes (") if you are doing variable interpolation. So for example:
> 
> ```diff
> diff --git a/extensions/PhabBugz/lib/Feed.pm b/extensions/PhabBugz/lib/Feed.pm
> index 06d677ced..520cc680f 100644
> --- a/extensions/PhabBugz/lib/Feed.pm
> +++ b/extensions/PhabBugz/lib/Feed.pm
> @@ -368,11 +368,11 @@ sub readable_answer {
>  
>    # Return "yes" for `1`.
>    if ($answer) {
> -    return "yes";
> +    return 'yes';
>    }
>  
>    # Return "no" for `0`.
> -  return "no";
> +  return 'no';
>  }
>  
>  sub format_uplift_request_as_markdown {
> ```
> 
> I realize Python just uses double quotes most of the time which is fine. We can leave those unless you want to change them to clean up the warnings.

Fixed both issues. :+1:

extensions/PhabBugz/lib/Feed.pm
```diff
@@ -358,6 +358,141 @@ sub _is_uplift_request_form_change {
...
+    # If we didn't find an existing `qe-verify` flag to update, add it now.
+    if (!@old_flags) {
+      my $qe_flag = Bugzilla::FlagType->new({name => 'qe-verify'});
+      if ($qe_flag) {
+        push @new_flags, {
+          flagtype => $qe_flag,
+          setter => $phab_bot_user,
+          status => '+',
+          type_id => $qe_flag->id,
+        };
```
dklawren, line 484: nit: line up =>

dklawren (changes requested):

extensions/PhabBugz/lib/Feed.pm
```diff
@@ -358,6 +358,141 @@ sub _is_uplift_request_form_change {
...
+  # Return 'no' for `0`.
+  return 'no';
+}
+
+sub format_uplift_request_as_markdown {
+  my ($question_answers_mapping) = @_;
+
+  my $comment = '# Uplift Approval Request\n';
+
+  while (my ($question, $answer) = each %{$question_answers_mapping}) {
```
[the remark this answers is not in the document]
cgsheeh: Hmmm yes it would be better to have the comments always in the same order. Maybe I could change the storage from a mapping to a list of `(question, answer)` pairs on the Phabricator side...

For now I think we can just ship this as-is and I can fix it when it becomes a problem. :)

dklawren (approved):

PR #1997, opened 2023-01-10
Bug 1809529: change `docker compose` to `docker-compose` globally

cgsheeh:


dklawren (commented):
`docker compose` is part of v2 of docker compose where it has been rewritten in Go and included as a plugin for Docker. It is installed by default in Docker Desktop for Windows and MacOS and can also be installed in Linux. https://github.com/docker/compose/tree/v2. I am using the docker-ce community edition on Linux and the latest for Fedora 37 has the ability to do `docker compose` so I am not sure we need to do this change. `docker compose` is the way going forward.

cgsheeh:
> `docker compose` is part of v2 of docker compose where it has been rewritten in Go and included as a plugin for Docker. It is installed by default in Docker Desktop for Windows and MacOS and can also be installed in Linux. https://github.com/docker/compose/tree/v2. I am using the docker-ce community edition on Linux and the latest for Fedora 37 has the ability to do `docker compose` so I am not sure we need to do this change. `docker compose` is the way going forward.

Oh, TIL. I should install that myself, it's annoying dealing with the Python version. This is a WONTFIX!

PR #2001, opened 2023-01-23
Bug 1811786 - Unable to expand an attachment comment after it has been tagged with `admin`

dklawren:
The expand/collapse code needs for a `<pre id="ct-X"></pre>` to be present for the code to execute properly. So when we attach an attachment without a comment, the `<pre>` was missing. This patch adds it back but leaves it empty so that expand/collapse can still work properly.


globau (approved):
lgtm

PR #2005, opened 2023-01-25
Bug 1811647 - Enforce consistent metadata for regression bugs

dklawren:
This pull request adds the regression keyword and also sets the type to defect if the regressed by field is populated with a bug id. The two methods are for both creation of new bugs, and for updating an existing bug.


extensions/MozChangeField/lib/Post/RegressedByTypeKeyword.pm
```diff
@@ -0,0 +1,77 @@
...
+}
+
+sub evaluate_change {
+  my ($self, $args) = @_;
+  my $bug       = $args->{bug};
+  my $timestamp = $args->{timestamp};
+  my $changes   = $args->{changes};
+  my $dbh       = Bugzilla->dbh;
+
+  if (@{$bug->regressed_by} && exists $changes->{regressed_by}) {
```
cgsheeh, line 46: Nit; all the code for this function is in the main `if` statement. Instead of having `if (condition)` with all the code inside the block, we should do `if (not condition) {return}` and then put everything else below, de-dented.

extensions/MozChangeField/lib/Post/RegressedByTypeKeyword.pm
```diff
@@ -0,0 +1,77 @@
...
+    if (!$bug->has_keyword('regression')
```
cgsheeh, line 47: Add a comment describing what this is doing.

extensions/MozChangeField/lib/Post/RegressedByTypeKeyword.pm
```diff
@@ -0,0 +1,77 @@
...
+        $changes->{keywords}->[0]
+          = ($changes->{keywords}->[0] ? $changes->{keywords}->[0] : '');
+        $changes->{keywords}->[1]
+          = ($changes->{keywords}->[1]
+          ? $changes->{keywords}->[1] . ', regression'
+          : 'regression');
+      }
+    }
+
+    if ($bug->bug_type ne 'defect'
```
cgsheeh, line 66: Add a comment describing what this is doing.

extensions/MozChangeField/lib/Post/RegressedByTypeKeyword.pm
```diff
@@ -0,0 +1,77 @@
...
+
+use Bugzilla::Keyword;
+
+sub evaluate_create {
+  my ($self, $args) = @_;
+  my $bug       = $args->{bug};
+  my $timestamp = $args->{timestamp};
+  my $dbh       = Bugzilla->dbh;
+
+  if (@{$bug->regressed_by}) {
```
cgsheeh, line 21: Same as the other comment, check for the inverse and `return`, and de-dent everything else.

extensions/MozChangeField/lib/Post/RegressedByTypeKeyword.pm
```diff
@@ -0,0 +1,77 @@
...
+    if (!$bug->has_keyword('regression')) {
```
cgsheeh, line 22: Add a comment describing what this is doing.

extensions/MozChangeField/lib/Post/RegressedByTypeKeyword.pm
```diff
@@ -0,0 +1,77 @@
...
+      my $keyword_obj = Bugzilla::Keyword->new({name => 'regression'});
+      if ($keyword_obj) {
+        $dbh->do('INSERT INTO keywords (bug_id, keywordid) VALUES (?, ?)',
+          undef, $bug->id, $keyword_obj->id);
+        delete $bug->{'keyword_objects'};    # Clear current keywords so they have to be reloaded.
+      }
+    }
+
+    if ($bug->bug_type ne 'defect') {
```
cgsheeh, line 31: Add a comment describing what this is doing.

cgsheeh (changes requested):

extensions/MozChangeField/lib/Post/RegressedByTypeKeyword.pm
```diff
@@ -0,0 +1,87 @@
...
+
+use Bugzilla::Keyword;
+
+sub evaluate_create {
+  my ($self, $args) = @_;
+  my $bug       = $args->{bug};
+  my $timestamp = $args->{timestamp};
+  my $dbh       = Bugzilla->dbh;
+
+  return unless @{$bug->regressed_by};
```
suhaibmujahid, line 21:
```suggestion
  return unless (@{$bug->regressed_by} || $bug->has_keyword('regression'));
```
In this way, the bug type will be enforced to be `defect` even if the `regressed_by` field is empty.

We could do a similar thing for `evaluate_change`.

cgsheeh (approved):
LGTM, you can carry forward review if you add Suhaib's suggestion. :)

extensions/MozChangeField/lib/Post/RegressedByTypeKeyword.pm
dklawren: Thanks. good point

PR #2006, opened 2023-01-25
Bug 1812384 - Error occurring when $revision->uplift_request in Phabbugz is being interpreted as true when it is really empty

dklawren:
Simplified the flag code which also fixes the multiple flag error we are getting sometimes. Also check $revision->uplift_request to make sure it is a HASH and not empty before proceeding.

### DIFF
--- a/extensions/PhabBugz/lib/Feed.pm
+++ b/extensions/PhabBugz/lib/Feed.pm
@@ -462,7 +462,7 @@ sub process_uplift_request_form_change {
       # Set the flag to `+`. If already '+', it will be non-change.
       INFO('Set `qe-verify` flag to `+`.');
       push @old_flags, {id => $flag->id, status => '+'};
-     last;
+      last;
     }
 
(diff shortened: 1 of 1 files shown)


extensions/PhabBugz/lib/Feed.pm
```diff
@@ -455,22 +457,12 @@ sub process_uplift_request_form_change {
...
-
-        INFO('Set `qe-verify` flag to `+`.');
-      }
-
-      last;
+      next if $flag->name ne 'qe-verify';
+      # Set the flag to `+`. If already '+', it will be non-change.
+      INFO('Set `qe-verify` flag to `+`.');
+      push @old_flags, {id => $flag->id, status => '+'};
+     last;
```
cgsheeh, line 465: Indentation looks off here.

cgsheeh (approved):
Thanks for the quick fix here, dkl. My Perl knowledge only extends so far. :)

PR #2007, opened 2023-01-25
Bug 1812457 - we are unconditionally setting release-managers review on the patch when we detect the change, even when release-managers has already given an r+ review.

dklawren:


cgsheeh (approved):
LGTM.

dklawren:
> LGTM.

Thanks. Tested with conduit-suite and works there as well

PR #2008, opened 2023-01-26
Bug 1812606 - github pull request with the same number but from different repos get confused

dklawren:
This pull request looks for attachments with the same pull request number in the filename but then also does a comparison of the attachment data containing the html url of the pull request. If they match then it is skipped, otherwise it moves the attachment to the different bug as before. Before this change we were only looking at the pull request number which causes issues when two repos have the same number.

### DIFF
--- a/Bugzilla/API/V1/Github.pm
+++ b/Bugzilla/API/V1/Github.pm
@@ -151,6 +151,7 @@ sub pull_request {
     WHERE    => {'bug_id != ? AND NOT isobsolete' => $bug->id}
   });
   foreach my $attachment (@$other_attachments) {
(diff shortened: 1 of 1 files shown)


Bugzilla/API/V1/Github.pm
```diff
@@ -151,6 +151,7 @@ sub pull_request {
     WHERE    => {'bug_id != ? AND NOT isobsolete' => $bug->id}
   });
   foreach my $attachment (@$other_attachments) {
+    next if $attachment->data ne $html_url; # same pr number but different repo so skip it
```
cgsheeh, line 154: Comment should probably be on the above line.

cgsheeh (approved):

PR #2010, opened 2023-01-26
Bug 1803374 - Numbers in PR descriptions are considered bug numbers by the Bugzilla GitHub integration

dklawren:
This pull request gets rid of the more complex regex for finding IDs in commit titles and just simplifies it down to just looking for a specific format. In the past we were mistakenly linking to non bug IDs if an integer was found in the title such as a pull request id. This change will require the titles to contain "Bug XXX", "bug XXX", or "bug - XXX".

### DIFF
--- a/Bugzilla/API/V1/Github.pm
+++ b/Bugzilla/API/V1/Github.pm
@@ -18,19 +18,6 @@ use Bugzilla::User;
 use Digest::SHA qw(hmac_sha256_hex);
 use Mojo::Util  qw(secure_compare);
 
-# Stolen from pylib/mozautomation/mozautomation/commitparser.py from
-# https://hg.mozilla.org/hgcustom/version-control-tools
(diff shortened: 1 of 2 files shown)


Bugzilla/API/V1/Github.pm
```diff
@@ -151,7 +137,8 @@ sub pull_request {
     WHERE    => {'bug_id != ? AND NOT isobsolete' => $bug->id}
   });
   foreach my $attachment (@$other_attachments) {
-    next if $attachment->data ne $html_url; # same pr number but different repo so skip it
+    # same pr number but different repo so skip it
+    next if $attachment->data ne $html_url;
```
cgsheeh, line 141: Nit: newline here.

cgsheeh (approved):
Nice. We have a similar bug in v-c-t with that regex that I never tracked down.

PR #2012, opened 2023-02-02
Bug 1802052 - GitHub API endpoint for GitHub/Bugzilla push hook to link bugs to GitHub commits

dklawren:
* If the bug does not have the “leave-open” keyword:
  * If the bug status is not RESOLVED or VERIFIED, resolve the bug with status RESOLVED FIXED.
  * Set the bug’s “status-firefox###” tracking flag to “fixed”.
  * Bug’s have “status-firefox###” flags for the currently Nighty, Beta, and Release versions.
  * The hook should be able to identify the correct “status-firefox###” flag version depending on whether the commit is
landing on “main” or a “releases_v###” release branch as that's information that'll be readily available to whatever bot is  doing the marking.
	* https://github.com/mozilla-mobile/firefox-android/branches
* The “qe-verify” flag has three values: empty, “+”, or “-”.
  * If the bug’s “qe-verify” flag is empty (not “+” or “-”), then set “qe-verify” to “+” so QA will test the fix and then manually reset the flag to empty.
  * We only want to set this flag after the bug has been RESOLVED FIXED and not “leave-open”.


Bugzilla/API/V1/Github.pm
```diff
@@ -164,7 +171,167 @@ sub pull_request {
...
+  my %update_bugs;
+
+  # Create a separate comment for each commit
+  foreach my $commit (@{$commits}) {
+    my $message = $commit->{message};
+    my $url     = $commit->{url};
+
+    if (!$url || !$message) {
+      return $self->code_error('github_pr_invalid_json');
+    }
```
cgsheeh, line 239: Can we log and/or return information about what is wrong with the JSON here?

Bugzilla/API/V1/Github.pm
```diff
@@ -164,7 +171,167 @@ sub pull_request {
...
+
+    # Find bug ID in the title and see if bug exists
+    my ($bug_id) = $message =~ /\b[Bb]ug[ -](\d+)\b/;
+    next if !$bug_id;
+
+    my $comment_text = "Authored by https://github.com/$pusher\n$url\n$message";
```
cgsheeh, line 245: Looking at the requirements doc, I think we want to use `$commit->author->username` instead of `$pusher`. In the example response it is the same user, but someone could author a patch and a different user could push it to the repo.

The reqs doc also requests just the first line of the commit message, so we need to parse that out of `$message` in case it is multi-line.

It's also unclear from that doc if we are supposed to strip the bug number from the commit message's first line when adding to the comment, we may want to ask for clarification on that.

Bugzilla/API/V1/Github.pm
```diff
@@ -164,7 +171,167 @@ sub pull_request {
...
+  $auto_user->{bless_groups} = [Bugzilla::Group->get_all];
+  Bugzilla->set_user($auto_user);
+
+  my $dbh = Bugzilla->dbh;
+  $dbh->bz_start_transaction;
+
+  # Actually create the comments in this loop
+  foreach my $bug_id (keys %update_bugs) {
+    my $bug = Bugzilla::Bug->new({id => $bug_id, cache => 1});
+    next if $bug->{error};
```
cgsheeh, line 268: What does this `next if` catch? Invalid bug number?

Bugzilla/API/V1/Github.pm
```diff
@@ -164,7 +171,167 @@ sub pull_request {
...
+
+    # Create a single comment if one or more commits reference the same bug
+    my $comment_text;
+    foreach my $comment (@{$update_bugs{$bug_id}}) {
+      $comment_text .= $comment->{text} . "\n\n";
+    }
+
+    $bug->add_comment($comment_text);
+
+    # If the bug does not have the keywork 'leave-open',
```
cgsheeh, line 278:
```suggestion
    # If the bug does not have the keyword 'leave-open',
```

Bugzilla/API/V1/Github.pm
```diff
@@ -164,7 +171,167 @@ sub pull_request {
...
+            [],
+            [{
+              flagtype => $qe_flag,
+              setter   => Bugzilla->user,
+              status   => '+',
+              type_id  => $qe_flag->id,
+            }]
+          );
+        }
+      }
```
cgsheeh, line 310: Setting a boolean flag and then running the code we want in a separate `if` seems unnecessary at first glance. We could eliminate the extra flag with the following change, any reason this wouldn't work?

```suggestion
        # Ignore for all flags except `qe-verify`.
        next if $flag->name ne 'qe-verify';

        my $qe_flag = Bugzilla::FlagType->new({name => 'qe-verify'});
        if ($qe_flag) {
          $bug->set_flags(
            [],
            [{
              flagtype => $qe_flag,
              setter   => Bugzilla->user,
              status   => '+',
              type_id  => $qe_flag->id,
            }]
          );
        }
        last;
      }
```

Bugzilla/API/V1/Github.pm
```diff
@@ -173,4 +340,49 @@ sub verify_signature {
...
+  my $version;
+  if ($branch eq 'main' || $branch eq 'master') {
+    my $versions  = fetch_product_versions('firefox');
+    return if (!%$versions || !exists $versions->{FIREFOX_NIGHTLY});
+    ($version) = split /[.]/, $versions->{FIREFOX_NIGHTLY};
+  }
+  else {
+    ($version) = $branch =~ /^releases_v(\d+)$/;
+  }
+  return if !$version;
```
cgsheeh, line 363: Can you add more comments describing this?

cgsheeh (changes requested):

Bugzilla/API/V1/Github.pm
```diff
@@ -164,7 +171,167 @@ sub pull_request {
...
+sub push_comment {
+  my ($self) = @_;
+  my $template = Bugzilla->template;
+  Bugzilla->usage_mode(USAGE_MODE_MOJO_REST);
+
+  # Return early if push commenting is not allowed
+  return $self->code_error('github_push_comment_disabled')
+    if !Bugzilla->params->{github_push_comment_enabled};
+
+  # Return early if not a push or ping event
```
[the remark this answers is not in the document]
dklawren: Not sure how this would work since they caller still uses the full endpoint even for pings (i.e. `/rest/github/{push_comment,pull_request}`). This only happens once that I know of when the admin first sets up the webhook in Github. The amount of code that is duplicated is so small I am also reluctant to create a separate ping function to call from both functions. Ideas?

Bugzilla/API/V1/Github.pm
dklawren: I will add a logging call that we can use for debugging if needed but I am reluctant to return the Perl error message verbatim so as to not open a potential security issue in the future. They can contact a BMO admin to help with debugging if needed.

dklawren: > I will add a logging call that we can use for debugging if needed but I am reluctant to return the Perl error message verbatim so as to not open a potential security issue in the future. They can contact a BMO admin to help with debugging if needed.

Scratch that. I am not failing if the JSON parser fails, only if a required key(s) is missing. I will update the error to be more descriptive if possible.

Bugzilla/API/V1/Github.pm
dklawren: If the bug is not visible or does not exist, etc. then $bug->{error} will be populated. We then pass on to the next commit silently and not cause an exception. If we throw an exception if will not get to the other commits which may have valid IDs. We just treat the invalid IDs as if they commit did not match our bug id format rules.

Bugzilla/API/V1/Github.pm
dklawren: Sure

Bugzilla/API/V1/Github.pm
dklawren: That would work.

dklawren: > That would work.

Oh wait. The reason I did it that way is I only want to add the qe-verify flag if it is empty and not already set to either + or -.  So if $found_flag is true, then I don't want to do anything. So with your change I would be trying to add the new flag even if the flag is already set.

cgsheeh (approved):

PR #2015, opened 2023-02-23
Bug 1818541 - add bug keywords to Lando uplift GET response

dklawren:
Adds keywords to the bug data returned.

### DIFF
--- a/extensions/PhabBugz/lib/API/V1/Lando.pm
+++ b/extensions/PhabBugz/lib/API/V1/Lando.pm
@@ -51,7 +51,11 @@ sub get {
   foreach my $id (@ids) {
     my $bug = Bugzilla::Bug->check($id);
 
-    my $result = {id => $bug->id, whiteboard => $bug->status_whiteboard};
+    my $result = {
+      id         => $bug->id,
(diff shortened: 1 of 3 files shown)


cgsheeh (approved):
Looks good.

extensions/PhabBugz/lib/API/V1/Lando.pm
```diff
@@ -27,6 +27,7 @@ sub setup_routes {
   $r->get('/lando/uplift')->to('PhabBugz::API::V1::Lando#get');
   $r->get('/lando/uplift/:id')->to('PhabBugz::API::V1::Lando#get');
   $r->put('/lando/uplift')->to('PhabBugz::API::V1::Lando#update');
+  $r->put('/lando/uplift/:id')->to('PhabBugz::API::V1::Lando#update');
```
cgsheeh, line 30: Let's just support having the bug IDs in the JSON body. I've already removed it on the Lando side so having a single supported way to pass the IDs would make the code a bit simpler.

cgsheeh (changes requested):

cgsheeh (approved):
LGTM!

PR #2019, opened 2023-02-24
Bug 1799913 - Remove parameters from crash signatures in bug summaries

dklawren:


scripts/bug1799913.pl
```diff
@@ -0,0 +1,84 @@
...
+
+my $updated = 0;
+foreach my $rh_bug (@$bugs) {
+  my $bug_id    = $rh_bug->{bug_id};
+  my $summary   = $rh_bug->{summary};
+  my $signature = $rh_bug->{cf_crash_signature};
+
+  # update summary
+  my $update_summary = $summary;
+  $update_summary =~ s/\[@ (.*)[(].*[)]\]/\[@ $1\]/g;
```
globau, line 52: Are there crash signatures with multiple `[..]` instances?  If so, this regex won't process them correctly.

scripts/bug1799913.pl
```diff
@@ -0,0 +1,84 @@
...
+getc;
+
+my $updated = 0;
+foreach my $rh_bug (@$bugs) {
+  my $bug_id    = $rh_bug->{bug_id};
+  my $summary   = $rh_bug->{summary};
+  my $signature = $rh_bug->{cf_crash_signature};
+
+  # update summary
+  my $update_summary = $summary;
```
globau, line 51: This should be `$updated_summary`

scripts/generate_bmo_data.pl
```diff
@@ -908,7 +908,10 @@
   my $values   = delete $flag_data->{values};
   my $products = delete $flag_data->{products};
 
-  my $flag_obj = Bugzilla::Extension::TrackingFlags::Flag->create($flag_data);
+  my $flag_obj = Bugzilla::Extension::TrackingFlags::Flag->new({name => $flag_data->{name}});
+  next if $flag_obj;
+
+  $flag_obj = Bugzilla::Extension::TrackingFlags::Flag->create($flag_data);
```
globau, line 914: Appears to be unrelated

scripts/bug1799913.pl
globau: For matching the `(..)` part I recommend `\([^)]*\)`

globau (changes requested):

scripts/generate_bmo_data.pl
globau: This still appears to be unrelated.

scripts/bug1799913.pl
```diff
@@ -0,0 +1,86 @@
...
+sub is_same {
+  my ($old, $new) = @_;
+  $old =~ s/[\015\012]+/ /g;
+  $new =~ s/[\015\012]+/ /g;
+  return trim($old) eq trim($new);
+}
+
+sub update_crash_signatures {
+  my $text = shift;
+  $text =~ s/\[@ (.*?)\([^)]*\)\]/\[\@$1\]/igx;
```
globau, line 84: The `/x` here makes reading the regex unintuitive.

At first glance the regex should be stripping the space after the `@`, as there's a space being matched on, but not in the replacement.  However, what's actually happening is the space is being stripped from the regex before parsing due to `/x`, making the regex actually `s/\[@(.*?)\([^)]*\)\]/\[\@$1\]/ig`, so `$1` captures the space after `@`.

The `/i` is unnecessary; there's nothing that's case sensitive.

Can you try the following suggestion against your test cases:

`s/\[@ (.+?)\([^)]*\)\]/[\@ $1]/g`

globau (changes requested):

globau (approved):

PR #2022, opened 2023-03-02
Bug 1820042 - Calling bug->update twice on the same bug object is generally a bad idea so we load a second copy during the uplift process

dklawren:


cgsheeh (approved):
Thanks for fixing this!

PR #2024, opened 2023-03-07
Bug 1820859 - Wrap code that looks for a latest nightly/stable versions of Firefox in try/catch block in case of failure

dklawren:
* Wrap the call to the product versions API in a try block.
* This allows bug page and search page to load properly in the even the endpoint is unavailable or misconfigured.
* Code in BMO extension needed updating to not check for a hash key before determining if the hash has values..

### DIFF
--- a/Bugzilla/Util.pm
+++ b/Bugzilla/Util.pm
@@ -972,28 +972,35 @@ sub extract_nicks {
 }
 
 sub fetch_product_versions {
-  my ($product)  = @_;
-  my $key      = "${product}_versions";
-  my $versions = Bugzilla->request_cache->{$key}
-    || Bugzilla->memcached->get_data({key => $key});
+  my ($product) = @_;
+  my $key = "${product}_versions";
+  my $versions
+    = Bugzilla->request_cache->{$key}
+    || Bugzilla->memcached->get_data({key => $key})
+    || {};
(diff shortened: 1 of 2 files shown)


Bugzilla/Util.pm
```diff
@@ -972,28 +972,35 @@ sub extract_nicks {
...
-    Bugzilla->memcached->set_data({
-      key   => $key,
-      value => $versions,
-
-      # Cache for 1 day if the data is available, otherwise retry in 5 min
-      expires_in => $response->is_success ? 86_400 : 300,
-    });
+    catch {
+      WARN("ERROR: Unable to retrieve product versions: $_");
+    };
```
cgsheeh, line 1003: Nit: you could split this into a few smaller `try/catch` blocks so we know which specific piece caused a problem. Was there a network error connecting to the service, did we fail to decode JSON, did we fail to populate memcached, etc.

cgsheeh (approved):

Bugzilla/Util.pm
dklawren: Actually I think the single try/catch will accomplish similar as it will WARN about whatever the last failure was in the process. The variable `$_` contains the last error message and will be added to the log.

PR #2026, opened 2023-03-08
Bug 1821175 - Push comment support is too strict when validating JSON data from Github and throwing error when sending event

dklawren:


qa/t/rest_github_push_comment.t
```diff
@@ -98,14 +98,21 @@ $t->post_ok(
...
+  created    => false,
+  deleted    => false,
+  forced     => false,
   commits    => [{
-    author => {username => 'foobar'},
-    url => 'https://github.com/mozilla-bteam/bmo/commit/abcdefghijklmnopqrstuvwxyz',
-    message => "Bug $bug_id - Test Github Push Comment",
+    author   => {username => 'foobar'},
+    url      => 'https://github.com/mozilla-bteam/bmo/commit/abcdefghijklmnopqrstuvwxyz',
+    message  => "Bug $bug_id - Test Github Push Comment",
```
cgsheeh, line 107: This is out of scope for this PR, but we do parse the first line out of the commit message IIRC. We probably want to add a multi-line commit message to the test data to ensure that parsing is working correctly.

cgsheeh (approved):

qa/t/rest_github_push_comment.t
dklawren: Good idea. I will file a bug for that.

PR #2028, opened 2023-03-09
Bug 1821251 - fix accessibilty to search field in header, and add a shortcut to focus it

globau:
- move `<h2>` outside of the quicksearch combo role; it being there breaks the ability for screen readers to navigate to the search field
- add a keyboard shortcut to focus quicksearch: alt+s / opt+s


dklawren (approved):
LGTM. r=dkl

PR #2029, opened 2023-03-10
Bug 1821521 - Only perform actions to a bug if branch is main, master, or a release branch. Ignore feature, bugfix branches, etc.

dklawren:


Bugzilla/API/V1/Github.pm
```diff
@@ -221,9 +221,11 @@ sub push_comment {
   my $ref        = $payload->{ref};
   my $repository = $payload->{repository}->{full_name};
   my $commits    = $payload->{commits};
+  my ($branch)   = $ref =~ /refs\/heads\/(.*)$/;
 
-  # Return success early if there are no commits or the ref is not a branch
-  if (!@{$commits} || $ref !~ /refs\/heads\//) {
+  # Return success early if there are no commits
+  # or the branch is not one we are interested in.
+  if (!@{$commits} || $branch !~ /^(?:main|master|releases_v\d+)$/) {
```
cgsheeh, line 228: I think we should be checking the value of the `default_branch` key on the repo object instead of hard-coding `master` and `main` here.

Also, the [behaviour requested in the bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1821521#c2) is for landings to `release_*` to only set flags and **not** close the bug.

cgsheeh (changes requested):

Bugzilla/API/V1/Github.pm
dklawren: > I think we should be checking the value of the `default_branch` key on the repo object instead of hard-coding `master` and `main` here.

Cool. Will change to use that.

> Also, the [behaviour requested in the bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1821521#c2) is for landings to `release_*` to only set flags and **not** close the bug.

Yeah it was brought to my attention yesterday by ryan that needed to be changes so I filed it as a separate bug just now https://bugzilla.mozilla.org/show_bug.cgi?id=1821694

dklawren:
> Also another nit: the first line of the commit message was copied from the bug, and on the bug there was a typo of the braces: `[GITHUB}`. We should fix that here before landing. :)

Heh. Sorry bout that. Fixed.

cgsheeh (approved):
r=sheehan

PR #2030, opened 2023-03-10
Bug 1821679 - [GITHUB] Display the name of the branch the commit was made in the comment added to the bug

dklawren:


Bugzilla/API/V1/Github.pm
```diff
@@ -355,10 +356,7 @@ sub _verify_signature {
 # If the ref matches a certain branch pattern for the repo we are interested
 # in, then we also need to set the appropriate status flag to 'fixed'.
 sub _set_status_flag {
-  my ($self, $bug, $ref) = @_;
-
-  my ($branch) = $ref =~ /refs\/heads\/(.*)$/;
-  return if !$branch;
```
cgsheeh, line 361: Note: this removes the `return if !$branch;` line, and it doesn't look like we check that we parsed a branch from the ref anywhere else. Is this expected/okay?

cgsheeh (changes requested):
Just have a question regarding the possible case when `$branch` is empty.

Bugzilla/API/V1/Github.pm
dklawren: Normally, yes, we would return if !$branch, but the change in #2029 will make this not necessary.

```perl
  if (!@{$commits} || $branch !~ /^(?:$default_branch|releases_v\d+)$/) {
    return $self->render(json => {error => 0});
  }
```

PR #2032, opened 2023-03-10
Bug 1821739 - [GITHUB] When a status flag is set from a commit, the flag is stored in the flags table, but it is not recorded properly in the bugs_activity table

dklawren:


cgsheeh (changes requested):
The commit message doesn't make sense ("the flag is stored by it is not recorded..."). Can you clarify it?

The commit message says the flag isn't being stored properly, but the bug only explains what problem this is causing. Can you explain how the bug isn't being properly stored and how this PR fixes the problem in the commit message also? The code itself looks good, I just want to understand what's happening.

dklawren:
> The commit message doesn't make sense ("the flag is stored by it is not recorded..."). Can you clarify it?
> 
> The commit message says the flag isn't being stored properly, but the bug only explains what problem this is causing. Can you explain how the bug isn't being properly stored and how this PR fixes the problem in the commit message also? The code itself looks good, I just want to understand what's happening.

Sorry for the fumble fingers when filling out the summary initially on the bug report. I need to get better at going back reading what I am typing :) 

Anyway, as coded, I was creating the flag for the bug properly in the tracking flags table, but neglected to also put the proper entry into the bugs_activity table like we do in other places. This means that when viewing a bug report, the change was not visible in line with the comments along with other bug changes. Normal bug changes handled by $bug->update() when $bug is a Bugzilla::Bug object, these entries get added automatically. TrackingFlags is a BMO add on that we do out side of normal Bugzilla::Bug changes so we have to make sure we add the entries manually when adding/removing tracking flags. I originally forgot to do so in the Github API and this adds that function.

cgsheeh (approved):
Makes sense to me.

Looks like we have a few places where we do this different behaviour when `$removed eq '---'`. It might make sense to have a function for the logic around which table to update so we don't have to remember in the future.

PR #2033, opened 2023-03-10
Bug 1821478 - [GITHUB] Fall back to author.name if author.username is missing

dklawren:
According to https://docs.github.com/en/webhooks-and-events/webhooks/webhook-events-and-payloads#push, author.name is required but author.username is not required to be included in the event JSON. So no need to use the committer object as a last resort.

If we have to user author.name, then we omit the `https://github.com/<username>` part in the comment.


cgsheeh (approved):
Makes sense to me.

PR #2034, opened 2023-03-12
Bug 1821386 - Set the Milestone field when resolving bugs after PR merges

dklawren:
* If a bug is being closed and it is firefox-android, then we also set the milestone to the latest nightly branch
* generate_bmo_data.pl adds the 111 Branch milestone needed for testing.
* Update the test script to check that the milestone has been set properly if bug is closed.

### DIFF
--- a/Bugzilla/API/V1/Github.pm
+++ b/Bugzilla/API/V1/Github.pm
@@ -14,6 +14,7 @@ use Bugzilla::Bug;
 use Bugzilla::BugMail;
 use Bugzilla::Constants;
 use Bugzilla::Group;
+use Bugzilla::Milestone;
 use Bugzilla::User;
 use Bugzilla::Util qw(fetch_product_versions);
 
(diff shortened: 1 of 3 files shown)


Bugzilla/API/V1/Github.pm
```diff
@@ -423,4 +428,28 @@ sub _set_status_flag {
...
+  # returns all of the current product versions so we can use that.
+  my $versions = fetch_product_versions('firefox');
+  return if (!%$versions || !exists $versions->{FIREFOX_NIGHTLY});
+  my ($version) = split /[.]/, $versions->{FIREFOX_NIGHTLY};
+  return if !$version;
+
+  # Load the appropriate milestone.
+  my $value = "$version Branch";
+  my $milestone
+    = Bugzilla::Milestone->new({product => $bug->product_obj, name => $value});
```
cgsheeh, line 448: Out of scope for this change, but `Bugzilla::Milestone->new` should take some kind of `version` argument and do this `"$version Branch"` string interpolation in the function call instead of being a responsibility of the caller.

cgsheeh (approved):
LGTM.

PR #2042, opened 2023-03-29
Bug 1823503 - After add comment, sometimes all status flags are changed to `fixed` by itself.

dklawren:


dklawren:
Although I would still like to see these changes happen at some point i am going to make this WIP for now as I want to add some test cases and some other improvements.

PR #2043, opened 2023-04-01
Bug 1825848 - LogActivityEntry is be called twice instead of once when the github push comment API code is executed

dklawren:
* Update the test script to look for proper history entries to make sure we are adding them properly and no duplicates
* Switch to using $bug->set_all before $bug->update. This is how process_bug.cgi work and we should us that so as to not duplicate by performing update steps manually. This also allows to use the LogActivityEntry that is in TrackingFlags extension instead of logging ourselves as before causing the duplicate issue.

### DIFF
--- a/Bugzilla/API/V1/Github.pm
+++ b/Bugzilla/API/V1/Github.pm
@@ -308,8 +308,6 @@ sub push_comment {
       }
     };
 
-    # $bug->add_comment($comment_text);
-
     # If the commit is on the default branch and the bug does not have
     # the keyword 'leave-open', we can also close the bug as RESOLVED/FIXED.
     if ($ref =~ /refs\/heads\/$default_branch/ && !$bug->has_keyword('leave-open')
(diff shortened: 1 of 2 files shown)


Bugzilla/API/V1/Github.pm
```diff
@@ -303,7 +301,14 @@ sub push_comment {
...
 
-    $bug->add_comment($comment_text);
+    # Set all parameters
+    my $set_all = {
+      comment => {
+        body => $comment_text
+      }
+    };
+
+    # $bug->add_comment($comment_text);
```
cgsheeh, line 311: Looks like this was leftover from development.

Bugzilla/API/V1/Github.pm
```diff
@@ -312,7 +317,10 @@ sub push_comment {
       && $bug->status ne 'VERIFIED')
     {
       # Set the bugs status to RESOLVED/FIXED
-      $bug->set_bug_status('RESOLVED', {resolution => 'FIXED'});
+      $set_all->{bug_status} = 'RESOLVED';
+      $set_all->{resolution} = 'FIXED';
+
+      #$bug->set_bug_status('RESOLVED', {resolution => 'FIXED'});
```
cgsheeh, line 323: Leftover from dev here.

cgsheeh (changes requested):

cgsheeh (approved):

PR #2045, opened 2023-04-07
Bug 1825899 - BMO push hook shouldn't set the status flag to fixed if the leave-open keyword is set

dklawren:
For a better idea of why this change is being made and what the flow should be now with the latest fix, see https://bugzilla.mozilla.org/show_bug.cgi?id=1825899#c1 Thanks!

### DIFF
--- a/Bugzilla/API/V1/Github.pm
+++ b/Bugzilla/API/V1/Github.pm
@@ -338,17 +338,17 @@ sub push_comment {
         }
       }
 
-      # Update the milestone to the nightly branch if default branch
       # Currently tailored for mozilla-mobile/firefox-android only
-      if ($ref =~ /refs\/heads\/$default_branch/) {
-        $self->_set_nightly_milestone($bug, $branch)
-          if $repository eq 'mozilla-mobile/firefox-android';
-      }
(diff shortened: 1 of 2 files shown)


Bugzilla/API/V1/Github.pm
```diff
@@ -339,14 +338,18 @@ sub push_comment {
         }
       }
 
-      # Update the milestone to the nightly branch if closing the bug.
+      # Update the milestone to the nightly branch if default branch
       # Currently tailored for mozilla-mobile/firefox-android only
-      $self->_set_nightly_milestone($bug, $branch) if $repository eq 'mozilla-mobile/firefox-android';
-    }
+      if ($ref =~ /refs\/heads\/$default_branch/) {
```
cgsheeh, line 343: Nit: you could move the `$repository eq 'mozilla-mobile/firefox-android'` conditional into this one instead of having two separate `if` checks.

cgsheeh (approved):

Bugzilla/API/V1/Github.pm
dklawren: Good point. Not sure why I didnt see that before :)

PR #2047, opened 2023-04-11
Bug 1824111 - Update group permission needed to view the groups using the `/rest/group` endpoint

dklawren:
This change makes it where anyone in mozilla-employee-confidential can see information about any group in BMO. Otherwise the user can only see if they are able to bless others into the group or have the empowered 'editusers' group.


dklawren:
> WFM - a test for this would be nice.

Adding a test for groups now.

PR #2052, opened 2023-04-19
Bug 1828677 - Invalid regex in BMO feed reader: `The regular expression you provided ':?:phab\\-bot[[:>:]]' is not valid`

dklawren:
https://stackoverflow.com/questions/59998409/error-code-3685-illegal-argument-to-a-regular-expression

"This question will probably become more popular as adoption of MySQL 8.0 increases and previously-stored SQL queries using REGEXP start to break.

According to [MySQL 8.0 Reference Manual / ... / Regular Expressions](https://dev.mysql.com/doc/refman/8.0/en/regexp.html), "MySQL implements regular expression support using International Components for Unicode (ICU)."

According to [MySQL 5.6 Reference Manual / ... / Regular Expressions](https://dev.mysql.com/doc/refman/5.6/en/regexp.html), "MySQL uses Henry Spencer's implementation of regular expressions."

Therefore, since you are using MySQL 8.0, rather than using [[:<:]] and [[:>:]], you can now use \b."

We are using 8.0 now with the move to GCP so I just updated the regexp to use `\b`.

### DIFF
--- a/extensions/PhabBugz/lib/Feed.pm
+++ b/extensions/PhabBugz/lib/Feed.pm
@@ -717,7 +717,7 @@ sub process_new_user {
   # CHECK AND WARN FOR POSSIBLE USERNAME SQUATTING
   INFO("Checking for username squatters");
   my $dbh     = Bugzilla->dbh;
-  my $regexp  = $dbh->quote(":?:" . quotemeta($phab_user->name) . "[[:>:]]");
+  my $regexp  = $dbh->quote($dbh->WORD_START . ':?:' . quotemeta($phab_user->name) . $dbh->WORD_END);
   my $results = $dbh->selectall_arrayref("
         SELECT userid, login_name, realname
           FROM profiles


cgsheeh (approved):
Nice, LGTM.

Sanity check, we also have some `WORD_START` and `WORD_END` constants [here](https://github.com/mozilla-bteam/bmo/blob/b078475b56a58443f1b55deceaf50527991895f7/Bugzilla/DB.pm#L122). Do those need to be updated also?

dklawren:
> Nice, LGTM.
> 
> Sanity check, we also have some `WORD_START` and `WORD_END` constants [here](https://github.com/mozilla-bteam/bmo/blob/b078475b56a58443f1b55deceaf50527991895f7/Bugzilla/DB.pm#L122). Do those need to be updated also?

actually i should have been using the WORD_START and WORD_END all along instead of doing it explicitly using `[[:<:]]` so I have converted to use those instead. The comments word WORD_START do a decent job of explaining why not to use `[[:<:]]` and it was written 13 years ago :)

cgsheeh:
> > > > Looks like the file changes in `conf/` were added in error.
> > > 
> > > 
> > > They are from a rebase from master from another commit I made for the a different conduit-suite fix.
> > 
> > 
> > I'm confused - why are they included in this PR?
> 
> Not sure exactly. Basically I wanted this PR to be based on the latest changes on master, so I did:
> 
> ```
> $ cd ../master
> $ git fetch upstream
> $ git merge upstream/master
> $ cd ../1828677
> $ git stash
> $ git rebase master
> $ git stash pop
> [make changes in extensions/PhabBugz/lib/Feed.pm]
> $ git add extensions/PhabBugz/lib/Feed.pm
> $ git commit -m "Updated to use WORD_START/WORD_END instead of \b"
> $ git push origin 1828677
> ```
> 
> the 1828677 directory is a git work-tree off of the master branch so I am not sure that makes things more prone to breakage but has worked fine for me over time. If this is a problem, I can create a new fresh PR and we can continue the review there. Lesson is to not rebase into a PR until it is ready to be merged I suppose.

Can you create a commit with only the relevant changes, move the branch to it and force push here? It's a one file change so I think a force-push would be fine.

I would r+ this but I imagine it won't land correctly with the other irrelevant changes already in the PR.

dklawren:
> Can you create a commit with only the relevant changes, move the branch to it and force push here? It's a one file change so I think a force-push would be fine.
> 
> I would r+ this but I imagine it won't land correctly with the other irrelevant changes already in the PR.

Done. Please checkout new version of this PR and review. Thanks!

cgsheeh (approved):

PR #2053, opened 2023-04-26
Bug 1825092 - Pulsebot should comment the changeset hash of the landed changesets to the bug for uplift landings

dklawren:
* Moves PhabBugz/lib/API/V1/Lando.pm to PhabBugz/lib/API/V1/PhabBugz.pm
* Adds Pulsebot related code PhabBugz.pm since Lando and Pulsebot needs are very similar.
* Depending on which user is logged in, the fields accepted and returned are different
* Added a qa/t/rest_pulsebot.t test case for testing the Pulsebot functionality and rest_lando.t for proper Lando functionality.
*  If the commit message contains a=someone, then the uplift comment tag is added.


dklawren:
Actually I think I might have left a piece out and so no review yet.

"Pulsebot should leave comments on the bug with links to the uplifted revisions, and tag the comments it leaves with the uplift tag. See https://bugzilla.mozilla.org/show_bug.cgi?id=1818762#c28 as an example."

dklawren:
> Actually I think I might have left a piece out and so no review yet.
> 
> "Pulsebot should leave comments on the bug with links to the uplifted revisions, and tag the comments it leaves with the uplift tag. See [bugzilla.mozilla.org/show_bug.cgi?id=1818762#c28](https://bugzilla.mozilla.org/show_bug.cgi?id=1818762#c28) as an example."

This is ready for review. Any other changes needed will be made on the Pulsebot side. This can be deployed even before the Pulsebot changes are live in production. Thanks!

extensions/PhabBugz/lib/Constants.pm
```diff
@@ -32,4 +37,22 @@ use constant PHAB_USER_POLL_SECONDS  => $ENV{PHAB_USER_POLL} // 60;
 use constant PHAB_GROUP_POLL_SECONDS => $ENV{PHAB_GROUP_POLL} // 300;
 use constant PHAB_TIMEOUT            => $ENV{PHAB_TIMEOUT} // 60;
 
+use constant LANDO_AUTOMATION_USER   => 'lobot@bmo.tld';
+use constant LANDO_BUG_UPDATE_FIELDS => qw(
+  cf_status_firefox
```
cgsheeh, line 42: When we want to add new fields to the Lando endpoint we can just add them here, right?? That would be quite convenient.

cgsheeh (changes requested):
Nit: this PR does a `git rm` of `Lando.pm` and then creates a new `PhabBugz.pm` with the same contents. In the future we should do a `git mv` for renaming to keep Git history and blame intact.

dklawren:
> Nit: this PR does a `git rm` of `Lando.pm` and then creates a new `PhabBugz.pm` with the same contents. In the future we should do a `git mv` for renaming to keep Git history and blame intact.
> 
> ETA: it should be fine for this PR since it's a small file with just a few changes, but wanted to mention for the future. +1

Ah yeah `git mv` Haven't used that in a while and forgot about it. I will definitely use that in the future.

extensions/PhabBugz/lib/Constants.pm
```diff
@@ -32,4 +37,22 @@ use constant PHAB_USER_POLL_SECONDS  => $ENV{PHAB_USER_POLL} // 60;
...
+  status_whiteboard
+);
+
+use constant PULSEBOT_AUTOMATION_USER   => 'pulsebot@bmo.tld';
+use constant PULSEBOT_UPLIFT_REPOS => qw(
+  integration/autoland
```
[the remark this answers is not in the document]
dklawren: Yeah I originally was going to handle the checking of repo on the bmo side but have changed to that on the pulsebot end. So this constant just needs to be removed.

extensions/PhabBugz/lib/Constants.pm
dklawren: > When we want to add new fields to the Lando endpoint we can just add them here, right?? That would be quite convenient.

That is the goal, yes.

cgsheeh (approved):

PR #2054, opened 2023-04-26
Bug 1825961: set `approval-mozilla-{repo}` flags on uplift revision attachments

cgsheeh:
When a new Phabricator revision attachment is being added to a bug,
check if the repo associated with the revision is an uplift repo and
set the appropriate `approval-mozilla-{repo}` flag if necessary.

To determine if a repo is an uplift repository, we add the `projects`
attachment from Phabricator to the Conudit API query which returns
repository data. We add a `projects_raw` and `projects` field which
is the raw API response and an array of built `Project` objects in
Bugzilla, respectively. Then we add an `is_uplift_repo` function which
iterates over the `project` attribute and checks for a project with
the name matching `uplift`.

### DIFF
--- a/extensions/PhabBugz/lib/Repository.pm
+++ b/extensions/PhabBugz/lib/Repository.pm
@@ -12,6 +12,8 @@ use Moo;
 use Types::Standard -all;
 
 use Bugzilla::Util qw(trim);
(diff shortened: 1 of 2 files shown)


dklawren (approved):
Otherwise code looks great and all my testing checked out. r=dkl

PR #2056, opened 2023-04-26
Bug 1825961: sync revision status with `approval-mozilla-{repo}` flag

cgsheeh:
When processing revision attachments changes to bugs in phab-bot
feed reader, check if the attachment belongs to an uplift revision
and sync the approval flag statuses as appropriate. To do this
we make the `set_approval_flag_statuses` function more generic
by accepting a flag status instead of hard-coding the `?` status.
We also move the call to update the attachment to the caller.


dklawren (approved):
LGTM. r=dkl

PR #2060, opened 2023-05-05
Bug 1831576 - Modify web bug bounty form

dklawren:
Added messages (from bug report) to the web bounty form to inform about HackerOne. I also changed the product from mozilla.org to Websites in Extension.pm since Websites is not the proper product and was not changed when the restructuring happened a while back.


cgsheeh (approved):
CI is red here FYI.

dklawren:
> CI is red here FYI.

Thanks. Fixed.

PR #2062, opened 2023-05-10
Bug 1832364 - Add support for AWS SES configuration sets to Mailer

dklawren:
Simple change that adds a special SES header to all emails that has a value that can be set with an environment variable. This allows for BMO emails to be uniquely identified in SES when grouped with other sites and dealing with bounces, etc.


cgsheeh (approved):
LGTM.

PR #2064, opened 2023-05-11
Bug 1832391 - Phabricator feed process crashes when a revision is in draft status and setting approval-mozilla-* flag on the new attachment

dklawren:
This patch adds the '?' status for revisions that are also in the draft state which is the state the revision is set initially when it is first created. THis error occurred when a new revision was created against a repository (mozilla-beta) in Phabricator that had the uplift tag associated with it.

### DIFF
--- a/extensions/PhabBugz/lib/Feed.pm
+++ b/extensions/PhabBugz/lib/Feed.pm
@@ -644,6 +644,7 @@ sub process_revision_change {
         'abandoned'       => '-',
         'accepted'        => '+',
         'changes-planned' => '?',
+        'draft'           => '?',
         'needs-review'    => '?',
         'needs-revision'  => '-',
       };
--- a/extensions/PhabBugz/lib/Feed.pm
+++ b/extensions/PhabBugz/lib/Feed.pm
@@ -643,8 +643,8 @@ sub process_revision_change {
       my $revision_status_flag_map = {
         'abandoned'       => '-',
         'accepted'        => '+',
-        'changes-planned' => '?',
-        'draft'           => '?',
+        'changes-planned' => 'X',
+        'draft'           => 'X',
         'needs-review'    => '?',
         'needs-revision'  => '-',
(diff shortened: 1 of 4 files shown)


extensions/PhabBugz/lib/Feed.pm
```diff
@@ -644,6 +644,7 @@ sub process_revision_change {
         'abandoned'       => '-',
         'accepted'        => '+',
         'changes-planned' => '?',
+        'draft'           => '?',
```
cgsheeh, line 647: Could we set the flag to be cleared when the state is `draft`? We should probably also do so for `changes-planned`.

I should note that I copied this mapping from [here](https://github.com/mozilla-bteam/bmo/blob/master/extensions/PhabBugz/lib/WebService.pm#L151) where the draft status isn't present - is that a bug?

cgsheeh (changes requested):

extensions/PhabBugz/lib/Feed.pm
dklawren: > Could we set the flag to be cleared when the state is `draft`? We should probably also do so for `changes-planned`.
> 

Yes we would just set the flag status to 'X' if it already exists, otherwise if the revision is new and in the draft state, then we would just not set a flag at all.

> I should note that I copied this mapping from [here](https://github.com/mozilla-bteam/bmo/blob/master/extensions/PhabBugz/lib/WebService.pm?rgh-link-date=2023-05-11T15%3A27%3A03Z#L151) where the draft status also isn't present - is that a bug?

Not reall but we can add it there for consistency. That mapping is for human readable version of the statuses. If the mapping did not have an entry for the state it would just show the raw status from Phabricator. In this case it would just show 'draft'. If we add a new entry it would just be 'Draft' ;)

cgsheeh (approved):
r+, thanks for fixing this 👍

PR #2067, opened 2023-05-15
Bug 1829061 - Replace access-s whiteboard tags with a drop-down field

dklawren:


scripts/bug1829061.pl
```diff
@@ -0,0 +1,100 @@
...
+    print "  severity $severity does not exist ... skipping\n";
+    next;
+  }
+
+  $whiteboard =~ s/\[access-$severity\]//;
+  $set_all->{status_whiteboard} = $whiteboard;
+  print "  updating whiteboard to '$whiteboard'\n";
+
+  $bug->set_all($set_all);
+  $bug->update();
```
globau, line 97: Shouldn't we be setting `lastdiffed` to `delta_ts` so this change isn't bundled into the next notification for this bug?

scripts/bug1829061.pl
```diff
@@ -0,0 +1,100 @@
...
+Bugzilla->set_user($auto_user);
+
+Bugzilla::Extension::TrackingFlags::Flag->get_all;    # preload
+
+# Load fields information (assumes already created)
+my $flag_field = Bugzilla::Extension::TrackingFlags::Flag->new(
+  {name => 'cf_accessibility_severity'});
+$flag_field || die "Can't find field cf_accessibility_severity\n";
+my $whiteboard_field = Bugzilla::Field->new({name => 'status_whiteboard'});
+$whiteboard_field || usage("Can't find field status_whiteboard\n");
```
globau, line 43: Use `->check` instead of `->new` for both of these

scripts/bug1829061.pl
```diff
@@ -0,0 +1,100 @@
...
+
+BEGIN {
+  Bugzilla->extensions;
+}
+
+Bugzilla->usage_mode(USAGE_MODE_CMDLINE);
+
+# User to make changes as automation@bmo.tld
+my $auto_user = Bugzilla::User->check({name => 'automation@bmo.tld'});
+$auto_user || usage("Can't find user 'automation\@bmo.tld'\n");
```
globau, line 31: `->check` will never return undefined.

globau (changes requested):

scripts/bug1829061.pl
```diff
@@ -0,0 +1,101 @@
...
+
+BEGIN {
+  Bugzilla->extensions;
+}
+
+Bugzilla->usage_mode(USAGE_MODE_CMDLINE);
+
+# User to make changes as automation@bmo.tld
+my $auto_user = Bugzilla::User->new({name => 'automation@bmo.tld'});
+$auto_user || usage("Can't find user 'automation\@bmo.tld'\n");
```
globau, line 31: It would be better to use `->check` and delete the line checking `$auto_user`

scripts/bug1829061.pl
```diff
@@ -0,0 +1,101 @@
...
+
+  $whiteboard =~ s/\[access-$severity\]//;
+  $set_all->{status_whiteboard} = $whiteboard;
+  print "  updating whiteboard to '$whiteboard'\n";
+
+  $bug->set_all($set_all);
+  $bug->update();
+
+  # Updated lastdiffed timestamp so the change is not included in email notifications
+  $dbh->do('UPDATE bugs SET lastdiffed = NOW() WHERE bug_id = ?', undef, $bug->id);
```
globau, line 98: This should use `delta_ts` instead of `NOW()`

`$dbh->do('UPDATE bugs SET lastdiffed=delta_ts WHERE bug_id=?', undef, $bug->id);`

globau (changes requested):

globau (approved):

PR #2071, opened 2023-05-17
Bug 1833442 - When a revision is abandoned for an uplift repo in Phabricator, Phabbugz is trying to set the approval-mozilla-beta flag to '-' and failing

dklawren:
- Added $user->can_change_flag checks before making flag changes and logging when permissions are not enough

### DIFF
--- a/extensions/PhabBugz/lib/Feed.pm
+++ b/extensions/PhabBugz/lib/Feed.pm
@@ -636,26 +636,8 @@ sub process_revision_change {
     }
 
     # Sync the `approval-mozilla-{repo}+` flags.
-    if (
-      $revision->repository &&
-      $revision->repository->is_uplift_repo()
-    ) {
-      my $revision_status_flag_map = {
-        'abandoned'       => '-',
-        'accepted'        => '+',
-        'changes-planned' => 'X',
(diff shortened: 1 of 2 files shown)


cgsheeh (changes requested):
These changes make sense to me.

We call `set_attachment_approval_flags` in `Util.pm::create_revision_attachment `, the call signature will need to be updated there as well. In that section of code we set the flag to `?` unconditionally, we should make sure the new method of setting the flag will still work.

dklawren:
> These changes make sense to me.
> 
> We call `set_attachment_approval_flags` in `Util.pm::create_revision_attachment `, the call signature will need to be updated there as well. In that section of code we set the flag to `?` unconditionally, we should make sure the new method of setting the flag will still work.

Ah good catch. I think we can just remove the call in create_revision_attachment altogether as we call it again later from Feed.pm.

Also since we have a status for draft => '?' in set_attachment_approval_flags so it should just work.

dklawren:
I should probably also move the part that sets the bugs assignee value in create_revision_attachment as it too also should be a function on its own. Another bug.

cgsheeh (approved):

PR #2076, opened 2023-05-28
Bug 1835092 - Create script to mass import a list of comment tags and urls to the new comment_tag_url table

dklawren:
Script useful for mass importing comma separated files of tag and urls to the longdescs_tags_url table.


scripts/bug1835092.pl
```diff
@@ -0,0 +1,66 @@
...
+  if ($url !~ /^https?:/) {
+    print "URL $url invalid...skipping\n";
+    next;
+  }
+
+  # Do not populate if already present
+  my $exists
+    = $dbh->selectrow_array(
+    'SELECT 1 FROM longdescs_tags_url WHERE tag = ? AND url = ?',
+    undef, $tag, $url);
```
suhaibmujahid, line 54: What about when a tag exists but with a different URL?

dklawren: > What about when a tag exists but with a different URL?

Good question. Two options. 1) I skip it, or 2) I update it with the possibly new URL value. I personally think 1) would be best for now as we can always update that one manually in the UI if needed and also it might be someone else's tag with a different meaning and there should be some discussion about who gets to set the URL for it.

suhaibmujahid (commented):
I'm not familiar with Perl, but LGTM. Thank you @dklawren!

scripts/bug1835092.pl
```diff
@@ -0,0 +1,65 @@
...
+foreach my $line (@csv_data) {
+  $csv->parse($line);
+  my @values = $csv->fields();
+  next if !@values;
+  my ($tag, $url) = @values;
+
+  $tag = trim($tag);
+  $url = trim($url);
+
+  # Skip if not an url
```
cgsheeh, line 44: Does the URL have to be HTTP?

cgsheeh (approved):

scripts/bug1835092.pl
dklawren: > Does the URL have to be HTTP/HTTPS?

Technically it should as we create link using the value for URL in the comment tag itself on the show bug page. So if we have something that is not http(s) then it will be invalid. The code that manages entering new values for URLs in the longdescs_tags_url table is a generic table CRUD page so it doesn't do any validation at this time. Not too worried as only admins have access to the generic table UI to enter new values for now.

PR #2077, opened 2023-05-30
Bug 1831632 - Update group membership report to more prominently display users are in groups that not Mozilla email addresses (i.e. personal accounts)

dklawren:
* Some cleanup of the group member report by adding a table for the different query fields
* Extend Bugzilla::Group->members_complete to take an argument that will filter results based on search type
* non-mozilla search type filters the users to only show email addresses that do not match the groups regex if one exists.
* The regex is used by Mozilla to automatically place users in specific confidential groups.
* Figured this would be better than hard-coding '.*@mozilla(foundation)?\.(org|com)' and all of the other possibilities.

### DIFF
--- a/Bugzilla/Group.pm
+++ b/Bugzilla/Group.pm
@@ -135,7 +135,7 @@ sub members_complete {
   }
 
   # If we want a list of users that are members but do not match the regexp
-  # if the current group, then we need to filter the user list.
+  # of the current group, then we need to filter the user list.
   # Regexp are normally used to automatically add users to a group such
   # as mozilla-employee-confidential, etc.
   if ($type eq 'non-mozilla' && $self->user_regexp) {
--- a/Bugzilla/Group.pm
+++ b/Bugzilla/Group.pm
@@ -110,7 +110,7 @@ sub members_non_inherited {
 # returns all possible members of groups, keyed by the group name or _direct
 # a user present in multiple groups will be returned multiple times
 sub members_complete {
-  my ($self) = @_;
+  my ($self, $type) = @_;
   my $dbh = Bugzilla->dbh;
   require Bugzilla::User;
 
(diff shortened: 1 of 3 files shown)


Bugzilla/Group.pm
```diff
@@ -119,13 +119,36 @@ sub members_complete {
...
+  # If we want a list of users that are members but do not match the regexp
+  # if the current group, then we need to filter the user list.
+  # Regexp are normally used to automatically add users to a group such
+  # as mozilla-employee-confidential, etc.
+  if ($type eq 'non-mozilla' && $self->user_regexp) {
+    my $regexp = $self->user_regexp;
+    foreach my $type (keys %{$result}) {
+      my @filtered;
+      foreach my $user (@{$result->{$type}}) {
+        push @filtered, $user if $user->login !~ m/$regexp/i;
```
cgsheeh, line 146: Crafting a regex from user-input is a tad dangerous. In theory someone could create a very special regex that consumes a ton of memory, causing a DoS.

Is this report something that runs only for trusted users? Can we limit them to a selection, or maybe say "search for this string in the email" or something along those lines instead?

extensions/BMO/lib/Reports/Groups.pm
```diff
@@ -185,18 +185,22 @@ sub members_report {
...
   my @group_names
     = sort grep { !/^(?:bz_.+|canconfirm|editbugs|editbugs-team|everyone)$/ }
     @grouplist;
   unshift(@group_names, '');
   $vars->{'groups'} = \@group_names;
 
+  my $include_disabled = $cgi->param('include_disabled') ? 1 : 0;
+  $vars->{'include_disabled'} = $include_disabled;
+
+  my $search_type = $cgi->param('search_type') || 'all';
```
cgsheeh, line 198: Can we limit `search_type` to only a set of allowed values here? Or perhaps in the `members_complete` subroutine.

cgsheeh (changes requested):

Bugzilla/Group.pm
dklawren: The `$self->user_regexp` value is not coming from user input but rather from the Group object and is only edited by a group admin. For example the group `mozilla-corporation` has a $self->user_regexp value of `@mozilla\.com$ ` so the above line would be the same as `dkl@mozilla !~ m/@mozilla\.com$/`

The user_regexp value requires a user with 'editgroups` permissions to update the value.

extensions/BMO/lib/Reports/Groups.pm
dklawren: Sure. Will do that.

Bugzilla/Group.pm
```diff
@@ -119,13 +119,36 @@ sub members_complete {
...
+
   foreach my $group_id (@{$self->flatten_group_membership($self->id)}) {
     next if $group_id == $self->id;
     my $group_name = Bugzilla::Group->new({id => $group_id, cache => 1})->name;
     my $user_ids = $dbh->selectcol_arrayref($sth, undef, $group_id);
     $result->{$group_name} = Bugzilla::User->new_from_list($user_ids);
   }
+
+  # If we want a list of users that are members but do not match the regexp
+  # if the current group, then we need to filter the user list.
```
cgsheeh, line 138: I think this should say "of the current group"?

cgsheeh (approved):

PR #2079, opened 2023-06-01
Bug 1836202 - Add the ability to list bugs in the order provided by the `bug_id` parameter

globau:
Adds support for `order=bug_list`


dklawren (approved):
Looks good to me. I wonder if we would hit any sort of SQL length limit with the bug ids being duplicated in the ORDER BY section. Also FIELD() Is not support by PostgreSQL which I am not worried about but something to think about another day. r=dkl

PR #2080, opened 2023-06-01
Make a data review form when requesting a data review instead of requiring copy-paste

padenot:


extensions/BMO/template/en/default/hook/flag/type_comment-form.html.tmpl
```diff
@@ -225,3 +225,97 @@
...
+      </tr>
+      <tr>
+        <th id="_ar_sec_i5_label">List all proposed measurements and indicate the category of data collection for each measurement, using the
+          <a href="https://wiki.mozilla.org/Data_Collection">Firefox data collection categories</a> found on the Mozilla wiki:</th>
+        <td>
+          <textarea aria-labelledby="_ar_sec_i5_label">
+For each measurements, please provide:
+* Measurement Description
+* Data Collection Category
+* Tracking Bug #
```
dklawren, line 264: `* Tracking [% terms.Bug %]`

Any mention of bug, Bug, Bugzilla, etc. has a matching [% terms.Bug %] variable that must be used. This is the reason for the sanity test failure.

dklawren (changes requested):

chutten (commented):
I had to play with `approval-mozilla-beta` to figure out how this would intersect with the [existing data review process](https://wiki.mozilla.org/Data_Collection#Step_1:_Submit_Request). It appears as though the existing path of

1. Add attachment that is a plaintext data collection review request
2. Flag `data-review?<a steward>`

Will continue to work, though the plaintext version of this template will autofill the comment (it'd be nice if it didn't, but we can live with it).

The augmented flow would be...
1. Mark an existing attachment `data-review?` (likely the patch adding the data collection)
2. Fill in the form as presented by BMO (or dismiss the form with the `x` and supply the full text yourself)

Is that correct? 

(( we'll have to update the wiki to make this clear. ))

--

Overall this appears to be a lovely augmentation for anyone who is filling out these forms piecemeal today on BMO. The additional maintenance burden of updating the data review request in one more place is very small.

dklawren:
Sorry that this slipped off my radar. Is there anything else I can do to help resolve this change?

padenot:
No it slipped of mine as well, will fix.

PR #2082, opened 2023-06-07
Bug 1797620 - Create modular framework for adding use case specific search endpoints to the BMO API

dklawren:
* Add a new extension where new search modules can live that allow for customized SQL and results depending in the needs of the customer.
* Added an simple example endpoint to make sure it is working that just returns possible bug states.
* Added a LastSeen module that provides two endpoints, one for bugs that have a needinfo set with a requestee that has not been seen in X days, and another similar for the assignee of the bug, This is for bug 1527841.
* The list of bugs is filtered based on permissions to view the bugs. 
* A Util module provides utility functions for returning hashes for bugs, comments, users, and flags.
* Util also provides a function that implements named placeholders in SQL. Not strictly needed for the SQL that is current being used for LastSeen but will allow for more complicated queries in the future.


extensions/SearchAPI/lib/Util.pm
```diff
@@ -0,0 +1,320 @@
...
+        |
+        \n                      # newline
+        |
+        [\t\ ]+                 # any kind of white spaces
+    )
+}smx;
+
+# Some workflow borrowed from
+# https://metacpan.org/pod/DBIx::Placeholder::Named
+sub named_params {
```
cgsheeh, line 67: Can you import this from the place it was borrowed from instead of maintaining it ourselves? Same with `$tokenizer_re`. I think we should try to avoid maintaining this code if we can.

cgsheeh (changes requested):

extensions/SearchAPI/lib/Util.pm
```diff
@@ -0,0 +1,320 @@
...
+    }
+  }
+
+  # Create new SQL stratement from tokens
+  my $new_query = join '', @query_tokens;
+
+  return ($new_query, \@values);
+}
+
+sub bug_to_hash {
```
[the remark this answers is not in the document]
dklawren: It is a good point. The OOP setup used by Bugzilla is more or less home grown and does not have some of the cool built in functionality of Python for displaying of an object in different formats. I could had new methods to Bugzilla::Bug, Bugzilla::Flag, and Bugzilla::User such as $bug->to_hash or similar that would give something like this. We also have functions in Bugzilla::WebService::Bug that do the same thing but the code there is heavily integrated with that old style of API (XMLRPC, JSONRPC) and it would take some kludgery to make it work in the newer API framework. Let me try some things.

extensions/SearchAPI/lib/Util.pm
dklawren: I was trying to avoid bring in any more dependencies that needed and this was something that I only needed small bits of the larger library. Therefore I try to make sure I give credit the original author. We do this in several other places in the code base already. Notably Bugzilla::S3 which is a copied version of an abandoned module.

dklawren: Actually for DBIx::Placeholder::Named it would have been a lot more work that just updating a query string as the module is a wrapper around the larger DBI library methods which I wanted to avoid. So I just pulled out the small bit that I needed and then altered it some more to add ability to pass in a list of values instead of just single values.

cgsheeh (approved):
LGTM. With the caveat that the review of `named_params` is just a rubber stamp, heh.

PR #2084, opened 2023-06-16
Bug 1802188 - Sitemap extension is not working due to XML content be disallowed by Mojoicious

dklawren:
This patch updates the Mojolicous framework specific parts of the BMO code to allow test/xml documents (needed for the SiteMap index file with links to the individual xml.gz files) and to allow static file serving from the data/SiteMapIndex directory. The index file is autogenerated from a link in the robots.txt file. All of which were not allowed before when we moved from Apache to Mojolicious so this feature has been broken now for almost 5 years :(


cgsheeh (approved):

zzzeid:
Note typo in commit message.

dklawren:
Closing as this does not fully solve the current problem and I am working on a different solution.

PR #2086, opened 2023-06-25
Bug 1835941 - Create a new report for "What should I work on next?" to help engineers see bugs matching a specific search criteria

dklawren:
Follow goals for this PR:
* New report that shows several tables listing what bugs a developer should focus on next according to requirements laid out in https://docs.google.com/document/d/1X3VU0ggVKAFdnUSls3pcwpnrZa_Eb_-5tVj1rRvN1p8/edit#heading=h.8z0sho6pod67
* Report is accessible to users with editbugs or higher from the standard Reports page
* Each table is generated with a custom SQL instead of using Search.pm in order to match specific fields not possible with Search.pm
* The bugs are filtered based on permissions so nothing sensitive can be leaked.
* The report defaults to the current user, but another user can be entered and show their results instead.
* The SQL can be tweaked if we need finer detail or the results are not what is expected.
* There is a new qa/t/1_test_whats_next.t test script that makes sure that the report is working and that unprivileged users cannot see private bugs.

Let me know if you have any more questions. I will also upload this to bugzilla-dev.allizom.org for testing as well.

 Thanks!

### DIFF
--- a/extensions/BMO/lib/Reports/WhatsNext.pm
+++ b/extensions/BMO/lib/Reports/WhatsNext.pm
@@ -32,7 +32,7 @@ use constant CLASSIFICATIONS => (
 );
 
 use constant SELECT =>
-  'SELECT bugs.bug_id, bugs.bug_status, bugs.priority, bugs.short_desc, bugs.delta_ts';
+  'SELECT bugs.bug_id, bugs.bug_status, bugs.priority, bugs.bug_severity, bugs.short_desc, bugs.delta_ts';
 
 # Wrap the sql execution in a try block so we can see any SQL errors in debug output
 sub get_bug_list {
@@ -57,6 +57,7 @@ sub format_bug_list {
       id          => $row->{bug_id},
       status      => $row->{bug_status},
       priority    => $row->{priority},
+      severity    => $row->{bug_severity},
       summary     => $row->{short_desc},
       changeddate => $row->{delta_ts},
     };
--- a/extensions/BMO/lib/Reports/WhatsNext.pm
+++ b/extensions/BMO/lib/Reports/WhatsNext.pm
@@ -1,0 +1,377 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+#
+# This Source Code Form is "Incompatible With Secondary Licenses", as
+# defined by the Mozilla Public License, v. 2.0.
+
+package Bugzilla::Extension::BMO::Reports::WhatsNext;
+
+use 5.10.1;
+use strict;
+use warnings;
+
+use Bugzilla;
+use Bugzilla::Error;
+use Bugzilla::Logging;
+use Bugzilla::Search;
+use Bugzilla::Status qw(BUG_STATE_OPEN);
+use Bugzilla::User;
+use Bugzilla::Util qw(datetime_from fetch_product_versions time_ago);
+
+use DateTime;
+use Mojo::Util qw(dumper);
+use Try::Tiny;
+
+use constant CLASSIFICATIONS => (
+  'Client Software',
+  'Developer Infrastructure',
+  'Components',
+  'Server Software',
+  'Other',
+);
+
+use constant SELECT =>
+  "SELECT bugs.bug_id, bugs.bug_status, bugs.priority, bugs.short_desc, bugs.delta_ts";
+
+# Wrap the sql execution in a try block so we can see any SQL errors in debug output
+sub get_bug_list {
+  my ($query, @values) = @_;
+  try {
+    return Bugzilla->dbh->selectall_arrayref($query, {Slice => {}}, @values);
+  }
+  catch {
+    DEBUG($_);
+    return [];
+  };
+}
+
+sub format_bug_list {
+  my ($bugs, $user) = @_;
+
+  my $datetime_now = DateTime->now(time_zone => $user->timezone);
+
+  my @formatted_bugs;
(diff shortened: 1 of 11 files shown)


dklawren:
Removing review as a I need to complete a few more things based on initial feedback from glob:

* Only show report for mozilla employees instead of editbugs
* New icon and link in the Bugzilla header of all pages to the report
* Auto refresh the page after 10 mins like we do with mydashboard
* Get rid of the extra text on the report since it will live elsewhere eventually and we can link to it if needed.

dklawren:
Here are the new changes with this revision
* Only show report for mozilla employees instead of editbugs
* New icon and link in the Bugzilla header of all pages to the report
* Auto refresh the page after 10 mins like we do with mydashboard
* Get rid of the extra text on the report since it will live elsewhere eventually and we can link to it if needed.

extensions/BMO/lib/Reports/WhatsNext.pm
```diff
@@ -0,0 +1,381 @@
...
+sub sec_crit_bugs {
+  my $user = shift;
+  my $dbh  = Bugzilla->dbh;
+
+  my $query = SELECT . '
+         FROM bugs JOIN products ON bugs.product_id = products.id
+              JOIN keywords ON bugs.bug_id = keywords.bug_id
+        WHERE products.classification_id IN (SELECT id FROM classifications WHERE name IN ('
+    . join(', ', map { $dbh->quote($_) } CLASSIFICATIONS) . "))
+              AND keywords.keywordid = (SELECT id FROM keyworddefs WHERE name = 'sec-critical')
```
globau, line 177: same caching for all keywords IDs

extensions/BMO/lib/Reports/WhatsNext.pm
```diff
@@ -0,0 +1,381 @@
...
+         FROM bugs JOIN products ON bugs.product_id = products.id
+              LEFT JOIN flags AS requestees_login_name ON bugs.bug_id = requestees_login_name.bug_id
+                AND COALESCE(requestees_login_name.requestee_id, 0) = ?
+                AND COALESCE(requestees_login_name.setter_id, 0) != ?
+        WHERE products.classification_id IN (SELECT id FROM classifications WHERE name IN ('
+    . join(', ', map { $dbh->quote($_) } CLASSIFICATIONS) . '))
+              AND bugs.bug_status IN ('
+    . join(', ', map { $dbh->quote($_) } BUG_STATE_OPEN) . ")
+              AND (requestees_login_name.bug_id IS NOT NULL
+                    AND requestees_login_name.type_id = (SELECT id FROM flagtypes WHERE name = 'needinfo'))
```
globau, line 342: cache flagtypes too

globau (changes requested):

extensions/BMO/lib/Reports/WhatsNext.pm
```diff
@@ -111,32 +111,34 @@ sub get_tracking_status_flags {
...
   return {} if !$nightly_field && !$beta_field && !$release_field;
 
-  $flag_data->{tracking} = {
-    nightly => $nightly_field,
-    beta    => $beta_field,
-    release => $release_field,
-  };
+  $flag_data->{tracking}
+    = {nightly => $nightly_field, beta => $beta_field, release => $release_field,
+    };
```
globau, line 124: Please revert these formatting changes that drop the `=` onto a new line; that isn't in alignment with our perlstyle config, nor is it used anywhere else.

extensions/BMO/lib/Reports/WhatsNext.pm
```diff
@@ -146,16 +148,19 @@ sub s1_bugs {
   my $user = shift;
   my $dbh  = Bugzilla->dbh;
 
-  my $query = SELECT . '
+  # Preselected values for inserting into SQL
+  my $cache      = Bugzilla->request_cache->{whats_next};
```
globau, line 152: A request-cache isn't sufficient; as per my previous review it needs to be **per-process or similar**.

globau (changes requested):

extensions/BMO/lib/Reports/WhatsNext.pm
```diff
@@ -0,0 +1,443 @@
...
+  my $input = Bugzilla->input_params;
+
+  $user->in_group('mozilla-employee-confidential') || ThrowUserError(
+    'auth_failure',
+    {
+      group  => 'mozilla-employee-confidential',
+      action => 'run',
+      object => 'whats_next'
+    }
+  );
```
globau, line 403: Slight miscommunication here, when I said the report should only be visible to employees, I meant just the icon in the header.  There's no need for the dashboard itself to be restricted.

extensions/BMO/lib/Reports/WhatsNext.pm
```diff
@@ -0,0 +1,443 @@
...
+  my $who
+    = $input->{who} ? Bugzilla::User->check({name => $input->{who}}) : $user;
+
+  # Here we load some values into cache that will be used later
+  # by the various queries.
+  my $cache = Bugzilla->process_cache->{whats_next} = {};
+  my $dbh   = Bugzilla->dbh;
+
+  # classifications
+  $cache->{classification_ids} = $dbh->selectcol_arrayref('
```
globau, line 416: Values in the cache should only be set if they aren't already present.

globau (changes requested):

extensions/BMO/lib/Reports/WhatsNext.pm
dklawren: Will fix.

dklawren:
Severity column added and changes pushed to bugzilla-dev.

globau (approved):
ship it!

PR #2087, opened 2023-06-30
Bug 1839929 - Need to update SiteMap extension to upload xml files to S3 and to do nightly as a cron task

dklawren:
This PR updates the SiteMapIndex extension to no longer dynamically generate the needed sitemap xml files needed for search engine indexing. Instead the robots.txt file will not redirect the indexer to the xml files located on an AWS S3 bucket. The sitemap xml files will be generated by a cron container that will run once per day and be set up by Cloud Ops.

* Added a SiteMapIndex configuration panel for setting the aws credentials and bucket name.
* robots.txt has been updated to display the `Sitemap: url` if enabled and the bucket name exists. instead of checking for hostname.
* A new script `rebuild_index.pl` has been added that generates the xml files and uploads to S3.

### DIFF
--- a/extensions/SiteMapIndex/lib/Util.pm
+++ b/extensions/SiteMapIndex/lib/Util.pm
@@ -44,7 +44,7 @@ use IO::Compress::Gzip qw(gzip $GzipError);
 
 sub too_young_date {
   my $hours_ago = DateTime->now(time_zone => Bugzilla->local_timezone);
-  $hours_ago->subtract(hours => 0); #SITEMAP_DELAY);
+  $hours_ago->subtract(hours => SITEMAP_DELAY);
   return $hours_ago;
 }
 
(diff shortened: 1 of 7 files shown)


extensions/SiteMapIndex/lib/Util.pm
```diff
@@ -145,11 +123,15 @@ sub _generate_sitemap_index {
 <sitemapindex xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">
 END
 
+  my $sitemap_url
+    = 'https://'
+    . Bugzilla->params->{sitemapindex_s3_bucket}
+    . '.s3-us-west-2.amazonaws.com';
```
cgsheeh, line 129: Is `us-west-2` always guaranteed? Maybe that should be configurable in the same way as the bucket name.

cgsheeh (approved):

extensions/SiteMapIndex/lib/Util.pm
dklawren: dlactin states that it will unlikely change unless we need to provision resources in another region due to a major outage. I will merge this and open a new bug for adding a config option for region.

PR #2089, opened 2023-07-07
Bug 1842217 - Add configuration parameter to SiteMapIndex extension for configuring AWS S3 region

dklawren:
Add new configuration parameter for the SiteMapIndex extension that allows changing the aws region in the future.

### DIFF
--- a/extensions/SiteMapIndex/lib/Util.pm
+++ b/extensions/SiteMapIndex/lib/Util.pm
@@ -125,8 +125,12 @@ END
 
   my $sitemap_url
     = 'https://'
-    . Bugzilla->params->{sitemapindex_s3_bucket}
-    . '.s3-us-west-2.amazonaws.com';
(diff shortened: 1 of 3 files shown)


extensions/SiteMapIndex/lib/Util.pm
```diff
@@ -194,6 +198,7 @@ END
 
 sub _upload_s3 {
   my ($filename, $data) = @_;
+  return;
```
cgsheeh, line 201: Testing leftover?

cgsheeh (changes requested):

extensions/SiteMapIndex/lib/Util.pm
dklawren: Yeah :(

cgsheeh (approved):

PR #2092, opened 2023-07-20
Bug 1841823 - entering correct second factor number (2FA pin) after first one was incorrect fails: "User Not Specified: You must select/enter a user."

dklawren:
* Now instead of going to a new page to show that the user entered an incorrect TOTP code, it will show an error in page.
* This allows the tokne cookie to still be used that is set initially which was not being set when the user clicked back to enter another TOTP code.
* A REST API endpoint was added to verify the code initially. If the endpoint successfully verifies the code then the form is submitted as usual and the user is logged in. If not, the form is not submitted and an error is displayed.
* I also created a MFA test case which was not there before that tests enrolling in TOTP, entering an incorrect code, entering a correct code, and then un-enrolling in TOTP.


template/en/default/mfa/totp/verify.html.tmpl
```diff
@@ -18,15 +19,19 @@
...
-<div class="verify-totp">
-  <form method="POST" action="[% postback.action FILTER none %]">
-    [% FOREACH field IN postback.fields.keys %]
-      <input type="hidden" name="[% field FILTER html %]" value="[% postback.fields.item(field) FILTER html %]">
-    [% END %]
-    <input type="text" name="code" id="code"
-           placeholder="123456" maxlength="9" pattern="\d{6,9}" size="10"
-           autocomplete="off" inputmode="numeric" required autofocus><br>
-    <br>
+<div class="verify_totp">
```
cgsheeh, line 22: The class name changed here, is that expected?

cgsheeh (changes requested):

cgsheeh (approved):

PR #2093, opened 2023-07-21
Bug 1844801 - Updates charting UI to allow users in special chartpublicgroup to allow setting charts as public

dklawren:
Instead of only admin users being able to make a query public, we added a config parameter so any arbitrary group can be used. Now we can add additional users or groups to be able to make queries public.

### DIFF
--- a/chart.cgi
+++ b/chart.cgi
@@ -99,7 +99,7 @@ $user->in_group(Bugzilla->params->{"chartgroup"}) || ThrowUserError(
 );
 
 # Only admins may create public queries
-$user->in_group('admin') || $cgi->delete('public');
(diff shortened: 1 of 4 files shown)


chart.cgi
```diff
@@ -99,7 +99,7 @@ $user->in_group(Bugzilla->params->{"chartgroup"}) || ThrowUserError(
 );
 
 # Only admins may create public queries
```
cgsheeh, line 101: This comment needs to be updated too.

cgsheeh (changes requested):

cgsheeh (approved):

PR #2097, opened 2023-08-01
Bug 1846398 - UI tweaks to the What Should I Work On Next page

globau:
- tables should be full page width
- bug summary should wrap
- tables should be sortable by clicking on column headers
- show a count of the number of bugs
- add a link to buglist.cgi
- add total bug counter to favicon
- use css over table attributes (eg. nowrap -> white-space:nowrap)
- fix broken html (eg. duplicate ids)


dklawren (approved):
All changes make sense and work as expected.

PR #2105, opened 2023-08-14
Bug 1847640 - Users cannot clear their needinfo in a bug that's closed, if they lack editbugs

dklawren:
* Updates the code that blocks commenting on closed bugs for those without editbugs
* Allow someone who is the needinfo on the bug to comment. 
* Once the comment is made that clears the needinfo, the same user cannot then comment again.


cgsheeh (approved):
Makes sense to me. Can we test this somehow?

dklawren:
> Makes sense to me. Can we test this somehow?

I would like to eventually add testing for all of these *rules* but since it is an extension, testing is not critical as core code would need. But I do want to get some wider testing in place.

PR #2106, opened 2023-08-17
Bug 1849069 - Real name change doesn't work if "New email address" is pre-filled with different email by password manager

LLyaudet:
To make it case insensitive like it is the case for login. See https://bugzilla.mozilla.org/show_bug.cgi?id=1849069

### DIFF
--- a/template/en/default/account/prefs/account.html.tmpl
+++ b/template/en/default/account/prefs/account.html.tmpl
@@ -107,7 +107,7 @@
           <th align="right">Current password:</th>
           <td>
             <input type="hidden" name="old_login" value="[% user.login FILTER html %]">
-            <input autocomplete="off" type="password" name="old_password" id="old_password">
+            <input autocomplete="current-password" type="password" name="old_password" id="old_password">
             <a href="#" id="forgot-password">I forgot my password</a>
           </td>
         </tr>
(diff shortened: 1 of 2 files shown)


dklawren (changes requested):
This will work and thanks for the pull request. We also want to add the autocomplete="off" to the new email form field as mentioned in the bug. 

https://github.com/mozilla-bteam/bmo/blob/master/template/en/default/account/prefs/account.html.tmpl#L99

LLyaudet:
I just committed another commit for the autocomplete="off". Tell me if there is something else to modify :)

template/en/default/account/prefs/account.html.tmpl
```diff
@@ -107,7 +107,7 @@
           <th align="right">Current password:</th>
           <td>
             <input type="hidden" name="old_login" value="[% user.login FILTER html %]">
-            <input autocomplete="current-password" type="password" name="old_password" id="old_password">
+            <input autocomplete="off" type="password" name="old_password" id="old_password">
```
dklawren, line 110: Actually after seeing that we already do autocomplete for current password, let's change this back to `current-password` as looking this up that is correct already.

https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete

dklawren (changes requested):
After that this should be all good.

template/en/default/account/prefs/account.html.tmpl
LLyaudet: You are right. Both were syntactically correct but keeping "current-password" value here is a feature and yields no bug.

LLyaudet:
You're welcome :)

PR #2114, opened 2023-09-19
Bug 1853462 - [EXT_CLEANUP] Remove any extension code from ./extensions that are currently disabled

dklawren:
This patch removes several extensions from the BMO code base that have not been (or never were) used in a long time and were already marked as disabled. They are not depended on by other extensions, etc.


cgsheeh (approved):
LGTM.

PR #2117, opened 2023-09-21
Bug 1853466 - [EXT_CLEANUP] Remove ZPushNotify extension as it is no longer being used (part 2)

dklawren:
Not that the table has been dropped, this removes the rest of the extension code.


cgsheeh (approved):
Ship it!

PR #2121, opened 2023-09-26
no bug - Update to Perl 5.38.0 and other dependencies

dklawren:
* Upgrade to latest Perl 5.38.0 
* Updated several other dependencies as part of the image build process
* Updated Dockerfile to pull bmo-perl-slim from ghcr.io instead of Docker Hub


zzzeid (approved):
lgtm (apart from failing tests which I assume will be fixed before merge!)

PR #2127, opened 2023-10-05
Bug 1857138 - update /favicon.ico

globau:
Update the icon served for non-html pages.


globau:
This file isn't referenced by an in-tree code, so it was missed by the initial landing of the new icon.

dklawren (approved):

PR #2129, opened 2023-10-10
Bug 1848412 - [GCS] Refactor S3 Bugzilla code to allow ability to support both S3 and GCP and share common code

dklawren:
This PR takes the old S3 code and refactors it to work with either S3 or Google Cloud Storage (gcs). I remove a lot of the old unused S3 that was not needed which made things smaller. There was some functions to manage buckets, ACLs, etc. All of that is managed by CloudOps outside of BMO. 

So there is a base Storage.pm class and then S3.pm and Google.pm inherit from that and add their specific API calls for get/set/delete. S3 uses standard client_id/secret_key added to a header for each API call. For GCP, we are using Workload Identity which ties a k8s service account to a standard GCP service account. This allows the BMO container to query an API to obtain an access token and then we just include that in an Authorization header. 

I have also added two test cases, one which uses the UI for adding attachments for both S3 and gcs. Another one is functional and just loads the libraries and access s3 and gcs directly.  To accomplish this, I have added mock S3 and GCS containers to the docker-compose.test.yml file that runs with the test suite. They do not test authentication but do mock the standard API calls for get/set/delete actions.

Let me know if you have questions


Bugzilla/Net/Google.pm
```diff
@@ -0,0 +1,169 @@
...
+    ThrowCodeError('net_storage_invalid_action', {action => $action});
+  }
+
+  if ($action eq 'add') {
+    $format
+      = 'upload/'
+      . $PATH_BASE
+      . '/%s/o?uploadType=media&name='
+      . uri_escape_utf8($key);
+    $method = 'POST';
```
cgsheeh, line 62: Can we build the format string here with `@args` in the same way as the other branches of this `if` statement? It's consistent that way, and we can also get rid of all the `push @args, $key;` lines and just have one after the `if` block.

Bugzilla/Net/Google.pm
```diff
@@ -0,0 +1,169 @@
...
+  return undef if !$self->service_account;
+
+  if (!$self->access_token || time > $self->access_token_expiry) {
+    $self->_get_access_token;
+  }
+
+  $headers->header(Authorization => 'Bearer ' . $self->access_token);
+}
+
+# Google Kubernetes allows for the user of Workload Identity. This allows
```
cgsheeh, line 102:
```suggestion
# Google Kubernetes allows for the use of Workload Identity. This allows
```

Bugzilla/Net/Google.pm
```diff
@@ -0,0 +1,169 @@
...
+# us to link two serice accounts together and give special access for applications
+# running under Kubernetes. We use the special access to get an OAuth2 access_token
+# that can then be used for accessing the the Google API such as Cloud Storage.
+sub _get_access_token {
+  my ($self) = @_;
+  my $url
+    = sprintf
```
cgsheeh, line 109: Does this URL change depending on the value of `google_storage_host`?

Is this supposed to be `http://` or should it be `https://`?

Bugzilla/Net/Google.pm
```diff
@@ -0,0 +1,169 @@
...
+# }
+
+# When an error occurs, Google API adds additional information
+# in the response body in the form of JSON that we can use.
+sub _remember_errors {
+  my ($self, $response) = @_;
+
+  try {
+    my $content = $response->decoded_content;
+    my $r       = ref $content ? $content : decode_json($content);
```
cgsheeh, line 156: Change `$r` to something more useful, maybe `$decoded_response` etc?

Bugzilla/Net/S3.pm
```diff
@@ -0,0 +1,192 @@
...
+# a given bucket name can be safely used as a DNS name.
+sub _is_dns_bucket {
+  my ($self, $bucketname) = @_;
+
+  if (length $bucketname > 63) {
+    return 0;
+  }
+
+  if (length $bucketname < 3) {
+    return undef;
```
cgsheeh, line 76: Assuming this is supposed to be a truthy value, can we return `0` for consistency? If we are returning `undef` here for some other reason, let's leave a comment explaining why.

Bugzilla/Net/S3.pm
```diff
@@ -0,0 +1,192 @@
...
+  my $content = $response->content;
+
+  # If not XML then just return status and message
+  unless (ref $content || $content =~ m/^[[:space:]]*</) {    # if not xml
+    $self->error_code($response->code);
+    $self->error_string($response->message);
+    return 1;
+  }
+
+  my $r = ref $content ? $content : $self->_xpc_of_content($content);
```
cgsheeh, line 181: Change `$r` to something else.

Bugzilla/Report/Net.pm
```diff
@@ -0,0 +1,76 @@
...
+  my ($self, $product) = @_;
+  $self->driver->delete_key($product)
+    or warn "Failed to remove data for product $product from S3: "
+    . $self->driver->error_string . "\n";
+  return $self;
+}
+
+sub data_exists {
+  my ($self, $product) = @_;
+  return !!$self->driver->head_key($product);
```
cgsheeh, line 73: Double negative here?

cgsheeh (changes requested):

Bugzilla/Report/Net.pm
dklawren: Converts to a real true/false boolean. just in case it returns a falsey value. So basically a double not or another way to remember it is "bang, bang, boolean". Works the same in Javascript. Not sure if Python has something similar.

Bugzilla/Net/Google.pm
dklawren: No that URL will stay the same, but yes it should have been https which I have fixed.

_get_access_token will not be called when running under CI as I am not mocking the API for getting access tokens. Also the fake gcs server used in docker-compose.test.yml will just ignore Authorization header anyway. So no need to have a param for an alternate access token endpoint.

Bugzilla/Net/Google.pm
```diff
@@ -0,0 +1,169 @@
...
+    $method = 'GET';
+  }
+  elsif ($action eq 'get') {
+    $format = $PATH_BASE . '/%s/o/%s?alt=media';
+    $method = 'GET';
+  }
+  elsif ($action eq 'delete') {
+    $format = $PATH_BASE . '/%s/o/%s';
+    $method = 'DELETE';
+  }
```
cgsheeh, line 78: Nit: if we can make these format strings non-positional that would be better.

Bugzilla/Net/Storage.pm
```diff
@@ -0,0 +1,200 @@
...
+  my ($self, $key) = @_;
+
+  ThrowCodeError('net_storage_invalid_key') unless $key && length $key;
+
+  my ($method, $path) = $self->_get_method_path('get', $key);
+
+  return $self->_send_request($method, $path);
+}
+
+sub delete_key {
```
cgsheeh, line 68: Add a comment for this method.

Bugzilla/Net/Storage.pm
```diff
@@ -0,0 +1,200 @@
...
+
+  # Update some settings if we are running in a test environment
+  if ($self->host eq 's3') {
+    $protocol = 'http';
+    $host     = 's3:9000';
+  }
+  elsif ($self->host eq 'gcs') {
+    $protocol = 'http';
+    $host     = 'gcs:4443';
+  }
```
cgsheeh, line 140: These pieces should be in the subclasses with a common method that is overridden.

cgsheeh (changes requested):
LGTM, just a few nits as discussed in our video review.

Also as we mentioned at the start of our call, we may be able to avoid a closure window if we started uploading attachments to GCP and S3 on attachment creation, defaulting to S3 on retrieval. Then Ops could do the S3->GCP migration in the background, and we could fully swap to GCP once the data is synced over with little or no downtime. Just something to consider before finalizing the PR, not a blocker. :)

Have a great holiday weekend, Dave!

cgsheeh (approved):

PR #2130, opened 2023-10-14
Bug 1853465 - [EXT_CLEANUP] Remove REMO extension as it is no longer being used

MichaelKohler:
This removes the `REMO` extensions, as the Reps program was shut down at the beginning of September: https://discourse.mozilla.org/t/important-announcement-decommissioning-of-the-mozilla-reps-program/122291. We do not need these forms anymore, however please note that we want to keep the access group and the actual bugs around until around the end of the year. At that point we will create a separate ticket to do these tasks.

I didn't find any registration of the extension, so I'm somewhat assuming that this is done automatically. If not, please shout and I can remove that part of the code as well.

Reps GitHub issue: https://github.com/mozilla/Reps/issues/443


dklawren (approved):
Looks good. Thanks for the help with this. r=dkl

PR #2131, opened 2023-10-17
Bug 913911 - add a webservice method to expose a bug graphs (including dependency and duplicates)

dklawren:
Ok. I feel like this is finally ready for review. Basically it adds a new API endpoint for the sole purpose of showing a tree of bug relationships. We have dependencies, regressions, and duplicates right now. It creates a directed graph from the DB tables that map the relationships of bug IDs. It takes the table data and creates the different connections with the main bug id as the root node. The code will filter out the privates bugs that the user cannot see and also hide resolved bugs unless they are requested. I have created a new test script to test the REST API for the three different relationships and there is also a small functional test that just tests the graphing part that was included in the original work done by Dylan long ago. Let me know if you have any further questions and I will be happy to explain. Thanks!


Makefile.PL
```diff
@@ -64,6 +64,7 @@ my %requires = (
   'Email::Sender'                       => 0,
   'FFI::Platypus'                       => 0,
   'Future'                              => '0.34',
+  'Graph'                               => 0,
```
cgsheeh, line 67: What do the number values represent here?

docs/en/rst/api/core/v1/bug.rst
```diff
@@ -1122,6 +1122,77 @@ This method can throw all the same errors as :ref:`rest_single_bug`, plus:
...
+         "alias" : null,
+         "changes" : {
+           "keywords" : {
+             "added" : "funny, stupid",
+             "removed" : ""
+           },
+             "status" : {
+               "added" : "IN_PROGRESS",
+               "removed" : "CONFIRMED"
+           }
```
cgsheeh, line 1170: Nit: spacing is a little off here.

docs/en/rst/api/core/v1/bug.rst
```diff
@@ -1122,6 +1122,77 @@ This method can throw all the same errors as :ref:`rest_single_bug`, plus:
...
+
+  GET /rest/bug/1156/graph?type=bug_tree
+
+The default relationship is the dependson to blocked relationship. To build the graph in the opposite direction:
+
+.. code-block:: text
+
+  GET /rest/bug/1156/graph?type=json_tree&relationship=dependencies:blocked,dependson
+
+**Response**
```
cgsheeh, line 1154: Can you add a more detailed example of how this works? ie something like "if you have a bug 123 and you want to see the tree of blockers for the bug, and bug 123 is blocked by bug 234 and bug 345, this is how you make the API call and what it will look like". The JSON example only contains a tree with a single element so it's hard to visualize what this will look like in more complicated examples.

cgsheeh (changes requested):

Bugzilla/API/V1/BugGraph.pm
```diff
@@ -0,0 +1,133 @@
...
+sub _bug_to_hash {
+  my ($self, $bug) = @_;
+  return {
+    summary    => $bug->short_desc,
+    status     => $bug->bug_status,
+    resolution => $bug->resolution,
+    milestone  => $bug->target_milestone,
+    assignee   => $bug->assigned_to->name,
+  };
+}
```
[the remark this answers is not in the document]
dklawren: Actually you caused me to remember that we implemented a to_hash method for Bug.pm as part of a recent change to add new Search API functionality. I will switch to use that instead.

Bugzilla/API/V1/BugGraph.pm
```diff
@@ -0,0 +1,133 @@
...
+
+  if (none { $relationship eq $_ } keys %relationships) {
+    ThrowCodeError('param_invalid',
+      {function => 'bug/<id>/graph ', param => 'relationship'});
+  }
+
+  my $result = {};
+  try {
+    foreach my $fields (@{$relationships{$relationship}}) {
+      Bugzilla->switch_to_shadow_db();
```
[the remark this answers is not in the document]
dklawren: switch_to_shadow_db() causes any SQL to follow to hit the readnly-replica if one is available. If one is not available then it just uses the main DB.  In production we have a primary and secondary DB and heavy read SQL goes to the secondary. Such as bug queries, etc. When loading a bug and making edits, we always hit the primary DB.

Makefile.PL
dklawren: in Makefile.pl 0, means any version will do meaning the latest (if installing from scratch) or whatever is currently installed if not.  If an actual number is defined, it must be that version or higher. So if you run it on a system with a version less than, it will generate an error and stop.

docs/en/rst/api/core/v1/bug.rst
dklawren: I need to fix the docs anyway as the last refactor I did made the docs no longer match what is actually being returned. I will expand on the docs when I edit them as well. Thanks.

Bugzilla/Report/Graph.pm
```diff
@@ -0,0 +1,118 @@
...
+  my $limit  = defined $self->limit ? 'LIMIT ' . $self->limit : '';
+
+  # WITH RECURSIVE is available in MySQL 8.x and newer as
+  # well as recent versions of PostgreSQL and SQLite.
+  my $query = "WITH RECURSIVE RelationshipTree AS (
+    SELECT t.$source, t.$sink, 1 AS depth FROM $table t WHERE t.$source = ?
+    UNION ALL
+    SELECT t.$source, t.$sink, rt.depth + 1 AS depth FROM $table t
+      JOIN RelationshipTree rt ON t.$source = rt.$sink WHERE rt.depth <= $depth $limit)
+    SELECT rt.$source, rt.$sink FROM RelationshipTree rt";
```
cgsheeh, line 71: Can we either build this query with a pre-existing query builder library? This seems like a target for a SQL injection given we are splatting strings directly into the query. If source/sink etc are all pre-defined values perhaps we could at least validate them here.

cgsheeh (changes requested):

cgsheeh (approved):
Nice, limiting the inputs to a select few choices is a good middle ground. :+1:

PR #2132, opened 2023-10-23
Bug 1549621 - Redevelop Enter Bug page

kyoshino:
[Bug 1549621 - Redevelop Enter Bug page](https://bugzilla.mozilla.org/show_bug.cgi?id=1549621)

I’m so sorry to take this long 🙇🏻 🙇🏻 🙇🏻 Here’s the new, modal version of the Enter Bug page! For now, I’ll keep the legacy form, meaning some hooks/templates are moved to new files so they can be included in both the modal and legacy forms. Otherwise, the change could be simpler. I’ve added the Use Legacy Form button at the top of the modal form; we could remove it once satisfied with the new form.

As discussed before, searchable component selector, mobile optimization, multiple attachment support, and changing the product will be done in the following phases. I’ll start working on these things right away on top of this work.

## Screenshots

image

image

image


dklawren:
Thanks Kohei! I am going to start looking at this today.

dklawren:
> Thanks! I’ll try to solve test failures today but it should work with a BMO database dump.

Yeah the test failures were to be expected as the tests heavily rely on the layout of the old enter bug form. Most of the changes will be in `qa/t/lib/QA/Util.pm` for file_bug_in_product().

kyoshino:
Finally, all the tests have passed!

dylanwh:
I just wanted to stop by and say this is great work and it looks fantastic.

dklawren (changes requested):

dklawren (changes requested):
Also found the following items of interest:

- Do we need the 'File a New Bug' header at all? Seems redundant. I would  say maybe 'File New Bug in <product>" but the product name is already displayed below the header. Not sure.
- Save changes button needs to be more to the left in line with the other fields
- Collapsed modals not remembered for each page load (do we need?)
- Maybe a quick link button to the comment field for advanced view? (Similar to Top button at bottom?)
- Use Guided Form as whitespace in the link which shows up as blue line outside of the button box
- When filing a new bug with description and then filing another bug (separate product) the old description is enterered again
- Clicking Etiquette above description opens a new tab where clicking on the writing guidelines loads in the same tab. Should be consistent.
- Substitute 'All' for 'All Platforms'
- Regressions field should be hidden from new bugs similar to current form
- When adding `&component=General` or other valid component name to the URL, the component is not automatically selected.
- Also when passing in `&component=General` and maybe others,  switching to Legacy form ignores everything except `product=Firefox` and `format=legacy`.
- Some fields are not auto-filled such as summary, component, etc. when using a bookmarked template from the enter bug form.

dklawren:
I will also be looking at the JS changes now in more detail. The above comments are from using the UI with some looks at the templates, etc.

kyoshino:
Thank you for looking at the details!

> Do we need the 'File a New Bug' header at all? Seems redundant. I would  say maybe 'File New Bug in " but the product name is already displayed below the header. Not sure.

I think the title is needed especially because the new Enter Bug page looks very similar to normal bug pages. Otherwise, people may get confused.

> Save changes button needs to be more to the left in line with the other fields

Modifed the layout.

> Collapsed modals not remembered for each page load (do we need?)

Looks like only People module didn’t remember the state. Fixed the issue.

> Maybe a quick link button to the comment field for advanced view? (Similar to Top button at bottom?)

Added the button to the header toolbar.

> Use Guided Form as whitespace in the link which shows up as blue line outside of the button box

Hmm, not sure what you mean. Maybe related: I’ve realized that there is a focus indicator issue due to nested `<a>` and `<button>` but it’s not specific to this page.

> When filing a new bug with description and then filing another bug (separate product) the old description is enterered again

Fixed. The cached is 

> Clicking Etiquette above description opens a new tab where clicking on the writing guidelines loads in the same tab. Should be consistent.

Fixed. These links are now open in a new tab.

> Substitute 'All' for 'All Platforms'

Fixed. The button with a longer label actually overflows the flexbox column but is displayed fine.

> Regressions field should be hidden from new bugs similar to current form

Removed.

> When adding `&component=General` or other valid component name to the URL, the component is not automatically selected.

Fixed.

> Also when passing in `&component=General` and maybe others,  switching to Legacy form ignores everything except `product=Firefox` and `format=legacy`.

Fixed. Now all the current field values will be copied to the legacy form for convenience.

> Some fields are not auto-filled such as summary, component, etc. when using a bookmarked template from the enter bug form.

Fixed.

extensions/BugModal/template/en/default/bug_modal/field.html.tmpl
```diff
@@ -124,6 +145,22 @@ END;
     [% aria_labelledby_html = BLOCK %]aria-label="[% aria_label || label FILTER html %]"[% END %]
   [% END %]
 
+  [% IF field.is_mandatory || required %]
+    [% aria_required = BLOCK %]
+      aria-required="true" aria-invalid="false" aria-errormessage="field-[% name FILTER html %]-error"
+    [% END %]
+    [% error_message = BLOCK %]
+      <div id="field-[% name FILTER html %]-error" class="input-error bz_default_hidden" aria-live="assertive">
+        [% IF field_type == constants.FIELD_TYPE_SINGLE_SELECT %]
```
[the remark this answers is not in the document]
kyoshino: I probably overlooked it because there is no multi-select field to be validated at least on BMO. Added the support just in case.

dklawren:
Looking better.

> > Use Guided Form as whitespace in the link which shows up as blue line outside of the button box
> 
> Hmm, not sure what you mean. Maybe related: I’ve realized that there is a focus indicator issue due to nested `<a>` and `<button>` but it’s not specific to this page.
> 

If you hover the pointer over 'Use Guided Form' there is a link underline (blue or purple depending on visited) that is seen between 'Use Guided Form' and 'Use Legacy Form'. If I inspect the 'Use Guided Form' element, the source view shows whitespace after `<button type="button" class="secondary">Use Guided Form</button>` and if I remove it manually it looks correct.

Also, there are some new warnings in the logs introduced with your changes that we will need to clean up before final release.

Use of uninitialized value in numeric eq (==) at /app/Bugzilla/Bug.pm line 5218.
Use of uninitialized value in numeric eq (==) at /app/Bugzilla/Bug.pm line 5300.

```
    diff --git a/Bugzilla/Bug.pm b/Bugzilla/Bug.pm                                                                                                                          
    index 73480de4c..1b645e5f4 100644                                                                                                                                       
    --- a/Bugzilla/Bug.pm                                                                                                                                                   
    +++ b/Bugzilla/Bug.pm                                                                                                                                                   
    @@ -5201,8 +5201,11 @@ sub check_can_change_field {                                                                                                                     
       if (!$self->{'error'}) {                                                                                                                                             
                                                                                                                                                                            
       # Allow the assignee to change anything else.                                                                                                                      
    -    if ( $self->{'assigned_to'} == $user->id                                                                                                                           
    -      || $self->{'_old_assigned_to'} && $self->{'_old_assigned_to'} == $user->id)                                                                                      
    +    if (                                                                                                                                                               
    +      (defined $self->{'assigned_to'} && $self->{'assigned_to'} == $user->id)                                                                                          
    +      || (defined $self->{'_old_assigned_to'}                                                                                                                          
    +        && $self->{'_old_assigned_to'} == $user->id)                                                                                                                   
    +      )                                                                                                                                                                
         {                                                                                                                                                                  
           return {allowed => 1};                                                                                                                                           
         }                                                                                                                                                                  
    @@ -5283,7 +5286,9 @@ sub check_can_change_field {                                                                                                                      
       }                                                                                                                                                                    
                                                                                                                                                                            
      # The reporter is allowed to change anything else.                                                                                                                   
    -  if (!$self->{'error'} && $self->{'reporter_id'} == $user->id) {                                                                                                      
    +  if (!$self->{'error'}                                                                                                                                                
    +    && (defined $self->{reporter_id} && $self->{'reporter_id'} == $user->id))                                                                                          
    +  {                                                                                                                                                                    
         return {allowed => 1};                                                                                                                                             
       } 
```

Argument "UNCONFIRMED" isn't numeric in numeric eq (==) at /app/extensions/BMO/lib/FakeBug.pm line 40.
Use of uninitialized value in numeric eq (==) at /app/extensions/BMO/lib/FakeBug.pm line 40.    

```
diff --git a/extensions/BMO/lib/FakeBug.pm b/extensions/BMO/lib/FakeBug.pm                                                                                              
    index f0624fc66..8b6efd5f2 100644                                                                                                                                       
    --- a/extensions/BMO/lib/FakeBug.pm                                                                                                                                     
    +++ b/extensions/BMO/lib/FakeBug.pm                                                                                                                                     
    @@ -37,7 +37,7 @@ sub _changes_everconfirmed {                                                                                                                          
                                                                                                                                                                            
     sub everconfirmed {                                                                                                                                                    
       my $self = shift;                                                                                                                                                    
    -  return ($self->{'status'} == 'UNCONFIRMED') ? 0 : 1;                                                                                                                 
    +  return (!$self->{'status'} || $self->{'status'} == 'UNCONFIRMED') ? 0 : 1;                                                                                           
     }                                                                                                                                                                      
                                                                                                                                                                           
    1;
```

Use of uninitialized value in pattern match (m//) at /app/Bugzilla/FlagType.pm line 698.

```
    diff --git a/Bugzilla/FlagType.pm b/Bugzilla/FlagType.pm                                                                                                                
    index 7acc860ff..ece447223 100644                                                                                                                                       
    --- a/Bugzilla/FlagType.pm                                                                                                                                              
    +++ b/Bugzilla/FlagType.pm                                                                                                                                              
    @@ -695,7 +695,7 @@ sub sqlify_criteria {                                                                                                                               
         my $is_active = $criteria->{is_active} ? "1" : "0";                                                                                                                
         push(@criteria, "flagtypes.is_active = $is_active");                                                                                                               
       }                                                                                                                                                                    
    -  if (exists($criteria->{active_or_has_flags})                                                                                                                         
    +  if (defined $criteria->{active_or_has_flags}                                                                                                                         
         && $criteria->{active_or_has_flags} =~ /^\d+$/)                                                                                                                    
       {                                                                                                                                                                    
         push(@$tables,
```

dklawren:
After discussion, we should also remove the Legacy Form button completely as we should not have an issue with pushing people to use this new form as the default. We should definitely leave the Guided button as it would be nice for people to see that it exists. Users with improved permissions never even see that form. 

Also please remove the 'Restrict comments...' checkbox as it is not needed on the enter bug page.

https://bugzilla-dev.allizom.org/page.cgi?id=restrict_comments_guidelines.html
' Restrictions should not be used as a preemptive measure against comments which have not yet occurred.'

Also I have the latest code installed on the test instance at https://bugzilla-dev.allizom.org/enter_bug.cgi?product=Firefox

Kohei, let me know if you need your password reset on bugzilla-dev as we have refreshed the database recently.

Looks good!

kyoshino:
Thank you! The fixes are in. Looks like I need a new password for the staging server as GitHub auth doesn’t work with 2FA enabled.

> Unable to log in with GitHub because two-factor authentication is enabled on your account.
> Please log in using your username and password.

dklawren (changes requested):

extensions/BugModal/Extension.pm
```diff
@@ -33,6 +33,15 @@ use constant READABLE_BUG_STATUS_PRODUCTS => (
...
 );
 
+sub enter_bug_format {
+  my ($self, $args) = @_;
+  my $cgi  = Bugzilla->cgi;
+
+  # Use the modal or custom format unless `format=legacy` is given as a URL param
+  my $format = $cgi->param('format') || 'modal';
+  $args->{format} = $format eq 'legacy' ? '' : $format;
+}
```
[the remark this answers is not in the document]
kyoshino: Added a user setting for this.

dklawren (changes requested):
We are getting very close to done! And it is looking awesome. Thanks for the user setting.

Here are a few more things found when testing on bugzilla-dev:

- User Story: Text area appears and the cursor position is not at position 0 but several characters to the right when expanding the modal and clicking in the textarea.
- When certain tracking flags are set to `?` the text `(comment updated)` appears and pushes the right side status flags outside of the modal box a small amount. I am not sure how this can be fixed unless we shrink the left hand side field container to `width: 45%;` only for the Tracking modal box.  
image
- `[Summarize time (including time for bugs blocking this bug)` should not display for enter bug for the Time Tracking modal box.
- Bookmarkable template issues: Some variables are not getting prefilled properly. This could be due to so much now being shared with the bug_modal edit form and fields that is it is not looking for the default values when available. For groups, could be related to the new separate groups template that is also shared between enter bug and show bug. Basically any key=values passed to enter_bug.cgi? should prefill the form field when possible if it was originally working on the old form.
	- qa contact
	- assignee
	- status whiteboard
	- qa whiteboard (cf_qa_whiteboard)
	- no tracking flags or bug flags (but doesn't work on old form either)
	- groups checked
	- user story
	- crash signature
	- time tracking `original estimate` and `hours left` (deadline is pre-filled correctly)

dklawren:
> Thank you! The fixes are in. Looks like I need a new password for the staging server as GitHub auth doesn’t work with 2FA enabled.
> 
> > Unable to log in with GitHub because two-factor authentication is enabled on your account.
> > Please log in using your username and password.

I have removed TOTP on your account on bugzilla-dev.allizom.org See if that works.

kyoshino:
Hmm, the bookmarkable template button seems working for me. Not sure why it doesn’t work because it’s also submitting form data to `post_bug.cgi`. Well, this feature can be replaced with a JavaScript implementation like what I did for the [removed Use Legacy Form button](https://github.com/mozilla-bteam/bmo/pull/2132/commits/26837daa6fd50423308269119859eae422ab236c), but it would be out of the scope of this change. (I assume that the `URLSearchParams` and `FormData` APIs didn’t exist when the button was initially added to Bugzilla 😉)

I can now sign in to https://bugzilla-dev.allizom.org now 👍🏼

kyoshino:
This is what the updated Tracking Fields section would look like:

image

dklawren (commented):
After some recent discussion, we decided have a user setting for the old form is the wrong option as we want to remove the ability to use the old form completely. By having a user setting, users will be able to still set their preferences back to the old form if desired. This would be best as a config parameter instead that just picks one form or the other with no option for the user to change. I can do this change myself after this lands unless you want to make it. I am not fussed either way.

I am working on today verifying the issue with bookmarked templates not working with the latest changes pushed to bugzilla-dev. If it turns out to be a non-issue for me then this should be ready to go.

kyoshino:
I can do it now 🙂

kyoshino:
Param added. Wasn’t sure if there’s a standard way to add a new one. If it doesn’t exist in the database, the code [dies here](https://github.com/mozilla-bteam/bmo/blob/master/Bugzilla/Config.pm#L80) and no change can be made through the admin panel. I just added it manually and rebooted the servers.

```sql
INSERT INTO params (name, value) VALUES ('use_modal_create', '1');
```

dklawren:
> Param added. Wasn’t sure if there’s a standard way to add a new one. If it doesn’t exist in the database, the code [dies here](https://github.com/mozilla-bteam/bmo/blob/master/Bugzilla/Config.pm?rgh-link-date=2023-12-04T22%3A07%3A01Z#L80) and the value cannot be changed through the admin panel. I just added it manually and rebooted the servers.
> 
> ```sql
> INSERT INTO params (name, value) VALUES ('use_modal_create', '1');
> ```

That  works but the way I designed it was that when you run ./checksetup.pl, it will create any missing params that have been added. I will need to do that as soon as the new code hits the production servers which is not a problem.

dklawren:
I am still seeing the issue with fields not being pre-populated based on query parameters passed in. Try the following simple tests.

Old form:
https://bugzilla-dev.allizom.org/enter_bug.cgi?product=Firefox&qa_contact=dkl@mozilla.com&assigned_to=dkl@mozilla.com&format=legacy

(assignee and qa contact have dkl@mozilla.com)

New form:
https://bugzilla-dev.allizom.org/enter_bug.cgi?product=Firefox&qa_contact=dkl@mozilla.com&assigned_to=dkl@mozilla.com

(assignee and qa contact are empty)

Does this still work for you?

kyoshino:
Problem solved 🙇🏼 Assignee and QA Contact are the only fields with `FIELD_TYPE_USER` and required a tweak.

extensions/BugModal/template/en/default/bug_modal/create.html.tmpl
```diff
@@ -0,0 +1,812 @@
...
+[%
+  USE Bugzilla;
+
+  bug_modal = 1;
+
+  # use the placeholder bug
+  bug = default;
+
+  # reset variables to avoid unexpected inheritance
+  default = undef;
```
dklawren, line 18: Thanks for the fix for assignee and qa_contact. Those and other user fields now pre-fill properly. Other fields form my earlier comment still do not fill but I think I see the issue. For the security groups being not being automatically checked if for example if we pass `groups=core-security-release&groups=mozilla-employee-confidential`  it is related to bug_modal/create.html.tmpl setting default = undef in the top of this file. Then bug/create/security.html.tmpl is looking for `default.groups.contains(group.name)` to see if the checkbox should be checked which of course is empty. This same file on the legacy form does work properly. I would imagine that any form fields in the new modal form that look for default.* would not see any values because of the undef. Converting to `bug.groups.contains(group.name)` will work for modal form but will break the checkboxes for the legacy form. A hack would be to not undef default and allow using either bug.* or default.* but maybe you have a better idea?

This might also fix the other fields that are not being pre-filled:
 * status whiteboard
  * qa whiteboard (cf_qa_whiteboard)
  * user story
  * crash signature
  * time tracking original estimate and hours left (deadline is pre-filled correctly)

dklawren (changes requested):

extensions/BugModal/template/en/default/bug_modal/create.html.tmpl
kyoshino: QA whiteboard (`cf_qa_whiteboard`) not being prefilled is an administration issue. Go to `/editfields.cgi?action=edit&name=cf_qa_whiteboard`, tick “Can be set on bug creation”, submit, and I see the field prefilled. It’s [fillered here](https://github.com/mozilla-bteam/bmo/blob/master/enter_bug.cgi#L244). The old form didn’t have the field, actually.

kyoshino: - Status Whiteboard: Fixed!
- QA whiteboard: See above. Most of the custom fields, including Rank and flags are actually not prefilled due to the same admin issue. Open `/editfields.cgi` to see that almost all fields come with `0` on the “Editable on Bug Creation“ column.
- User Story: Fixed!
- Crash Signature: It was prefilled but wasn’t visible at first glance. The field now behaves like the old form: it shows a textbox with a value prefilled or hides the textbox if no prefill value is set.
- Time Tracking: Fixed! I have removed the Hours Worked and Hours Left fields which seem unnecessary and aren’t on the old form.

PR #2135, opened 2023-10-25
Bug 1861260 - Update the Push extension to add additional status check code in order to restart the pod if an error is occurring

dklawren:
This adds additional checking code to the Push extension daemon. When running the command with the `check` parameter, instead of just returning a PID as is the default, we also check to make sure the database is up and accessible. In the recent past, the database disappeared and the daemon continued to run and checkout but the log was full of MySQL connection errors. This code change allows for Kubernetes to do a liveness check and restart the POD if the daemon can no longer access the database.

### DIFF
--- a/extensions/Push/lib/Daemon.pm
+++ b/extensions/Push/lib/Daemon.pm
@@ -14,6 +14,8 @@ use warnings;
 use Bugzilla::Constants;
 use Bugzilla::Extension::Push::Push;
 use Bugzilla::Extension::Push::Logger;
+use Bugzilla::Util qw(get_text);
+
 use Carp qw(confess);
 use Daemon::Generic;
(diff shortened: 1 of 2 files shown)


cgsheeh (approved):
How does this function run? Is it running somewhere due to the `gd_` prefix?

dklawren:
> How does this function run? Is it running somewhere due to the `gd_` prefix?

It will be run by kubernetes from the command line at regular intervals. It is called as `bugzilla-pushd.pl check` and exits with a 0 if successful or something else if it was not. `gd_check()` is the method looked for by Daemon::Generic to decide what to do when the `check` parameter is given. If not `gd_check` method is defined, then it just defaults to displaying the current PID and exiting with 0.

PR #2137, opened 2023-11-02
Bug 1862316 - Update Thunderbird logo

dklawren:
Updated thunderbird logo to newer logo.


cgsheeh (approved):
Nice! The new image is 1024x1024 whereas the old one was 128x128, that won't be a problem I assume?

dklawren:
> Nice! The new image is 1024x1024 whereas the old one was 128x128, that won't be a problem I assume?

Should be fine. The html that use the image is `` so all of the icons will look the same size.

PR #2140, opened 2023-11-16
Bug 1864995 - Changes to fix bug 1841823 broke the use of recovery codes for accessing an account with MFA enabled 

dklawren:


dklawren:
Moving over zeid r+ review from bug.

PR #2145, opened 2023-12-08
Bug 1861888 - Requiring Staff To Use Duo 2FA On Bugzilla

dklawren:
* This pull request adds two new config parameters, `duo_required_group` and `duo_required_excluded_group`. 
* Any user that is added to the `duo_required_group` will automatically be redirected to the MFA preferences page if their current MFA is not set to `Duo`.
* There they will be able to configure `Duo` for their account using the standard enrollment procedure.
* If the user account is a member of the `duo_required_excluded_group`, then they will not be redirected to the MFA page and will operate normally.
* Only bot accounts (addresses ending in tld or bugs) should be allowed to be added to the excluded group.
* If a member of the required group is removed from that group, they are logged out, their password is cleared, and their MFA is cleared. They can then reset their password and setup TOTP instead.
* If they were previously not a member and are added, then they are logged out and the next time they log in they, they are redirected to the MFA page if they are not already Duo enrolled. Otherwise the operate normally.

### DIFF
--- a/Bugzilla.pm
+++ b/Bugzilla.pm
@@ -308,6 +308,20 @@ sub login {
       $authenticated_user->update();
     }
   }
+  
+  # Require Duo Security as MFA provider if user is in the duo_required_group
+  elsif (!i_am_webservice()
+    && Bugzilla->params->{duo_required_group}
+    && $authenticated_user->in_duo_required_group
+    && $authenticated_user->mfa ne 'duo')
+  {
+    my $on_mfa_page
+      = $script_name eq '/userprefs.cgi' && $cgi->param('tab') eq 'mfa';
+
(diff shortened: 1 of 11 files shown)


cgsheeh (approved):
Makes sense to me.

dklawren:
I updated the code in Bugzilla/WebService/User.pm get() to have more consistent variable naming when dealing with the logged in user and the user objects being returned. Hopefully the code is less confusing now and less error prone.

dklawren:
Re-requesting review: Test script has been updated to cover more functionality. I am now setting the mfa_required_date to a date in the past (immediate change needed) which should require the user to setup TOTP when removed from the duo_required_group.

cgsheeh (approved):

PR #2151, opened 2023-12-12
Bug 1869601 - GitHub push endpoint should stop setting the `qe-verify` flag when resolving a Fenix or Focus bug

dklawren:


cgsheeh (approved):

Bugzilla/BugMail.pm
```diff
@@ -366,6 +366,7 @@ sub sendMail {
     # Only display bug ids that the user is allowed to see for certain fields
     if ($diff->{field_name} =~ /^(?:dependson|blocked|regress(?:ed_by|es))$/) {
       foreach my $field ('new', 'old') {
+        next if !defined $diff->{$field};
```
cgsheeh, line 369: Is this intended?

cgsheeh (changes requested):

Bugzilla/BugMail.pm
dklawren: Yeah snuck that in there as it causes a stderr warning when running the API tests. I can take it if you want and add it to a separate bug

cgsheeh (approved):

PR #2156, opened 2023-12-18
Bug 1870549 - When grabbing the next story id, sometimes it is invalid so we should try to grab the next one up to 5 tries

dklawren:
This PR updates the subroutine that is used to get the next list of feed stories from Phabricator based on the last successful feed ID stored. It will try the next ID but if it fails with the familiar `does not identify a valid object` then it will increment and try the next ID up to 5 increments. If after 5 then it will just return an empty list like we do currently and an admin will need to step in.

### DIFF
--- a/extensions/PhabBugz/lib/Feed.pm
+++ b/extensions/PhabBugz/lib/Feed.pm
@@ -878,16 +878,28 @@ sub process_new_user {
 sub new_stories {
   my ($self, $after) = @_;
   my $data = {view => 'text'};
-  $data->{after} = $after if $after;
+  $data->{after} = ($after ? $after : 0);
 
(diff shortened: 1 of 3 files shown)


cgsheeh (approved):
LGTM - maybe we could add some logging as well, so we know if we skip an entry/we hit the `last`/we exhaust all the retries.

dklawren:
> LGTM - maybe we could add some logging as well, so we know if we skip an entry/we hit the `last`/we exhaust all the retries.

Good idea. Will add that.

dklawren:
Actually I need to revise this as request() throws an exception early on if 'error_code' is set so I need to wrap this in a try/catch block to even work.

PR #2157, opened 2023-12-20
Bug 1870891 - Replace native <select> with custom element with search bar

kyoshino:
[Bug 1870891 - Replace native `<select>` with custom element with search bar](https://bugzilla.mozilla.org/show_bug.cgi?id=1870891)

This PR replaces most of the native `<select>` elements on the enter_bug and show_bug pages with new custom elements. A search bar appears when there are 10 or more options for easier filtering. The component dropdown displays the component description in a popover.

Implementation note: These dropdown lists and popovers are created using `<dialog>` (modal [top layer](https://developer.chrome.com/blog/what-is-the-top-layer)) instead of the [Popover API](https://developer.chrome.com/blog/introducing-popover-api/) (non-modal top layer) because the latter is [not yet enabled](https://bugzilla.mozilla.org/show_bug.cgi?id=1866993) in Firefox other than Nightly. It’s not a problem but just in case some wonder.

image

image

### DIFF
--- a/js/components/select.js
+++ b/js/components/select.js
@@ -1,0 +1,1242 @@
+// @ts-check
+
+/**
+ * Implement a `<select>` element alternative that shows a searchbar for easier data input. This
+ * partially implements the {@link HTMLSelectElement} DOM API and the `combobox` WAI-ARIA role.
+ * @see https://developer.mozilla.org/en-US/docs/Web/API/HTMLSelectElement
+ * @see https://w3c.github.io/aria/#combobox
+ * @see https://www.w3.org/WAI/ARIA/apg/patterns/combobox/
+ */
+class BzSelectElement extends HTMLElement {
+  /**
+   * Make it the custom element part of the outer form.
+   * @type {boolean}
+   */
+  static formAssociated = true;
+
+  /**
+   * A list of attributes to be captured with {@link attributeChangedCallback}.
+   * @type {string[]}
+   */
+  static get observedAttributes() {
+    return ['name', 'disabled', 'required', 'aria-required', 'aria-invalid', 'aria-errormessage'];
+  }
+
+  /**
+   * Whether to allow selecting multiple options. This component doesn’t support multiple selection,
+   * so the value is fixed with `false`.
+   * @type {boolean}
+   */
+  multiple = false;
+
+  /**
+   * The `type` property. This component doesn’t support multiple selection, so the value is fixed.
+   * @type {string}
+   */
+  type = 'select-one';
+
+  /**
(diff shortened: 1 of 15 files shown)


kyoshino:
Hmm, all the Selenium tests failed. I assume the headless Firefox instance is too old to support `<dialog>`, which was shipped with [Firefox 98](https://www.mozilla.org/en-US/firefox/98.0/releasenotes/) in March 2022. [selenium/standalone-firefox:3.141.59](https://github.com/mozilla-bteam/bmo/blob/master/docker-compose.test.yml#L64) was [released in September 2021](https://registry.hub.docker.com/layers/selenium/standalone-firefox/3.141.59/images/sha256-738eeec4eba6e44d0e5f0697e36b77b907e3c6fb647759a82fe47dec0d188abf?context=explore).

kyoshino:
Tests are still failing, of course — because I have replaced `<select>` and `<option>` with `<bz-select>` and `<bz-option>`. I’m updating relevant XPaths in the tests accordingly.

kyoshino:
All tests passed 🟢

dklawren:
Is this ready for review? I can start looking at it right away if so.

kyoshino:
> Is this ready for review? I can start looking at it right away if so.

Yes!

dklawren (changes requested):
Mostly everything checks out and looks great! I did see one issue that will need to be resolved. When changing the product on a current bug (show_bug.cgi) the component, version and milestone fields need to be highlighted the same way the current bug page does. This is not happening with the new select. For example:

image

I have installed your latest changes on bugzilla-dev.allizom.org if you want to see it in action with a large list of components, etc.

kyoshino:
Thanks @dklawren! I’ve added the fix for field highlights.

js/components/select.js
```diff
@@ -0,0 +1,1246 @@
...
+      (this.hasAttribute('aria-labelledby')
+        ? document
+            .getElementById(this.getAttribute('aria-labelledby'))
+            ?.textContent?.trim()
+            .replace(/:$/, '')
+        : this.getAttribute('aria-label')) || this.getAttribute('name');
+
+    this.attachShadow({ mode: 'open' }).innerHTML = `
+      <style>
+        :host {
```
dklawren, line 345: Would the css be better in a separate file outside of the js code or was there a reason it needed to be here?

kyoshino: This is the way custom components are encapsulated. I personally don’t like this, but CSS is often written in JS. Should be fine.

dklawren (approved):

PR #2158, opened 2023-12-20
no bug - Update Selenium headless Firefox to latest

kyoshino:
Not sure if this will work but we need to update [selenium/standalone-firefox](https://registry.hub.docker.com/r/selenium/standalone-firefox/tags) to a newer version as per #2157. If `latest` doesn’t work, we could try an older version, maybe Firefox 115 ESR.

### DIFF
--- a/docker-compose.test.yml
+++ b/docker-compose.test.yml
@@ -62,7 +62,7 @@ services:
 
   selenium:
     image: selenium/standalone-firefox:102.0.1
-    shm_size: "512m"
+    shm_size: "2gb"
     ports:
       - "59000:5900"
 
--- a/docker-compose.test.yml
+++ b/docker-compose.test.yml
@@ -61,7 +61,7 @@ services:
     image: ghcr.io/ghcr-library/memcached:latest
 
   selenium:
-    image: selenium/standalone-firefox:3.141.59
+    image: selenium/standalone-firefox:latest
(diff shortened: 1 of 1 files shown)


dklawren:
Strange thing is the tests are running fine up until the very end and then the selenium just timesout waiting for something else. Normally the window would close and the next script would start. So the last screen is the logged out screen.

kyoshino:
2 Stack Overflow threads hit when I search for "[Net/HTTP/Methods.pm line 274](https://github.com/libwww-perl/Net-HTTP/blob/master/lib/Net/HTTP/Methods.pm#L274)"

https://stackoverflow.com/questions/75737243/read-timeout-error-in-nethttp-live-t-test-in-perl-5-32
https://stackoverflow.com/questions/74756646/problem-running-selenium-with-firefox-geckodriver-and-perl

One says the problem is solved by increasing CPU and memory. 🤔

dklawren:
I went a different route for this in pull request https://github.com/mozilla-bteam/bmo/pull/2160 so please check that out and see if it works better for your testing. It is using Mozilla Firefox 115.5.0esr which should be new enough.

kyoshino:
Wow, thanks! Firefox 115 should be fine. I’ll rebase #2157 once #2160 is merged. And if there are still failures, I’ll take care of them.

PR #2160, opened 2023-12-21
Bug 1871414 - Update Selenium test suite to use a more recent version of Firefox to support new features

dklawren:
This allows our tests to use Mozilla Firefox 115.5.0esr instead of the older version in selenium/standalone-firefox

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -24,13 +24,26 @@ defaults:
 
   build_image: &build_image
     run:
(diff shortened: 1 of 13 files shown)


cgsheeh (approved):
Makes sense to me.

IMO there should be a bug on file for this change, changes to tests infrastructure aren't appropriate for no bug workflows.

cgsheeh (approved):

PR #2163, opened 2024-01-08
Bug 1873447 - On the new bug form, can't use control-E on Mac to go to the end of fields

kyoshino:
[Bug 1873447 - On the new bug form, can't use control-E on Mac to go to the end of fields](https://bugzilla.mozilla.org/show_bug.cgi?id=1873447)

This is not actually a regression but due to a typo in `nodeName`. I’m simply replacing it with modern DOM 🙂


kyoshino:
Looks like there are some random failures. 🤔

dklawren (approved):

PR #2165, opened 2024-01-10
Bug 1873826 - Avoid hiding scrollbar when opening drop down menu.

arai-a:
Instead of suppressing scroll by hiding scroll bar, let the drop down menu follow the scroll,
so that the page content doesn't shift by hiding/showing scrollbar for each click.

### DIFF
--- a/js/components/select.js
+++ b/js/components/select.js
@@ -66,6 +66,12 @@ class BzSelectElement extends HTMLElement {
   #dialog;
 
   /**
+   * A function to remove scroll event listener when hiding drop down.
+   * @type {function}
+   */
+  #removeScrollEventListener;
+
+  /**
    * A reference to the `<input>` element that serves as a searchbar.
    * @type {HTMLInputElement}
    */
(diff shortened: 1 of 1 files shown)


arai-a:
`3_test_net_attachments.t` intermittently fails even without the patch.
so I'll leave it for other bug.

dklawren:
> `3_test_net_attachments.t` intermittently fails even without the patch. so I'll leave it for other bug.

I am running it again and will probably pass the second time around. I am looking at it as a timing issue and will hopefully fix it for good soon.

dklawren:
I will let @kyoshino review this one

kyoshino:
@arai-a Thanks for the heads-up! I thought there would be a simpler solution to this, and [found this article](https://blog.mayank.co/is-dialog-enough) introducing [`scrollbar-gutter`](https://developer.mozilla.org/en-US/docs/Web/CSS/scrollbar-gutter). So we just need this:

```css
html {
  scrollbar-gutter: stable both-edges;
}
```

kyoshino (changes requested):
@arai-a Thanks for the heads-up! I thought there would be a simpler solution to this, and [found this article](https://blog.mayank.co/is-dialog-enough) introducing [`scrollbar-gutter`](https://developer.mozilla.org/en-US/docs/Web/CSS/scrollbar-gutter). So we just need this probably in [`global.css`](https://github.com/mozilla-bteam/bmo/blob/master/skins/standard/global.css#L340):

```css
html {
  scrollbar-gutter: stable both-edges;
}
```

PR #2171, opened 2024-01-12
Bug 1874037 - Type to search options doesn’t work with custom <bz-select> element having no search bar

kyoshino:
[Bug 1874037 - Type to search options doesn’t work with custom <bz-select> element having no search bar](https://bugzilla.mozilla.org/show_bug.cgi?id=1874037)

Restore the native `<select>` element’s type-to-search functionality. For example,

- Click on the Platform field
- Hit the A key to select All
- Hit the A key again to select ARM

### DIFF
--- a/js/components/select.js
+++ b/js/components/select.js
@@ -318,6 +318,15 @@ class BzSelectElement extends HTMLElement {
   }
 
   /**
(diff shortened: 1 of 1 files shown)


dklawren (approved):
LGTM and works as expected when testing. r=dkl

PR #2175, opened 2024-01-23
Bug 1875961 - Update attachment migration script by adding option to only update the class and not actually copy the data

dklawren:
- Adds --class option to migrate-attachments.pl to just update the class names in the atachment_storage_class table from one value to another.
  - This is useful for when the data is already migrated from net storage to another using external means and we just want the data to be loaded from the new location
- Also made some fixes to the migration script that properly calculates the number of attachments that will be affected. Before if was mistakenly just counting all attachments regardless of source.
- Updated Storage.pm to properly encode the returned content into UTF8 to fix a bug with using the fake GCS server as test platform.

### DIFF
--- a/scripts/migrate-attachments.pl
+++ b/scripts/migrate-attachments.pl
@@ -91,38 +91,33 @@ if ($options{migrate}) {
   }
   my ($source, $dest) = @{$options{migrate}};
 
+  my $query = '
+    FROM  attachments 
+          JOIN attachment_storage_class ON attachments.attach_id = attachment_storage_class.id 
+    WHERE attachment_storage_class.storage_class = ? AND attachments.attach_size != 0 ';
+
   # Do not migrate from database to net storage if data is less than minsize
-  my $where = '';
   if ( $source eq 'database'
     && $dest eq 's3'
     && Bugzilla->params->{attachment_s3_minsize})
   {
-    $where .= ' AND attachments.attach_size > '
+    $query .= ' AND attachments.attach_size > '
       . int Bugzilla->params->{attachment_s3_minsize};
   }
   elsif ($source eq 'database'
     && $dest eq 'google'
     && Bugzilla->params->{attachment_google_minsize})
   {
-    $where .= ' AND attachments.attach_size > '
+    $query .= ' AND attachments.attach_size > '
       . int Bugzilla->params->{attachment_google_minsize};
   }
 
-  my ($total) = $dbh->selectrow_array('
-    SELECT COUNT(*) 
-      FROM attachments 
-           JOIN attachment_storage_class ON attachments.attach_id = attachment_storage_class.id 
-     WHERE attachment_storage_class.storage_class = ? ' . $where, undef, $source);
+  my ($total) = $dbh->selectrow_array("SELECT COUNT(*) $query", undef, $source);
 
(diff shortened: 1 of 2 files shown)


scripts/migrate-attachments.pl
```diff
@@ -75,21 +92,38 @@
   my ($source, $dest) = @{$options{migrate}};
 
   # Do not migrate from database to net storage if data is less than minsize
-  my $where;
-  if ($source eq 'database' && $dest eq 's3' && Bugzilla->params->{attachment_s3_minsize}) {
-    $where = 'WHERE attach_size > ' . int Bugzilla->params->{attachment_s3_minsize};
+  my $where = '';
```
cgsheeh, line 95: Nit: since this is now just the attach size clause, we could probably rename this variable.

scripts/migrate-attachments.pl
```diff
@@ -110,33 +144,80 @@
...
+
+    my $attachment = Bugzilla::Attachment->new({id => $attach_id, cached => 1});
+
+    # Update the class of the attachment
+    $attachment->current_storage($dest)->set_class();
+
+    $updated++;
+  }
+  print "\n";
+  print "Attachments updated: $updated\n";
```
cgsheeh, line 180: Looks like `$count` and `$updated` are the same thing here? We can remove one of them.

scripts/migrate-attachments.pl
```diff
@@ -185,25 +270,32 @@
...
+    undef, $source);
+
   confirm(sprintf 'DELETE %s attachments from %s?', $total, $options{delete});
 
-  my $sth
-    = $dbh->prepare(
-    "SELECT attach_id FROM attachments WHERE attach_size != 0 ORDER BY attach_id DESC"
-    );
-  $sth->execute();
+  my $sth = $dbh->prepare('
```
cgsheeh, line 284: `$sth` is an ambiguous variable name.

dklawren: $sth means "statement handle" and $dbh means "database handle". The statement handle is so we can iterate over the rows as they are returned from the database when executing a query. Instead of pulling in all rows at once into memory. $sth is used throughout the BMO code and this script will not be used very often.

scripts/migrate-attachments.pl
dklawren: $count is the number of rows currently process from the total rows returned from the database. $updated, $deleted, $migrated, are the number of rows that actually had some action taken with the row. So they need to be different values.

scripts/migrate-attachments.pl
dklawren: Latest revision changes the script to reuse SQL in more places so this variable has been renamed as part of that.

scripts/migrate-attachments.pl
cgsheeh: In this case, `$updated` and `$count` will always be the same value when they are accessed.

cgsheeh (approved):

scripts/migrate-attachments.pl
dklawren: > In this case, `$updated` and `$count` will always be the same value when they are accessed.

Ah I see what you mean. Yes for the 'class' section, since there is no conditional in the while loop, then the two values are always the same anyway. Well rather than push a new change and since it is no harm, I will leave it for consistency :)

PR #2186, opened 2024-02-02
Bug 1157227 - redesign the "attachment details" page

kyoshino:
I apologize for keeping you waiting. Here’s the attachment overlay suggested in my Figma prototype that replaces the current attachment details page.

[Bug 1157227 - redesign the "attachment details" page](https://bugzilla.mozilla.org/show_bug.cgi?id=1157227)

As described in the bug:

- Removed the preview iframe
- Removed the Edit as Comment feature
- Always show the metadata
- Make action links more prominent

Other things:

- Renamed the View action to Raw, just like GitHub’s file viewer
- Removed the GitHub PR diff preview as well, because it’s not useful
- The overlay also replaces the lightbox for images
- I’ve kept the current page’s layout showing the metadata on the left, not on the right as shown in Figma
- You can navigate multiple attachments with the right/left arrow keys

Screen Shot 2024-02-02 at 10 40 19
Screen Shot 2024-02-02 at 10 40 16
Screen Shot 2024-02-02 at 10 40 12


kyoshino:
All tests passed. @dklawren Please have a look when you have time. 🙏🏼

dklawren:
> All tests passed. @dklawren Please have a look when you have time. 🙏🏼

Thanks Kohei! I will try to start looking at it today.

dklawren (changes requested):
* Sorry for the delay, Overall this looks really good. Just a few comments/questions.
* For consistency with show bug, can we move the [Save] button from the bottom right where it is not easy to see to the top right? Also the buttons on the bottom could be consolidated to the top as well maybe. How about if the top bar looked like
    * [View] [Splinter Review] [Hide Details] [Cancel] [Save]
    * The [View] button would be a drop down with the choices of Raw and Diff. We similar on show bug with XML and JSON.
    * The [Delete] button and maybe another [Save] button near the comment area could stay at the bottom. Most people will never see the [Delete] button anyway.
    * Thoughts?
* Please add tool tips to the arrow buttons such as "Next attachment in list" and "Previous attachment in list" or something like that. Would help to describe what those are for.
* Found an issue that is result of the attachment details displaying in page with show bug.
  1. Logout.
  2. Login using header login form while viewing a bug with attachments
  3. Click on an attachment details, change anything on the attachment such Obsolete, etc.
  4. Get an error message "To display this page, Nightly must send information that will repeat any action (such as a search or order confirmation) that was performed earlier. [Cancel] [Resend]"
  5. Hit Resend, and then make another change to attachment and same error popup.
  6. If you reload the bug page while logged in, you can make changes to attachments without error.
* Show created/update timestamps in friendly format similar to show bug with the exact timestamps when hovering with mouse.
  * Created: 3 hours ago
  * Updated: 26 minutes ago

kyoshino:
Fixed a few things:

- Renamed the hook file (oops I forgot to rename it)
- Added a tooltip to previous/next buttons
- Made dates relative format

Buttons:

- Generally speaking, the primary/submit action button should always be at the (right) bottom and there should be only one button for one action. There is a Save Changes button at the right bottom of the Show Bug page’s detail form, so it’s kinda consistent already. 🙂
- The header gets crowded when all the buttons are placed. There are already the attachment title and previous/next buttons, and there is actually a status bar (empty by default) next to the Save button. I think the current button placement is organized.
- I don’t know why the Show Bug page has three Save buttons. It could be solved by having a floating/sticky/fixed action toolbar at the bottom [like this](https://experience.sap.com/fiori-design-web/footer-toolbar/), but of course, it’s not the scope of this bug.

Reload warning dialog:

- I was able to reproduce this, but looks like it’s not specific to the attachment overlay. The dialog appears on any page when you try to (manually) reload the page right after logging in with the header login form.
- Also the situation may not happen often, so I would defer the fix.

kyoshino:
> Reload warning dialog

No worries, I’ve just worked around the issue.

dklawren (changes requested):
Everything is mostly looking good except I found a couple bugs when doing my final testing on bugzilla-dev.allizom.org

* When setting a flag such as review when the requestee field is mandatory, the field has the red border properly, but you can still submit and the flag is just not set.
* When a review/feedback, etc flag is set to ?, you cannot clear the flag. If you set the flag to empty or even accept it (+|-), and the submit, the flag(s) are unchanged.

kyoshino:
Thanks @dklawren! I’ve solved the remaining issues.

dklawren (approved):

PR #2187, opened 2024-02-09
Bug 1879358 - Fix a few remaining issues with code changes enforcing duo 2fa for certain members

dklawren:
- Fixed issue where required date was not being cleared properly.
- Added additional text to the MFA preference page explaining why MFA is required.

### DIFF
--- a/template/en/default/account/prefs/mfa.html.tmpl
+++ b/template/en/default/account/prefs/mfa.html.tmpl
@@ -75,7 +75,7 @@
       [% IF user.mfa == 'Duo' && user.in_duo_required_group %]
         <p class="mfa-warning-msg">
           You are in a group requiring Duo Security for two-factor authentication and 
-          therefore it cannot be disabled. Talk to admin if you need assistance.
+          therefore it cannot be disabled. Talk to an admin on Slack (#bmo) if you need assistance.
         </p>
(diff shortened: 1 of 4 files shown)


template/en/default/account/prefs/mfa.html.tmpl
```diff
@@ -61,7 +72,12 @@
     <input type="hidden" name="mfa_action" id="mfa-action" value="disable">
 
     <div class="mfa-buttons">
-      [% IF user.mfa != 'Duo' || (user.mfa == 'Duo' && !user.in_duo_required_group) %]
+      [% IF user.mfa == 'Duo' && user.in_duo_required_group %]
+        <p class="mfa-warning-msg">
+          You are in a group requiring Duo Security for two-factor authentication and 
+          therefore it cannot be disabled. Talk to admin if you need assistance.
```
cgsheeh, line 78:
```suggestion
          therefore it cannot be disabled. Talk to an admin if you need assistance.
```

Also, is it worth calling out where to go to contact an admin? Since this is employee only it could just be "come to #bmo on Slack".

cgsheeh (approved):

template/en/default/account/prefs/mfa.html.tmpl
dklawren: Good idea. I will change that.

PR #2193, opened 2024-02-21
Bug 1881253 - Hide the attachment if the corresponding comment is marked as spam.

arai-a:


arai-a:
This does the following:
  * (a) if the comment that creates the attachment has "spam" tag, do not show the attachment in the attachment list
  * (b) the comment's tag is the source of truth for the condition, and it's not mirrored to attachment data
  * (c) the attachment is not marked obsolete, private, or deleted
  * (d) there's no way to force-show the attachment in the table (in contrast to "obsolete" mark, which can be shown by button)
  * (e) comment that updates the attachment is not used as condition

(b) and (c) are to avoid increasing the chance of inconsistency, and also to avoid introducing attack vector around unmarking private or undoing the deletion, given tagging requires less privilege.

(d) is to reduce the complexity around "what's hidden and shown", given there's already "obsolete" mark. The attachment's "Details" page can still be reached by expanding the comment if necessary.

(e) is just for simplicity

PR #2194, opened 2024-02-22
Bug 1881309 - Create new admin parameter to disabled new account creation

dklawren:
* Remove the unused createemailregex parameter which was originally also used for disabling new account creation.
* Added a new allow_account_creation parameter which is a boolean and defaults to on.
* Now we can disable quickly and re-enable quickly account creation when needed for spammers, etc.
* Update test cases to use the new parameter.

### DIFF
--- a/Bugzilla/Auth.pm
+++ b/Bugzilla/Auth.pm
@@ -490,7 +490,7 @@ Returns:     C<true> if users can log themselves out, C<false> otherwise.
 
 Description: Tells you whether or not users are allowed to manually create
              their own accounts, based on the current login system in use.
-             Note that this doesn't check the C<createemailregexp>
+             Note that this doesn't check the C<allow_account_creation>
              parameter--you have to do that by yourself in your code.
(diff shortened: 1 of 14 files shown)


cgsheeh (approved):
Seems simple enough as an implementation, but what is the purpose of implementing this flag? Why would we ever disable new account creation on BMO? Have we considered that disabling new account creation might be disruptive for legitimate contributors? Without knowing why we need this it's hard to give any feedback about alternative solutions.

dklawren:
> Seems simple enough as an implementation, but what is the purpose of implementing this flag? Why would we ever disable new account creation on BMO? Have we considered that disabling new account creation might be disruptive for legitimate contributors? Without knowing why we need this it's hard to give any feedback about alternative solutions.

It will not be used that often but given the issue that happened yesterday, I was instructed to do this change. Hopefully we do not have to use it in the future. And it would always be temporary of course and we would reactivate after the abuse has subsided.

PR #2197, opened 2024-02-29
Bug 1882734 - Use new CVE website for linking CVE-IDs

maltejur:
The new CVE website from mitre is https://www.cve.org/, replacing https://cve.mitre.org/. From https://cve.mitre.org/:

>    NOTICE: Transition to the all-new CVE website at WWW.CVE.ORG and CVE Record Format JSON are underway.
>
>    NOTICE: Legacy CVE download formats deprecation is now underway and will end on June 30, 2024. New CVE List download format is available now.

As far as I know, new CVEs also have a delay before they appear on https://cve.mitre.org/, while they appear instantly on https://www.cve.org/. The autolinks from BMO for CVE-IDs should probably be updated to that new site.

### DIFF
--- a/extensions/BMO/Extension.pm
+++ b/extensions/BMO/Extension.pm
@@ -600,7 +600,7 @@ sub bug_format_comment {
         my $args  = shift;
         my $match = html_quote($args->{matches}->[0]);
         return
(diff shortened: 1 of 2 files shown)


dklawren (approved):
LGTM. Thanks!

PR #2198, opened 2024-02-29
Bug 1882551 - Create new admin page just for adjusting anti-spam parameters and create special group that can administer it

dklawren:


Bugzilla/App/BMO/AntiSpam.pm
```diff
@@ -0,0 +1,143 @@
...
+    my $item
+      = {name => lc $panel, param_list => \@module_param_list, module => $module,};
+    push @panels, $item;
+  }
+
+  $vars->{panels} = \@panels;
+
+  if ($self->req->method eq 'POST') {
+
+    # Check token data
```
cgsheeh, line 64: Rather than copy this parameter handler code from `editparams.cgi`, could we move some of that code into a function somewhere and re-use it? That would make it much easier to parse what is actually happening here, and avoid duplication of code.

cgsheeh (changes requested):

Bugzilla/Config.pm
```diff
@@ -117,11 +124,26 @@ sub update {
     delete Bugzilla->request_cache->{params};
     delete Bugzilla->request_cache->{params_obj};
     $self->{params} = $params;
+
+    # We want log any param changes in the audit_log table
+    # But Bugzilla::Object->audit_log assumes each instance of
+    # a class has its own unique ID. Such as if Bugzilla::Config 
+    # had a separate instance for each param separatly. This is
```
cgsheeh, line 131:
```suggestion
    # had a separate instance for each param separately. This is 
```

cgsheeh (approved):
LGTM.

PR #2199, opened 2024-02-29
no bug - Remove the need for a proxy container and just access localhost directly

dklawren:
- Much faster page load time without proxy
- Most docker desktops allow accessing container ports on localhost
- No longer need to update Network preferences in browser to use the proxy at localhost:1080
- Updated README to direct people to access http://localhost:8000/ instead of http://bmo.test.


cgsheeh (approved):
This doesn't change how this service runs in Conduit Suite, right? Assuming that is the case this is r+, but if it would impact suite we should fix this there as well.

dklawren:
> This doesn't change how this service runs in Conduit Suite, right? Assuming that is the case this is r+, but if it would impact suite we should fix this there as well.

Correct. conduit-suite has it's own docker-compose configuration and requires a proxy container since you need to be able to access several sites in the cluster. The bmo repo only needs access to a single entry point.

PR #2202, opened 2024-03-14
Bug 1871065 - [DUO] Update BMO to move away from the deprecated iframe Duo prompt to the suggested Duo Universal Prompt

dklawren:
Connor and I will be doing a in-person code review for this due to the size and timing of it but I will add some details here as well.

* Instead of showing the Duo MFA dialog inside an iframe we are moving to a redirection scheme which sends the user to a Duo hosted verification page, and then redirects back to BMO once they complete the verification. 
* BMO will then use a temporary auth code provided by Duo to send a request with a client secret and get back the final status of the users verification. Failing if the user was not successful.
* This flow is very similar to how standard OAuth2 works and uses similarly named endpoints at Duo as OAuth2.
* I have added some test cases to verify that the Duo part inside BMO is working properly. One test case is for functional testing of the DuoClient module. The other is a UI test that mocks a very basic the Duo site and redirects back to BMO.
* The new Duo Universal Prompt uses a client_id and client_secret value instead of the previous duo_ikey, duo_akey, and duo_skey values. The duo_ikey and duo_skey values are interchangeable with the client_id and client_secret respectively.
* The admin params page for Duo has been updated to show client_id and client_secret params.
* DuoAPI and DuoWeb modules removed.


Bugzilla/MFA/Duo.pm
```diff
@@ -65,23 +46,16 @@ sub prompt {
...
-  $template->process('mfa/duo/verify.html.tmpl', $vars)
-    || ThrowTemplateError($template->error());
+  my $username     = $self->property_get('user');
+  my $redirect_uri = $duo->create_auth_url($username, $token);
+
+  print $cgi->redirect($redirect_uri);
 }
 
+# Always return 0 (failure) in case check() for Duo called just to be safe
+# This should now happen.
```
cgsheeh, line 56: Should this be "This should not happen."? Any reason we can't just remove this?

external_test_api.pl
```diff
@@ -69,43 +80,43 @@ sub startup {
...
+  #     $c->render(
+  #       json => {
+  #         email          => 'oauth2-user@example.com',
+  #         name           => 'OAuth2 Test User',
+  #         email_verified => 1,
+  #       },
+  #       status => 200
+  #     );
+  #   }
+  # );
```
cgsheeh, line 119: Is commenting this out intentional?

Bugzilla/DuoClient.pm
```diff
@@ -0,0 +1,282 @@
...
+  $uri->host($self->host);
+  $uri->path($path);
+
+  # If we are running tests, then use http
+  if ($self->host =~ /(externalapi|localhost)/) {
+    $uri->scheme('http');
+  }
+  else {
+    $uri->scheme('https');
+  }
```
cgsheeh, line 90: I've noticed we are checking for `localhost` or `externalapi` as the hostname in various places and making changes to parameters based on that. Can we somehow remove this "test environment only" condition and instead have parameters that we can change on a per-environment basis?

cgsheeh (changes requested):

Bugzilla/App/MFA/Duo.pm
```diff
@@ -0,0 +1,90 @@
...
+
+  # Get authorization code to trade for 2FA
+  my $duo_code = $self->param('duo_code');
+
+  # Also grab the mfa cookie to compare with the state from Duo
+  my $mfa_cookie = $self->cookie('mfa_verification_token');
+
+  if (!$state || !$duo_code || !$mfa_cookie || $mfa_cookie ne $state) {
+    return $self->code_error('duo_client_error', {reason => ERR_BAD_CALLBACK_INFO});
+  }
```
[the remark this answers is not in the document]
dklawren: Ok will do.

Bugzilla/DuoClient.pm
```diff
@@ -0,0 +1,282 @@
...
+
+use Mojo::JWT;
+use Mojo::URL;
+use Try::Tiny;
+use Types::Standard -types;
+
+use constant JTI_LENGTH              => 36;
+use constant MINIMUM_STATE_LENGTH    => 22;
+use constant MAXIMUM_STATE_LENGTH    => 1024;
+use constant FIVE_MINUTES_IN_SECONDS => 300;
```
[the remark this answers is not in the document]
dklawren: Will do

Bugzilla/MFA/Duo.pm
dklawren: I will look into it as Bugzilla::MFA expects the downstream class to implement a check() method but I can also make it first make sure the method exists. And if not then return 0 by default.

external_test_api.pl
dklawren: Yeah. I am not sure how to make this work quite yet as we have builtin code in BMO that allows BMO to authenticate against an external OAuth2 provider such as Github, etc. And the test case for that code expects externalapi.test to mock the OAuth2 endpoints. But the Duo Security API also uses the same endpoint naming and so I made the Duo Security testing take priority over the older OAuth2 authentication tests which we not currently use in production for anything. The other alternative would be to add config parameters to make the endpoints configurable for both OAuth2 and Duo Security so that they can be changed just when running the tests. I can do that now or as a later fix.

Bugzilla/DuoClient.pm
dklawren: This is because when I run the docker cluster locally to use BMO myself on my workstation, the duo host needs to be different if we are redirecting the user to the Duo (mocked) site or if the DuoClient code is calling Duo on the backend to exchange the auth code for a token. In production, the external host will be the same in both cases.  I can do what you suggest and pass in an environment variable when running under the test environment and look for that instead of matching on hostname. Or if you think of a better way.

Bugzilla/MFA/Duo.pm
cgsheeh: Why does the downstream class provide the check? What is it doing that we can always return false here knowing that would cause a failed check?

Bugzilla/DuoClient.pm
```diff
@@ -0,0 +1,264 @@
...
+
+use Mojo::JWT;
+use Mojo::URL;
+use Try::Tiny;
+use Types::Standard -types;
+
+use constant JTI_LENGTH             => 36;
+use constant MINIMUM_STATE_LENGTH   => 22;
+use constant MAXIMUM_STATE_LENGTH   => 1024;
+use constant JWT_DEFAULT_EXPIRATION => 300;
```
cgsheeh, line 27: Nit: should probably add `_SECONDS` so it's obvious what the units are here. I realize that this name was my own suggestion, sorry about that. :sweat_smile:

Bugzilla/DuoClient.pm
```diff
@@ -0,0 +1,264 @@
...
+#    Private Methods    #
+#########################
+
+# Standard JWT args to include with all JWT tokens
+sub _create_jwt_args {
+  my ($self, $endpoint) = @_;
+
+  my $jwt_args = {
+    'iss' => $self->client_id,
+    'sub' => $self->client_id,
```
cgsheeh, line 68: Question: The `sub` claim usually refers to `subject`, ie the user. Is passing the client ID here correct?

Bugzilla/DuoClient.pm
```diff
@@ -0,0 +1,264 @@
...
+    'redirect_uri'           => $self->redirect_uri,
+    'client_id'              => $self->client_id,
+    'iss'                    => $self->client_id,
+    'aud'                    => $self->uri,
+    'exp'                    => time() + JWT_DEFAULT_EXPIRATION,
+    'state'                  => $state,
+    'response_type'          => 'code',
+    'duo_uname'              => $username,
+    'use_duo_code_attribute' => 1
+  };
```
cgsheeh, line 162: Could we re-use `_create_jwt_args` here?

token.cgi
```diff
@@ -509,6 +509,12 @@ sub mfa_event_from_token {
 
   # verify
   my $event = $user->mfa_provider->verify_token($token);
+
+  # If we got this far and MFA is Duo, we should be verified
+  if ($user->mfa eq 'Duo' && !$event->{duo_verified}) {
+    ThrowUserError('duo_user_error', {reason => 'Invalid Duo Security MFA Code'});
+  }
+
```
cgsheeh, line 517: Let's make a function for this and re-use it in `userprefs.cgi`.

cgsheeh (changes requested):
A few nits, but doing "request changes" just to ensure the `sub` claim is correct.

Bugzilla/DuoClient.pm
dklawren: Originally I was just going by https://github.com/duosecurity/duo_universal_python/blob/main/duo_universal/client.py#L120-L129 and what the Duo folks did in their example Python modules.

Also according to the Duo docs at https://duo.com/docs/oauthapi#json-web-tokens, it states that the 'iss' value should be 'The Client ID (or Integration key) from the application's page in the Duo Admin Panel'. And that the 'sub' value is described as `This should match the above iss value'.

token.cgi
dklawren: I am not sure that creating a function for this will gain much as it would be a very small check and is only in two places currently. Would the utility function throw the error or would it just return true or false. Would it do the $mfa eq 'Duo' check or would we do that before calling it. Just not sure it is worth the code line savings yet.

cgsheeh: The whole `if` block should be in one function like `assert_duo_verified($user->mfa, $event)`, since we're checking the exact same condition and throwing the exact same error in both places. It's mostly a maintainability suggestion, so if you're fine with maintaining the same logic in multiple places, it's no problem.

cgsheeh (approved):
Very nice!

PR #2204, opened 2024-03-19
Bug 1834454 - Webhooks should fire if bug removed from matching product

justdave:
See [bug 1834454](https://bugzilla.mozilla.org/show_bug.cgi?id=1834454)

Caveat: I have no way to test this.

### DIFF
--- a/extensions/Push/lib/Connector/Webhook.pm
+++ b/extensions/Push/lib/Connector/Webhook.pm
@@ -101,7 +101,7 @@ sub should_send {
     }
     if ($removed_product || $removed_component) {
       if ($removed_product eq '') {
-        $removed_product eq $bug->product;
+        $removed_product eq $bug_data->{'product'};
       }
       if ($product eq $removed_product
           && ($component eq $removed_component || $component eq 'any'))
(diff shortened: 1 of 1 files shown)


extensions/Push/lib/Connector/Webhook.pm
```diff
@@ -84,6 +84,32 @@ sub should_send {
...
+          $removed_product = $change->{'removed'};
+        }
+        if ($change->{'field'} eq 'component') {
+          $removed_component = $change->{'removed'};
+        }
+      }
+    }
+    if ($removed_product || $removed_component) {
+      if ($removed_product eq '') {
+        $removed_product eq $bug_data->{'product'};
```
dklawren, line 103: $removed_product = $bug_data->{'product'};

dklawren (changes requested):
Looks good except for one change.

extensions/Push/lib/Connector/Webhook.pm
justdave: Bwahahaha, I can't believe I did that.

justdave: Ok, fixed.

dklawren (approved):

PR #2207, opened 2024-03-19
Bug 1886289 - Never attempt to send mail to nobody@mozilla.org

dklawren:
Similar to addresses @.(bugs|tld) we should also not send email to the [nobody@mozilla.org](mailto:nobody@mozilla.org) (default assignee, etc.) address as it is only a placeholder and isn't meant to be delivered to.

### DIFF
--- a/Bugzilla/Mailer.pm
+++ b/Bugzilla/Mailer.pm
@@ -154,8 +154,10 @@ sub MessageToMTA {
       WARN("Attempted to send email to address in badhosts: $to");
       $email->header_set(to => '');
     }
-    elsif ($recipient->host =~ /\.(?:bugs|tld)$/) {
-      WARN("Attempted to send email to fake address: $to");
+    elsif ($recipient->host =~ /\.(?:bugs|tld)$/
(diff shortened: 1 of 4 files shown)


Bugzilla/Mailer.pm
```diff
@@ -154,8 +154,10 @@ sub MessageToMTA {
       WARN("Attempted to send email to address in badhosts: $to");
       $email->header_set(to => '');
     }
-    elsif ($recipient->host =~ /\.(?:bugs|tld)$/) {
-      WARN("Attempted to send email to fake address: $to");
+    elsif ($recipient->host =~ /\.(?:bugs|tld)$/
+      || $recipient->address eq 'nobody@mozilla.org')
```
cgsheeh, line 158: Can you put this logic into a separate function and then use it from both locations? Maybe `is_fake_email(email_address)` or something like that.

cgsheeh (changes requested):

cgsheeh (approved):
LGTM! We could test the new function too, as a follow up.

PR #2216, opened 2024-03-27
Bug 1861776: improve usability of the uplift request form comment

cgsheeh:
Add an array which contains the correct order of uplift request form
fields to ensure they are always presented in the correct order in
the comment. Add the repo's short name to the header so it is more
obvious which request the form corresponds to. Use a smaller header
size to more closely align with the existing Bugzilla uplift request
form.

### DIFF
--- a/extensions/PhabBugz/lib/Feed.pm
+++ b/extensions/PhabBugz/lib/Feed.pm
@@ -378,11 +378,25 @@ sub readable_answer {
 }
 
 sub format_uplift_request_as_markdown {
-  my ($question_answers_mapping) = @_;
+  my ($repo_short_name, $question_answers_mapping) = @_;
 
-  my $comment = "# Uplift Approval Request\n";
+  # Form content will come across a JSON object. Ensure the question/response pairs
+  # are added to the markdown output in the correct order.
+  my @uplift_questions_order = (
+    "User impact if declined",
+    "Code covered by automated testing",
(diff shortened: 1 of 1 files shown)


extensions/PhabBugz/lib/Feed.pm
```diff
@@ -378,11 +378,25 @@ sub readable_answer {
...
+    "User impact if declined",
+    "Code covered by automated testing",
+    "Fix verified in Nightly",
+    "Needs manual QE test",
+    "Steps to reproduce for manual QE testing",
+    "Risk associated with taking this patch",
+    "Explanation of risk level",
+    "String changes made/needed",
+    "Is Android affected?",
+  );
```
dklawren, line 395: Only issue I can see if additional fields are added in Phabricator, they will not automatically show up in the Bugzilla comment due to the hard coded list. We would just need to make sure to update this code as well if additional questions are ever added in Phabricator.

dklawren (approved):

extensions/PhabBugz/lib/Feed.pm
cgsheeh: Ideally we could have the output come across as an ordered list of `(question, answer)` pairs or something along those lines, but I don't want to muck about with the Phabricator code and running Phab in suite is broken at the moment. For now this is a quick-and-dirty solution. :)

PR #2224, opened 2024-03-30
Bug 1887858 - What needs my attention: remove duplicate needinfos

dklawren:
This pull request uses caches to make sure that a bug is only displayed once per page load of the attention report. The local cache is for removing duplicates for a single section and the global cache is for all sections of the page. The way the cache works is the duplicates will be removed for the lower sections which are the least important and display them in the upper more critical sections.

### DIFF
--- a/extensions/BMO/lib/Reports/Attention.pm
+++ b/extensions/BMO/lib/Reports/Attention.pm
@@ -52,12 +52,10 @@ sub format_bug_list {
   my $datetime_now = DateTime->now(time_zone => $user->timezone);
 
   my @formatted_bugs;
-  my %local_seen;
   foreach my $row (@{$bugs}) {
 
-    # Skip if we have already seen this bug locally or globally
+    # Skip if we have already seen this bug globally
     my $bug_id = $row->{bug_id};
-    next if $local_seen{$bug_id};
     next if $global_seen->{$bug_id};
 
     my $bug = {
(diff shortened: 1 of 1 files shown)


extensions/BMO/lib/Reports/Attention.pm
```diff
@@ -48,24 +48,37 @@ sub get_bug_list {
...
+
     my $datetime = datetime_from($bug->{changeddate});
     $datetime->set_time_zone($user->timezone);
     $bug->{changeddate}       = $datetime->strftime('%Y-%m-%d %T %Z');
     $bug->{changeddate_epoch} = $datetime->epoch;
     $bug->{changeddate_fancy} = time_ago($datetime, $datetime_now);
+
+    # We only want to see a bug id once per page load so mark them seen
+    $local_seen{$bug_id} = 1;
+    $global_seen->{$bug_id} = 1;
```
cgsheeh, line 80: Since we always set both `local_seen` and `global_seen` at the same time, and then check both caches at the start of the loop over the bugs at the same time, isn't `local_seen` unnecessary? When is `local_seen` ever used in a condition where `global_seen` isn't used?

cgsheeh (changes requested):

extensions/BMO/lib/Reports/Attention.pm
dklawren: Valid point. They are actually performing the same task now that you point it out. I will remove the local_seen part.

cgsheeh (approved):

PR #2228, opened 2024-04-09
Bug 1888810 - Hide editing option of user profiles for users without corresponding authorization

dklawren:
Since the values are now considered strings instead of boolean values from vcard.dataset, we use strict value checking to see if the value is "0' or "1".

### DIFF
--- a/extensions/BMO/web/js/edituser_menu.js
+++ b/extensions/BMO/web/js/edituser_menu.js
@@ -47,7 +47,7 @@ function show_usermenu(vcard) {
             }
         });
     }
-    if (hide_profile == 0) {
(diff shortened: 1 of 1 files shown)


dklawren:
> Could we map these to `true`/`false` somehow instead?

Yeah. One of my iterations was doing something similar in the beginning of the function:
```javascript
const hide_profile_bool = (hide_profile === "1" ? true : false);
const show_edit_bool = (show_edit === "1" ? true : false);
```

but then I thought 1) the checks only happen in one place per value and 2) the change is larger than just doing the === in place where it is used. So went the simpler route.

PR #2232, opened 2024-04-18
Bug 1884724 - Standard bug entry form  for non-canconfirm users is missing fields that are available from the guided form

dklawren:
When a user who does not have canconfirm, normally they see the guided bug entry form by default. The guided form has a drop down for selecting product version. If they click the link at the bottom to switch over to the standard bug form, they are unable to select a version as it is hidden.

This PR changes the standard bug form to no longer consider the version drop down to be an *advanced* field and also make it required. Users without canconfirm cannot show advanced field so they always see a very simple version of the standard bug entry form.

### DIFF
--- a/Bugzilla/Bug.pm
+++ b/Bugzilla/Bug.pm
@@ -5138,6 +5138,7 @@ sub check_can_change_field {
     if (
          $field eq 'short_desc'
       || $field eq 'component'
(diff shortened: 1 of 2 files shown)


cgsheeh (approved):
Makes sense to me.

PR #2234, opened 2024-04-19
Bug 1891527 - In Bugmail & Bug History for User Story edits, there's a warning about 'No newline at end of file' for user-entered values, even though the submitted value included a trailing newline

dklawren:
I updated the code to add back the stripped newlines from the user story value before performing the diff if the newlines are missing. This fixes the issue with the warning being added to the diff output.

### DIFF
--- a/extensions/UserStory/Extension.pm
+++ b/extensions/UserStory/Extension.pm
@@ -68,11 +68,15 @@ sub bug_update_before_logging {
   my ($self, $args) = @_;
   my $changes = $args->{changes};
   return unless exists $changes->{cf_user_story};
-  my $diff = diff(
-    \$changes->{cf_user_story}->[0],
-    \$changes->{cf_user_story}->[1],
-    {CONTEXT => 0,},
-  );
+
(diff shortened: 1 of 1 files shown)


extensions/UserStory/Extension.pm
```diff
@@ -68,11 +68,15 @@ sub bug_update_before_logging {
...
   my $changes = $args->{changes};
   return unless exists $changes->{cf_user_story};
-  my $diff = diff(
-    \$changes->{cf_user_story}->[0],
-    \$changes->{cf_user_story}->[1],
-    {CONTEXT => 0,},
-  );
+
+  my $old_value = $changes->{cf_user_story}->[0];
+  my $new_value = $changes->{cf_user_story}->[1];
```
cgsheeh, line 73: If the newline is stripped by the time it reaches this variable in Perl, could the issue be that the JS on the front-end is stripping the newline? Fixing it there is probably the more correct solution, but this works too and the bug probably doesn't warrant much more of a time investment. :)

cgsheeh (approved):

PR #2236, opened 2024-04-23
Bug 1892972 - In some cases when obtaining the ldap_email for a user over API, a recovery code will be returned instead of the ldap email address

dklawren:


dklawren:
Moving forward r+ from bug 1892972.

PR #2241, opened 2024-05-02
Bug 1893924 - Tabular reports return blank tables

dklawren:
Second try at fixing this the correct way. 

What was happening is when a single table of the report was being rendered, report.cgi was setting the table name to " " which the JS changes didn't like. So I now set the table name to "single" and look for that instead of " ". This fixes one issue occurring. On the same note there was also an issue with table identifiers have spaces in the names which I had to remove otherwise the tables were misrendered.

Another issue was that linkify and linkifyTotal formatters were not getting the column name value properly and only the row name and count. By passing in $column element (a <td>) to the formatters I am able to access the column and name and remove the undefined values showing up in the links to buglist.cgi.

And lastly, a third issue was when displaying multiple tables on a single report. The tables had data appearing in the wrong places when the JS was rendering them since each table was not distinct and was using the same "report_table" identifier. Changing to "report_table_[% tbl_id FILTER html %]" made each one distinct and all were rendered properly.

### DIFF
--- a/template/en/default/reports/report-table.html.tmpl
+++ b/template/en/default/reports/report-table.html.tmpl
@@ -37,6 +37,9 @@
 [% col_field_disp = field_descs.$col_field || col_field %]
 [% row_field_disp = field_descs.$row_field || row_field %]
 
+[%# Class and ID names need to have whitespace removed. %]
+[% tbl_id = tbl FILTER css_class_quote %]
+
 [% urlbase = BLOCK %][% basepath FILTER none %]buglist.cgi?[% buglistbase FILTER html %][% END %]
 [% IF tbl == "-total-" %]
   [% IF tbl_vals %]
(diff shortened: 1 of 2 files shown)


template/en/default/reports/report-table.html.tmpl
```diff
@@ -47,34 +50,33 @@
...
+  str = str.replace(/&lt;/g, "<").replace(/&gt;/g, ">").replace(/&quot;/g, '"')
+           .replace(/&nbsp;/g, " ").replace(/&amp;/g, "&").replace(/\s+$/,"");
 
   // encodeURIComponent() doesn't escape single quotes.
   return encodeURIComponent(str).replace(/'/g, escape);
 };
 
 window.addEventListener('DOMContentLoaded', () => {
-  const linkify = ({ value, data }) => {
+  const linkify = ({ $column, value, data }) => {
```
cgsheeh, line 67: Nit: I think this can just be `column` without the `$`, right?

cgsheeh (approved):

template/en/default/reports/report-table.html.tmpl
dklawren: > Nit: I think this can just be `column` without the `$`, right?

You are correct. Technically $column and column are no different to Javascript but in style of JS used in Bugzlla, variables with $ are used to denote a DOM element so I used it to be consistent.

PR #2243, opened 2024-05-07
Bug 1895526 - Create Bugzilla/App/Controller directory where modules are auto loaded instead of being hard coded in the App.pm module

dklawren:
Currently in Bugzilla/App.pm whenever a new Mojo controller module is added to add new functionality, we have to include the loading of the specific module and call its setup_routes() function. This bug is to create a new Bugzilla/App/Controller directory that we can place new controller modules and they will be loaded automatically instead of being hard coded. The change will create the code to auto load the modules as well as move the existing modules to the new location.

### DIFF
--- a/Bugzilla/App.pm
+++ b/Bugzilla/App.pm
@@ -201,16 +201,11 @@ sub setup_routes {
   # Load controller modules and their routes
   foreach my $module (find_modules('Bugzilla::App::Controller', {recursive => 1}))
   {
-    try {
-      require_module($module);
-      my $controller = $module->new;
-      if ($controller->can('setup_routes')) {
-        $controller->setup_routes($r);
-      }
+    require_module($module);
+    my $controller = $module->new;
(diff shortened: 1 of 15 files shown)


Bugzilla/App.pm
```diff
@@ -205,17 +196,17 @@ sub startup {
...
-  Bugzilla::App::MFA::Duo->setup_routes($r);
+
+  # Load controller modules and their routes
+  foreach my $module (find_modules('Bugzilla::App::Controller', {recursive => 1}))
+  {
+    require_module($module);
+    my $controller = $module->new;
+    if ($controller->can('setup_routes')) {
+      $controller->setup_routes($r);
+    }
```
cgsheeh, line 208: Should the absence of a `setup_routes` function be an error? Or maybe we should log it for debugging purposes etc?

cgsheeh (approved):

PR #2247, opened 2024-05-14
Bug 1896389 - Uplift approval granted by non-relman revision review

dklawren:


dklawren:
Actually this is incorrect. We should only be checking group membership for approval + and - flags and not for ?, etc. I will update this.

dklawren:
Ready for review

extensions/PhabBugz/lib/Util.pm
```diff
@@ -80,6 +80,22 @@ sub set_attachment_approval_flags {
       . $approval_flag_name
       . $status);
 
+  # If we are setting to + or -, we need to make sure
+  # Phabricator user is a member of the release-managers project
+  my $is_release_manager = 0;
+  my $release_manager_project
+    = Bugzilla::Extension::PhabBugz::Project->new_from_query(
+    {name => 'release-managers'});
+  if ($release_manager_project) {
```
cgsheeh, line 89: Nit: maybe if we don't find the `#release-managers` group that should be a fatal error? Seems we might not find it and then we will continue on.

cgsheeh, line 89: Nit: this section of code to check if a user is a release manager could be a function.

cgsheeh (approved):

extensions/PhabBugz/lib/Util.pm
dklawren: In that case $is_release_manager would still be 0. This would block setting of + or - for approval flags later so would just never set those statuses.

dklawren: I will add a method to the Phab User class that checks for release manager.

dklawren:
Maybe should do a quick re-review since I moved the code to a dedicated method.

cgsheeh (approved):
Thanks for fixing this so quickly!

PR #2254, opened 2024-06-07
Bug 1900932 - Create API for adding/removing/listing current bug reminders based on work from bug 1894968

dklawren:


Bugzilla/API/V1/Reminders.pm
```diff
@@ -0,0 +1,117 @@
...
+  my ($self, $reminder) = @_;
+  return {
+    id          => $reminder->id,
+    bug_id      => $reminder->bug_id,
+    note        => $reminder->note,
+    reminder_ts => $reminder->reminder_ts,
+    creation_ts => $reminder->creation_ts,
+    sent        => ($reminder->sent ? true : false),
+  };
+}
```
cgsheeh, line 115: I think this would be better as a method on the `Reminder` class, like `$reminder->to_hash()`.

Bugzilla/API/V1/Reminders.pm
```diff
@@ -0,0 +1,117 @@
...
+sub add {
+  my ($self) = @_;
+  Bugzilla->usage_mode(USAGE_MODE_MOJO_REST);
+  my $user = $self->bugzilla->login;
+  $user->id || return $self->user_error('login_required');
+
+  # Do not allow adding of reminders if feature is not enabled
+  # or user does not have permission to do so
+  return $self->render(json => {})
+    unless Bugzilla->params->{reminders_enabled}
```
cgsheeh, line 61: Could we avoid setting up the routes in `setup_routes` when reminders are disabled instead? That seems like a more thorough way to gate the feature behind a flag.

Bugzilla/API/V1/Reminders.pm
```diff
@@ -0,0 +1,117 @@
...
+  my $user = $self->bugzilla->login;
+  $user->id || return $self->user_error('login_required');
+
+  # You can only delete your own so pass in user_id as well
+  my $reminder
+    = Bugzilla::Reminder->new({user_id => $user->id, id => $self->param('id')});
+
+  my $success = 0;
+  if ($reminder) {
+    $reminder->remove_from_db();
```
cgsheeh, line 98: I'm surprised `remove_from_db()` doesn't return something to indicate success. Also, if the removal fails, what is the behaviour? Should we add retries or error handling?

cgsheeh (changes requested):
Looks great, just a few questions on the PR.

cgsheeh (approved):

PR #2262, opened 2024-06-17
Bug 1903078: remove extra `"` in `Update Reminders` link

cgsheeh:


template/en/default/email/reminder.html.tmpl
```diff
@@ -39,7 +39,7 @@
     </p>
     <p>
       Thank you for using [%terms.Bug %] Reminders!<br>
-      <a href=[% urlbase FILTER none %]userprefs.cgi?tab=reminders">Update Reminders</a>
+      <a href=[% urlbase FILTER none %]userprefs.cgi?tab=reminders>Update Reminders</a>
```
dklawren, line 42: Sorry for the typo. But I think the fix would be to add a " and not remove. So let's make it:

`<a href="[% urlbase FILTER none %]userprefs.cgi?tab=reminders">Update Reminders</a>`

dklawren (changes requested):

dklawren (approved):

PR #2264, opened 2024-06-21
Bug 1903334 - Convert from using graphviz (webdot) to Mermaid.js for rendering dependency graphs in showdependencygraph.cgi

dklawren:
I feel like this ready for review now.

- Remove any reference to the webdot function of BMO. This is the old method of rendering graph png files onto the local file system and then serving them to the client.
- Add the latest mermaid.js lib to the BMO instance locally
- Update showdependencygraph.cgi to generate properly formatted mermaid data into the pre block in the template.
- mermaid.js will then render the graph based on the text data on the client side. 
- We no longer need to store temporary files on the web server file system using this new method as everything is rendered on the fly. 
- Added mermaid_quote() and truncate_string() utility functions to Util.pm for use by showdependencygraph.cgi and Template.pm
- Updated the graph form and page to remove unused fields and moved the form to the top.
- This is all installed on bugzilla-dev if you need to see it working.
- Example: https://bugzilla-dev.allizom.org/showdependencygraph.cgi?id=1852288&showsummary=on&display=web&rankdir=TB


showdependencygraph.cgi
```diff
@@ -167,166 +176,52 @@ else {
...
-$vars->{'multiple_bugs'} = ($cgi->param('id') =~ /[ ,]/);
-$vars->{'display'}       = $display;
-$vars->{'rankdir'}       = $rankdir;
-$vars->{'showsummary'}   = $cgi->param('showsummary');
+my @bugs = grep { detaint_natural($_) } split /[\s,]+/, $cgi->param('id');
+$vars->{'bug_id'}      = join ', ', @bugs;
+$vars->{'display'}     = $display;
+$vars->{'rankdir'}     = $rankdir;
+$vars->{'showsummary'} = $cgi->param('showsummary');
+$vars->{'debug'}       = ($cgi->param('debug') ? 1 : 0);
```
cgsheeh, line 222: Do we still need `debug`?

cgsheeh (approved):

showdependencygraph.cgi
dklawren: Yeah I think so. I feel like there may be more corner cases later on that some characters, etc. in bug summaries will break the graph markup text and will need to be debugged. Having the debug=1 available, I can just ask people to paste the graph text to a bug report and I can then debug it locally.

PR #2265, opened 2024-06-21
Bug 1904151 - Updates and improvements to Bugzilla/Report/Graph.pm and Bugzilla/API/V1/BugGraph.pm

dklawren:
- Use breadth first instread of depth first for generating the JSON tree for /rest/bug/123/graph
  - https://metacpan.org/pod/Graph::Traversal::BFS
- Move the code that filters secure bugs and resolved bug to Bugzilla/Report/Graph.pm
  - This will allow later for showdependencytree.cgi and showdependencygraph.cgi to switch over to using Bugzilla/Report/Graph instead of each generating the trees themselves. This will remove duplication.
  - Will file separate bugs for the work in above

### DIFF
--- a/Bugzilla/API/V1/BugGraph.pm
+++ b/Bugzilla/API/V1/BugGraph.pm
@@ -74,14 +74,10 @@ sub graph {
       );
 
       # Remove any secure bugs that user cannot see
-      $report->prune_graph(sub { $user->visible_bugs($_[0]) });
+      $report->prune_secure;
 
-      # If we do not want resolved bugs (default) then filter those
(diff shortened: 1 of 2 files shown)


cgsheeh (approved):
Seems reasonable, what's the motivation behind the switch to BFS?

dklawren:
> Seems reasonable, what's the motivation behind the switch to BFS?

I mistakenly was using depth-first which meant the tree was being returned in the wrong order. breadth-first starts from the root bug and id and the order is more correct and matches the showdependencytree.cgi output.

PR #2266, opened 2024-06-26
Bug 1904753: return immediately when approval flag should not be changed

cgsheeh:
In `set_attachment_approval_flags` we check for several conditions
that indicate the flag cannot be updated by the user, but we do not
immediately `return` when these conditions are hit. This causes the
code to attempt to create the flag once it exits the loop, which
is not allowed since approval flags are not multiplicable. Add `return`
statements to branches which indicate the flag should not be updated.

Since we are now returning early from these branches, we can reverse the
logic of the main `if` branch to clean up the loop by de-denting the code.


dklawren (approved):
Thanks for the fix.

PR #2269, opened 2024-06-28
Bug 1905292 - For comments tagged as spam, remove the content

dklawren:
This needed to be added to the older legacy comment tag API endpoint as well as the one that is used by the newer Bug Modal show bug UI.


globau (changes requested):
With this change if a comment is accidentally tagged as spam its contents will immediately be removed, without any feedback to the user that anything has happened - they won't notice the content is gone until the next time they reload the page.

The content removal should happen after a delay of at least 5 minutes.

Additionally, as the goal of this bug is to hide spam content from search engines, viewing a comment's history _if the comment is tagged as spam_ should require authentication.

dklawren:
Closing this pull request as per https://bugzilla.mozilla.org/show_bug.cgi?id=1905292#c3

PR #2283, opened 2024-07-25
Bug 1909798 - Large dependency graphs do not work

dklawren:
I added quotes around the bug id and summaries which should with future special characters that Mermaid treats as special.

### DIFF
--- a/showdependencygraph.cgi
+++ b/showdependencygraph.cgi
@@ -76,7 +76,7 @@ my $add_link = sub {
 
       $link_text .= $dependson;
       $link_text .= ($dependson_status eq 'RESOLVED') ? '[' : '([';
-      $link_text .= "$dependson";
+      $link_text .= '"' . $dependson;
 
       if ($show_summary && $user->can_see_bug($dependson)) {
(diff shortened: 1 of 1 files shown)


showdependencygraph.cgi
```diff
@@ -76,16 +76,16 @@ my $add_link = sub {
 
       $link_text .= $dependson;
       $link_text .= ($dependson_status eq 'RESOLVED') ? '[' : '([';
-      $link_text .= "$dependson";
+      $link_text .= '"' . $dependson;
```
cgsheeh, line 79: This seems like it's likely to break in the future... could you build up the variable to be quoted in a separate variable and then just `$link_text .= '"$dependson_text"';` instead?

cgsheeh (changes requested):

dklawren:
Admittedly the add_link function is getting very complex and in the future I would like to break it apart into smaller manageable chunks.

I have pushed the latest fixes in this PR to bugzilla-dev for you to try out. Here is the original bug causing the breakage in its full, practically unreadable glory. Hopefully most dependency trees will not be this large.

cgsheeh (approved):

PR #2284, opened 2024-07-26
Bug 1909777 - Require new users to set their username during the GitHub account creation flow

dklawren:
This change adds an input field to the github (and other oauth2) login flow that allows a user to set a Real Name when the new account is created.


cgsheeh (approved):
Do you have a screenshot of what this looks like?

dklawren:
Yep. Sorry didn't add them before

This is when a user has more than one email associated with their Github account which I do. If they only have one, then this screen is skipped entirely:

Screenshot 2024-07-26 at 12-54-33 Choose Account

This is the screen they see after choosing an email:

Screenshot 2024-07-26 at 12-54-48 Verify Account Creation

The user then enters their desired realname and click Verify.

PR #2287, opened 2024-07-31
Bug 1910469 - Update sanitizeme script to truncate bugs_fulltext table to significantly reduce size of dump file

dklawren:


cgsheeh (approved):
How much smaller is the dump file without the fulltext table?

dklawren:
> How much smaller is the dump file without the fulltext table?

30G => 17G

PR #2288, opened 2024-07-31
Bug 1910693 - Update sanitizeme script to remove more sensitive user data from dump file

dklawren:
This pull request removes additional user specific data from the DB when doing a sanitized dump.

Tables cleared:
component_watch
anstispam_*
reminders
whine_*
watch
series_*
email_rates
email_bug_ignore
secbugs_*

Also the following:
Clears MFA values from the profiles table
Clears user public gpg keys from the profiles table

### DIFF
--- a/scripts/sanitizeme.pl
+++ b/scripts/sanitizeme.pl
@@ -253,6 +253,34 @@ sub delete_sensitive_user_data {
   $dbh->do("DELETE FROM ts_funcmap");
   $dbh->do("DELETE FROM ts_job");
   $dbh->do("DELETE FROM ts_note");
+
+  # Reminders
+  $dbh->do('TRUNCATE TABLE reminders');
+
+  # Whine events
+  $dbh->do('DELETE FROM whine_events');
+
+  # Watched users by other users
+  $dbh->do('TRUNCATE TABLE watch');
+
+  # Series data (charts)
+  $dbh->do('DELETE FROM series_categories');
+
+  # Email rate limiting
+  $dbh->do('TRUNCATE TABLE email_rates');
+
+  # Clear MFA for all users
+  $dbh->do("UPDATE profiles SET mfa = ''");
+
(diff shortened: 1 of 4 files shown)


scripts/sanitizeme.pl
```diff
@@ -253,6 +253,34 @@ sub delete_sensitive_user_data {
...
   $dbh->do("DELETE FROM ts_note");
+
+  # Reminders
+  $dbh->do('TRUNCATE TABLE reminders');
+
+  # Whine events
+  $dbh->do('DELETE FROM whine_events');
+
+  # Watched users by other users
+  $dbh->do('TRUNCATE TABLE watch');
```
cgsheeh, line 264: Any reason for `DELETE FROM` in some places but `TRUNCATE TABLE` in others?

cgsheeh (approved):

scripts/sanitizeme.pl
dklawren: The DELETE FROM  is required when the table you are deleting values for have foreign key relationships to that table and the other tables have ON DELETE CASCADE setup. So when we do the DELETE FROM it automatically deletes the rows in the other tables. If nothing else links to the rows in the table, you can just do TRUNCATE TABLE. TRUNCATE TABLE is preferable when you can because it resets the auto increment counters back to zero.

dklawren: > The DELETE FROM is required when the table you are deleting values for have foreign key relationships to that table and the other tables have ON DELETE CASCADE setup. So when we do the DELETE FROM it automatically deletes the rows in the other tables. If nothing else links to the rows in the table, you can just do TRUNCATE TABLE. TRUNCATE TABLE is preferable when you can because it resets the auto increment counters back to zero.

Actually, it will error as well if you try to TRUNCATE TABLE and their are rows that are referenced to by other tables.

PR #2289, opened 2024-07-31
Bug 1905296 - Clear needinfo if comment is marked as spam

dklawren:
- Create a antispam_comment_cleanup table for recording comment ids
- Hook into comment tag hooks to add comment ids to the table when spam is added and remove if tag removed
- Create a cron script to run every 30 mins to process the comment ids in the cleanup queue.
- Currently the script will just remove a needinfo that was added at the same time as a spam comment
- We can add other cleanup changes to the script in future pull requests.

### DIFF
--- a/extensions/AntiSpam/bin/spam_cleanup.pl
+++ b/extensions/AntiSpam/bin/spam_cleanup.pl
@@ -56,9 +56,10 @@ foreach my $comment_id (@{$comment_ids}) {
       && $flag->modification_date eq $comment->creation_ts;
     push @clear_flags, {id => $flag->id, status => 'X'};
   }
-  $bug->set_flags(\@clear_flags, []) if @clear_flags;
-
-  $bug->update;
+  if (@clear_flags) {
+    $bug->set_flags(\@clear_flags, []);
+    $bug->update;
+  }
 
   # Lastly we need to remove the entry from the cleanup queue
   $dbh->do('DELETE FROM antispam_comment_cleanup WHERE comment_id = ?',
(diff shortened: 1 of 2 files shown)


extensions/AntiSpam/bin/spam_cleanup.pl
```diff
@@ -0,0 +1,66 @@
...
+    next
+      if $flag->type->name ne 'needinfo'
+      && $flag->status ne '?'
+      && $flag->requestee_id != $comment->author->id
+      && $flag->modification_date eq $comment->creation_ts;
+    push @clear_flags, {id => $flag->id, status => 'X'};
+  }
+  $bug->set_flags(\@clear_flags, []) if @clear_flags;
+
+  $bug->update;
```
cgsheeh, line 61: Should we also gate this behind `if @clear_flags`?

cgsheeh (approved):

extensions/AntiSpam/bin/spam_cleanup.pl
```diff
@@ -0,0 +1,66 @@
...
+  my $comment = Bugzilla::Comment->new($comment_id);
+  my $bug     = $comment->bug;
+
+  # Clear a needinfo that was set on a comment tagged as spam
+  my @clear_flags = ();
+  foreach my $flag (@{$bug->flags}) {
+    next
+      if $flag->type->name ne 'needinfo'
+      && $flag->status ne '?'
+      && $flag->requestee_id != $comment->author->id
```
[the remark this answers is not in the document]
dklawren: In Perl, you use `ne` for comparing strings and `!=` for comparing integers. Technically `!=` will work on both cases but Perl will generate a warning if the values are not integers.

extensions/AntiSpam/bin/spam_cleanup.pl
dklawren: Originally I put the `$bug->update` outside as I plan to add more code to this script for other spam cleanup so I didn't see the need to call `$bug->update` multiple times. But for now I will move it inside the condition. If `@clear_flags` is empty then the `$bug->update` would just be a no-op.

PR #2294, opened 2024-08-14
Bug 1911202 - Remove support for iprepd from the BMO code base as it is not being used since the move to GCP

dklawren:


Bugzilla/Auth.pm
```diff
@@ -234,7 +234,6 @@ sub _handle_login_result {
   # to find account names by brute force)
   elsif ($fail_code == AUTH_LOGINFAILED or $fail_code == AUTH_NO_SUCH_USER) {
     my $remaining_attempts = MAX_LOGIN_ATTEMPTS - ($result->{failure_count} || 0);
-    Bugzilla->iprepd_report('bmo.username_password_mismatch', remote_ip());
```
cgsheeh, line 237: No `check_rate_limit` here?

cgsheeh (approved):
LGTM, just want to make sure that one location wasn't overlooked.

Bugzilla/Auth.pm
dklawren: Yeah. Not needed as Bugzilla already has a built in method for locking out a user if they enter the wrong password more than 5  times. They are then not allowed to log in again for 30 mins. It is keyed off of IP address.

PR #2295, opened 2024-08-14
Bug 1912884 - Users who cannot comment on a bug should not be able to change a bug's resolution

dklawren:


dklawren:
Users who are not in editbugs and do not have a role on the bug cannot edit the resolution anyway, same as commenting.

PR #2296, opened 2024-08-16
Bug 1891378 - meta-clicking on attachment name should open the attachment, not the details page

globau:
- aligns behaviour with opening via the context menu
- the details link can be meta-clicked for the current behaviour


kyoshino (approved):
Thanks for fixing the bug! Just tested locally and confirmed it fixed the stated issue.

PR #2302, opened 2024-08-23
Bug 1912884 - Users who cannot comment on a bug should not be able to change a bug's resolution

dklawren:


cgsheeh (approved):
Makes sense to me.

PR #2305, opened 2024-08-29
Bug 1914061 - Sitemap urls are using GCP storage links which are being rejected by Google search indexer since it is not a property we own

dklawren:


cgsheeh (approved):

dklawren:
Thanks. This is just waiting on https://github.com/mozilla-it/webservices-infra/pull/2641 to land.

PR #2306, opened 2024-08-30
Bug 1879792 - Add a cookie banner to BMO

dklawren:
From bug:
We should include a Cookie Banner on BMO to reduce our exposure to legal risks from changes in EU and US privacy law of the last several years.
BMO has a number of non-essential cookies that improve the UX of the site. Even though these non-essential cookies are not used for tracking purposes, Legal have advised that a cookie banner is still warranted.

* This pull request uses JS code written by the web team for use in Mozilla web properties that I have vendored into the codebase since we do not use webpack.
* I updated the BMO JS to also not set keys in localStorage if cookie consent is rejected.
* The changes also look at a special header that is already included by CloudOps which has the region code for the client. If the region is in the whitelist then it will show the banner, otherwise all works as before.
* GoogleAnalytics code is removed completely as we do not use it anymore.
* There is a new input in the user preferences to turn on or off cookie consent even after clicking the banner.

Let me know if you have any questions


Bugzilla/Constants.pm
```diff
@@ -678,6 +682,19 @@ use constant JOB_QUEUE_VIEW_MAX_JOBS => 2500;
...
 
+# Consent cookie name
+use constant CONSENT_COOKIE => 'moz-consent-pref';
+
+# List of essential cookies that cannot be opted out
+use constant ESSENTIAL_COOKIES =>
+  qw(bugzilla Bugzilla_login Bugzilla_logincookie Bugzilla_login_request_cookie
+     bugzilla github_state github_token sudo moz-consent-pref);
+
+# List of countries the require cookie consent
```
cgsheeh, line 693:
```suggestion
# List of countries that require cookie consent
```

userprefs.cgi
```diff
@@ -72,6 +72,10 @@ sub DoAccount {
       }
```
cgsheeh, line 72: For clarification: the changes in this file handle the comments in [this section](https://github.com/mozmeao/consent-banner?tab=readme-ov-file#how-can-visitors-change-their-cookie-settings-after-interacting-with-the-banner) of the consent-banner README, correct? Just want to make sure that was properly addressed.

extensions/BugModal/web/bug_modal.js
```diff
@@ -21,7 +21,7 @@ function slide_module(module, action, fast) {
             'aria-label': is_visible ? latch.data('label-expanded') : latch.data('label-collapsed'),
         });
         if (BUGZILLA.user.settings.remember_collapsed && module.is(':visible'))
-            localStorage.setItem(module.attr('id') + '.visibility', is_visible ? 'show' : 'hide');
+            Bugzilla.Storage.set(module.attr('id') + '.visibility', is_visible ? 'show' : 'hide');
```
cgsheeh, line 24: Question: are there any performance implications to using this Bugzilla storage vs localstorage?

extensions/BugModal/web/bug_modal.js
```diff
@@ -200,7 +200,7 @@ $(function() {
 
     function restoreSavedBugComment() {
         expireSavedComments();
-        let value = JSON.parse(localStorage.getItem(bugCommentCacheKey));
+        let value = Bugzilla.Storage.get(bugCommentCacheKey);
```
cgsheeh, line 203: This line lost the JSON parsing, is that expected?

cgsheeh (changes requested):
Mostly LGTM, just have a few questions and want to make sure that `JSON.parse` removal is expected.

extensions/BugModal/web/bug_modal.js
dklawren: Bugzilla.Storage is just a wrapper around localStorage so it should perform the same as localStorage
https://github.com/mozilla-bteam/bmo/blob/master/js/util.js#L848-L902

extensions/BugModal/web/bug_modal.js
dklawren: Bugzilla.Storage.get() performs the JSON.parse itself so bug_modal.js doesn't need to do so anymore.

userprefs.cgi
dklawren: I made some changes to follow the requirements. Including:

- A separate non-authenticated cookie settings page
- Links to the cookie settings page in easy to find places.

extensions/BugModal/web/bug_modal.js
```diff
@@ -213,10 +213,10 @@ $(function() {
     function expireSavedComments() {
         const AGE_THRESHOLD = 7 * 24 * 60 * 60 * 1000; // 7 days in milliseconds.
         let expiredKeys = [];
-        for (let i = 0; i < localStorage.length; i++) {
-            let key = localStorage.key(i);
+        for (let i = 0; i < window.localStorage.length; i++) {
+            let key = window.localStorage.key(i);
             if (key.match(/^bug-modal-saved-comment-/)) {
-                let value = JSON.parse(localStorage.getItem(key));
+                let value = Bugzilla.Storage.get(key);
```
cgsheeh, line 219: Out-of-scope nit: I don't really love the implicit JSON handling, it would be better to have a `get_json` method that handles JSON values instead.

cgsheeh (approved):

Bugzilla/Config/Admin.pm
```diff
@@ -65,6 +65,12 @@ sub get_param_list {
       type    => 't',
       default => 'https://product-details.mozilla.org/1.0',
     },
+
```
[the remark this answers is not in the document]
dklawren: I did that to separate the different config blocks and to be consistent with the others in this file.

extensions/BugModal/web/bug_modal.js
dklawren: Not sure why this is an issue as the Bugzilla.Storage class is handling the JSON parsing and stringification behind the scenes now. So the caller will use Bugzilla.Storage.get and the value returned will be a valid JSON object or a single value. So the caller will just need to check of value is typeof object if it is not sure. Bugzilla.Storage.set looks to see if the value passed is and object and if so, it will stringify it as JSON before storing.

PR #2316, opened 2024-09-06
Bug 1917348 - Update robots.txt hook for SiteMapIndex to add 'Allow: /sitemap/' to fix blocked sitemap issue

dklawren:
After waiting a while for Google to finally get around to indexing bugzilla-dev.allizom.org to test this fix, it seems like this is now working. Please review.


cgsheeh (approved):
PR title doesn't make sense, you should fix it before landing.

PR #2317, opened 2024-09-07
Bug 1917167 - Disable emoji reactions for users without editbugs on bugs that have commenting restricted

kyoshino:
[Bug 1917167 - Disable emoji reactions for users without editbugs on bugs that have commenting restricted](https://bugzilla.mozilla.org/show_bug.cgi?id=1917167)

Restrict emoji reactions as with bug comments. The reaction button will be hidden from non-editbugs users, and the API usage will also fail.

### DIFF
--- a/template/en/default/global/user-error.html.tmpl
+++ b/template/en/default/global/user-error.html.tmpl
@@ -384,6 +384,10 @@
     [% title = "Comment Reactions Closed" %]
     This [% terms.bug %] no longer accepts comment reactions.
 
+  [% ELSIF error == "comment_reaction_restricted" %]
+    [% title = "Comment Reactions Restricted" %]
+    You are not allowed to react on this [% terms.bug %].
+
   [% ELSIF error == "comment_tag_disabled" %]
     [% title = "Comment Tagging Disabled" %]
     The comment tagging is not enabled.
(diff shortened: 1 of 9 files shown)


template/en/default/global/user-error.html.tmpl
```diff
@@ -384,6 +384,10 @@
     [% title = "Comment Reactions Closed" %]
     This [% terms.bug %] no longer accepts comment reactions.
 
+  [% ELSIF error == "comment_reaction_restricted" %]
+    [% title = "Comment Reactions Restricted" %]
+    You are not allowed to react on this [% terms.bug %].
```
dklawren, line 389: Nit: Probably should be 'You are not allowed to react to comments on this [% terms.bug %].'

dklawren (approved):

template/en/default/global/user-error.html.tmpl
kyoshino: Changed!

globau (changes requested):
The text next to the restrict comments checkbox on show_bug should be updated to indicate this also restricts reactions (ie "Restrict commenting and reactions on this …")

The Impact section of extensions/RestrictComments/template/en/default/pages/restrict_comments_guidelines.html.tmpl also needs updating

PR #2321, opened 2024-09-13
Bug 1918771 - Support pronouns such as %user% in "changed by" searches

dklawren:
This backports some old fixes from upstream that allows for use of pronouns (%self%, %user%, etc.) for the cc and commenter fields of email searches.


Bugzilla/Search.pm
```diff
@@ -2573,6 +2519,15 @@ sub _triage_owner_pronoun {
   }
 }
 
+sub _get_user_id {
+  my ($self, $value) = @_;
+
+  if ($value =~ /^%\w+%$/) {
```
cgsheeh, line 2525: Nit: a comment to clarify what this is would be nice, or maybe put the logic in a well-named function so this becomes `if (value_is_pronoun($value))` or something like that.

cgsheeh (approved):

PR #2330, opened 2024-09-25
Bug 1920995 - wrong calculation of relative date (rel-time) caused by repeatedly rounding in time_ago

oldherl:
Avoid huge errors of rounding the seconds again and again.
Instead, every unit of time should be directly calculated from seconds and rounded just once.

### DIFF
--- a/Bugzilla/Util.pm
+++ b/Bugzilla/Util.pm
@@ -650,10 +650,12 @@ sub time_ago {
   # DateTime object or seconds
   my $ss = ref($param) ? time() - $param->epoch : $param;
   my $mm = round($ss / 60);
-  my $hh = round($mm / 60);
-  my $dd = round($hh / 24);
-  my $mo = round($dd / 30);
-  my $yy = round($mo / 12);
+  my $hh = round($ss / (60 * 60));
+  my $dd = round($ss / (60 * 60 * 24));
+  # They are not the best definition of month and year,
+  # but they should be good enough to be used here.
+  my $mo = round($ss / (60 * 60 * 24 * 30));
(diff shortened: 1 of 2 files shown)


Bugzilla/Util.pm
```diff
@@ -650,10 +650,12 @@ sub time_ago {
   # DateTime object or seconds
   my $ss = ref($param) ? time() - $param->epoch : $param;
   my $mm = round($ss / 60);
-  my $hh = round($mm / 60);
-  my $dd = round($hh / 24);
-  my $mo = round($dd / 30);
-  my $yy = round($mo / 12);
+  my $hh = round($ss / (60 * 60));
```
cgsheeh, line 653: Maybe we should convert to a [`DateTime::Duration`](https://metacpan.org/pod/DateTime::Duration) and use that object to populate the correct string instead of doing the calculation ourselves?

oldherl: Using a dedicated library to generate those is a good idea. However I have two concerns about that:

(1) With a quick skim, I don't find any half-rounding in that library and [DateTime::Format::Duration](https://metacpan.org/pod/DateTime::Format::Duration). They seem to be just rounding towards zero. Is that too much a semantic change from the old behaviour?
Personally I do not like half-rounding a time duration very much ( I won't call 20 months ago as 2 years ago), but the consistency matters more.

(2) We also need a JavaScript version and they should match. I'm not sure whether that's achievable if we use the above library.

dklawren (approved):
i agree that using another library may be overkill for this change and also I do like to keep the method consistent when trying to figure out what is happening in the code down the road. r=dkl

PR #2333, opened 2024-10-01
Bug 1921923 - add healthchecks to bmo.test

shtrom:


dklawren:
Thanks for the changes. For Bugzilla at least, we require a bug id associated with all commits (along with at least one review). Could you create a general bug to track these changes and then update the pull request summary? Thanks

shtrom:
done

PR #2334, opened 2024-10-02
Bug 1904321 - Make more user-facing pages mobile friendly

kyoshino:
[Bug 1904321 - Make more user-facing pages mobile friendly](https://bugzilla.mozilla.org/show_bug.cgi?id=1904321)

- Compact, fixed global header with mobile drawer navigation
    - Larger menu items on mobile for easier tapping
    - Minor UI refresh: new icons, menu padding
    - My Attention link is moved to the BMO extension
    - New Account and Forgot Password links are moved to Log In dropdown
- Responsive page layout: Home, Log In, New Account, Browse, User Preferences
    - Log In page gets a cleaner design
    - Removed the legacy mobile CSS for the Log In page
    - Some tables in the prefs pages are too big and simply overflow


dklawren (changes requested):
Please make the header icons and maybe the text a little brighter. They are difficult to see (at least for me). Also the text 'Search Bugs' placeholder text is too dim for me.This is true for both light and dark modes.
image

Also when in responsive mode, and touch simulation is on, the calendar widget will not work properly. For example, the reminder preference panel. You can see if by turning the simulation on/off and trying to enter a date in the calendar field.
image

Overall it looks very good. I will put this version on bugzilla-dev for feedback. A lot of the old legacy tables (admin pages, dashboard, etc,) are still awkward viewing on mobile but as long as you can scroll sideways to see the full data, it is not a showstopper for now.

dklawren:
> Please make the header icons and maybe the text a little brighter. They are difficult to see (at least for me). Also the text 'Search Bugs' placeholder text is too dim for me.This is true for both light and dark modes. image
> 

Another example is the current header for BMO
image

The text and icons are brighter to me on production. The 'Search Bug' placeholder text is the same so not a regression but would be nice if we could make that more readable as well in the process.

kyoshino:
Refined the header/drawer design as per feedback! The calendar not showing up seems to be a [Firefox bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1828168) 😅

PR #2336, opened 2024-10-02
Bug 1921383 - Bugzilla rich text editor shortcuts don't respect user layout (revised)

kyoshino:
[Bug 1921383 - Bugzilla rich text editor shortcuts don't respect user layout](https://bugzilla.mozilla.org/show_bug.cgi?id=1921383)

Sorry, #2331 was incomplete and broke shortcuts with non-alphanumeric keys like Space and Enter. Now I remember why I said “keyboard shortcuts not working sometimes” in my Sveltia UI change 😓 This does the right thing.

### DIFF
--- a/js/text-editor.js
+++ b/js/text-editor.js
@@ -93,7 +93,7 @@ Bugzilla.TextEditor = class TextEditor {
         'Accel+I': { handler: () => this.execCommand('italic') },
         'Accel+E': { handler: () => this.execCommand('code') },
         'Accel+K': { handler: () => this.execCommand('link') },
-        'Accel+Shift+Period': { handler: () => this.execCommand('quote') },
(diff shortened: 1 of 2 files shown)


dklawren (changes requested):
These changes break keyboard shortcuts for Linux on both stable and nightly. On production, I am able to do control-shift-. (period) to quote, as well as the shortcuts for number and un-ordered list. But with the latest changes, none of those actions work for me anymore.

kyoshino:
I’m seeing the same issue on Windows. 😞 A fix incoming.

Also, the (legacy) tab switching shortcut with `Ctrl+Shift+P` doesn’t work because it overlaps with Firefox’s native keyboard shortcut for opening a private window. I’ll remove it.

dklawren (approved):

PR #2343, opened 2024-10-11
Bug 1923840 - Searching for bugs containing a mentor times out (504 Gateway timeout)

dklawren:
When selecting a search include Mentors, it is using a non-existent function _commenter_pronoun and needs to instead use _contact_pronoun. This was missed in an earlier commit back porting pronoun support to more places.


cgsheeh (approved):
Does BMO/Perl have linters we could enable in CI to catch things like this?

dklawren:
> Does BMO/Perl have linters we could enable in CI to catch things like this?

It does but for some reason does not catch undefined functions :(

PR #2345, opened 2024-10-23
Bug 1926081 - Add modification_time datetime column to profiles table denoting when the profile was last updated

dklawren:
Currently we would have to look at the last entry in profiles_activity and/or audit_log to see when a user profile was last updated. It will be much simpler if we just create a new column to track this timestamp and not have to look in one or more other places. I will also need to create a script to go back and fill in the new timestamp with the current value from the different activity tables. This pull request adds the new column, updates it when the user is updated and also populates the column with the proper values.


Bugzilla/User.pm
```diff
@@ -448,6 +456,10 @@ sub _check_numeric {
   return $value;
 }
 
+sub _check_timestamp {
+  return Bugzilla->dbh->selectrow_array('SELECT LOCALTIMESTAMP(0)');
```
cgsheeh, line 460: Question: why `LOCALTIMESTAMP` instead of `UNIX_TIMESTAMP` here?

cgsheeh (approved):

Bugzilla/User.pm
dklawren: I am only using the unix timestamps for the migration part as it is easier to determine which is the larger value and use that for the modifcation_ts. For normal updating of users using ->update() we always want the timestamp in the standard `YYYY-MM-DD HH:MI:SS` format which is what we get from LOCALTIMESTAMP.

dklawren:
I have updated the migration code to fix some logic issues discovered during recent deployment. Please review the new changes.

- Fixed a bug in the migration code that was not looping over the userids from the profiles table properly.
- Also check for the existence of the creation_ts column and if it exist add it to the list of possible modification_ts values.
- Take the same code from the migration function and put into a run-once fixup script to run on production to finish the migration. This is required because the profiles.migration_ts column has been added to the profiles table and the migration code will not be ran. The fixup script will finish the migration that was aborted before and not need to be ran again.

cgsheeh (approved):

PR #2347, opened 2024-10-24
Bug 1926081 - Add modification_time datetime column to profiles table denoting when the profile was last updated (schema only)

dklawren:
The PR #2345  will need to have the schema change deployed first then the rest of the code later.


cgsheeh (approved):
LGTM.

PR #2348, opened 2024-10-28
Bug 1926574 - When changing product of a bug and the destination product only has one component, pre-select the one component

dklawren:
This change updates the webservice API that BugModal (show_bug.cgi) page uses to fetch a list of components when a user changes the product. If the new product selected only contains one component (i.e. Invalid Bugs) then it pre-selects the single component. If it has more than one, then it does the old behavior of adding a block component value and requires the user to select the proper component before submitting.


cgsheeh (approved):
Makes sense to me.

PR #2351, opened 2024-10-30
Bug 1928083 - Bugzilla’s emojis are displayed on the bottom of the page when activated creating a new scroll bar

kyoshino:
[Bug 1928083 - Bugzilla’s emojis are displayed on the bottom of the page when activated creating a new scroll bar](https://bugzilla.mozilla.org/show_bug.cgi?id=1928083)

The cause is due to a change in page layout (fixed global header) in #2334, but reverting the change would rather introduce other problems (I don’t remember exactly, but I saw several problems while working on mobile optimization).

The solution to detect the CSS Anchor Positioning API, which is not yet implemented in Firefox and Safari, and fall back to the legacy implementation if the API is not available.

This also fixes the anchor positioning itself. My initial implementation hardcoded `anchor-name` in CSS, but it doesn’t work well because there are usually multiple comments in a bug, and `anchor-name` needs to be unique for each. Not sure why it wasn’t caught earlier.

I’ve tested this fix with Firefox, Chrome and Safari.

### DIFF
--- a/extensions/BugModal/web/bug_modal.css
+++ b/extensions/BugModal/web/bug_modal.css
@@ -1249,7 +1249,6 @@ a.comment-tag-url {
   position: relative;
   padding: 0 12px 12px;
   background: var(--primary-text-background-color);
-  anchor-name: --comment-reactions;
 }
 
 .comment-text.bz_private + .comment-reactions {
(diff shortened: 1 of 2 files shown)


dklawren (approved):
Reproduced the error and verified that this pull request fixes the issue. r=dkl

kyoshino:
@dklawren Thank you for your quick review 🙏🏼

PR #2353, opened 2024-11-04
Bug 1929001 - Saved data for comment 0 overrides "comment" parameter

dklawren:
This simple fix just checks to make sure that the comment textarea is empty before pre-filling the stored comment text in localStorage.

### DIFF
--- a/extensions/BugModal/web/bug_modal.js
+++ b/extensions/BugModal/web/bug_modal.js
@@ -203,7 +203,8 @@ $(function() {
         let value = JSON.parse(localStorage.getItem(bugCommentCacheKey));
         if (value){
             let commentBox = document.querySelector("textarea#comment");
-            commentBox.value = value['text'];
+            if (commentBox.value === '')
+              commentBox.value = value['text'];
             // Resize the textarea and enable the Preview button
             commentBox.dispatchEvent(new InputEvent('input'));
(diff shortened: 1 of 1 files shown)


extensions/BugModal/web/bug_modal.js
```diff
@@ -203,7 +203,8 @@ $(function() {
         let value = JSON.parse(localStorage.getItem(bugCommentCacheKey));
         if (value){
             let commentBox = document.querySelector("textarea#comment");
-            commentBox.value = value['text'];
+            if (commentBox.value === '')
+              commentBox.value = value['text'];
```
cgsheeh, line 207: Just to make sure I'm following, this is "if the comment box hasn't been pre-filled from some external source (like POST args or query string parameters), use the existing local storage value"?

cgsheeh (approved):

extensions/BugModal/web/bug_modal.js
dklawren: Sorry wasn't more clear. Yes that is correct. If someone passes `&comment=foo` on the query string to the enter bug page, the text area for the description will have `foo` already in it. The JS to pull previously cached content for the description from localStorage will not be placed into the comment field, overwriting the `foo` value.

PR #2354, opened 2024-11-05
Bug 1922986 - Create script that exports BMO data as JSON suitable for import into a BigQuery instance in GCP

dklawren:
This pull request adds a new script and test suite changes to support exporting bug data from the DB to BigQuery. When this is merged, I will file a new PR to have a cron container run once daily to call export_bmo_etl.pl and import a snapshot of the bug data for that day. Each row of each table will have the same timestamp. The test suite does some basic tests against a bigquery emulator that does not require credentials. We have a bigquery instances already setup for production and dev and I am running a full import from bugzilla-dev currently to see how long it will take.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -286,7 +286,8 @@ jobs:
           name: run bmo specific tests
           command: |
             [[ -f build_info/only_version_changed.txt ]] && exit 0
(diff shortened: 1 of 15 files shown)


dklawren:
Closing to create a new pull request with cleaner changes.

PR #2358, opened 2024-11-14
Bug 1931232 - Add BMO vs Bugzilla warning block at top of new bug page for Bugzilla product

dklawren:
image

### DIFF
--- a/extensions/BMO/template/en/default/bug/create/bugzilla-admin.html.tmpl
+++ b/extensions/BMO/template/en/default/bug/create/bugzilla-admin.html.tmpl
@@ -6,28 +6,26 @@
   # defined by the Mozilla Public License, v. 2.0.
   #%]
 
-[% RETURN UNLESS product.name == 'Bugzilla' &&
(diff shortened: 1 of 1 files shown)


dklawren:
> Personally, I don't love the way the warning is worded. I would specify that this is not the "component" for filing issues against BMO, since the website (ie the "place") is technically correct.
> 
> It also seems a little weird to list things that are incorrect for the current component and then refer to a different component. I'd change this to describe what the Bugzilla component is for ("this component tracks changes to the upstream Bugzilla project"), then say "you may be looking for the Mozilla-specific `BMO :: Admin` component, which is used for <list of config/permissions/account changes>".
> 
> Anyways I realize this is already in the code and you're just changing the conditions where it is revealed, so it works for me. 👍

I agree but didn't want to get too wordy as to take up a large part of the page with the warning. I will see if I can tweak the wording some and let you take a look.

cgsheeh (approved):
LGTM!

PR #2361, opened 2024-11-19
Bug 1932257 - CircleCI configuration cleanups and removal of un-used code in entrypoint.pl

dklawren:
Rather than junk up the other pull request for the ETL work with these minor fixes, I am making them here instead. These are changes to the way images are built in CircleCI for running the automated tests as well as some code cleanup to remove unused code and get rid of some warnings in the test output that looks bad.

In order to build the bigquery-emulator container for testing in the ETL pull request, we need to build the images from the docker-compose.test.yml file instead of doing docker builds outside of docker-compose. Instead of `docker build -t bmo . ; docker-compose run bmo.test test_webservices`, I do `docker run --build bmo.test test_webservices`.

For github, I still need to do the `docker-compose build` before the test run since for some reason their implementation did not like using `docker-compose run --build ...`.

I removed some extra functions in entrypoint.pl that are never used by BMO or mozilla-conduit/suite.

The rest of the changes are small fixes to remove warnings and other extra stuff sent to STDOUT that was cluttering up the test logs.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -22,28 +22,6 @@ defaults:
       username: $DOCKER_USER
       password: $DOCKER_PASS
 
-  build_image: &build_image
-    run:
-      name: Build Docker BMO Image
-      command: |
-        docker build \
-          --build-arg CI="$CI" \
-          --build-arg CIRCLE_SHA1="$CIRCLE_SHA1" \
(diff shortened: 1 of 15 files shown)


cgsheeh (approved):
LGTM, appreciate you proactively splitting out the cleanups into a separate PR!

PR #2366, opened 2024-11-26
Bug 1933585 - Fix TOTP form layout on Account Verification page

kyoshino:
I haven’t filed a bug, but I believe this is a fallout from #2334. When you sign in, you’ll probably enter a 6-digit verification code on the Account Verification page. Currently, the Submit button is displayed in full width, which is not an intended layout. This PR fixes it.

(The UI could be [better than this](https://www.google.com/search?q=otp+input&udm=2), but it’s out of scope 😉)

### Before

image

### After

image

### DIFF
--- a/skins/standard/login.css
+++ b/skins/standard/login.css
@@ -74,8 +74,14 @@ button[type="submit"] {
   font-size: var(--font-size-h3);
 }
 
-#verify-totp-form div {
-  margin-bottom: 15px;
+#verify-totp-form {
+  display: flex;
+  align-items: center;
+  gap: 8px;
+}
+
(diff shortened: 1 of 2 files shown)


dklawren (changes requested):
The Submit button needs to be below the input field since there is a hidden span next to the input field. When a user enters their mfa code, it does a pre-check via API and displays an error next to the input field if the value is wrong. If the submit button is next to it, would it not slide over?

kyoshino:
I missed the error message. I’ve modified the markup and style so the message now appears below the input instead of to the right.

image

dklawren (approved):

dklawren:
Thanks!

PR #2372, opened 2024-12-12
Bug 1936911 - Add back CIRCLECI_SHA1 AND CIRCLE_BUILD_URL back to the build and deploy steps for CircleCI

dklawren:
before these changes, the `/app/version.json` file had empty values for CIRCLE_SHA1 and CIRCLE_BUILD_URI. This fixes the issue.

### DIFF
--- a/docker-compose.test.yml
+++ b/docker-compose.test.yml
@@ -8,6 +8,8 @@ services:
       context: .
       dockerfile: Dockerfile
       target: test
+      args:
+        - CI=1
     command: dev_httpd
     tmpfs:
       - /tmp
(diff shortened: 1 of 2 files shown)


dklawren:
Oops not quite right yet.

docker-compose.test.yml
```diff
@@ -8,6 +8,8 @@ services:
       context: .
       dockerfile: Dockerfile
       target: test
+      args:
+        - CI=1
```
cgsheeh, line 12: I'm surprised we need to set this and CircleCI doesn't set this when appropriate on its own?

cgsheeh (approved):

PR #2373, opened 2024-12-13
Bug 1934616 - Add comment to SQL that is generated by buglist.cgi to add userid, query string, and user agent

dklawren:
To help pin-pointing where some long running MySQL queries are coming from, we can add comment strings to the SQL statements that will show up in the long running query logs. The extra information will be helpful in finding out where/who they are coming from and we can work with them to find more efficient ways of getting the data they need like bespoke API calls using the Search API support we have.

### DIFF
--- a/Bugzilla/Search.pm
+++ b/Bugzilla/Search.pm
@@ -870,9 +870,19 @@ sub data {
     '_no_security_check' => 1
   );
 
+  # Add some user information to the SQL so we can pinpoint where some
+  # slow running queries originate and help to refine the searches.
+  my $sql_user_info
+    = ' /* userid: '
+    . Bugzilla->user->id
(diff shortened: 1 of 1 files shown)


cgsheeh (approved):
Seems reasonable. Assuming we log information about the queries we're running in a structured format somewhere, perhaps we could also add this information there also?

dklawren:
> Seems reasonable. Assuming we log information about the queries we're running in a structured format somewhere, perhaps we could also add this information there also?

Yeah. We can see the long running queries at a specific time in the Query Insights page Cloud SQL in the GCP console. When people are experiencing slowness on BMO and see that there is high load on the replica, we can look and see what the actual query was and the comment will give us more info on who is running the query.

PR #2385, opened 2025-02-05
Bug 1930197 - Add an option (or config) to not set the qe-verify+ flag when a GitHub push closes a bug in a given product

dklawren:
Before deployment, I will get with the Fenix and Focus people to add the ?no-qe-verify=1 to their webhooks.

### DIFF
--- a/Bugzilla/API/V1/Github.pm
+++ b/Bugzilla/API/V1/Github.pm
@@ -318,8 +318,8 @@ sub push_comment {
       $set_all->{resolution} = 'FIXED';
 
       # Update the qe-verify flag if not set and the bug was closed.
-      # Do not perform this change if Fenix or Focus product.
-      if ($bug->product ne 'Fenix' && $bug->product ne 'Focus') {
+      # Do not set the flag if the no-qe-verify parameter is true.
+      if (!$self->param('no-qe-verify')) {
         my $found_flag;
         foreach my $flag (@{$bug->flags}) {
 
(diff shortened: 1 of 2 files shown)


willdurand:
I can't speak for Focus but AFAIK Fenix lives in tree these days.

cgsheeh (approved):
Any tests we could add for this?

dklawren:
> I can't speak for Focus but AFAIK Fenix lives in tree these days.

@willdurand so I would only need to get Focus to change their webhook to match the new format?

willdurand:
> @willdurand so I would only need to get Focus to change their webhook to match the new format?

It looks like focus might also be in m-c: https://searchfox.org/mozilla-central/source/mobile/android/focus-android

dklawren:
Sigh. Ok this was not what I originally hoped for. I will have to continue the old route and look at the product of the bug ID that is in the push messsage and make the determination there. I cannot rely on setting webhook parameters for different repos as they are all on the same repo. I don't know why I didn't catch this before now. I will create a config param to add a list of product names in BMO to omit setting the qe-verify flag. At least that way I can make changes without needed a code deployment.

willdurand:
> I cannot rely on setting webhook parameters for different repos as they are all on the same repo.

That is still useful for all the other github repos, though. IIUC fenix/focus has not used this exclusion in the code for a while now (since all the repos have moved to m-c basically) so that won't change anything for focus/fenix. That said, for other repos, it'd be useful to be able to opt-in/opt-out, and the query param does make sense to me.

PR #2386, opened 2025-02-07
Bug 1946805 - Count is not being incremented properly when excluding bugs from the full list and therefore the export script is hanging

dklawren:
The new BMO ETL export script has a bug in that it does not properly increment the $count variable when encountering bugs that are to be skipped in the export. These include Legal bugs, etc. Then the while() loop runs forever as $count is always less than $total.

Bad code:
```perl
        if (any { $obj->product eq $_ } EXCLUDE_PRODUCTS) {
          $excluded_bugs{$obj->id} = 1;
          next;
        }
```

This should be instead:

```perl
        if (any { $obj->product eq $_ } EXCLUDE_PRODUCTS) {
          $excluded_bugs{$obj->id} = 1;
          $count++;
          next;
        }
```

I didn't not catch this unfortunately til we tried it in production as bugzilla-dev does not have any bugs to be excluded so the count was always correct. 

Any other functions where we skip excluded bugs will also need to be update to increment $count so I will fix those as well.

### DIFF
--- a/extensions/BMO/bin/export_bmo_etl.pl
+++ b/extensions/BMO/bin/export_bmo_etl.pl
@@ -157,6 +157,7 @@ sub process_bugs {
 
         if (any { $obj->product eq $_ } EXCLUDE_PRODUCTS) {
           $excluded_bugs{$obj->id} = 1;
+          $count++;
           next;
         }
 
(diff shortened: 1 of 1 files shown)


cgsheeh (approved):
It seems like we're always incrementing `count` when we have an excluded bug. Could we check the size of `excluded_bugs` to determine the count instead?

dklawren:
> It seems like we're always incrementing `count` when we have an excluded bug. Could we check the size of `excluded_bugs` to determine the count instead?

Not really as some of the other tables may not have any entries referring to the bugs that were excluded, attachments, etc. so using the excluded_bugs amount to determine true count would not work. So we need to do it explicitly for each function.

PR #2388, opened 2025-02-07
Bug 1941032 - "Always Enable Edit Mode" is broken

dklawren:
With the changes to suppose cookie banner, the old code was looking for the string 'true' instead of a proper true/false boolean.


cgsheeh:
Looks like there might be another instance of this [here](https://github.com/mozilla-bteam/bmo/blob/master/extensions/MyDashboard/web/js/query.js#L20)?

dklawren:
> Looks like there might be another instance of this [here](https://github.com/mozilla-bteam/bmo/blob/master/extensions/MyDashboard/web/js/query.js#L20)?

Thanks! Will check that out and file a different bug.

PR #2394, opened 2025-02-19
Bug 1949285 - BMO ETL: Update export script to temporaily not cache profiles daat until migration is complete from bug 1926081

dklawren:
Even with adding a default value for modification_ts, with the migration script from [bug 1926081](https://bugzilla.mozilla.org/show_bug.cgi?id=1926081) still running in the background (very slowly), when the script starts to export the profiles and hits a similar place in the profiles table that the migration script is hitting, it causes the script to crash.

So I want to temporarily not cache profile rows in the DB and just send the data directly. When the migration is complete, we can add the caching back in. I need this to get consistent nightly exports of BMO data without crashes. Also the profiles table is not huge like the other tables so it doesn't really effect the export time significantly.


cgsheeh (approved):
WFM. If you think you might need this in the future, you could also think about having the cache be disabled via a CLI flag.

PR #2398, opened 2025-03-04
Bug 1951478 - BMO ETL: Add a timestamp to lock table and remove lock if timestamp is greater than a day

dklawren:


extensions/BMO/bin/export_bmo_etl.pl
```diff
@@ -841,11 +841,17 @@ sub check_and_set_lock {
   return if $test;    # No need if just dumping test files
 
   my $dbh_main = Bugzilla->dbh_main;
+
+  # Clear out any locks that are greater than 24
```
cgsheeh, line 845:
```suggestion
  # Clear out any locks that are greater than 24h old.
```

cgsheeh (approved):

PR #2399, opened 2025-03-04
Bug 1951739 - BMO ETL: Add parameters to export script to turn off cache for one or more tables

dklawren:
Add parameters to the export script that allows for disabling caching of specific tables in case issues are found with regard to specific data causing the export script to fail consistently.


cgsheeh (approved):
You could consider having this variable feed into `store_cache` and `get_cache` directly instead of wrapping each call in an `if`, so they are simply no-ops if the flag is passed. This LGTM either way. :)

PR #2402, opened 2025-03-14
Bug 1196803 - Add a needinfo option for a user drop down for quick needinfo

dklawren:
image

image

### DIFF
--- a/extensions/BMO/web/js/edituser_menu.js
+++ b/extensions/BMO/web/js/edituser_menu.js
@@ -65,6 +65,20 @@ function show_usermenu(vcard) {
             }
         });
     }
+    if (email) {
+        items.push({
+            name: "Needinfo",
+            callback: function () {
+                $('#needinfo').prop('checked', true);
+                $('#needinfo_role').val('other');
+                $('#needinfo_from').val(email);
+                $('#needinfo_from_container', '#needinfo_container').removeClass('bz_default_hidden');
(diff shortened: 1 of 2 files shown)


extensions/BMO/web/js/edituser_menu.js
```diff
@@ -65,6 +65,20 @@ function show_usermenu(vcard) {
             }
         });
     }
+    if (email) {
+        items.push({
+            name: "Needinfo",
+            callback: function () {
+                $('#needinfo').prop('checked', true);
+                $('#needinfo_role').val('other');
```
cgsheeh, line 73: Is it mostly a UI inconsistency that this will always put "other" instead of "reporter"/"assignee" etc? Is there logic to detect that on the JS side?

cgsheeh (approved):

extensions/BMO/web/js/edituser_menu.js
dklawren: I thought just using Other for the needinfo from comment user name would make it more obvious as to what is going on cause the input field would be visible always. It is possible, with some more logic, to see if the comment user name is the reporter, assignee, etc. and then choose the option from the drop down. But when we do that the input field is hidden and may not be as obvious what change has occurred. I can look into doing it that way if you think it will be more consistent.

cgsheeh (approved):
LGTM.

PR #2412, opened 2025-03-21
Bug 1954310 - BMO ETL: Add more verbose debugging output to what is already logged

dklawren:
- Removes --verbose and prints standard output when no arguments
- Adds --debug to print even more verbose output than normal.
- New logger function to handle the printing of all output.


cgsheeh (approved):
LGTM.

PR #2418, opened 2025-03-26
Bug 1956136 - GITHUB: Add new bmo slim work flow to Github actions to allow manually building and pushing of new bmo-perl-slim images to GAR

dklawren:


emaydeck-mozilla:
You're missing your env declarations. For example, here are the ones from the deploy workflow: https://github.com/mozilla-bteam/bmo/blob/master/.github/workflows/deploy.yml#L10-L14

`IMAGE_NAME` doesn't look like it's being used, so you can probably omit that one. The rest can be kept the same as what you have in deploy.yaml.

emaydeck-mozilla (approved):

PR #2419, opened 2025-03-26
Bug 1956136 - GITHUB: Add new bmo slim work flow to Github actions to allow manually building and pushing of new bmo-perl-slim images to GAR (fix docker run)

dklawren:
- Removed -it a it caused a TTY error in Github Actions
- Copied cpanfile* into build_info after it was copied in current working directory.


dklawren:
This worked in a test run at https://github.com/mozilla-bteam/bmo/actions/runs/14094695551

PR #2424, opened 2025-04-09
Bug 1959485 - BMO ETL: Wrap DB calls for storing/retrieving cached data in try/catch to allow script to continue export to BQ if exception occurs

dklawren:
This wraps two DB calls for getting and storing cached data in the BMO database when exporting to BigQuery. There have been intermittent crashes in the past during these operations that would cause the entire export to fail. At least now the error will be logged but the export will continue on. The critical part is just getting the data to BigQuery each day, whether it is cached or not. I will still be able to look at logs for any errors and address them as they appear.

### DIFF
--- a/extensions/BMO/bin/export_bmo_etl.pl
+++ b/extensions/BMO/bin/export_bmo_etl.pl
@@ -17,6 +17,7 @@ use Bugzilla::Bug;
 use Bugzilla::Constants;
 use Bugzilla::Flag;
 use Bugzilla::Group;
+use Bugzilla::Logging;
 use Bugzilla::User;
 use Bugzilla::Extension::Review::FlagStateActivity;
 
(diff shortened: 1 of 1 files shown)


extensions/BMO/bin/export_bmo_etl.pl
```diff
@@ -744,15 +753,21 @@ sub store_cache {
...
-  $main_dbh->do(
-    'INSERT INTO bmo_etl_cache (id, table_name, snapshot_date, data) VALUES (?, ?, ?, ?)',
-    undef, $id, $table, $timestamp, $gzipped_data
-  );
+    # Enter new cached JSON
+    $main_dbh->do(
+      'INSERT INTO bmo_etl_cache (id, table_name, snapshot_date, data) VALUES (?, ?, ?, ?)',
+      undef, $id, $table, $timestamp, $gzipped_data
+    );
+  }
```
cgsheeh, line 766: Nit: you could put each of these in their own `try/catch` block, so the error message indicates which part failed for better debugging. The same goes for the other `try/catch` added above.

dklawren: The "$_" in the WARN text will be the actual error message of the command that failed so we will see at which part failed. Since the commands in the try {} block are all dependent on the previous one passing I figured just wrap it all together.

PR #2427, opened 2025-04-11
Bug 1957414 - Make it clear why the component field is empty when editing bugs in bulk that spam multiple products

dklawren:


cgsheeh (approved):
LGTM.

PR #2432, opened 2025-04-24
Bug 1962078 - Clicking the "Duplicate" button scrolls down to the bottom of the page

dklawren:


cgsheeh (approved):
Sounds like this doesn't change behaviour for when the new comment box is at the bottom, so it LGTM.

We could also find the "duplicate id form element" and do `element.scrollIntoView`, but if the field is already in the right place, this seems reasonable. :)

dklawren:
> Sounds like this doesn't change behaviour for when the new comment box is at the bottom, so it LGTM.
> 
> We could also find the "duplicate id form element" and do `element.scrollIntoView`, but if the field is already in the right place, this seems reasonable. :)

Good idea. Had forgotten about the scrollIntoView() method so updated the code to use that instead of taking out completely. Please take a look.

cgsheeh (approved):

PR #2433, opened 2025-04-25
Bug 1962762 - Use of uninitialized value $user_agent in concatenation (.) or string at Bugzilla/Search.pm line 966 when running collectstats.pl daily

dklawren:


Bugzilla/Search.pm
```diff
@@ -961,7 +961,7 @@ sub _sql {
   # slow running queries originate and help to refine the searches.
   my $user_id = Bugzilla->user->id;
   my $remote_ip = remote_ip();
-  my $user_agent = Bugzilla->cgi->user_agent;
+  my $user_agent = Bugzilla->cgi->user_agent || 'None';
```
cgsheeh, line 964: Could `collectstats.pl` set a `User-Agent` instead? Seems any other tools/bots/etc which don't set a user agent will now show up as a `None` user agent, which would make determining the source of a slow query more difficult.

cgsheeh (changes requested):

Bugzilla/Search.pm
dklawren: Good call. I updated the Search.pm to use the script_name to handle other cases of no user_agent. And then updated collectstats.pl to set the script_name value appropriately. Some other changes were needed in Search.pm such as proper user id value, etc.

cgsheeh (approved):
LGTM!

PR #2439, opened 2025-05-13
Bug 1960099 - When a revision has enters the needs-review state with a reviewer group assigned, query Phab for list of group users and update the reviewer

dklawren:


dklawren:
Please review the tech outline as well to make sure I didn't miss anything or implement incorrectly
https://docs.google.com/document/d/1-O27gB5ckWyHrL8mOt7mS9JTH1mChtOjzBwE7FHLyWs/edit?tab=t.0#heading=h.v57iutue8rat

extensions/PhabBugz/lib/Util.pm
```diff
@@ -305,4 +306,135 @@ sub set_phab_user {
...
+          undef, $rev_phid, $blocking_project->phid, $member->phid
+        );
+      }
+
+      next;
+    }
+
+ # 7. Once a potential reviewer has been found, look to see if they can see the bug,
+ # and they are not set to away (not accepting reviews). If both are are negative,
+ # we choose the next person in the list.
```
cgsheeh, line 403:
```suggestion
 # 7. Once a potential reviewer has been found, look to see if they can see the bug,
 # and they are not set to away (not accepting reviews). If both are are positive, we have found our reviewer and exit the loop.
```

extensions/PhabBugz/lib/Util.pm
```diff
@@ -305,4 +306,135 @@ sub set_phab_user {
...
+  # and 2) ends in "-reviewers". Normally Herald will set this if certain
+  # conditions are met. If a blocking reviewer group cannot be found then
+  # do nothing.
+  my $blocking_project;
+  my @blocking_users;
+
+  foreach my $reviewer (@{$revision->reviews}) {
+    next if !$reviewer->{is_blocking};    # Only interested in blocking
+    if ($reviewer->{is_project}) {
+      next if $reviewer->{user}->name !~ /-reviewers$/;    # Only interested in reviewer groups
```
cgsheeh, line 330: From the spec it looks like the intention was for only review groups with `-reviewers-rotation` to be considered.

```suggestion
      # Only interested in opt-in reviewer rotation groups.
      next if $reviewer->{user}->name !~ /-reviewers-rotation$/;
```

extensions/PhabBugz/lib/Util.pm
```diff
@@ -305,4 +306,135 @@ sub set_phab_user {
...
+    if ( $member->bugzilla_user->can_see_bug($revision->bug->id)
+      && $member->bugzilla_user->settings->{block_reviews}->{value} ne "on")
+    {
+      INFO("$rev_identifier: Found new reviewer " . $member->id);
+      $found_reviewer = $member;
+      last;
+    }
+  }
+
+  if ($found_reviewer) {
```
cgsheeh, line 413: Looking at this logic, it seems it is possible for this system to *not* find a reviewer, and thus leave the blocking review group set on the revision. AIUI the goal of this system is to always set a specific person as the reviewer on the revision when a blocking review group is set. We should have some logic here to assign a reviewer even if one isn't found with the current logic.

extensions/PhabBugz/lib/Util.pm
```diff
@@ -305,4 +306,135 @@ sub set_phab_user {
...
+      undef, $blocking_project->phid, $member->phid
+    );
+
+    if ($rev_phid) {
+      INFO("$rev_identifier: Previous review found for " . $member->id);
+
+      my $rev_obj = Bugzilla::Extension::PhabBugz::Revision->new_from_query(
+        {phids => [$rev_phid]});
+
+# 6. If the user is already a reviewer or they were but their revision is now closed, skip to the next
```
cgsheeh, line 386: If their revision is now closed, couldn't they be the `$found_reviewer` for this revision? They no longer have a revision to review.

cgsheeh (changes requested):

extensions/PhabBugz/lib/Util.pm
```diff
@@ -305,4 +306,135 @@ sub set_phab_user {
...
+
+  # 2. Once the blocking reviewer group is determined, query Phabricator for
+  # list of group members and match up the BMO user account. Sort them by user
+  # id descending.
+  my $project_members
+    = [sort { $a->id <=> $b->id } @{$blocking_project->members}];
+
+  # 3. Make sure that none of the individual group members are not already
+  # set as a blocking reviewer. If so, then remove the blocking group and return.
+  foreach my $member (@{$project_members}) {
```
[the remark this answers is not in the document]
dklawren: Well we have two options here:
1. Fix all herald rules to only add the group once instead of every time.
2. Or if a blocking groups member is already set as a reviewer from a previous run, then we do nothing.

extensions/PhabBugz/lib/Util.pm
```diff
@@ -305,4 +306,135 @@ sub set_phab_user {
...
+    # 10. Store the data in the phab_reviewer_rotation table so they will be
+    # next time.
+    INFO("$rev_identifier: Adding new blocking reviewer to database.");
+    $dbh->do(
+      'INSERT INTO phab_reviewer_rotation (revision_phid, project_phid, user_phid, user_id) VALUES (?, ?, ?, ?)',
+      undef,
+      $revision->phid,
+      $blocking_project->phid,
+      $found_reviewer->phid,
+      $found_reviewer->bugzilla_user->id
```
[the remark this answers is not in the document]
dklawren: Decided not needed. Originally thought it would be useful to have a link to the BMO user in the DB but it isn't really necessary.

extensions/PhabBugz/lib/Util.pm
dklawren: The idea was that even though their revision is now closed and they are no longer reviewing it, I would still pick the next person in the list since the previous just reviewed something. This is to not favor any single person and make sure everyone gets a fair amount.

cgsheeh (changes requested):
Looking good, the DB schema is much easier to follow.

dklawren:
> LGTM! Excited to see this in action. :)
> 
> One final nit: instead of tracking the `$found_reviewer` variable throughout the loop, you could immediately call `set_new_reviewer()`, push the found reviewer to the list of reviewers and go to the next loop iteration. That would eliminate the `!$found_reviewer` checks and make it obvious that once the reviewer is found and set on the revision, we move to processing the next project.

Good idea. Will make that change.

dklawren:
Since it is not a simple 1-2 line change, I will request one last look-over from you and we can call this done. I have tested it and it works as expected.

extensions/PhabBugz/lib/Util.pm
```diff
@@ -332,4 +332,284 @@ sub set_phab_user {
...
+    INFO('Last reviewer phid found: '
+        . ($last_reviewer_phid ? $last_reviewer_phid : 'None'));
+
+    # If a last reviewer exists, we want to rotate the list to where the
+    # last reviewer is at the front of the list while maintaining order.
+    # move them to the front of the member list
+    @project_members = rotate_reviewer_list(\@project_members, $last_reviewer_phid);
+
+    # If there is only one member, then guess what? You are always getting it.
+    my $found_reviewer = 0;
```
cgsheeh, line 392: You can remove this variable and all references to it. All we do is set it here and then check `!$found_reviewer`, which is always True.

cgsheeh (approved):

PR #2441, opened 2025-05-15

dklawren:


cgsheeh (changes requested):

extensions/PhabBugz/lib/Util.pm
```diff
@@ -94,6 +94,13 @@ sub set_attachment_approval_flags {
       . $approval_flag_name
       . $status);
 
+  # Set value if the revision has been accepted by member of release managers
+  my $relman_accepted = 0;
+  foreach my $reviewer (@{$revision->reviews}) {
+    next if $reviewer->status ne 'accepted';
+    $relman_accepted = 1 if $reviewer->is_release_manager;
+  }
+
```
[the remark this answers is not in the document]
dklawren: Will try that out

extensions/PhabBugz/lib/Util.pm
```diff
@@ -64,7 +65,13 @@ sub set_attachment_approval_flags {
     'needs-revision'  => '-',
   };
 
-  my $status = $revision_status_flag_map->{$revision->status};
+  # Find the current review status of the revision changer
+  my $status = undef;
+  foreach my $reviewer (@{$revision->reviews}) {
+    if ($reviewer->id == $phab_user->id) {
+      $status = $reviewer->status;
```
cgsheeh, line 72: Question: looking at `extensions/PhabBugz/lib/Revision.pm`, it seems we have a custom `accepted-prior` status that we use when the revision has been accepted on a diff that is not the latest. Do you think we should add `accepted-prior` to the `$revision_status_flag_map`, with the same `+` BMO flag as `accepted`?

I would like to avoid having the flag be set to `-`, or unset, when a release manager has approved the revision already.

extensions/PhabBugz/lib/Util.pm
```diff
@@ -103,14 +110,14 @@ sub set_attachment_approval_flags {
...
-    if (!$flag_setter->can_change_flag($flag->type, $flag->status, $status)) {
+    if (!$bmo_user->can_change_flag($flag->type, $flag->status, $status)) {
       INFO(
         "Unable to set existing `$approval_flag_name` flag to `$status` due to permissions."
       );
       return;
     }
 
-    # If setting to + or - then user needs to be a release manager in Phab.
+    # If setting to + or - then user needs to a release manager.
```
cgsheeh, line 120:
```suggestion
    # If setting to + or - then user needs to be a release manager.
```

extensions/PhabBugz/lib/Util.pm
```diff
@@ -128,10 +135,18 @@ sub set_attachment_approval_flags {
   if (!@old_flags && $status ne 'X') {
     my $approval_flag = Bugzilla::FlagType->new({name => $approval_flag_name});
     if ($approval_flag) {
-      if ($flag_setter->can_change_flag($approval_flag, 'X', $status)) {
+      # If setting to + then at least one accepted reviewer needs to a release manager.
```
cgsheeh, line 138:
```suggestion
      # If setting to + then at least one accepted reviewer needs to be a release manager.
```

cgsheeh (changes requested):
LGTM, just have a question about the custom `accepted-prior` status.

cgsheeh (approved):
LGTM, thank you for fixing this! :)

PR #2442, opened 2025-05-15
Bug 1937196 - Add database timeouts to search queries

globau:


globau:
Uses https://dev.mysql.com/doc/refman/5.7/en/optimizer-hints.html#optimizer-hints-execution-time

You can test this by editing Bugzilla::Search and inserting a SLEEP statement into the generated query:

`SELECT $timeout_comment SLEEP(10), $select`

dklawren (approved):
LGTM and works as expected locally.

PR #2443, opened 2025-05-19
Bug 1967247 - BMO ETL: Create transaction around entire export process with  REPEATABLE READ  isolation to ensure data integrity

dklawren:
- Wrapped entire export in transaction ensuring consistent read data in case rows are changed or deleted during export
- Added additional debugging output


cgsheeh (approved):
LGTM.

PR #2445, opened 2025-05-20
Bug 1967330 - Animated menu bar when small screen size should not be animated if user prefers reduced motion

dklawren:


dklawren:
> Any way we could reduce the animations on more than just the menu bar? I see we have the `transition` CSS property set on a few other objects.

Filed https://bugzilla.mozilla.org/show_bug.cgi?id=1967776

PR #2446, opened 2025-05-20
Bug 1967336 - The error message generated by timeout of long running searches needs to also work with REST API in addition to the web UI

dklawren:
- Removed the MySQL setting ONLY_FULL_GROUP_BY due to some BMO code still not listing full group by column list. MySQL changed to checking by default in recent versions and this change now crashes certain queries that worked before.
- Created a single _sql_execute() function in Search.pm that can be used in multiple places to check for time out error instead of doing in both buglist.cgi and REST API.

### DIFF
--- a/Bugzilla/WebService/Bug.pm
+++ b/Bugzilla/WebService/Bug.pm
@@ -693,21 +693,7 @@ sub search {
   my $search = new Bugzilla::Search(%options);
 
   # Execute the query.
-  my $data = [];
-  do {
-    local $SIG{__DIE__}  = undef;
-    local $SIG{__WARN__} = undef;
-    ($data) = eval { $search->data };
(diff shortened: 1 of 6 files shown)


Bugzilla/WebService/Bug.pm
```diff
@@ -689,7 +691,23 @@ sub search {
   $options{params} = $match_params;
 
   my $search = new Bugzilla::Search(%options);
-  my ($data) = $search->data;
+
+  # Execute the query.
```
cgsheeh, line 695: This seems like something which should go in a well-named function, since it's a little confusing at first glance. It's also copied code from `buglist.cgi`, so I think it makes sense to put it somewhere that it can be re-used in both places.

cgsheeh (approved):

PR #2447, opened 2025-05-23
Bug 1968044 - BMO ETL: Change the way export loops through the row data so as to not allow on total and current count when tables may actually shrink before the table is fully exported

dklawren:
Creating transactions around the SELECTS in [bug 1967247](https://bugzilla.mozilla.org/show_bug.cgi?id=1967247) failed and was reverted. It was generating ERROR: Unable to create lock in database: Another process has set a lock. Exiting errors on different tables after 3 retries of running the job. My original thinking was that wrapping the SELECTs in a transaction, I would get a consistent view of each table regardless of changes being made on the production DB.

The current code performs a SELECT on a table to get the total count of rows and then a variable $count is incremented for each row. When $total == $count we exit the loop and move on to the next table. This would work if we had a consistent snapshot of the current table and the count of rows does not change during the loop.

I reverted the changes in [bug 1967247](https://bugzilla.mozilla.org/show_bug.cgi?id=1967247) and propose in this bug to change the way we loop over the data that does not rely on total rows and just exits the loop when there is no more rows to fetch. This way if a tables total rows can decrease over time, such as the flags table, we still exit the loop properly at the right time and not loop forever.

Also added addition debugging

### DIFF
--- a/extensions/BMO/bin/export_bmo_etl.pl
+++ b/extensions/BMO/bin/export_bmo_etl.pl
@@ -136,7 +136,7 @@ delete_lock();
 
 sub process_bugs {
   my $table_name  = 'bugs';
-  my $count       = 0;
+  my $total_count = 0;
   my $last_offset = 0;
   my @results     = ();
 
@@ -172,7 +172,7 @@ sub process_bugs {
 
       if (any { $obj->product eq $_ } EXCLUDE_PRODUCTS) {
         $excluded_bugs{$obj->id} = 1;
-        $count++;
+        $total_count++;
         next;
       }
 
@@ -239,14 +239,14 @@ sub process_bugs {
 
     push @results, $data;
 
-    $count++;
+    $total_count++;
 
   # Send a batch of data, reset counts and then run query for the next batch of rows
-    if ($count % API_BLOCK_COUNT == 0) {
-      logger("count: $count, last offset: $last_offset", DEBUG_OUTPUT);
+    if (scalar @results >= API_BLOCK_COUNT) {
+      logger("total count: $total_count, last offset: $last_offset", DEBUG_OUTPUT);
 
       # Send the rows to the server
-      send_data($table_name, \@results, $count);
+      send_data($table_name, \@results, $total_count);
 
       @results = ();
       $last_offset += API_BLOCK_COUNT;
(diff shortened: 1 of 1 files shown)


dklawren:
I installed this new code on bugzilla-dev for testing today. After a successful run, I went into BigQuery and did a simple count comparison of all tables with the changes and before the changes. The numbers matched up.

dklawren:
It seems like a lot of change but half of it is indention changes. The other half is a reduction in number of lines as I am only doing a single while loop for each table instead of one inside another. If you want, we can get on a zoom call and go through it in more detail.

extensions/BMO/bin/export_bmo_etl.pl
```diff
@@ -461,49 +506,55 @@ sub process_tracking_flags {
...
-      }
+    push @results, $data;
 
-      push @results, $data;
+    $count++;
 
-      $count++;
-    }
+  # Send a batch of data, reset counts and then run query for the next batch of rows
+    if ($count % API_BLOCK_COUNT == 0) {
```
cgsheeh, line 535: Can we avoid using `$count` and just check the size of `$results` instead? It seems that when we have enough results, we want to submit. Something like Python's `len(results)` seems like a better check here.

Also, could we directly check the result size instead of using modulo, like `len($results) >= API_BLOCK_COUNT`? If the logic ever changed in the future and more than one entry was added in a loop, this would skip submission (ie if `API_BLOCK_COUNT = 100` and then we had 101 results, we wouldn't submit).

cgsheeh (approved):
LGTM, I left a few suggestions but since you've already tested I won't block you on them. :+1:

extensions/BMO/bin/export_bmo_etl.pl
dklawren: > Can we avoid using `$count` and just check the size of `$results` instead? It seems that when we have enough results, we want to submit. Something like Python's `len(results)` seems like a better check here.
> 
> Also, could we directly check the result size instead of using modulo, like `len($results) >= API_BLOCK_COUNT`? If the logic ever changed in the future and more than one entry was added in a loop, this would skip submission (ie if `API_BLOCK_COUNT = 100` and then we had 101 results, we wouldn't submit).

Actually I need to keep $count for running in test mode as it uses the current count to create unique file names on the file system. I will change it to $total_count and change the module operator to (scalar @results >= API_BLOCK_COUNT) instead

In the future I may just get rid of test mode altogether and then will remove $total_count

cgsheeh (approved):
LGTM.

PR #2448, opened 2025-05-27
Bug 1967334 - BMO ETL: Refactor bmo etl export script to run as a service instead a cron job

dklawren:


dklawren:
No longer going to do this.

PR #2451, opened 2025-05-29

dklawren:
Redo of pr #2441 This with proper references.

### DIFF
--- a/extensions/PhabBugz/lib/Util.pm
+++ b/extensions/PhabBugz/lib/Util.pm
@@ -53,18 +53,26 @@ use constant LEGACY_APPROVAL_MAPPING => {
 
 # Set approval flags on Phabricator revision bug attachments.
 sub set_attachment_approval_flags {
-  my ($attachment, $revision, $flag_setter, $phab_user) = @_;
+  my ($attachment, $revision, $phab_user) = @_;
+  my $bmo_user = $phab_user->bugzilla_user;
 
   my $revision_status_flag_map = {
     'abandoned'       => '-',
     'accepted'        => '+',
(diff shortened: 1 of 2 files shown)


extensions/PhabBugz/lib/Util.pm
```diff
@@ -53,28 +53,38 @@ use constant LEGACY_APPROVAL_MAPPING => {
...
   };
 
-  my $status = $revision_status_flag_map->{$revision->status};
+  # Find the current review status of the revision changer
+  my $status          = undef;
+  my $reviewer_status = undef;
+  foreach my $reviewer (@{$revision->reviews}) {
+    if ($reviewer->{user}->id == $phab_user->id) {
+      $reviewer_status = $reviewer->{status};
+      $status          = $revision_status_flag_map->{$reviewer_status};
```
cgsheeh, line 75: Nit: we can `break` here.

extensions/PhabBugz/lib/Util.pm
```diff
@@ -128,10 +138,18 @@ sub set_attachment_approval_flags {
   if (!@old_flags && $status ne 'X') {
     my $approval_flag = Bugzilla::FlagType->new({name => $approval_flag_name});
     if ($approval_flag) {
-      if ($flag_setter->can_change_flag($approval_flag, 'X', $status)) {
+      # If setting to + then at least one accepted reviewer needs to be a release manager.
```
cgsheeh, line 141: Nit: comment is slightly incorrect:

```suggestion
      # If setting to + then the Phabricator user needs to be a release manager.
```

cgsheeh (approved):

PR #2456, opened 2025-06-12
Bug 1971281 - Webhooks fail with anything other than 200 response code

dklawren:
This uses $response->is_success which under the hood considers any code 2xx to be a success. This is better than just looking for 200.


andrew-est:
Thanks looks good to me

cgsheeh (approved):

PR #2457, opened 2025-06-13

dklawren:
Let's try this again. Last time it was not setting the approval-mozilla-*? flag when the revision was first created as it was only looking at the reviewers statuses and there are none when the revision is first created. So this time we pass in $is_new to the function and if that is true, we take the revision status instead of looking at the reviewer list. In this case, the revision status will be 'draft' and maps properly to '?'.

### DIFF
--- a/extensions/PhabBugz/lib/Util.pm
+++ b/extensions/PhabBugz/lib/Util.pm
@@ -53,28 +53,46 @@ use constant LEGACY_APPROVAL_MAPPING => {
 
 # Set approval flags on Phabricator revision bug attachments.
 sub set_attachment_approval_flags {
-  my ($attachment, $revision, $flag_setter, $phab_user) = @_;
+  my ($attachment, $revision, $phab_user, $is_new) = @_;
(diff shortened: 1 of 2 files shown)


extensions/PhabBugz/lib/Util.pm
```diff
@@ -53,28 +53,46 @@ use constant LEGACY_APPROVAL_MAPPING => {
...
-  my ($attachment, $revision, $flag_setter, $phab_user) = @_;
+  my ($attachment, $revision, $phab_user, $is_new) = @_;
+  my $bmo_user = $phab_user->bugzilla_user;
 
   my $revision_status_flag_map = {
     'abandoned'       => '-',
     'accepted'        => '+',
+    'accepted-prior'  => '+',
     'changes-planned' => 'X',
     'draft'           => '?',
```
cgsheeh, line 64: Looking at this again, I wonder if `draft` should map to `X`, since the `draft` state implies the revision isn't ready for approval.

cgsheeh (approved):

PR #2476, opened 2025-08-20
Bug 1896310 - Attached images are not displayed inline when the bug has no keywords

kyoshino:
[Bug 1896310 - Attached images are not displayed inline when the bug has no keywords](https://bugzilla.mozilla.org/show_bug.cgi?id=1896310)

Remove the unnecessary check for `bug_keywords` introduced in #2132 and simplify the keyword test condition.

### DIFF
--- a/skins/standard/buglist.css
+++ b/skins/standard/buglist.css
@@ -213,15 +213,18 @@ td.bz_total {
   }
 
   .bz_buglist.responsive tbody td {
-    display: flex;
+    display: flow-root;
     gap: 4px;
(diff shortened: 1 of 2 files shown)


dklawren:
> Will figure out why the selenium test fails. Seems unrelated to the change, but not sure.

Yeah was looking at that myself just now. Something to with the check where we turn off attachment display in the admin params, then the attachment should only be downloadable and not viewable. Maybe we should also not display attachments inline if the attachment display parameter is set to off. If you click on the details link, it already blocks displaying the attachment.

PR #2478, opened 2025-08-21
Bug 1984452 - When processing the list of reviewers to assign to a new revision, PhabBugz needs to skip a reviewer if they are the author of the revision

dklawren:
Before any call to set_new_reviewer we also check to make sure the reviewer is not the revision author.

### DIFF
--- a/extensions/PhabBugz/lib/Util.pm
+++ b/extensions/PhabBugz/lib/Util.pm
@@ -392,7 +392,8 @@ PROJECT: foreach my $project (@review_projects) {
     if (scalar @project_members == 1) {
       my $lone_reviewer = $project_members[0];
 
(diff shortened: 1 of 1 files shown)


extensions/PhabBugz/lib/Util.pm
```diff
@@ -428,7 +430,9 @@ PROJECT: foreach my $project (@review_projects) {
 
     # If we still have not found a reviewer and there is no member that was the
     # last reviewer, then just pick the first member in the list
-    if (!$last_reviewer_phid) {
+    if (!$last_reviewer_phid
```
cgsheeh, line 433: As an aside: do we actually need this check? If we removed this, the first reviewer would still be set in the loop over the group members below.

cgsheeh (approved):
LGTM.

PR #2491, opened 2025-09-26
Bug 1988115 - In dependency graph show VERIFIED FIXED same as RESOLVED

dklawren:


dklawren:
> After reading the commit message, I don't see where we check if resolution "also includes VERIFIED", it seems like we're only looking to see if the value is non-empty. Is that part still accurate?

Any closed status such AS RESOLVED, VERIFIED, etc. will always have a non-empty value for resolution so I am just checking to see if resolution is non-empty. We definitely change the bug summary and commit message to make it more obvious what is happening.

cgsheeh (approved):

PR #2492, opened 2025-09-26
Bug 1987854 - Make name for advisory credits more clear

dklawren:
add a note to the sign-up/profile editing page to let people know that whatever they put in the "Your real name:" field will be used on any advisory credits. Also on the user preferences page where they can change their current real name.


cgsheeh (changes requested):
Maybe we should link to the advisory credits as an example?

cgsheeh (approved):

PR #2551, opened 2026-02-23
Bug 1995464 - Overhaul guided bug entry form

kyoshino:
[Bug 1995464 - Overhaul guided bug entry form](https://bugzilla.mozilla.org/show_bug.cgi?id=1995464)

Apologies for the long delay. 🙇🏼 🙇🏼 🙇🏼  Here’s a complete overhaul of the guided bug entry form.

- Fresh, responsive design
- Additional product options
- Standard Markdown editor
- Redesigned attachment selector — This also applies to the standard bug entry form and the new attachment page, solving [Bug 1982282](https://bugzilla.mozilla.org/show_bug.cgi?id=1982282)

These features that require backend changes are missing in this PR. I’ll follow up later.

- User preference to change the default entry form
- Ability to add multiple attachments


extensions/GuidedBugEntry/web/js/guided.js
```diff
@@ -5,259 +5,395 @@
...
+    GuidedBugEntryWebDevPage.onInit();
+    GuidedBugEntryProductPage.onInit();
+    GuidedBugEntryOtherDupesPage.onInit();
+    GuidedBugEntryFormPage.onInit();
+    GuidedBugEntryFormPage.initHelp();
 
-      if (_component) {
-        params.set('component', _component);
-      }
+    const noSetHistory = !window.history.state;
```
dklawren, line 161: **Blocking — this condition is inverted, and it breaks all deep links.**

`master` had `if (!window.history.state) { ...replaceState... }` — seed the state *when there is none*. Here it becomes:

```js
const noSetHistory = !window.history.state;
if (!noSetHistory) {   // === if (window.history.state)
```

So the seeding block now runs only when state already exists (where it's redundant) and is skipped on a cold load (where it's needed). On a fresh page load `history.state` stays `null`, `onStateChange` reads `history.state ?? {}` → `{}`, and the `product` / `component` query params are silently dropped.

What this breaks:
- The link this PR adds in `products.html.tmpl` — `enter_bug.cgi?format=guided&product=Focus&component=Security%3A+iOS` ("use this form for security issues") — lands on the product picker instead. `enter_bug_start` (`Extension.pm:47`) serves the guided template regardless of `product`, so this JS is the only thing that can honour it.
- Any existing bookmark or external link to `enter_bug.cgi?format=guided&product=X`.
- Reloading mid-flow loses your place.
- `setStep(undefined)` falls through to `default:`, which recurses as `setStep(this.defaultStep)` **without forwarding `noSetHistory`** — so a spurious `pushState` fires on every page load and Back needs two presses to leave the page.

Suggested fix: `if (noSetHistory) { ...replaceState... }`, and pass `noSetHistory` through the `default:` branch of `setStep`.

Separately, `master`'s legacy `#h=step|product|component` hash parsing was dropped here. `extensions/BMO/template/en/default/bug/create/create-mdn.html.tmpl:198` still emits one of those links.

extensions/GuidedBugEntry/web/js/guided.js
```diff
@@ -5,259 +5,395 @@
...
-    var result = [];
+  /**
+   * Get the currently selected product name and related products.
+   */
+  static get productNameAndRelated() {
+    const { productName } = this;
 
-    var name = this.getName();
-    result.push(name);
+    return [productName, ...(products[productName].related ?? [])];
```
dklawren: **Blocking — throws for any product not listed in `products.js`.**

`products[productName]` is `undefined` for every product that isn't one of the ~8 entries in `products.js`, so `.related` raises a `TypeError`. `master` guarded this:

```js
if (products[name] && products[name].related) { ... }
```

This getter is evaluated inside `doSearch`'s inner `try` (line ~767), so the throw is swallowed into `data = { error: true }` and the user just sees *"An error occurred while searching for similar issues"*. In other words duplicate detection is dead for every product reached through **Other Products** — i.e. most of BMO.

It also hits a path this PR touches directly: `Core` was removed from `products.js` in this PR, but the webdev step still selects it (`product_name="Core"` in `guided.html.tmpl`).

```suggestion
    return [productName, ...(products[productName]?.related ?? [])];
```

extensions/GuidedBugEntry/web/js/guided.js
```diff
@@ -266,501 +402,825 @@ var product = {
...
-    if (document.getElementById('component_select').length == 0)
+
+    if (this.$componentSelect.length === 0) {
       this.onProductUpdated();
-    this.onFileChange();
-    this._mandatoryFields.forEach((id) => {
-      document.getElementById(id).classList.remove('missing');
+    }
+
+    new Bugzilla.AttachmentSelector({
```
dklawren: **Blocking — a new `AttachmentSelector` is constructed on every `onShow()`, silently dropping the user's attachment.**

`onShow()` runs on every transition into the form step, including `popstate`. Each construction calls `#renderUI()`, which does `$placeholder.innerHTML = ...` and destroys `#att-data` (the base64 payload) and `#att-filename`.

Repro: attach a file → click **Change** next to Product → pick the product again → the file is gone, but `#att-desc-section` is still visible with the stale filename sitting in `description`.

Each pass also leaks a `submit` listener on `#bug-form` plus a `click` listener on the submit button (`attachment.js:406`), which are never removed and end up operating on detached nodes.

Suggested fix — construct it once:

```js
this.selector ??= new Bugzilla.AttachmentSelector({ ... });
```

extensions/BugModal/web/bug_modal.js
```diff
@@ -757,13 +757,7 @@ $(function() {
...
                     let invalid = false;
 
-                    if ($input.id === 'att-data') {
-                        invalid = !$input.value.trim()
-                            && !document.querySelector('#att-textarea').value.trim()
-                            && !document.querySelector('#att-file').files.length;
-                        document.querySelector('#att-dropbox')?.classList
-                            .toggle('attention', invalid);
-                    } else if ($input.type === 'text') {
+                    if ($input.type === 'text') {
```
dklawren, line 806: **Blocking — removing the `#att-data` branch means a required attachment no longer blocks submit on the BugModal New Bug form.**

The deleted branch was the only thing enforcing this here, and `AttachmentSelector` didn't carry over `master`'s `update_validation()` / `setCustomValidity()`. The replacement (`AttachmentSelector.validate`) hooks the form's `submit` event and the submit button's `click` — but neither fires on this path:

- `$form` is a **native** element (`const $form = document.querySelector('#changeform')`, line 119), and this handler ends in `$form.submit()` (line ~804). `HTMLFormElement.submit()` fires no `submit` event and skips constraint validation entirely.
- `editMode` is `true` on the create page (`let editMode = isNewBug || false`, line 138), so that's the branch taken.
- The `click` listener does run, but this handler already calls `event.preventDefault()` unconditionally and then proceeds regardless of what other listeners did.

Repro: **Attach New File** → fill in a Description → leave the dropbox empty → **Submit**. "You must provide an attachment." appears and the bug is created anyway, with no attachment.

Either restore an equivalent check here, or have `AttachmentSelector` reinstate `setCustomValidity()` so `$form.checkValidity()` on line 784 catches it.

js/attachment.js
```diff
@@ -281,226 +281,334 @@ var bz_attachment_form;
...
-    if (!this.description_override) {
-      this.$description.value = file.name;
+      this.$file.value = '';
+      this.$data.value = '';
+      this.$filename.value = '';
     }
 
-    this.$description.select();
-    this.$description.focus();
+    this.showPreview(file, isText);
```
dklawren, line 590: **An oversized file shows a preview and then files a bug with no attachment.**

When `checkFileSize` fails, `$data` / `$filename` are cleared — but `showPreview(file, isText)` runs unconditionally, so the user sees a preview and reasonably assumes the file is attached.

`master` compensated with `this.$data.required = true` (old `attachment.js:435`), which blocked native form submission. That's gone, and on any form where the attachment is optional (guided form, BugModal/standard New Bug without "wants attachment") `validate()` computes `invalid === false` and actively *wipes* the size error on its way through:

```js
this.$errorMessage.textContent = invalid ? '...' : '';
this.$errorMessage.hidden = !invalid;
this.$dropbox.classList.toggle('invalid', invalid);
```

So the warning disappears at submit time and the bug is filed with nothing attached. Suggest skipping `showPreview` (or clearing it) when `checkFileSize` returns `false`, and not clearing an existing size error in `validate()`.

js/attachment.js
```diff
@@ -512,280 +620,627 @@ Bugzilla.AttachmentForm = class AttachmentForm {
...
+   * the submit button.
+   * @returns {boolean} `true` if the form is valid and can be submitted, `false` otherwise.
    */
-  textarea_onpaste(event) {
-    const image = [...event.clipboardData.items].find(item => item.type.match(/^image\/(?!vnd)/));
+  validate(event) {
+    const invalid =
+      this.required &&
+      !this.$data.value.trim() &&
+      !this.$file.length &&
```
dklawren: `$file` is the `<input type="file">`, and `HTMLInputElement` has no `length` property — so this is `!undefined`, always `true`, and the clause is dead. `master`'s equivalent in `bug_modal.js` (removed in this PR) had it right: `document.querySelector('#att-file').files.length`.

```suggestion
      !this.$file.files.length &&
```

This matters because `processFile` clears `$file.value` and relies on the **asynchronous** `dataReader` to populate `$data`. Pick a large file and submit immediately and `$data.value` is still `''`, so the intended `$file` fallback never fires: on `attachment.cgi?action=enter` (`required=1`) you get a spurious "You must provide an attachment", and on the guided form the POST goes through with an empty `data_base64` and the attachment is lost.

skins/standard/global.css
```diff
@@ -1766,6 +1817,13 @@ input.requestee {
   line-height: var(--line-height-comfortable);
 }
 
+[aria-invalid="true"],
```
dklawren, line 1835: Moving this block out of `extensions/BugModal/web/bug_modal.css` (two pages) into `global.css` (every page), and adding a `box-shadow`, has a much wider blast radius than I think is intended.

`input:invalid` matches a `required` empty input **on first paint, before the user has touched it**. So every page now renders these in the error state:

- `template/en/default/account/auth/login-small.html.tmpl:114` — the header login dropdown's `required` email input, which is present on **every page** for logged-out users
- `account/auth/login.html.tmpl:70`, `account/prefs/mfa.html.tmpl:145,222`, `account/reset-password.html.tmpl:111,119,125`, `account/email/confirm-new.html.tmpl:52`, `pages/new_release.html.tmpl`, plus BMO's `create-crm` / `create-legal` forms

Suggest `input:user-invalid` instead — it only matches after interaction or a submit attempt, it's baseline since 2023, and this file already relies on `:has()`. Alternatively, scope the rule to the forms that opt into it.

extensions/GuidedBugEntry/web/style/guided.css
```diff
@@ -65,170 +220,304 @@ ul.product-list > li {
...
+  width: 100%;
 }
 
-#submitTD {
-  text-align: right;
+#bug-form .required-star {
+  color: var(--required-label-color);
+}
+
+#bug-form :is(.checkbox, .radio) {
```
dklawren, line 415: This author rule (specificity `1,1,0`) beats the UA stylesheet's `[hidden] { display: none }`, so `hidden` has no effect on anything carrying `.checkbox` or `.radio` inside `#bug-form`.

`#firefox-android-row` is `<div class="checkbox" id="firefox-android-row">`, which means `conditionalDetails.forEach(cond => document.getElementById(cond.id).hidden = !cond.check())` (`guided.js:1010`) can never hide it. "This is a problem with Firefox on mobile or tablet" shows up for Thunderbird, Bugzilla, and every other product.

Same fix as elsewhere in this file (`#dupe-header .messages li:not([hidden])`, `#att-selector [hidden]`) — add:

```css
#bug-form :is(.checkbox, .radio)[hidden] {
  display: none;
}
```

extensions/GuidedBugEntry/web/js/guided.js
```diff
@@ -266,501 +402,825 @@ var product = {
...
+   * Perform the duplicate search.
+   */
+  static async doSearch() {
+    if ([...this.summary].length < 4) {
+      const message = 'The summary must be at least 4 characters.';
+      this.$summary.setAttribute('aria-invalid', 'true');
+      this.$summary.setAttribute('aria-errormessage', 'dupe-summary-error');
+      this.$summary.insertAdjacentHTML(
+        'afterend',
+        `<div id="dupe-summary-error" class="error-message">${message}</div>`,
```
dklawren: Two problems in this error path:

1. No dedupe guard — unlike `checkMissing()` (line ~1300), which checks `if (!$field.parentElement.querySelector('.error-message'))` first. Click **Find similar issues** three times with `abc` and you get three identical `<div id="dupe-summary-error">` elements, i.e. duplicate IDs and a stack of repeated messages.
2. The cleanup on the success path (line 727) looks for `.error`, not `.error-message`, so it never matches — see the comment there.

Net effect: the "must be at least 4 characters" message can never be cleared once shown.

extensions/GuidedBugEntry/web/js/guided.js
```diff
@@ -266,501 +402,825 @@ var product = {
...
-  _doSearch: async function() {
-    if (dupes.getSummary().length < 4) {
-      alert('The summary must be at least 4 characters long.');
       return;
     }
-    dupes._elSummary.blur();
+
+    this.$summary.setAttribute('aria-invalid', 'false');
+    this.$summary.removeAttribute('aria-errormessage');
+    this.$summary.parentElement.querySelector('.error')?.remove();
```
dklawren: Class mismatch — the error div is inserted with `class="error-message"` on line 719, so `.error` never matches and this cleanup is a no-op. Same bug at line 1014.

```suggestion
    this.$summary.parentElement.querySelector('.error-message')?.remove();
```

extensions/GuidedBugEntry/web/js/guided.js
```diff
@@ -266,501 +402,825 @@ var product = {
...
-        visibleCount++;
-        document.getElementById(cond.id).classList.remove('hidden');
-      }
-      else {
-        document.getElementById(cond.id).classList.add('hidden');
-      }
+    this.requiredFields.forEach((el) => {
+      el.removeAttribute('aria-invalid');
+      el.removeAttribute('aria-errormessage');
+      el.parentElement.querySelector('.error')?.remove();
```
dklawren: Same class mismatch as line 727 — `checkMissing()` inserts `class="error-message"` (line 1303), so `.error` never matches.

Because of this, failing validation, navigating back to the dupes step and returning to the form leaves the stale "This field is required." messages on screen — and since `aria-invalid` *is* correctly removed just above, the visible message now contradicts the field's state.

```suggestion
      el.parentElement.querySelector('.error-message')?.remove();
```

extensions/GuidedBugEntry/web/js/guided.js
```diff
@@ -266,501 +402,825 @@ var product = {
...
     } else {
-      // no default version, select an empty value to force a decision
-      var opt = new Option('', '');
-      try {
-        // standards
-        elVersions.add(opt, elVersions.options[0]);
-      } catch(ex) {
-        // IE only
-        elVersions.add(opt, 0);
+      // Fallback to 'unspecified' if available
```
dklawren, line 1400: Behaviour change worth a second look: when no version was detected and there's no `unspecified` option, the `<select>` keeps its implicit first-option selection, and `onVersionChange($versions.value)` a few lines down writes that into the hidden `version` field. `checkMissing` then sees a non-empty value, so the bug is filed against an arbitrary version the user never picked.

`master` inserted a blank option and set `value = ''` specifically to force a decision here. Suggest restoring that.

Related: `#version-section` is hidden with the `hidden` attribute, but `requiredFields` filters on `:not([aria-hidden="true"])`. If a product ever has no active versions, the hidden-and-empty select fails validation and `focus()` on it is a no-op — the form silently refuses to submit with nothing shown to the user.

extensions/GuidedBugEntry/web/js/guided.js
```diff
@@ -266,501 +402,825 @@ var product = {
...
+      this.$summary.select();
+    });
+  }
+
+  /**
+   * Show the product support message.
+   */
+  static showProductSupport() {
+    const { productName } = GuidedBugEntryProductPage;
+    const elSupportId = `product-support-${productName.replace(' ', '-').toLowerCase()}`;
```
dklawren: Dead code carried over from `master` — nothing has a `product-support` class. The template has `id="product-support"` (singular element), and the real show/hide logic lives in `setProduct()` at line ~371.

Safe to delete this method along with its callers in `reset()` and `onShow()`.

js/attachment.js
```diff
@@ -281,226 +281,334 @@ var bz_attachment_form;
...
+              <span id="att-text-remove-button" class="att-remove-button" tabindex="0" role="button"
+                  aria-label="Remove attachment">
+                <span class="icon" aria-hidden="true"></span>
+              </span>
+            </div>
+            <div hidden id="att-preview">
+              <input id="att-filename" type="hidden" name="filename">
+              <textarea hidden id="att-data" name="data_base64" aria-hidden="true"
+                  aria-invalid="false" aria-errormessage="att-data-error"></textarea>
+              <figure role="img" aria-labelledby="att-preview-name" itemscope
```
dklawren, line 369: Two dangling ARIA references in the generated markup:

1. `aria-labelledby="att-preview-name"` — line 361 gives the figcaption `class="att-preview-name"`, not `id`. `master` used `id="att-preview-name"`, so the figure has lost its accessible name.
2. Line 355's `aria-errormessage="att-data-error"` points at a `<div id="att-data-error">` that this PR deletes from `createformcontents.html.tmpl`.

Either restore the `id` on the figcaption and re-add the error div, or drop the two ARIA attributes.

extensions/GuidedBugEntry/template/en/default/guided/guided.html.tmpl
```diff
@@ -121,109 +122,85 @@ tabbed browsing or the location bar)
...
-      </span>
-      [% IF security_bug_link %]
-        <span class="product-item">
-          If you are reporting a security issue that puts users at risk,
-          <b><a href="[% security_bug_link FILTER none %]">use this form</a>
-          instead.</b>
-        </span>
-      [% END %]
+    <li role="button" tabindex="0" class="product-item product-link"
+        data-link="[% link FILTER none %]" data-product="[% product_name FILTER html %]"
```
dklawren: Nit / defence in depth: `FILTER none` on an attribute value that feeds a navigation sink. `guided.js:186` does `location.href = link` with this value, so a `javascript:` URL here would execute.

Every `link` today is a template literal from `products.html.tmpl`, so this isn't exploitable as written — but it's one dynamic value away from being an XSS. Suggest `FILTER html` here plus a scheme check in the JS before assigning to `location.href`.

extensions/GuidedBugEntry/template/en/default/guided/guided.html.tmpl
```diff
@@ -11,108 +11,109 @@
...
+        <span>2</span>
+        Find similar issues
+      </div>
+      <div id="stepper-form" role="listitem">
+        <span>3</span>
+        Enter details
+      </div>
+    </div>
+  </header>
+  <div id="steps" hidden>
```
dklawren, line 59: The `<noscript>` block from `master` was dropped. `#steps` is `hidden` and only JS unhides it, so with JS disabled users now get a page header, a stepper and a footer link with no explanation of why the form is missing.

`master` pointed them at `enter_bug.cgi?format=__default__`. Cheap to keep — worth re-adding unless it was deliberate.

dklawren (changes requested):

extensions/GuidedBugEntry/web/js/guided.js
kyoshino: `create-mdn.html.tmpl` was removed in https://github.com/mozilla/bmo/commit/07495fc508e2a9f11a27b1a8f53a9395c46b4277 so it’s safe to drop the legacy hash parser.

extensions/GuidedBugEntry/template/en/default/guided/guided.html.tmpl
kyoshino: The page header already has a `<noscript>`:

> Please enable JavaScript in your browser to use all the features on this site.

And the standard bug form no longer works without JavaScript. Safe to drop the irrelevant message.

extensions/GuidedBugEntry/template/en/default/bug/create/comment-guided.txt.tmpl
```diff
@@ -8,21 +8,19 @@ Firefox for Android
 [% END %]
 
 [% IF cgi.param('bug_steps') %]
-Steps to reproduce:
+### Steps to reproduce
```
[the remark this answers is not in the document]
kyoshino: The GuidedBugEntry extension is only used on BMO and Markdown is enabled by default.

js/attachment.js
```diff
@@ -512,280 +649,650 @@ Bugzilla.AttachmentForm = class AttachmentForm {
...
+
+    if (!this.descriptionOverridden) {
+      this.description = isPatch ? 'patch' : isGhpr ? 'GitHub Pull Request' : '';
+    }
+
+    // `processFile()` runs this with `hasText: false`, so don't let it clear the required state
+    // on a form that demands an attachment
+    this.$description.setAttribute('aria-required', this.required || hasText);
+    this.$typeInput.value = isGhpr ? 'text/x-github-pull-request' : '';
+    this.updateIsPatch(isPatch);
```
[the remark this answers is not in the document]
kyoshino: Claude says this comment is false positive.

js/attachment.js
```diff
@@ -512,280 +661,652 @@ Bugzilla.AttachmentForm = class AttachmentForm {
...
-    const is_patch = !!text.match(/^(?:diff|---)\s/);
-    const is_ghpr = !!text.match(/^https:\/\/github\.com\/[\w\-]+\/[\w\-]+\/pull\/\d+\/?$/);
+    const hasText = !!text;
+    const isPatch = !!text.match(/^(?:diff|---)\s/);
+    const isGhpr = !!text.match(/^https:\/\/github\.com\/[\w\-]+\/[\w\-]+\/pull\/\d+\/?$/);
 
-    if (has_text) {
-      this.$file.value = this.$data.value = this.$filename.value = '';
-      this.update_content_type('text/plain');
+    if (hasText) {
```
dklawren, line 821: `textareaOnInput()` clears `$file` and `$data` but not `$filename`; the old `textarea_oninput()` cleared all three (`this.$file.value = this.$data.value = this.$filename.value = ''`). `validate()` (lines 922-926) treats a non-empty `$filename` as proof an attachment exists.

On the Create New Attachment page: browse to `foo.png`, then drag & drop text onto the dropbox (clears `$data`, keeps `$filename`), then empty the textarea. `validate()` passes, the form posts with no `data`/`attach_text`/`data_base64`, and `attachment.cgi:592` falls through to `$data = $filename = $data_fh` with `$data_fh` undef, so `Bugzilla::Attachment->create` dies and the user loses the form.

js/attachment.js
```diff
@@ -512,280 +661,652 @@ Bugzilla.AttachmentForm = class AttachmentForm {
...
+   * @param {string} [text] Text to be inserted.
    */
-  update_text(text = '') {
+  updateText(text = '') {
     this.$textarea.value = text;
-    this.textarea_oninput();
+    this.textareaOnInput();
 
-    if (text) {
+    if (text.trim()) {
```
dklawren, line 708: `updateText()` clears `readingFile` without calling `#abortReads()`, and the text branch of `dropboxOnDrop()` (688-695) doesn't abort either — only `processFile()` and `resetFields()` do.

Pick a large file (read starts), then immediately drop text: `textareaOnInput` zeroes `$data`, this line drops the `readingFile` guard, then the still-running `dataReader` completes and `dataReaderOnLoad()` (624) writes the old file's base64 back into `$data`. Emptying the textarea afterwards doesn't clear it (`hasText` is false), so submitting silently attaches the file the user thought they replaced.

js/attachment.js
```diff
@@ -512,280 +661,652 @@ Bugzilla.AttachmentForm = class AttachmentForm {
...
-    this.$preview_image.src = file.type.match(/^image\/(?!vnd|svg)/) ? URL.createObjectURL(file) : '';
-    this.$preview.hidden = false;
+  validate(event) {
+    // Nothing to validate when the selector is not in use, e.g. when the user opted out of adding
+    // an attachment on the New Bug page
+    if (this.$placeholder.closest('[hidden]')) {
+      return true;
+    }
+
+    if (this.sizeError) {
```
dklawren, line 907: `sizeError` is sticky: it's only reset by `processFile()`, `updateText()` with non-empty text, and `clearError()`. Typing goes through `textareaOnInput()`, which never touches it.

Pick an over-limit file (error shown), click "Enter Text", type valid content, press Submit: `validate()` returns false here without setting a message, so nothing visibly happens — recoverable only via Remove, which discards the typed text. Worse on the BugModal New Bug page: "Don't Attach File" calls `updateRequirements(false)` (885) which hides `$errorMessage` but leaves `sizeError` true, so after re-opening the section submission fails with no visible error at all.

js/attachment.js
```diff
@@ -281,226 +281,375 @@ var bz_attachment_form;
...
+            <span class="icon" aria-hidden="true"></span>
+            ${isDesktop ? `<span>Drag &amp; drop a file here, or</span>` : ''}
+           <span class="button-group">
+              <button type="button" id="att-browse-button">Browse File</button>
+              <button type="button" id="att-paste-button">Paste Text or Image</button>
+              <button type="button" id="att-enter-button">Enter Text</button>
+              <button type="button" id="att-capture-button">Take Screenshot</button>
+            </span>
+          </div>
+          <input hidden id="att-file" type="file">
```
dklawren, line 353: `#att-file` lost the `name="data"` it had in `createformcontents.html.tmpl`, and `processFile()` clears `$file.value` for every input method (586), so file-picker uploads no longer use native multipart upload — they go into `data_base64`. The old `file_onchange()` passed `transferred = false` precisely to keep the native path.

The payload grows ~33% and the whole file is buffered as a JS string. Against Mojo's default 16 MiB `max_message_size` (see `Bugzilla/App/Controller/CSPReport.pm:47`), a `maxattachmentsize` above ~11.5 MB now 413s on uploads that previously worked.

extensions/GuidedBugEntry/web/js/guided.js
```diff
@@ -266,501 +531,886 @@ var product = {
...
+   * Called whenever a file is processed by `AttachmentSelector`. Update the Description, Content
+   * Type, Patch checkbox, etc. based on the file properties and content.
+   * @param {object} params An object with the following properties:
+   * @param {File} params.file A processed `File` object.
+   * @param {string} params.type A MIME type to be selected in the Content Type field.
+   * @param {boolean} params.isPatch `true` if the file is detected as a patch, `false` otherwise.
+   */
+  static onAttachmentProcessed({ file, type, isPatch }) {
+    this.$attDescSection.hidden = false;
+    this.$attDescription.value = file.name;
```
dklawren, line 1218: This assigns `file.name` without checking `attDescOverridden`, unlike `onAttachmentTextUpdated()` (1238) and `AttachmentForm.onAttachmentProcessed()` (`js/attachment.js:1050`).

Enter attachment text, type a custom File Description, then drag a file onto the dropbox: the description silently reverts to the filename, defeating the flag's purpose.

extensions/GuidedBugEntry/web/js/guided.js
```diff
@@ -5,259 +5,524 @@
...
+    // init steps
+    GuidedBugEntryProductPage.onInit();
+    GuidedBugEntryOtherDupesPage.onInit();
+    GuidedBugEntryFormPage.onInit();
+    GuidedBugEntryFormPage.initHelp();
 
-      if (_component) {
-        params.set('component', _component);
-      }
+    const noSetHistory = !window.history.state;
```
dklawren, line 161: `!window.history.state` is false on reload once the user has left the first step, so `onStateChange(false)` -> `setStep(step, false)` calls `history.pushState()` during page load, appending a duplicate entry.

Reload while on the dupes step: the first Back press re-renders the same step instead of returning to product selection.

dklawren (changes requested):
Everything checks out from manual testing. Review the following claude comments and once resolved or disputed, we will be ready to go with this.

kyoshino:
Claude said all the comments made sense and solved them 😅

dklawren:
One last item from the final review pass. Everything else checks out, with all 27 threads verified as actually applied in code, and both disputes holding up.

**`js/attachment.js:1122`: `onAttachmentPasted()` unconditionally unchecks *and disables* the patch checkbox.**

This was Copilot's *suppressed* comment #2, so it never became a thread and was never triaged either way.

Repro: paste a diff with the **Paste Text or Image** button. `AttachmentTextUpdated` fires first and correctly checks the patch box. `AttachmentPasted` then fires immediately after and clears it with `disabled = true`. The attachment is filed as plain `text/plain` with description "patch", and the checkbox is disabled so it cannot be corrected without Remove plus retype. Dragging the same text in works fine (no `AttachmentPasted`), and the guided form is unaffected since it does not wire that handler.

The override only makes sense for image clipboard items, where the content genuinely cannot be inspected:

```js
  onAttachmentPasted({ items }) {
    if (items.some((item) => item.types.includes('image/png'))) {
      this.updateIsPatch(false, true);
    }
  }
```

Happy to take this as a follow-up bug instead if you would rather not respin.

kyoshino:
Done. Claude confirmed it was a regression from the rewrite.

dklawren (approved):

PR #2570, opened 2026-03-17
Bug 2023761 - [GITHUB] Allow use of individual api keys for pull requests and push comments instead of single share secret

dklawren:
Bug 2023761 - Replace single shared GitHub webhook secret with per-bot API keys

Instead of a single `github_pr_signature_secret` parameter shared across all
GitHub webhook integrations, each bot account now creates a Bugzilla API key
and uses it as their webhook secret. If a key is compromised, only that bot
account needs to be disabled — other integrations are unaffected.

Changes:
- Add `github-webhook-bot` system group; bot accounts in this group have their
  non-revoked API keys accepted as valid webhook secrets
- Rewrite `_verify_signature` in `Bugzilla/API/V1/Github.pm` to query all API
  keys for `github-webhook-bot` group members, with a fallback to the legacy
  `github_pr_signature_secret` parameter for backward compatibility
- Mark `github_pr_signature_secret` as deprecated in the admin UI
- Add API key for `github-automation@bmo.tld` and assign it to the new group
  in `generate_bmo_data.pl`
- Update webservices tests to sign webhook payloads with the bot account API
  key instead of the hardcoded shared secret
- Update setup documentation for both webhook endpoints

### DIFF
--- a/extensions/PhabBugz/Extension.pm
+++ b/extensions/PhabBugz/Extension.pm
@@ -137,7 +137,7 @@ sub object_before_create {
 
   # Only allow phab-bot to set content type to text/x-phabricator-request
   my $content_type = $params->{mimetype};
-  if ($content_type eq PHAB_CONTENT_TYPE
+  if ($content_type && $content_type eq PHAB_CONTENT_TYPE
     && !Bugzilla->request_cache->{allow_phab_revision_attachment})
   {
     Bugzilla->audit(
--- a/qa/t/1_test_bug_edit.t
+++ b/qa/t/1_test_bug_edit.t
@@ -40,7 +40,7 @@ check_page_load($sel, q{http://HOSTNAME/editgroups.cgi});
 $sel->title_is("Edit Groups");
 $sel->click_ok("link=Master");
 check_page_load($sel,
-  q{http://HOSTNAME/editgroups.cgi?action=changeform&group=31});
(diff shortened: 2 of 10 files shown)


extensions/PhabBugz/Extension.pm
```diff
@@ -137,7 +137,7 @@ sub object_before_create {
 
   # Only allow phab-bot to set content type to text/x-phabricator-request
   my $content_type = $params->{mimetype};
-  if ($content_type eq PHAB_CONTENT_TYPE
+  if ($content_type && $content_type eq PHAB_CONTENT_TYPE
```
cgsheeh, line 140: Unrelated change? In the future this should be in a separate PR.

qa/t/1_test_bug_edit.t
```diff
@@ -40,7 +40,7 @@ check_page_load($sel, q{http://HOSTNAME/editgroups.cgi});
 $sel->title_is("Edit Groups");
 $sel->click_ok("link=Master");
 check_page_load($sel,
-  q{http://HOSTNAME/editgroups.cgi?action=changeform&group=31});
+  q{http://HOSTNAME/editgroups.cgi?action=changeform&group=32});
```
cgsheeh, line 43: As a follow-up, can we get the group number with code so we don't need to bump it when we add a new group? This feels quite brittle and tedious to maintain.

cgsheeh (approved):

PR #2581, opened 2026-04-05
Bug 2028240 - Pre-check security group when cloning bugs across products

loganrosen:
When cloning a security bug into a different product (via "New/Clone... as a clone, in a different product"), the source product's security group name doesn't match the target product's security group, so the security checkbox ends up unchecked — creating a public clone of a security-sensitive bug.

**Fix:** When cloning across products, if the source bug had its product's default security group set, also add the target product's default security group to the form defaults. This pre-checks the security checkbox on the clone form.

The fix is guarded so it only activates when:
- Cloning across products (same-product clones are unaffected)
- The source bug actually had its security group set (public bugs don't get a false positive)
- The BMO extension's `default_security_group` method exists (`can()` + `eval` guards)

**Test:** `t/bmo/clone-security-groups.t` covers cross-product security clone, same-product clone (no duplication), and public bug clone (no false positive).

### DIFF
--- a/Bugzilla/Bug.pm
+++ b/Bugzilla/Bug.pm
@@ -4352,6 +4352,25 @@ sub in_group {
   return grep($_->id == $group->id, @{$self->groups_in}) ? 1 : 0;
 }
 
+# Returns extra group names to add when cloning $self into $target_product.
+# If the bug is in its source product's default security group and the target
+# product has a different one, the target group name is returned so the clone
+# stays secure (Bug 2028240).
+sub extra_security_groups_for_clone {
+  my ($self, $target_product) = @_;
+  return () if $self->product_id == $target_product->id;
+  return () unless $target_product->can('default_security_group');
+
+  my @clone_groups = map { $_->name } @{$self->groups_in};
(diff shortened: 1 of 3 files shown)


dklawren (changes requested):
The test just duplicates production logic instead of invoking it: The test copy-pastes the 12-line fix block from enter_bug.cgi into the test file. The test will pass even if the production code is deleted, because it's testing its own copy. The test should either:
- Extract the logic to a subroutine in enter_bug.cgi that the test can call, or
- Use the selenium test harness to exercise the actual enter_bug.cgi page
 
The test manually deletes hash keys to clear the product cache:
 ```
 delete $prod_b->{security_group_id};
 delete $prod_b->{default_security_group_obj};
 ```
This pokes directly at the internal hash structure of Bugzilla::Object. If the cached key   names change, this silently stops working. A safer approach would be to re-fetch the product object: $prod_b = Bugzilla::Product->new({id => $prod_b->id}).

loganrosen:
@dklawren Addressed your feedback, thanks!

dklawren (approved):
Thanks!

PR #2587, opened 2026-04-06
Bug 2029523 - Selenium tests flake with element click intercepted errors in CI

dklawren:


Bugzilla/Test/Selenium.pm
```diff
@@ -300,6 +307,12 @@ sub select_ok {
     $locator =~ s/\@id/\@name/;
     $element = $self->find_element($locator);
   }
+  # Scroll select element to center of viewport before clicking options to
+  # prevent fixed header/nav elements from intercepting the click.
+  if ($element) {
+    $element->execute_script(
```
cgsheeh: Ideally we put this logic in a function like `scroll_select_to_center` and re-use it instead of duplicating the code in both places.

cgsheeh (approved):

PR #2591, opened 2026-04-09
Bug 2030581 - Pass is_markdown when adding comment via attachment update REST API

ChunMinChang:
When using PUT /rest/bug/attachment/{id} with a 'comment' field, the resulting comment is always rendered as plain text, even when the system has use_markdown enabled.

**Root cause:** the update_attachment method in Bugzilla/WebService/Bug.pm does not pass is_markdown to $bug->add_comment(). The web UI path in attachment.cgi correctly sets is_markdown based on Bugzilla->params->{use_markdown}.

This means there is no way via the REST API to get a Markdown-rendered comment when updating an attachment (e.g. for sec-approval requests where both the questionnaire comment and the sec-approval? flag need to appear as a single entry).

**Fix:** accept an optional is_markdown parameter in update_attachment, default to the system use_markdown setting when not provided (matching attachment.cgi behavior), and pass it through to $bug->add_comment().

**Steps to reproduce:**
1. PUT /rest/bug/attachment/{id} with {"comment": "### Heading\n* **bold**: text", "flags": [...]}
2. The comment is stored and displayed as plain text, not Markdown

**Expected:** The comment should respect is_markdown (or default to the system use_markdown setting).

**Actual:** is_markdown is never passed to add_comment(); passing it as a top-level parameter causes: "The requested method 'Bugzilla::Attachment::set_is_markdown' was not found."

### What this patch does

The update_attachment WebService method does not pass is_markdown to add_comment(), so comments added through PUT /rest/bug/attachment/{id} are always rendered as plain text. The web UI path in attachment.cgi correctly sets is_markdown based on the system use_markdown parameter.

This patch:
- Accepts an optional is_markdown parameter in update_attachment
- Defaults to the system use_markdown setting when not provided (matching attachment.cgi behavior)
- Passes it through to $bug->add_comment()
- Documents the new parameter in the REST API docs

### DIFF
--- a/Bugzilla/WebService/Bug.pm
+++ b/Bugzilla/WebService/Bug.pm
@@ -1228,8 +1228,11 @@ sub add_comment {
     {
       isprivate => $params->{is_private},
       work_time => $params->{work_time},
-      is_markdown =>
-        ( defined $params->{is_markdown} ? $params->{is_markdown} : 0 )
+      is_markdown => (
+        defined $params->{is_markdown}
+          ? ($params->{is_markdown} ? 1 : 0)
+          : (Bugzilla->params->{use_markdown} ? 1 : 0)
+      )
     }
   );
 
--- a/Bugzilla/WebService/Bug.pm
+++ b/Bugzilla/WebService/Bug.pm
@@ -1006,11 +1006,17 @@ sub add_attachment {
   $dbh->bz_start_transaction();
   my $timestamp = $dbh->selectrow_array('SELECT LOCALTIMESTAMP(0)');
 
-  my $flags     = delete $params->{flags};
-  my $comment   = delete $params->{comment};
-  my $bug_flags = delete $params->{bug_flags};
+  my $flags       = delete $params->{flags};
+  my $comment     = delete $params->{comment};
+  my $is_markdown = delete $params->{is_markdown};
+  my $bug_flags   = delete $params->{bug_flags};
 
   $comment = $comment ? trim($comment) : '';
+
+  # Default to the system use_markdown setting, matching attachment.cgi.
+  if (!defined $is_markdown) {
+    $is_markdown = Bugzilla->params->{use_markdown} ? 1 : 0;
+  }
 
   foreach my $bug (@bugs) {
     my $attachment = Bugzilla::Attachment->create({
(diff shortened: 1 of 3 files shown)


Bugzilla/WebService/Bug.pm
```diff
@@ -1138,9 +1144,10 @@ sub update_attachment {
       $bug->add_comment(
         $comment,
         {
-          isprivate  => $attachment->isprivate,
-          type       => CMT_ATTACHMENT_UPDATED,
-          extra_data => $attachment->id
+          isprivate   => $attachment->isprivate,
+          type        => CMT_ATTACHMENT_UPDATED,
+          extra_data  => $attachment->id,
+          is_markdown => $is_markdown,
```
dklawren: Explicit validation (e.g., $is_markdown = $is_markdown ? 1 : 0) would make this robust against non-boolean input like "true" or 1.5.

` is_markdown => ($is_markdown ? 1 : 0)`

dklawren (changes requested):
While we are working on this, we should go ahead and fix other comment related methods that do not handle markdown params correctly. 

The add_comment WebService method (line 1224-1225) defaults is_markdown to 0 when not provided:
is_markdown => (defined $params->{is_markdown} ? $params->{is_markdown} : 0)

But this PR defaults update_attachment to Bugzilla->params->{use_markdown}. This means the same API caller gets different markdown behavior depending on whether they're using POST /rest/bug/{id}/comment vs PUT /rest/bug/attachment/{id}. The PR
description justifies it as "matching attachment.cgi behavior", which is correct.

Would you be able to make the other places honor the Bugzilla->params->{use_markdown} value as well?

Bugzilla/WebService/Bug.pm
ChunMinChang: thanks for the review! I've addressed this.

Bugzilla/WebService/Bug.pm
```diff
@@ -1088,12 +1088,18 @@ sub update_attachment {
     $bugs{$bug->id} = $bug;
   }
 
-  my $flags     = delete $params->{flags};
-  my $comment   = delete $params->{comment};
-  my $bug_flags = delete $params->{bug_flags};
+  my $flags       = delete $params->{flags};
+  my $comment     = delete $params->{comment};
+  my $is_markdown = delete $params->{is_markdown};
```
dklawren, line 1100: This version is better. But we also need to do the same thing for the bug->add_comment in the add_attachment() function as well juts like in update_attachment().

dklawren (commented):
After this requested change, we should be good to go.

Bugzilla/WebService/Bug.pm
ChunMinChang: Thanks for the check, I've added similar `$is_markdown` usage in `add_attachment()` in the latest commit.

dklawren:
Thanks for the work!

PR #2592, opened 2026-04-10
Bug 2018260 - "Fields You Can Search On" is blocking people from making it through quicksearch.html doc

kyoshino:
[Bug 2018260 - "Fields You Can Search On" is blocking people from making it through quicksearch.html doc](https://bugzilla.mozilla.org/show_bug.cgi?id=2018260)

Revised version of #2564.

- Include an active flag in the page data rather than modifying `quicksearch_map`
- Hide inactive fields by default
- Add a checkbox to toggle inactive fields
- Use the standard table styles

### DIFF
--- a/extensions/TrackingFlags/Extension.pm
+++ b/extensions/TrackingFlags/Extension.pm
@@ -90,9 +90,12 @@ sub page_before_template {
         $quicksearch_extra_field_names{$key} = {
           nicknames => $field_names->{$key}->{nicknames},
           db_field  => $db_field,
-          # A field is active if it’s an alias like `tracking-firefox-release` or if it’s a real
-          # field that’s active in the database
-          active    => $db_field =~ /_[a-z]+$/ || $active_flag_names{$db_field} ? 1 : 0,
+          # Check if the field is active by looking for a corresponding active tracking flag, e.g.
+          # `cf_tracking_firefox135`, or if it's one of the channel pronouns like
+          # `cf_tracking_firefox_release`, which are created by the BMO extension in
+          # install_update_db but not stored in the tracking flags table. The `$flag_pronoun_re`
+          # regex matches these pronoun fields precisely.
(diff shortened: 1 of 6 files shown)


extensions/TrackingFlags/Extension.pm
```diff
@@ -76,6 +76,31 @@ sub page_before_template {
...
+    foreach my $key (keys %$field_names) {
+      my $db_field = $field_names->{$key}->{db_field};
+
+      if ($db_field =~ /^cf_(tracking|status)_/) {
+        $quicksearch_extra_field_names{$key} = {
+          nicknames => $field_names->{$key}->{nicknames},
+          db_field  => $db_field,
+          # A field is active if it’s an alias like `tracking-firefox-release` or if it’s a real
+          # field that’s active in the database
+          active    => $db_field =~ /_[a-z]+$/ || $active_flag_names{$db_field} ? 1 : 0,
```
dklawren: Nit: To be more clear about precedence use parens here.
`active => ($db_field =~ /_[a-z]+$/ || $active_flag_names{$db_field}) ? 1 : 0,`

Also, the comment says this matches aliases like tracking-firefox-release, but the db_field being tested has underscores (e.g., cf_tracking_firefox_release). The pattern /_[a-z]+$/ matches any field ending with
  _[letters] — e.g., cf_tracking_firefox. It is worth adding a comment explaining why this pattern identifies an alias vs. a real field.

dklawren (changes requested):

extensions/TrackingFlags/Extension.pm
kyoshino: A simple regex was incorrectly matching `cf_tracking_firefox_relnote` and `cf_tracking_firefox_sumo`. I have exported the pronoun regex and used it here.

dklawren (approved):

PR #2593, opened 2026-04-10
Bug 2028222 - Pasting multi-line text after selecting multi-line text does not overwrite, but applies markup for link

kyoshino:
[Bug 2028222 - Pasting multi-line text after selecting multi-line text does not overwrite, but applies markup for link](https://bugzilla.mozilla.org/show_bug.cgi?id=2028222)

Adjust the regex to only match single-line URLs.


dklawren (approved):
Thanks!

PR #2594, opened 2026-04-10
Bug 2012634 - "Phabricator Revisions" table overflows on X axis on mobile

kyoshino:
[Bug 2012634 - "Phabricator Revisions" table overflows on X axis on mobile](https://bugzilla.mozilla.org/show_bug.cgi?id=2012634)

Modify the style so that the module can be scrolled horizontally rather than overflowing.


dklawren (approved):
Thanks!

PR #2598, opened 2026-04-15
Bug 1950453 - Cookie banner shows up as "light" even in dark color-scheme

kyoshino:
[Bug 1950453 - Cookie banner shows up as "light" even in dark color-scheme](https://bugzilla.mozilla.org/show_bug.cgi?id=1950453)

- Use CSS variables to support dark mode
- Prune CSS so the style is consistent with other parts of BMO

image

image


dklawren (approved):
Thanks Kohei

PR #2599, opened 2026-04-15
Bug 1976263 - Cookie banner hides cookie page

kyoshino:
[Bug 1976263 - Cookie banner hides cookie page](https://bugzilla.mozilla.org/show_bug.cgi?id=1976263)

- Add a var to hide the banner only on the cookie settings page
- Fix JS-embedded config
  - `Param('cookie_consent_enabled')` is a string, either `'0'` or `'1'`, so cast it to boolean
  - Also cast `Bugzilla.cgi.cookie_consent_required`

### DIFF
--- a/js/global.js
+++ b/js/global.js
@@ -244,6 +244,11 @@ const scroll_element_into_view = ($target, complete) => {
 }
 
 const openBanner = () => {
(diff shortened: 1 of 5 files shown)


dklawren (approved):
Thanks Kohei!

PR #2602, opened 2026-04-16
Bug 2032230 - Cookie consent needs to be updated to look for the country code that Fastly is providing which is different then before

dklawren:


Bugzilla/CGI.pm
```diff
@@ -696,7 +696,10 @@ sub cookie_consented {
 sub cookie_consent_required {
   my ($self) = @_;
   return 1 if $ENV{CI};
-  my $client_region = $self->http('X-Client-Region') || '';
+  my $client_region
+    = $self->http('X-Sigsci-Client-Geo-Country-Code')
```
cgsheeh, line 700: LGTM. We may want to confirm this returns the 2-char country code, as that is the format we use in the `COOKIE_CONSENT_COUNTRIES`.

cgsheeh (approved):

PR #2603, opened 2026-04-21
Bug 2033529 - Security Bugs Report: exclude 'unsupported-config' bugs

dklawren:


Bugzilla/Report/SecurityRisk.pm
```diff
@@ -281,7 +289,7 @@ sub _build_results {
         }
       }
 
-      # Undo stalled keyword changes
+      # Undo stalled and unsupported-config keyword changes
       if ($event->{field_name} eq 'keywords') {
         if ($event->{added} =~ /\b\stalled\b/) {
```
cgsheeh: Pre-existing, but shouldn't this regex be `/\bstalled\b/`? As written now, the `s` in `stalled` is instead `\s`, which matches on whitespace. I can see this in the syntax highlighting, and from how the below regex doesn't have a `\` after `\b`.

Bugzilla/Report/SecurityRisk.pm
```diff
@@ -417,8 +437,8 @@ sub _bugs_by_sec_keyword {
 }
 
 sub _is_bug_open {
-  my ($self, $status, $is_stalled) = @_;
-  return ($self->check_open_state->($status)) && !($is_stalled);
+  my ($self, $status, $is_stalled, $is_unsupported_config) = @_;
+  return ($self->check_open_state->($status)) && !($is_stalled) && !($is_unsupported_config);
```
cgsheeh: Pre-existing issue: we check for the open state regardless of `$is_stalled` or `$is_unsupported_config`, we should check the two variables first to avoid an unnecessary check.

t/security-risk.t
```diff
@@ -174,4 +178,102 @@ catch {
...
+    },
+    events => [
+      {
+        bug_id     => 5,
+        bug_when   => DateTime->new(year => 2000, month => 1, day => 12),
+        field_name => 'keywords',
+        removed    => '',
+        added      => 'unsupported-config',
+      },
+    ],
```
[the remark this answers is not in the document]
cgsheeh: Agreed, we should check the `removed` path since we have added code for it. 👍

cgsheeh (approved):
LGTM, it would be best to add test coverage of the `removed` code path.

PR #2610, opened 2026-04-29
Bug 2035819 - Pasting a link while a link is currently selected adds an extra markdown entry

kyoshino:
[Bug 2035819 - Pasting a link while a link is currently selected adds an extra markdown entry](https://bugzilla.mozilla.org/show_bug.cgi?id=2035819)

Check if the selected text is not a URL to prevent double-pasting, similarly to #2607.

### DIFF
--- a/js/util.js
+++ b/js/util.js
@@ -963,7 +963,7 @@ Bugzilla.String = class String {
    * @returns {boolean} `true` if the string is a valid URL, `false` otherwise.
    */
   static isURL(string) {
-    return /^(?:https?|mailto):\S+$/.test(string) && URL.canParse(string);
+    return /^(?:https?:\/\/|mailto:)\S+$/.test(string) && URL.canParse(string);
   }
 
   /**
--- a/js/util.js
+++ b/js/util.js
@@ -958,6 +958,15 @@ Bugzilla.Storage = class LocalStorage {
  */
 Bugzilla.String = class String {
   /**
+   * Check if a string is a valid URL.
+   * @param {string} string The string to check.
+   * @returns {boolean} `true` if the string is a valid URL, `false` otherwise.
+   */
+  static isURL(string) {
+    return /^(?:https?|mailto):\/\/\S+$/.test(string) && URL.canParse(string);
+  }
+
+  /**
    * Escape special characters in a string so it can be used for `new RegExp()`.
    * @param {string} string Input string.
    * @returns {string} Escaped string.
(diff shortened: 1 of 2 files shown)


dklawren (changes requested):
Issue: mailto: regression in insertLink

  This is a bug introduced by the PR.

  isURL regex: /^(?:https?|mailto):\/\/\S+$/

  The \/\/ (literal //) is required — but valid mailto URLs are mailto:user@example.com, not mailto://user@example.com. So isURL() never matches mailto URLs.

  Old insertLink condition:
  selectedText.match(/^(https?|mailto):/) || !selectedText
  → Correctly caught mailto:user@example.com as a URL.

  New condition:
  !selectedText || Bugzilla.String.isURL(selectedText)
  → isURL('mailto:user@example.com') returns false. The else branch runs instead, producing [mailto:user@example.com](url) — wrong.

  Fix: Either remove mailto from isURL's regex (since URL.canParse + :// makes it dead code anyway), or keep a separate check:

  static isURL(string) {
    return /^https?:\/\/\S+$/.test(string) && URL.canParse(string);
  }

  And in insertLink:
  } else if (!selectedText || Bugzilla.String.isURL(selectedText) || /^mailto:/.test(selectedText)) {

  Or add a separate isMailto helper. Whatever the approach, the mailto case needs to be preserved.

kyoshino:
Thanks, just removed the `\/\/` from the regex.

js/util.js
```diff
@@ -957,6 +957,15 @@ Bugzilla.Storage = class LocalStorage {
  * Provide static utility methods related to string parsing and manipulation.
  */
 Bugzilla.String = class String {
+  /**
+   * Check if a string is a valid URL.
+   * @param {string} string The string to check.
+   * @returns {boolean} `true` if the string is a valid URL, `false` otherwise.
+   */
+  static isURL(string) {
+    return /^(?:https?|mailto):\S+$/.test(string) && URL.canParse(string);
```
dklawren: isURL uses /^(?:https?|mailto):\S+$/ but the original paste-handler regex was /^https?:\/\/\S+$/. The new regex allows https:foo (no //) to pass the regex gate, relying solely on URL.canParse to reject it.

URL.canParse("https:foo") returns true in some browser versions (opaque path). 

Should be more explicit:
`return /^(?:https?:\/\/|mailto:)\S+$/.test(string) && URL.canParse(string);`

dklawren (changes requested):

js/util.js
kyoshino: Applied the suggestion.

PR #2612, opened 2026-05-01
Bug 2036191 - Crash Signature Field Mismatch in Bugzilla REST API

dklawren:


Bugzilla/WebService/Bug.pm
```diff
@@ -1731,20 +1731,26 @@ sub _bug_to_hash {
...
+  my @hidden_cfs = grep {
+    !$seen_cf{$_->name}
+      && $_->type != FIELD_TYPE_EXTENSION
+      && $_->type != FIELD_TYPE_MULTI_SELECT
+  } Bugzilla->active_custom_fields({skip_extensions => 1});
+  foreach my $field (@hidden_cfs) {
+    my $name = $field->name;
+    next if !filter_wants($params, $name, ['default', 'custom']);
+    my $raw = $bug->$name;
+    next if !defined($raw) || $raw eq '';
```
cgsheeh, line 1752: Claude thinks this line will allow `---` through as an option, and that this is the default field which should be skipped for `FIELD_TYPE_SINGLE_SELECT` fields. Is it safe to assume `''` is the field's default?

Bugzilla/Bug.pm
```diff
@@ -2700,7 +2700,9 @@ sub set_all {
 
   # And set custom fields.
   my @custom_fields
-    = grep { $_->type != FIELD_TYPE_EXTENSION } Bugzilla->active_custom_fields;
+    = grep { $_->type != FIELD_TYPE_EXTENSION } Bugzilla->active_custom_fields(
+        {product => $self->product_obj, component => $self->component_obj}
+    );
```
cgsheeh: Seems like it is possible to drop writes if the bug is moved to a product where the field isn't enabled? Is that intentional?

cgsheeh (approved):

PR #2614, opened 2026-05-08
Bug 2038298 - Render Phabricator revisions stack fork in the parent's row.

beurdouche:
Screenshot 2026-05-08 at 19 41 03

This PR supposedly fixes the line fork to be above the dot in the parent patch instead of below the dot in the child patch (the image is the current behavior).


beurdouche:
Untested.

dklawren (approved):
Thanks!

PR #2633, opened 2026-05-29
Bug 2043229 - Display a "Show External" button beneath attachments list to display all attachments that would redirect such as Github and Phabricator

dklawren:


extensions/BugModal/template/en/default/bug_modal/edit.html.tmpl
```diff
@@ -51,12 +51,24 @@
...
   FOREACH attachment IN bug.attachments;
     NEXT IF attachment.isprivate && !(user.is_insider || attachment.attacher.id == user.id || (attachment.is_bounty_attachment && bug.reporter.id == user.id)) ;
     IF attachment.isobsolete;
       obsolete_attachments = obsolete_attachments + 1;
     ELSE;
       active_attachments = active_attachments + 1;
+      IF attachment_hide_types.${attachment.contenttype};
+        external_attachments = external_attachments + 1;
+      END;
     END;
```
Xzzz, line 80: `external_attachment`s here only counts external-type attachments that are not obsolete.
`attachments.html.tmpl:108-110` gates the "Show External" button on that count:
```
[% IF external_attachments %]
  <button type="button" id="attachments-external-btn" class="secondary">Show External</button>
[% END %]
```
Following bug 2043229 requirements, a row which is both obsolete and external is only meant to show once both "Show External" and "Show Obsolete" are toggled on (`bug_modal.js:279-292`: function `updateAttachmentRows`).

So a bug with only obsolete external attachments never displays the "Show External" button, also we need that button to reveal them.
=> It could results in permanently hidden and then unreachable attachments

Quick-fix: I suggest to count obsolete and external attachments too when deciding whether to render the button, but to not include them in the subtitle number of external files

Xzzz (approved):
lgtm

PR #2634, opened 2026-05-29
Bug 2043322 - text/html attachments are downloaded instead of viewing in-browser

dklawren:


Bugzilla/Attachment.pm
```diff
@@ -290,17 +290,40 @@ sub isprivate {
...
+current deployment context. Delegates to C<is_safe_inline_content_type>.
 
 =back
 
 =cut
 
 sub is_viewable {
-  my $contenttype = $_[0]->contenttype;
-  return is_executable_content_type($contenttype) ? 0 : 1;
+  return is_safe_inline_content_type($_[0]->contenttype) ? 1 : 0;
```
cgsheeh: The `? 1 : 0` bit is redundant. You're basically saying `return 1 ? 1 : 0` which is the same as if you just returned the result of `is_safe_inline_content_type` directly.

cgsheeh (approved):
LGTM. Please try not to include unnecessary formatting changes in PRs in the future, it obfuscates the actual changes in the review.

longsonr:
What about svg? i.e. image/svg+xml can that be treated the same as text/html?

cgsheeh (approved):

PR #2635, opened 2026-05-30
Bug 2043733 - Live Github Status for Pull Requests

dklawren:
image

### DIFF
--- a/extensions/GitHubPullRequests/lib/API/V1/PullRequests.pm
+++ b/extensions/GitHubPullRequests/lib/API/V1/PullRequests.pm
@@ -5,48 +5,43 @@
 # This Source Code Form is "Incompatible With Secondary Licenses", as
 # defined by the Mozilla Public License, v. 2.0.
 
-package Bugzilla::Extension::GitHubPullRequests::WebService;
+package Bugzilla::Extension::GitHubPullRequests::API::V1::PullRequests;
 
 use 5.10.1;
-use strict;
-use warnings;
-
-use base qw(Bugzilla::WebService);
+use Mojo::Base qw( Mojolicious::Controller );
 
 use Bugzilla::Bug;
 use Bugzilla::Constants;
 use Bugzilla::Error;
 use Bugzilla::Logging;
-use Bugzilla::WebService::Constants;
 use Bugzilla::Extension::GitHubPullRequests::Constants;
-use Types::Standard qw(-types);
-use Type::Params qw(compile);
 
 use JSON qw(decode_json);
 use LWP::UserAgent;
+use URI::Escape qw(uri_escape);
 
-use constant READ_ONLY => qw(
-  bug_pull_requests
-);
-
-use constant PUBLIC_METHODS => qw(
-  bug_pull_requests
-);
+sub setup_routes {
+  my ($class, $r) = @_;
+  $r->get('/githubpr/bug_pull_requests/:bug_id')
+    ->to('GitHubPullRequests::API::V1::PullRequests#bug_pull_requests');
+}
 
 sub bug_pull_requests {
-  state $check = compile(Object, Dict [bug_id => Int]);
-  my ($self, $params) = $check->(@_);
+  my ($self) = @_;
+  Bugzilla->usage_mode(USAGE_MODE_MOJO_REST);
 
-  my $user = Bugzilla->login(LOGIN_REQUIRED);
+  my $user = $self->bugzilla->login(LOGIN_REQUIRED);
+  $user->id || return $self->user_error('login_required');
 
   # Kill switch: when disabled, return nothing rather than querying GitHub.
-  return {pull_requests => []}
+  return $self->render(json => {pull_requests => []})
     unless Bugzilla->params->{github_pr_status_enabled};
 
-  ThrowUserError('invalid_parameter', {name => 'bug_id', err => 'required'})
-    unless $params->{bug_id};
+  my $bug_id = $self->param('bug_id');
+  return $self->user_error('invalid_parameter', {name => 'bug_id', err => 'required'})
+    unless $bug_id;
 
-  my $bug = Bugzilla::Bug->check({id => $params->{bug_id}, cache => 1});
+  my $bug = Bugzilla::Bug->check({id => $bug_id, cache => 1});
 
   my $ua = LWP::UserAgent->new(timeout => GITHUB_API_TIMEOUT);
   $ua->agent('BMO-Bugzilla/1.0');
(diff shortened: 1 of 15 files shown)


extensions/GitHubPullRequests/lib/API/V1/PullRequests.pm
```diff
@@ -0,0 +1,241 @@
...
+  my ($reviews) = @_;
+
+  return () unless ref($reviews) eq 'ARRAY';
+
+  # Keep only the latest review state per reviewer.
+  # Reviews are returned in chronological order so we can just overwrite.
+  my %latest;
+  my @order;
+  for my $review (@{$reviews}) {
+    my $login = $review->{user}{login};
```
cgsheeh, line 227: Can `user` be `null` for a deleted user? Might want to check that here.

extensions/GitHubPullRequests/lib/API/V1/PullRequests.pm
cgsheeh, line 1: As this feature is making lots of (potentially rate-limited) requests to Github, should we add a flag to enable/disable it, in case we want to turn it off to debug issues?

extensions/GitHubPullRequests/web/js/github_pull_requests.js
```diff
@@ -0,0 +1,239 @@
...
+  async onLoad() {
+    const showClosedCheckbox = document.querySelector("#github-show-closed");
+    if (!showClosedCheckbox) return;
+
+    this.showClosed = showClosedCheckbox.checked;
+
+    const tbody = document.querySelector("tbody.github-prs-body");
+    if (!tbody) return;
+
+    const loadingRow = tbody.querySelector(".github-loading-row");
```
cgsheeh, line 189: Above we check that `if (!tbody) return;`, but we don't do the same check here. Can we make that check so we don't throw an exception later on if this returns null?

extensions/GitHubPullRequests/lib/WebService.pm
```diff
@@ -0,0 +1,241 @@
...
+use Types::Standard qw(-types);
+use Type::Params qw(compile);
+
+use JSON qw(decode_json);
+use LWP::UserAgent;
+
+use constant GITHUB_CONTENT_TYPE  => 'text/x-github-pull-request';
+use constant GITHUB_PR_REGEX      => qr{^https://github\.com/([^/]+)/([^/]+)/pull/(\d+)/?$};
+use constant GITHUB_API_BASE      => 'https://api.github.com';
+use constant GITHUB_API_TIMEOUT   => 10;
```
cgsheeh: There's room for de-duplication of these constants in this PR and elsewhere.

extensions/GitHubPullRequests/web/js/github_pull_requests.js
```diff
@@ -0,0 +1,239 @@
...
+      const errRow = tbody.querySelector(".github-loading-error-row");
+      if (!errRow) return;
+      errRow.querySelector(".github-load-error-string")?.replaceChildren(errStr);
+      errRow.classList.remove("bz_default_hidden");
+    };
+
+    try {
+      const { pull_requests } = await Bugzilla.API.get(
+        `githubpr/bug_pull_requests/${BUGZILLA.bug_id}`
+      );
```
cgsheeh, line 202: When we get a cache hit this seems fine. I think the cold cache scenario has a potential to cause a problem. Say we have a bug with 8 PRs linked in a stack, each PR makes 2 API calls in the BMO API backend, and each hits the 10s max timeout on Github. We would tie up the worker for 160s, then fail and re-try again in 60s. If multiple people visit such a bug at once, this could make BMO unresponsive as a result of a (very common these days) Github outage.

Does BMO have a background worker we could move the PR fetching to? Then we could populate the cache in memcached and always return results from there via the API.

cgsheeh (changes requested):

dklawren:
Thanks @cgsheeh! Addressed your comments 1-4 in the latest push:

1. **Deleted-user null check** (`WebService.pm`) — `_summarize_reviews` now guards `$review->{user}` (a review from a deleted GitHub account comes back with a null user) and skips it instead of dying.
2. **Enable/disable flag** — added a dedicated `github_pr_status_enabled` param (default off). When disabled the WebService returns nothing and the panel isn't rendered, so we can kill the feature quickly during a GitHub outage or while debugging.
3. **JS null check** — added `if (!loadingRow) return;` mirroring the existing `tbody` guard.
4. **Constant de-duplication** — moved the shared constants into a new `lib/Constants.pm` used by both `WebService.pm` and `Extension.pm`. The broader codebase-wide `text/x-github-pull-request` consolidation I've left as a follow-up to avoid churning unrelated modules in this PR.

5. **Background worker for fetching** — good call on the cold-cache/outage scenario. Leaving this one for now while I think through the job-queue + polling approach; will follow up.

extensions/GitHubPullRequests/lib/WebService.pm
```diff
@@ -0,0 +1,239 @@
...
+    $state = 'open';
+  }
+
+  my @labels = map { $_->{name} } @{$pr->{labels} // []};
+
+  my $reviews_response = _github_get($ua, $api_url . '/reviews');
+  my @reviews;
+  if ($reviews_response->{ok}) {
+    @reviews = _summarize_reviews($reviews_response->{data});
+  }
```
cgsheeh: This endpoint uses pagination but we don't implement that here. Is this okay for the dashboard?

extensions/GitHubPullRequests/lib/WebService.pm
```diff
@@ -0,0 +1,239 @@
...
+
+  return {pull_requests => \@pull_requests};
+}
+
+sub _fetch_pull_request {
+  my ($ua, $owner, $repo, $pr_number) = @_;
+
+  my $url     = "https://github.com/$owner/$repo/pull/$pr_number";
+  my $api_url = GITHUB_API_BASE . "/repos/$owner/$repo/pulls/$pr_number";
```
cgsheeh: Looks like we're splatting the Github owner/repo straight into this URL from the regex, we should probably escape it.

extensions/GitHubPullRequests/lib/API/V1/PullRequests.pm
```diff
@@ -0,0 +1,239 @@
...
+  # Kill switch: when disabled, return nothing rather than querying GitHub.
+  return {pull_requests => []}
+    unless Bugzilla->params->{github_pr_status_enabled};
+
+  ThrowUserError('invalid_parameter', {name => 'bug_id', err => 'required'})
+    unless $params->{bug_id};
+
+  my $bug = Bugzilla::Bug->check({id => $params->{bug_id}, cache => 1});
+
+  my $ua = LWP::UserAgent->new(timeout => GITHUB_API_TIMEOUT);
```
cgsheeh, line 46: Should this be using Mojo?

cgsheeh (approved):

extensions/GitHubPullRequests/lib/WebService.pm
dklawren: Summary

GitHub returns reviews oldest-first at 30/page by default, and the code keeps "latest state per reviewer" by overwriting as it iterates. So on a PR with >30 review events we were fetching the oldest 30 and computing latest-state over the wrong subset. COMMENTED reviews (which we filter out) still count toward the 30, making chatty PRs hit this sooner. The per-bug GITHUB_MAX_PULL_REQUESTS cap doesn't help since it bounds PRs, not reviews.

Fix (bounded, not unbounded pagination):
- Constants.pm — added GITHUB_REVIEWS_PER_PAGE => 100 (GitHub's max), exported it, with a comment explaining the oldest-first ordering and the deliberate tradeoff.
- PullRequests.pm — the reviews request now appends ?per_page=100.

This makes the review summary correct for any PR with ≤100 review events (effectively all real PRs) in a single bounded request, without reintroducing the amplification risk that full Link-header pagination would.

Consciously accepted limitation: a PR with >100 review events would still report the oldest 100 and could show a stale reviewer state. I chose not to paginate to the last page because it multiplies outbound requests per PR (undermining the DoS bounding) for a genuinely rare case. If you'd rather be fully correct there, the alternative is bounded Link-header pagination capped at a small number of pages — say ~5 — but I'd only add that if you expect PRs that active.

PR #2636, opened 2026-06-01
Bug 1995467 - Show dependency tree on meta bugs by default

kyoshino:
[Bug 1995467 - Show dependency tree on meta bugs by default](https://bugzilla.mozilla.org/show_bug.cgi?id=1995467)

Here’s an overdue PR for my assigned task 🙇🏼 

- Add the `is_meta` property to bug objects in the BMO extension as the `meta` keyword convention is BMO-specific.
- Overhaul the legacy Dependency Tree page to support dynamic loading and toggling of the tree view without a full page reload.
- For meta bugs, replace the flat dependency list in the References section with an embedded interactive tree widget loaded via `showdependencytree.cgi?embed=1`. This provides the same experience as the standalone tree page, but without leaving the bug page.
- Add a Hide Dependency Tree button to revert to the flat list; References module starts expanded for meta bugs.
- Fix the off-by-one `$realdepth` bug. It was incremented unconditionally on every recursive call, inflating the value by 1.

### DIFF
--- a/extensions/BMO/Extension.pm
+++ b/extensions/BMO/Extension.pm
@@ -73,6 +73,7 @@ our $VERSION = '0.1';
 BEGIN {
   *Bugzilla::Bug::last_closed_date               = \&_last_closed_date;
   *Bugzilla::Bug::reporters_hw_os                = \&_bug_reporters_hw_os;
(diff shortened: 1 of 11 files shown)


dklawren (approved):
LGTM. r=dkl

PR #2637, opened 2026-06-03
Bug 2044471 - Add loading indicator to embedded dependency tree

kyoshino:
[Bug 2044471 - Add loading indicator to embedded dependency tree](https://bugzilla.mozilla.org/show_bug.cgi?id=2044471)

Add a loading indicator in case the tree update takes more than 300ms.

### DIFF
--- a/extensions/BugModal/web/bug_modal.css
+++ b/extensions/BugModal/web/bug_modal.css
@@ -569,6 +569,11 @@ input[type="number"] {
   font-size: inherit;
 }
 
+#dependency-tree-container .tree-container {
+  max-height: 80dvh; /* Limit the height of the tree container to 80% of the viewport height */
+  overflow: auto;
+}
+
 #hide-dependency-tree-btn {
   margin-left: auto;
 }
--- a/js/dependency-tree.js
+++ b/js/dependency-tree.js
@@ -104,6 +104,11 @@ Bugzilla.DependencyTree = class DependencyTree {
   }
 
   /**
+   * Error message to display when the dependency tree fails to load.
+   */
+  #UPDATE_TREES_ERROR_MESSAGE = '<p class="error">Failed to load the dependency tree.</p>';
+
+  /**
    * Fetch and update the dependency tree HTML based on the given parameters, then inject it into
    * the page.
    * @param {object} params Parameters.
@@ -119,11 +124,26 @@ Bugzilla.DependencyTree = class DependencyTree {
     });
 
     const url = `${this.data.action}?${params}`;
-    const response = await fetch(`${url}&embed=1&tree_only=1`);
-    const html = response.ok ? await response.text() : undefined;
 
-    // Safe to inject HTML as is: same-origin fetch, Template Toolkit escapes all user-supplied data
-    this.$container.innerHTML = html ?? '<p class="error">Failed to load the dependency tree.</p>';
+    // Set up a delayed loading indicator — only show after 300ms to avoid flicker on fast loads
+    const loadingTimeout = setTimeout(() => {
+      this.$container.setAttribute('aria-busy', 'true');
+    }, 300);
+
+    try {
+      const response = await fetch(`${url}&embed=1&tree_only=1`);
+      const html = response.ok ? await response.text() : undefined;
+
+      // Safe to inject HTML as is: same-origin fetch, Template Toolkit escapes all user-supplied data
+      this.$container.innerHTML = html ?? this.#UPDATE_TREES_ERROR_MESSAGE;
+    } catch {
+      this.$container.innerHTML = this.#UPDATE_TREES_ERROR_MESSAGE;
+    } finally {
+      // Cancel the loading timeout if it hasn’t fired yet
+      clearTimeout(loadingTimeout);
+      // Remove the loading state if it was set
+      this.$container.removeAttribute('aria-busy');
+    }
 
     // Update the URL query parameters if we’re on the dependency tree page
     if (location.pathname === this.data.action) {
--- a/js/dependency-tree.js
+++ b/js/dependency-tree.js
@@ -290,6 +290,8 @@ Bugzilla.DependencyTree = class DependencyTree {
       console.error('Error fetching dependency tree:', ex);
       await this.showErrorMessage(generation);
     } finally {
+      // Increment generation to invalidate any in-flight message operations from this request
+      this.updateGeneration++;
       // Cancel the loading timeout if it hasn’t fired yet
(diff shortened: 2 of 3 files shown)


kyoshino:
Note on the CSS: for simplicity, we could use [CSS nesting](https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Nesting/Using), which is available in Firefox 117+. However, since Firefox 115 ESR is maintained [until August](https://whattrainisitnow.com/calendar/), I have avoided using it.

dklawren (commented):
Suggestions / Nits

- #UPDATE_TREES_ERROR_MESSAGE should be static. It's an immutable constant; as an instance field it's reallocated per DependencyTree instance. static #UPDATE_TREES_ERROR_MESSAGE = ... (referenced as DependencyTree.#UPDATE_TREES_ERROR_MESSAGE) is more accurate. Minor.
- Empty catch {} swallows the error silently. Adding console.error(...) would aid debugging of intermittent fetch failures without changing user-facing behavior:
  ```
  } catch (ex) {
    console.error('Failed to load the dependency tree', ex);
    this.$container.innerHTML = this.#UPDATE_TREES_ERROR_MESSAGE;
  }
  ```
- top: 100px is a magic number. For short trees (or the error \<p>, which isn't a [role="group"] and so won't be dimmed) the badge may sit oddly relative to content. Not blocking, but a comment or a content-relative value (e.g. centering within the visible area) would be more robust.
- Concurrency: rapid successive calls to updateTrees() each register their own timeout + finally. An earlier call's finally can removeAttribute('aria-busy') while a later in-flight call still expects the busy state. This is an edge case and the pre-existing code already raced on innerHTML, so it's not a regression — but if you want it airtight, track an in-flight counter or abort the prior fetch with AbortController.

kyoshino:
Resolved the feedback 🙏🏼

dklawren:
Some issues found during Claude code review just to see any JS issues that I am not an expert of. Please take a look and see if they are real issue that can be resolved before merge. 

---

Issues & Risks

1. Race condition can orphan the "Updating…" message (correctness — medium)
  showUpdatingMessage() is async and awaits getContainerHeight() (up to 100ms), but it is not awaited when invoked from the setTimeout callback. Consider this timeline:
  - t=300ms: timeout fires → showUpdatingMessage() starts, awaiting height
  - t=350ms: fetch completes → finally runs clearTimeout (no-op) + hideUpdatingMessage() (removes nothing — message not inserted yet, clears aria-busy)
  - t=400ms: height resolves → message is inserted into the container and aria-busy="true" is re-set

Result: a stuck "Updating…" element and a container left in the busy state. This triggers whenever the fetch finishes within ~100ms after the 300ms threshold — plausible in practice. Consider tracking the latest request (e.g. a generation counter or an aborted flag set in finally) and bailing out of showUpdatingMessage() if the fetch already completed, or awaiting the show and checking state before inserting.

2. disableControllers() lacks the defensive null-checks used elsewhere (correctness — low/medium)
  this.$numberInput is accessed with optional chaining everywhere else (this.$numberInput?.addEventListener line 58, this.$numberInput?.value line 93), implying it may be absent. But:
  
```
disableControllers() {
  this.$numberInput.disabled = true;   // throws if $numberInput is null
}
```
and the new this.$numberInput.disabled = false; in updateControllers() have the same exposure. If the input can legitimately be absent, these will throw a TypeError and break the whole update flow. Either confirm it's always present (and drop the ?. elsewhere for consistency) or guard these accesses.

3. Unawaited showErrorMessage() in catch/else (minor)
  Same not-awaited-async pattern as #1. Less harmful here since it ends with innerHTML = '' + insert, but ordering relative to finally's hideUpdatingMessage() is implicit. Worth confirming the error message reliably survives.

kyoshino:
Claude has solved the issue 1 (race condition) and 3 (unawaited promise). The issue 2 can be ignored because these input elements always exist, and the existence of the wrapper element is checked earlier with `if (!this.$trees) return`.

kyoshino:
Also fixed some bugs on the number input 🐛

dklawren:
### Code review

Found 1 issue:

1. The race condition the loading indicator was meant to solve can still leave a stuck "Updating…" message and a permanently dimmed, non-interactive tree.

`updateGeneration` is only incremented when a *new* request starts (`const generation = ++this.updateGeneration`), never when the current request completes. The `finally` block calls `clearTimeout` + `hideUpdatingMessage()` but does not bump the generation:

https://github.com/mozilla-bteam/bmo/blob/5e213933bef10dd449668feedb305e8e5ebf7a8a/js/dependency-tree.js#L291-L297

So the second generation check in `showMessage()` only guards against a newer request superseding an older one — not against the current request's own completion:

https://github.com/mozilla-bteam/bmo/blob/5e213933bef10dd449668feedb305e8e5ebf7a8a/js/dependency-tree.js#L179-L195

Timeline when a fetch lands shortly after the 300ms threshold (common on 300–400ms responses):
- t=300ms: timeout fires → `showUpdatingMessage(gen)` starts → passes check #1 → awaits `getContainerHeight()` (resolves ~1 frame to 100ms later)
- fetch completes → `innerHTML` set → `finally` runs `hideUpdatingMessage()` as a no-op (the `<p>` hasn't been inserted yet) → generation unchanged
- `getContainerHeight()` resolves → check #2 still passes (same generation) → the "Updating…" `<p>` is inserted on top of the loaded tree and `aria-busy="true"` is set, with no later `hideUpdatingMessage()`

The tree is then stuck at `opacity: 0.5; pointer-events: none` (per the `[aria-busy="true"] > [role="group"]` rule) and `aria-busy` stays set for screen readers until another update is triggered. Incrementing `this.updateGeneration` inside the `finally` block (before `hideUpdatingMessage()`) would make check #2 fail and close the window.

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

kyoshino:
Claude fixed the issue again 😅

dklawren:
> 🤖 Generated with [Claude Code](https://claude.ai/code)
> 
> - If this code review was useful, please react with 👍. Otherwise, react with 👎.

Sorry bout that. I must have hit Yes too many times as it submitted the review comment itself. But glad to see it was something worth looking into and was a quick fix. Normally I let Claude loose on a new pull request first, then decide if the change is worth worrying about and then summarize in my own words so I understand it myself. Of course if it is something I can do myself I will just do that.

dklawren (approved):
Issues all resolved and manual testing looks good. r=dkl

PR #2638, opened 2026-06-08
Bug 1764713 - Easily switch between open/closed/both combined bugs in bug search list

Xzzz:
#### Summary

This PR adds a quick-toggle widget at the top of the bug list results page (buglist.cgi), allowing users to switch between open, closed or all bugs without returning to the search form.

#### Details

- buglist.cgi

  - Computes a new template variable status_filter ('open', 'closed', or 'all') by inspecting the bug_status CGI parameters of the current search. Handles both the special aliases (__open__, __closed__, __all__) and explicit lists of individual status names (NEW, ASSIGNED, RESOLVED, etc.)
  - Computes a new template variable urlquerypart_no_status: the current search query string with all bug_status parameters stripped out, used to build toggle links that preserve every other search criterion

- template/en/default/list/list.html.tmpl

  - Adds a `<div class="bz_status_toggle">` immediately after the `<hr>` separator, containing three controls: Open, Closed, All
  - The active state is rendered as a non-clickable `<span>`; the other two are `<a>` links pointing to the current search with bug_status=__open__, bug_status=__closed__, or bug_status=__all__ appended
  - The toggle is hidden in the Change Several Bugs at Once (dotweak) mode

- skins/standard/buglist.css

  Adds styles for the toggle widget following the existing bz_ naming convention:
  
  - `.bz_status_toggle`: Flex container for the widget
  - `.bz_status_toggle_label`: "Show:" label
  - `.bz_status_toggle_btn`: Individual button (link or span)
  - `.bz_status_toggle_btn:hover`: Hover state for inactive buttons
  - `.bz_status_toggle_btn.active`: Active/selected button (filled with --link-color)

#### How it works

The search engine already understands bug_status=__open__, bug_status=__closed__, and bug_status=__all__ as special aliases that expand to the appropriate status sets at query time, so the toggle links are simple and require no server-side changes to the search logic.

#### Additional info

* [bug#1764713](https://bugzilla.mozilla.org/show_bug.cgi?id=1764713)

#### Test plan

1. Run a search returning open bugs (bug_status=__open__) -- Open button is highlighted
2. Click Closed -- redirects to the same search with bug_status=__closed__
3. Click All -- redirects with bug_status=__all__
4. Verify that other search parameters (product, assignee…) are preserved in the redirect
5. Verify display on a search with multiple bug_status values (no button highlighted)

### DIFF
--- a/buglist.cgi
+++ b/buglist.cgi
@@ -29,6 +29,7 @@ use Bugzilla::Status;
 use Bugzilla::Token;
 
 use Date::Parse;
+use List::Util qw(any);
 use List::MoreUtils qw(uniq);
 
 
@@ -890,13 +891,13 @@ $vars->{'closedstates'} = [map { $_->name } closed_bug_statuses()];
 # Determine which status filter tab is active for the toggle UI.
 {
   my @requested = $params->param('bug_status');
-  if (!@requested || grep { $_ eq '__all__' } @requested) {
+  if (!@requested || any { $_ eq '__all__' } @requested) {
     $vars->{'status_filter'} = 'all';
   }
-  elsif (grep { $_ eq '__open__' } @requested) {
+  elsif (any { $_ eq '__open__' } @requested) {
     $vars->{'status_filter'} = 'open';
   }
-  elsif (grep { $_ eq '__closed__' } @requested) {
+  elsif (any { $_ eq '__closed__' } @requested) {
     $vars->{'status_filter'} = 'closed';
   }
   else {
--- a/buglist.cgi
+++ b/buglist.cgi
@@ -889,7 +889,7 @@ $vars->{'closedstates'} = [map { $_->name } closed_bug_statuses()];
 
 # Determine which status filter tab is active for the toggle UI.
 {
-  my @requested = $cgi->param('bug_status');
+  my @requested = $params->param('bug_status');
   if (!@requested || grep { $_ eq '__all__' } @requested) {
     $vars->{'status_filter'} = 'all';
   }
--- a/buglist.cgi
+++ b/buglist.cgi
@@ -887,6 +887,35 @@ $vars->{'displaycolumns'} = \@displaycolumns;
 $vars->{'openstates'} = [BUG_STATE_OPEN];
 $vars->{'closedstates'} = [map { $_->name } closed_bug_statuses()];
 
+# Determine which status filter tab is active for the toggle UI.
+{
+  my @requested = $cgi->param('bug_status');
+  if (!@requested || grep { $_ eq '__all__' } @requested) {
+    $vars->{'status_filter'} = 'all';
+  }
+  elsif (grep { $_ eq '__open__' } @requested) {
+    $vars->{'status_filter'} = 'open';
+  }
+  elsif (grep { $_ eq '__closed__' } @requested) {
+    $vars->{'status_filter'} = 'closed';
+  }
+  else {
+    my %open_set   = map { $_ => 1 } BUG_STATE_OPEN;
+    my %closed_set = map { $_ => 1 } map { $_->name } closed_bug_statuses();
+    my $has_open   = grep { $open_set{$_}   } @requested;
+    my $has_closed = grep { $closed_set{$_} } @requested;
+    if ($has_open && !$has_closed) {
+      $vars->{'status_filter'} = 'open';
+    }
+    elsif ($has_closed && !$has_open) {
+      $vars->{'status_filter'} = 'closed';
+    }
+    else {
+      $vars->{'status_filter'} = 'all';
+    }
+  }
+}
+
 # The iCal file needs priorities ordered from 1 to 9 (highest to lowest)
 # If there are more than 9 values, just make all the lower ones 9
 if ($format->{'extension'} eq 'ics') {
@@ -1057,6 +1086,8 @@ else {
 # Set 'urlquerypart' once the buglist ID is known.
 $vars->{'urlquerypart'}
   = $params->canonicalize_query('order', 'cmdtype', 'query_based_on', 'token');
+$vars->{'urlquerypart_no_status'}
+  = $params->canonicalize_query('order', 'cmdtype', 'query_based_on', 'token', 'bug_status');
 
 if ($format->{'extension'} eq "csv") {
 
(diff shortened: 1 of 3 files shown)


buglist.cgi
```diff
@@ -887,6 +887,35 @@ $vars->{'displaycolumns'} = \@displaycolumns;
 $vars->{'openstates'} = [BUG_STATE_OPEN];
 $vars->{'closedstates'} = [map { $_->name } closed_bug_statuses()];
 
+# Determine which status filter tab is active for the toggle UI.
+{
+  my @requested = $cgi->param('bug_status');
+  if (!@requested || grep { $_ eq '__all__' } @requested) {
```
dklawren: The active-tab detection reads $cgi->param('bug_status'), but for named/saved searches and series, $params is reassigned to a fresh Bugzilla::CGI($buffer) (see the remaction=run/runseries handling) while $cgi keeps only the request params (e.g. namedcmd=Foo). So bug_status is empty there and the toggle always shows "All" active, even for an open-only saved search — and the adjacent urlquerypart_no_status correctly uses $params. Use $params->param('bug_status') for consistency.

skins/standard/buglist.css
```diff
@@ -9,6 +9,43 @@
...
+  cursor: pointer;
+}
+
+.bz_status_toggle_btn:hover {
+  background: var(--secondary-region-background-color);
+}
+
+.bz_status_toggle_btn.active {
+  background: var(--link-color, #0060df);
+  color: #fff;
```
dklawren, line 43: The CSS references var(--link-color), which is not defined anywhere in the codebase. The themed variable is --link-text-color (defined in skins/standard/global.css with a dark-mode variant). As written, inactive button text falls back to the browser default link color, and the active button background always uses the hardcoded #0060df, which has poor contrast in dark mode.

template/en/default/list/list.html.tmpl
```diff
@@ -123,6 +123,32 @@
...
+    [% IF status_filter == 'closed' %]
+      <span class="bz_status_toggle_btn active">Closed</span>
+    [% ELSE %]
+      <a class="bz_status_toggle_btn" href="[% base_status_url FILTER html %]&amp;bug_status=__closed__">Closed</a>
+    [% END %]
+    [% IF status_filter == 'all' %]
+      <span class="bz_status_toggle_btn active">All</span>
+    [% ELSE %]
+      <a class="bz_status_toggle_btn" href="[% base_status_url FILTER html %]&amp;bug_status=__all__">All</a>
+    [% END %]
```
dklawren, line 148: The toggle links are built from urlquerypart_no_status (which excludes order via canonicalize_query) but do not re-append &order=$qorder. Every other navigation link in this template re-adds the sort order (e.g. the "all search results", "Change Several Bugs at Once", and "Remember search" links). As a result, clicking Open/Closed/All silently drops the user's custom sort.

dklawren (changes requested):
Looks good but needs a few minor changes. Thanks!

buglist.cgi
Xzzz: Fixed: Following your advice, I have replaced $cgi->param('bug_status') with $params->param('bug_status') to ensure consistency with the way the urlquerypart_no_status function is constructed and to correctly handle named/saved queries in which $params is reset from the saved query buffer.

skins/standard/buglist.css
Xzzz: Fixed: replaced var(--link-color) with var(--link-text-color) everywhere (both the inactive button text and the active button background), and removed the hardcoded #0060df color fallback. The variable being defined in global.css with a dark-mode variant, no fallback is needed.

template/en/default/list/list.html.tmpl
Xzzz: Fixed: appended [%- "&order=$qorder" FILTER html IF order %] to all three toggle links (Open, Closed, All), consistent with the other navigation links in the template.

Xzzz:
BMO Test Suite / test_sanity failed cause of the 3 'grep' added at lines 893,896 and 899.
I propose to use List::Utils::any instead.

buglist.cgi
```diff
@@ -1057,6 +1087,8 @@ else {
 # Set 'urlquerypart' once the buglist ID is known.
 $vars->{'urlquerypart'}
   = $params->canonicalize_query('order', 'cmdtype', 'query_based_on', 'token');
+$vars->{'urlquerypart_no_status'}
+  = $params->canonicalize_query('order', 'cmdtype', 'query_based_on', 'token', 'bug_status');
```
dklawren: Minor finding and then we are good to go:

Lingering resolution filter can make "Open" return nothing. urlquerypart_no_status strips only bug_status, not resolution. If a search carries resolution=FIXED and the user clicks Open, the link becomes bug_status=__open__&resolution=FIXED → open bugs have no resolution → zero results. Edge case, but the most likely real-world surprise. Consider whether resolution should also be stripped from the toggle links (it's a product decision — stripping it changes the meaning of "preserve every other criterion").

The fix is a one-line change to also strip resolution from the toggle links' query string. Status and resolution are coupled (open bugs have resolution ---), so carrying a stale resolution=FIXED into an __open__ toggle is what produces the empty result — dropping it alongside bug_status is the correct scope for these links.

In buglist.cgi, change the urlquerypart_no_status line the PR adds:

# before
$vars->{'urlquerypart_no_status'}
  = $params->canonicalize_query('order', 'cmdtype', 'query_based_on', 'token', 'bug_status');

# after
$vars->{'urlquerypart_no_status'}
  = $params->canonicalize_query('order', 'cmdtype', 'query_based_on', 'token',
  'bug_status', 'resolution');

The variable name urlquerypart_no_status is now slightly under-descriptive (it also drops resolution). Optional: rename to something like urlquerypart_no_status_resolution, or just add a short comment above it explaining that resolution is stripped because it's coupled to status. I'd lean toward a one-line comment over a rename to keep the diff small.

dklawren (changes requested):

buglist.cgi
Xzzz: Added 'resolution' in canonicalize_query() and a comment to explain that 'resolution' and 'bug_status' must be dropped together from the query used by toggle buttons.

dklawren (approved):
LGTM r=dkl

PR #2643, opened 2026-06-10
Bug 2046434 - Dependency Tree: Hide Resolved doesn’t update Max Depth control

kyoshino:
[Bug 2046434 - Dependency Tree: Hide Resolved doesn’t update Max Depth control](https://bugzilla.mozilla.org/show_bug.cgi?id=2046434)

Make sure the depth control gets updated properly when clicking on the Show/Hide Resolved button or editing the max depth manually using the number input.

### DIFF
--- a/js/dependency-tree.js
+++ b/js/dependency-tree.js
@@ -353,8 +353,8 @@ Bugzilla.DependencyTree = class DependencyTree {
     // Update button states
     this.$toggleBtn.textContent = hideResolved ? 'Show Resolved' : 'Hide Resolved';
     this.$toggleBtn.disabled = false;
-    this.$setLimitBtn.disabled = realDepth === 1 || maxDepth === 1;
-    this.$removeLimitBtn.disabled = unlimited;
+    this.$setLimitBtn.disabled = !hasOpenBugs || realDepth === 1 || maxDepth === 1;
+    this.$removeLimitBtn.disabled = !hasOpenBugs || unlimited;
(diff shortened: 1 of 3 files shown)


dklawren (changes requested):
updateControls() can crash when a tree update fails

In onAction(), we always call updateControls() right after await updateTrees(...), regardless of whether the fetch succeeded:

this.disableControls();
await this.updateTrees({ maxDepth, hideResolved });
this.updateControls({ maxDepth, hideResolved });

When updateTrees() hits its error path — a network failure or a non-2xx response — it calls showErrorMessage(), which runs with clearContainer: true and does this.$container.innerHTML = ''. That wipes out the <meta name="real-depth"> tag along with the rest of the container.

Then updateControls() runs and does:

this.realDepth = Number(this.$trees.querySelector('meta[name="real-depth"]').content);

Since the meta tag is gone, querySelector(...) returns null and reading .content throws a TypeError. The user sees the "Failed to load" message, but the toolbar is left fully disabled (because disableControls() already ran and the exception prevents it from being re-enabled), so the widget is stuck and an error is logged to the console.

To reproduce: click any toolbar control while the server is returning an error (e.g. a 500 or a dropped connection).

Suggested fix: guard the meta lookup so a missing tag doesn't crash — e.g. only update realDepth when the meta tag is present, and/or skip updateControls() when the update failed so the controls get re-enabled instead of staying disabled.

dklawren (approved):
LGTM r=dkl

PR #2644, opened 2026-06-11
Bug 1619459 - Updated QA test to stop testing XMLRPC/JSONRPC and only test REST

dklawren:


qa/t/rest_group_get.t
cgsheeh: Out of scope of the PR, but this is doing an `=` assignment instead of a `==` equality check.

qa/t/rest_bug_create.t
```diff
@@ -210,54 +206,64 @@ my @tests = (
...
+  my $headers = api_headers($api_key);
+
+  if (my $error = $test->{error}) {
+    $t->post_ok($url . 'rest/bug' => $headers => json => $test->{args})
+      ->status_isnt(200);
+    like($t->tx->res->json->{message}, qr/$error/, "$test->{test}");
+  }
+  else {
+    $t->post_ok($url . 'rest/bug' => $headers => json => $test->{args})
+      ->status_is(200);
```
cgsheeh, line 264: Every other create route asserts 201 (`rest_user_create`, `rest_product_create`, `rest_group_create`, `rest_bug_add_comment`). `POST /rest/bug` returns 200 only because of a typo in `Bugzilla/WebService/Server/REST/Resources/Bug.pm:26`: it uses `status_code => STATUS_CREATED`, but `REST.pm` reads `success_code` (the other routes in that same file use `success_code` correctly). Ideally fix the production key and assert 201 here to match the rest of the suite.

qa/t/rest_bug_create.t
```diff
@@ -210,54 +206,64 @@ my @tests = (
...
+  my $bug = $t->tx->res->json->{bugs}->[0];
 
+  my $expect = dclone $test->{args};
   my $comment_is_private = delete $expect->{comment_is_private};
-  $expect->{creator} = $rpc->bz_config->{$t->{user} . '_user_login'};
+  $expect->{creator} = $config->{$test->{user} . '_user_login'};
 
   my @fields = keys %$expect;
-  $rpc->bz_test_bug(\@fields, $bug, $expect, $t);
+  test_bug(\@fields, $bug, $expect, $test);
```
cgsheeh, line 241: `test_bug` is called without a `$creation_time` argument. Its `creation_time`/`last_change_time` branch does `$creation_time->ymd`, so if a future create test ever includes those fields in its args, this dies with "Can't call method ymd on undefined value". Currently unreachable, but latent — consider passing the time or guarding the branch.

qa/t/lib/QA/REST/Util.pm
```diff
@@ -0,0 +1,150 @@
...
+  $private->{id} = _create_one($t, $url, $private_key, $private);
+
+  return ($public, $private);
+}
+
+sub _create_one {
+  my ($t, $url, $api_key, $fields) = @_;
+  $t->post_ok(
+    $url . 'rest/bug' => {'X-Bugzilla-API-Key' => $api_key} => json => $fields)
+    ->status_is(200);
```
cgsheeh, line 102: Two small things: (1) this asserts `status_is(200)` for `POST /rest/bug`, which bakes in the `status_code`/`success_code` typo noted over in `rest_bug_create.t` — it should be 201 once that's fixed; (2) the header `{'X-Bugzilla-API-Key' => $api_key}` is inlined even though this module defines `api_headers` for exactly this. Using `api_headers($api_key)` keeps it the single source of truth.

qa/t/lib/QA/REST/Util.pm
```diff
@@ -0,0 +1,150 @@
...
+  my ($api_key) = @_;
+  return $api_key ? {'X-Bugzilla-API-Key' => $api_key} : {};
+}
+
+# Build a Mojo::URL for a GET request to the REST API, turning a params
+# hashref into a query string. Array-ref values become repeated query keys
+# (e.g. names => ['a','b'] -> names=a&names=b), which is how the Bugzilla
+# WebService GET interface expects multi-valued parameters. undef values
+# are skipped.
+sub rest_get_url {
```
cgsheeh, line 45: `rest_get_url` largely re-implements `Mojo::URL->query`, which already turns a hashref with arrayref values into repeated keys (`names => ['a','b']` -> `names=a&names=b`). Only the undef-skipping and sorted ordering are genuinely additive — consider leaning on the framework for the rest.

qa/t/lib/QA/REST/Util.pm
```diff
@@ -0,0 +1,150 @@
...
+sub test_bug {
+  my ($fields, $bug, $expect, $test, $creation_time) = @_;
+
+  foreach my $field (sort @$fields) {
+
+    # "description" is used by Bug.create but comments are not returned
+    # by Bug.get or Bug.search.
+    next if $field eq 'description';
+
+    my @include = @{$test->{args}{include_fields} || []};
```
cgsheeh: `@include`/`@exclude` are loop-invariant but rebuilt from `$test` on every field iteration (and membership is a linear `grep` per field). Hoisting them above the `foreach` removes the repeated work and flattens the nesting. Minor.

qa/t/rest_bug_search.t
```diff
@@ -175,31 +139,26 @@ push(
...
-push(
-  @tests,
-  {
-    args => {votes => 1},
-    test => 'Search by votes',
-    bugs => -1,                  # We don't care how many it returns, for now.
-  }
-) if $config->{test_extensions};
+push(@tests,
+  {args => {votes => 1}, test => 'Search by votes', bugs => -1})
```
cgsheeh: `bugs => -1` makes `check_search` run `cmp_ok(scalar @$bugs, '>=', -1)`, which is always true — so this votes search asserts nothing about the result. It's a faithful port of the old weak assertion, but worth tightening to the expected count (or `exactly`) while migrating.

cgsheeh (approved):
Overall this is a clean migration — the new `POST /user/offer_account_by_email` route is ordered correctly (before `^/user/([^/]+)$`) and is `LOGIN_EXEMPT`, this glob change orphans nothing (all `webservice_*.t` are deleted, no `QA::RPC` refs remain), and interpolating `"@patterns"` into `glob` handles space-separated patterns fine.

Two coverage regressions worth a conscious sign-off, since they can't be commented inline (the files are deleted):
- JSONP `callback` sanitization (old `webservice_jsonp.t`) is gone, but `jsonrpc.cgi` still ships. That test guarded against XSS via the `callback` param and nothing in `rest_*.t` replaces it — the most notable loss.
- XML-RPC / JSON-RPC transport coverage is dropped entirely while `xmlrpc.cgi`/`jsonrpc.cgi` remain in production. Intentional per the commit message — just flagging it as a deliberate call.

PR #2648, opened 2026-06-20
Bug 1834454 part 2 - Webhooks - look at payload not bug_data

justdave:
Over in #2204 we attempted to get webhooks to fire for the matching product if a bug was moving OUT of that product. That patch failed. I finally got some time to try to debug this, and figured out that it was trying to pull the change data out of the wrong object.  Oops. We also had a case mismatch on a case-sensitive compare to test for Any as a component name.

This PR fixes those issues, and hopefully this works now.


dklawren (approved):
LGTM. r=dkl

PR #2650, opened 2026-06-23
Bug 1684509 - Add "Close as Invalid" button to close spam/test bugs in one action

Xzzz:
## Summary:
- New InvalidBugHelper extension that adds a "Close as Invalid" button to the bug modal
- Moves the bug to the "Invalid Bugs" product, resolves as INVALID, clears needinfo flags and strips security groups (all in one click)
- Spam-tagging the reporter's comments is opt-in via a checkbox in the confirmation dialog, with a warning message to avoid accidentally auto-disabling legitimate accounts
- Configurable warning comment posted to the reporter, editable via Administration > Parameters > Invalid Bug Helper
- Button visibility checks: editbugs group, product access and mandatory security group membership
- Server-side validation mirrors template checks: rejects already-resolved/verified bugs, enforces mandatory group membership and snapshots groups before the product move

## Test plan:
- Verify button appears on open bugs for users with editbugs + access to "Invalid Bugs" product + membership in mandatory security groups
- Verify button appears even when comment_taggers_group param is unset
- Verify button is hidden on RESOLVED/VERIFIED bugs and bugs already in "Invalid Bugs"
- Verify button is hidden for users without sufficient permissions
- Verify clicking the button with spam checkbox unchecked moves the bug and resolves it but does not tag comments as spam
- Verify clicking the button with spam checkbox checked moves the bug, resolves it and tags the reporter's comments as spam
- Verify security groups are stripped using the pre-move snapshot (not the post-move group list)
- Verify API rejects calls on already-resolved/verified bugs and bugs already in "Invalid Bugs"
- Verify API rejects calls from users who are not members of mandatory security groups on the bug
- Run t/bmo/invalid-bug-helper.t

### DIFF
--- a/extensions/InvalidBugHelper/lib/WebService.pm
+++ b/extensions/InvalidBugHelper/lib/WebService.pm
@@ -79,15 +79,7 @@ sub close_as_invalid {
     comment           => {body => $warning_text},
   });
 
-  # Strip non-mandatory security groups so the bug becomes public.
-  foreach my $group (@groups) {
-    next if $bug->product_obj->group_is_valid($group)
-      && $bug->product_obj->group_controls->{$group->id}->{membercontrol}
-         == CONTROLMAPMANDATORY
-      && $group->is_active;
-    $bug->remove_group($group);
-  }
-
+  # Non-mandatory security group stripping is handled by _set_product (called via set_all above)
   $bug->update();
 
   # Only tag reporter comments as spam when explicitly requested.
--- a/extensions/InvalidBugHelper/lib/WebService.pm
+++ b/extensions/InvalidBugHelper/lib/WebService.pm
@@ -53,11 +53,12 @@ sub close_as_invalid {
     ThrowUserError('bug_status_unresolvable', {bug => $bug});
   }
 
-  # Snapshot reporter comments and flags before any mutations.
+  # Snapshot reporter comments, flags, and groups before any mutations.
   my @reporter_comments
     = grep { $_->author->id == $bug->reporter->id } @{$bug->comments};
   my @clear_flags = map { {id => $_->id, status => 'X'} }
     grep { $_->type->name eq 'needinfo' } @{$bug->flags};
+  my @groups = @{$bug->groups_in};
 
   my $warning_text = Bugzilla->params->{invalidbughelper_warning_text}
     || 'This bug has been marked as invalid.';
@@ -78,9 +79,7 @@ sub close_as_invalid {
     comment           => {body => $warning_text},
   });
 
-  # Strip security groups so the bug becomes public after the product move.
-  # Copy the array — remove_group mutates groups_in in-place (see Bug.pm:3050).
-  my @groups = @{$bug->groups_in};
+  # Strip non-mandatory security groups so the bug becomes public.
   foreach my $group (@groups) {
     next if $bug->product_obj->group_is_valid($group)
       && $bug->product_obj->group_controls->{$group->id}->{membercontrol}
--- a/extensions/InvalidBugHelper/lib/WebService.pm
+++ b/extensions/InvalidBugHelper/lib/WebService.pm
@@ -90,10 +90,12 @@ sub close_as_invalid {
 
   $bug->update();
 
-  # Tag every reporter comment as spam (triggers AntiSpam user-disable logic).
-  foreach my $comment (@reporter_comments) {
-    $comment->add_tag('spam');
-    $comment->update();
+  # Only tag reporter comments as spam when explicitly requested.
+  if ($params->{mark_as_spam} && $user->can_tag_comments) {
+    foreach my $comment (@reporter_comments) {
+      $comment->add_tag('spam');
+      $comment->update();
+    }
   }
 
   $dbh->bz_commit_transaction();
--- a/extensions/InvalidBugHelper/lib/WebService.pm
+++ b/extensions/InvalidBugHelper/lib/WebService.pm
@@ -35,9 +35,17 @@ sub close_as_invalid {
 
   my $bug = Bugzilla::Bug->check({id => $bug_id});
 
-  ThrowUserError('auth_failure',
-    {action => 'modify', object => 'bug'})
-    unless $user->can_tag_comments;
+  # Block non-members from closing bugs with mandatory security groups.
+  foreach my $group (@{$bug->groups_in}) {
+    my $gc = $bug->product_obj->group_controls->{$group->id};
+    if ($gc
+      && $gc->{membercontrol} == CONTROLMAPMANDATORY
+      && !$user->in_group($group->name))
+    {
+      ThrowUserError('auth_failure',
+        {group => $group->name, action => 'modify', object => 'bug'});
+    }
+  }
 
   if ($bug->bug_status eq 'RESOLVED' || $bug->bug_status eq 'VERIFIED'
       || $bug->product eq 'Invalid Bugs')
--- a/extensions/InvalidBugHelper/lib/WebService.pm
+++ b/extensions/InvalidBugHelper/lib/WebService.pm
@@ -1,0 +1,94 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+#
+# This Source Code Form is "Incompatible With Secondary Licenses", as
+# defined by the Mozilla Public License, v. 2.0.
+
+package Bugzilla::Extension::InvalidBugHelper::WebService;
+
+use 5.10.1;
+use strict;
+use warnings;
+
+use base qw(Bugzilla::WebService);
+
+use Bugzilla;
+use Bugzilla::Bug;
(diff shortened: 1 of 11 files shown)


extensions/InvalidBugHelper/template/en/default/hook/bug_modal/edit-top_actions.html.tmpl
```diff
@@ -0,0 +1,29 @@
...
+   IF can_close_invalid;
+     FOREACH group IN bug.groups_in;
+       gc = bug.product_obj.group_controls.${group.id};
+       IF gc AND gc.membercontrol == constants.CONTROLMAPMANDATORY
+          AND NOT user.in_group(group.name);
+         can_close_invalid = 0;
+         LAST;
+       END;
+     END;
+   END;
```
dklawren, line 24: We will need to add this same mandatory group check to the permissions checks in WebService.pm in close_as_invalid() so a confidential bug can not be made public by a non-member.

extensions/InvalidBugHelper/lib/WebService.pm
```diff
@@ -0,0 +1,113 @@
...
+  # Strip security groups so the bug becomes public after the product move.
+  # Copy the array — remove_group mutates groups_in in-place (see Bug.pm:3050).
+  my @groups = @{$bug->groups_in};
+  foreach my $group (@groups) {
+    next if $bug->product_obj->group_is_valid($group)
+      && $bug->product_obj->group_controls->{$group->id}->{membercontrol}
+         == CONTROLMAPMANDATORY
+      && $group->is_active;
+    $bug->remove_group($group);
+  }
```
dklawren: The explicit group-stripping loop reads groups_in after set_all already moved the product, making it
redundant.  The comment says "Strip security groups so the bug becomes public after the product move," but by line 64 set_all→_set_product (Bugzilla/Bug.pm:3049-3057) has already recomputed groups_in for Invalid Bugs and removed the groups invalid there. The loop therefore iterates the post-move list and the actual stripping is done implicitly by _set_product, not by this code. To make the intent explicit and deterministic, snapshot @{$bug->groups_in} before set_all (consistent with the comment/flag snapshots at lines 39-42) — or drop the loop and document the reliance on _set_product.

extensions/InvalidBugHelper/template/en/default/hook/bug_modal/edit-top_actions.html.tmpl
```diff
@@ -0,0 +1,29 @@
...
+%]
+[% IF can_close_invalid %]
+| <button type="button" id="close-as-invalid-btn" class="minor">Close as Invalid</button>
+[% END %]
```
dklawren, line 28: Whole button silently disappears when comment_taggers_group is unset. (CONFIRMED) can_close_invalid ANDs in user.can_tag_comments, which returns false whenever the comment_taggers_group param is empty (Bugzilla/User.pm:2614). On any instance that doesn't configure comment tagging, even an admin never sees the button, with no explanation. Since the spam-tagging step is already made conditional inside the WebService, gating the entire close action on this unrelated param is likely unintended.

extensions/InvalidBugHelper/lib/WebService.pm
```diff
@@ -0,0 +1,113 @@
...
+
+  $bug->update();
+
+  # Tag every reporter comment as spam (triggers AntiSpam user-disable logic).
+  foreach my $comment (@reporter_comments) {
+    $comment->add_tag('spam');
+    $comment->update();
+  }
```
dklawren, line 91: Reporter-account auto-disable triggered with only an editbugs check. (PLAUSIBLE) Tagging the reporter's comments as spam feeds AntiSpam's spammer logic (extensions/AntiSpam/Extension.pm), which can auto-disable the reporter's account once a threshold is reached. Server-side this is gated only by comment_taggers_group being set and editbugs — not by the template's fuller guard set, and with no per-comment judgment that the comments are actually abusive. An editbugs user closing any reporter's bug can mass-tag their comments and potentially auto-disable a legitimate account. I wonder if we should add a checkbox to the confirmation that must be checked to mark the comment as spam in addition to the other changes. That way we do not disable the account if the bug is invalid but not abusive. What did the others think about this in their feedback?

dklawren (changes requested):

extensions/InvalidBugHelper/template/en/default/hook/bug_modal/edit-top_actions.html.tmpl
Xzzz: Modified WebService.pm: Added check to close_as_invalid (commit "Bug 1684509 - Enforce mandatory group check server-side in close_as_invalid")

extensions/InvalidBugHelper/lib/WebService.pm
Xzzz: Modified WebService.pm: moved "my @groups = @{$bug->groups_in};" line before set_all and updated comments (commit "Bug 1684509 - Snapshot groups_in before set_all in close_as_invalid")

extensions/InvalidBugHelper/template/en/default/hook/bug_modal/edit-top_actions.html.tmpl
Xzzz: Removed user.can_tag_comments from conditions at line 11 in edit-top_actions.html.tmpl (commit "Bug 1684509 - Remove can_tag_comments from Close as Invalid button visibility check"

extensions/InvalidBugHelper/lib/WebService.pm
Xzzz: Finally added checkbox with a warning message just near. For this I needed to change confirmation window type cause window.confirm() only supports plain text, then added new showCloseDialog(bugId) function to extensions/InvalidBugHelper/web/js/invalid_bug_helper.js, created new css file and added line to load it in header-end.html.tmpl (commit Bug 1684509 - Make spam-tagging opt-in when closing a bug as invalid)

dklawren (changes requested):
The snapshot was moved before set_all (now WebService.pm:61), but the strip loop (lines 83-89) was kept as-is:

```perl
  foreach my $group (@groups) {        # @groups = PRE-move snapshot
    next if $bug->product_obj->group_is_valid($group)
      && ...->{membercontrol} == CONTROLMAPMANDATORY
      && $group->is_active;
    $bug->remove_group($group);        # <-- throws for already-removed groups
  }
```

The bug: set_all → _set_product (Bugzilla/Bug.pm:3049-3057) already removes every group not valid in "Invalid Bugs" during the move. The loop then iterates the pre-move snapshot, which still contains those groups. For each one, group_is_valid is now false (not valid in Invalid Bugs), so the next is skipped and remove_group is called — but the bug is no longer in that group, so remove_group hits Bugzilla/Bug.pm:3385:

```perl
$self->in_group($group) || ThrowUserError('group_invalid_removal', $args);
```

This throws and rolls back the whole transaction. Any confidential bug (i.e., one carrying a security group not configured for "Invalid Bugs" — the primary target of this feature) will fail to close.

Minimal fix — guard against already-stripped groups:
```perl
  foreach my $group (@groups) {
    next unless $bug->in_group($group);   # _set_product already removed it
    next if $bug->product_obj->group_is_valid($group)
      && $bug->product_obj->group_controls->{$group->id}->{membercontrol}
         == CONTROLMAPMANDATORY
      && $group->is_active;
    $bug->remove_group($group);
  }
```
(Or, per your original alternative, drop the loop entirely and document reliance on _set_product.)

Note: t/bmo/invalid-bug-helper.t only checks routing/config/PUBLIC_METHODS — it never executes close_as_invalid against a real bug, so it will not catch this. A functional test that moves a grouped bug would.

All other changes look good to me.

Xzzz:
Following your advice, I finally removed entire loop and added comment.
Adding `next unless $bug->in_group($group);` just keeps dead code alive: it makes the loop a no-op for group _set_product already handled and only touches groups that are mandatory, which are then skipped via the `next if` anyway.
Thanks!

extensions/InvalidBugHelper/template/en/default/hook/bug_modal/edit-top_actions.html.tmpl
```diff
@@ -0,0 +1,28 @@
...
```
dklawren, line 28: Didn't see this before but it shows up during manual testing. Please remove the '|' character before the button. After the small change I think we are good to go.

dklawren (changes requested):

extensions/InvalidBugHelper/template/en/default/hook/bug_modal/edit-top_actions.html.tmpl
Xzzz: Indeed, it's a remnant from my tests displaying different types of links and buttons. A "dw" that was missed during the final cleanup... Fixed, we should be good to go now

dklawren (approved):
LGTM Thanks! r=dkl

wsmwk:
As the person who instigated the creation of the "Invalid Bugs" product category, I want to chime in to say thank you for improving the process around this workflow. 

Well done.

PR #2652, opened 2026-06-23
Bug 2049554 - Cloned security bugs should default to being secure even if they aren't in the default security group

dklawren:
Root cause

The bug 2028240 fix in extra_security_groups_for_clone only pre-checked the target product's security group when the source bug was in its own product's default security group:

```perl
my $source_sec = eval { $self->product_obj->default_security_group };
return () unless $source_sec && grep { $_ eq $source_sec } @clone_groups;
```

For a Core bug in dom-core-security / javascript-core-security / gfx-core-security (Core's default is plain core-security), this check failed, so no target security group was added — exactly the case :sfink hits.

The fix (Bugzilla/Bug.pm)

Replaced the "is the bug in this product's default security group?" test with "is the bug in any security group?":

```perl
return () unless grep { $self->_is_security_group($_) } @{$self->groups_in};
```

New helper _is_security_group treats a group as a security group if it is the default security group of some product (SELECT DISTINCT security_group_id FROM products), plus the insidergroup fallback for products without an explicit one. That set covers dom-core-security et al., since each is some product's default. The set is request-cached so the cross-product clone only does one query.

The downstream logic is unchanged: it still adds the target product's default security group (e.g. mobile-core-security for Firefox for Android) and skips it if the bug already has it.

Test (t/bmo/clone-security-groups.t)

Added a SKIP block reproducing the exact STR: a bug filed into a non-default security group (one owned by a third product, not prod_a's default), then cloned to prod_b. Asserts the bug is not in prod_a's default group yet the clone still pre-checks prod_b's default security group. It clears the security_group_ids request cache around its setup/teardown and restores all mutated rows.

A couple of things worth your eye when you run it:
- The new subtest needs ≥3 products in the test DB (it skips otherwise) so it can give the non-default group a distinct owning product.
- It relies on Bugzilla::Bug->create accepting the temp group in prod_a via the group_control_map (membercontrol=1) + user_group_map rows, same pattern the existing test uses

### DIFF
--- a/Bugzilla/Bug.pm
+++ b/Bugzilla/Bug.pm
@@ -4368,22 +4368,49 @@ sub in_group {
 }
 
 # Returns extra group names to add when cloning $self into $target_product.
-# If the bug is in its source product's default security group and the target
-# product has a different one, the target group name is returned so the clone
-# stays secure (Bug 2028240).
+# If the bug is in any security group and the target product has a different
+# default security group, the target group name is returned so the clone stays
+# secure. A group counts as a security group if it is the default security
+# group of some product, which covers non-default groups such as
+# dom-core-security and not just the source product's own default
+# (Bug 2028240, Bug 2049554).
 sub extra_security_groups_for_clone {
   my ($self, $target_product) = @_;
   return () if $self->product_id == $target_product->id;
   return () unless $target_product->can('default_security_group');
 
   my @clone_groups = map { $_->name } @{$self->groups_in};
-  my $source_sec = eval { $self->product_obj->default_security_group };
-  return () unless $source_sec && grep { $_ eq $source_sec } @clone_groups;
+  return () unless grep { $self->_is_security_group($_) } @{$self->groups_in};
 
   my $target_sec = eval { $target_product->default_security_group };
   return () unless $target_sec && !grep { $_ eq $target_sec } @clone_groups;
 
   return ($target_sec);
+}
+
+# Returns true if $group is the default security group of some product, and is
+# therefore acting as a security group somewhere in the installation.
+sub _is_security_group {
+  my ($self, $group) = @_;
+
+  my $cache = Bugzilla->request_cache->{security_group_ids} ||= do {
+    my $dbh = Bugzilla->dbh;
+    my %ids = map { $_ => 1 } @{
+      $dbh->selectcol_arrayref(
+        'SELECT DISTINCT security_group_id FROM products
+          WHERE security_group_id IS NOT NULL')
+    };
+
+    # Products without an explicit security group fall back to the insider
+    # group, so treat that as a security group too.
+    my $insider
+      = Bugzilla::Group->new({name => Bugzilla->params->{insidergroup}, cache => 1});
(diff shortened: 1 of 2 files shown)


dveditz:
> New helper _is_security_group treats a group as a security group if it is the default security group of some product [...]  That set covers dom-core-security et al., since each is some product's default.

That is not true: Core covers so many components and security bugs that we sub-divided into security teams to limit access risk. Most of them are only ever used within the Core product and are never the default group of another product: dom-core-security, gfx-core-security, javascript-core-security, layout-core-security, media-core-security, network-core-security.

It's this set that is causing the problem in bug 2049554.

Does it matter if a group is a "security" group or not? If a bug is hidden for any reason and you clone it, the clone should default hidden to the best of BMO's ability. Of course if a cloned bug's groups exist in the new Product we should clone them, but If you have to drop any group then adding the new Product's default group is a reasonable fallback. For example, if a Core bug is _both_ dom-core-security and mozilla-employee-confidential, if you clone that bug into the Firefox product the clone should be firefox-core-security (the default) and mozilla-employee-confidential.

Moves should work the same way so maybe there's already logic that could be shared?

dklawren:
> > New helper _is_security_group treats a group as a security group if it is the default security group of some product [...]  That set covers dom-core-security et al., since each is some product's default.
> 
> That is not true: Core covers so many components and security bugs that we sub-divided into security teams to limit access risk. Most of them are only ever used within the Core product and are never the default group of another product: dom-core-security, gfx-core-security, javascript-core-security, layout-core-security, media-core-security, network-core-security.
> 
> It's this set that is causing the problem in bug 2049554.
> 
> Does it matter if a group is a "security" group or not? If a bug is hidden for any reason and you clone it, the clone should default hidden to the best of BMO's ability. Of course if a cloned bug's groups exist in the new Product we should clone them, but If you have to drop any group then adding the new Product's default group is a reasonable fallback. For example, if a Core bug is _both_ dom-core-security and mozilla-employee-confidential, if you clone that bug into the Firefox product the clone should be firefox-core-security (the default) and mozilla-employee-confidential.
> 
> Moves should work the same way so maybe there's already logic that could be shared?

Thanks for the feedback. I have made the following changes in response:

The fix in bug 2028240 only kicked in when the source bug was in its product's default security group. That missed the common case for Core: bugs in the non-default security groups (dom-core-security, gfx-core-security, javascript-core-security, etc.) which aren't any product's default. Cloning one of those into another product silently dropped the restriction and the clone came out public.

Reframed the check around what actually matters: if cloning would drop any group the bug currently has, fall back to the target product's default security group. This reuses get_invalid_groups - the same primitive the product-move path uses - so clone and move now behave consistently. Groups that are valid in the target carry over as before, and we only add the default when something would otherwise be lost (e.g. a Core bug in dom-core-security + mozilla-employee-confidential cloned to Firefox becomes firefox-core-security + mozilla-employee-confidential).

Rewrote the test to be deterministic: instead of leaning on whatever group controls the seed products happen to have, it sets up controlled groups with known validity and covers the dropped, carried-over, mixed, same-product, and public-bug cases.

cgsheeh (approved):

PR #2653, opened 2026-06-24
Bug 1726931 - Add "All Reported by You" query to My Dashboard

Xzzz:
## Summary
- Add a new "reportedbugs" query in the My Dashboard dropdown listing all bugs reported by the user, regardless of status (open, resolved, closed)
- Auto-fallback: when "Assigned to You" returns no results and the user has no saved query preference, the dashboard switches to "All Reported by You" so new users see their bugs instead of an empty list

## Test plan
- Select "All Reported by You" in the dropdown - verify it shows all reported bugs including resolved/closed ones
- Log in as a user with no assigned bugs and no saved dashboard preference - verify the dashboard falls back to "All Reported by You"
- Log in as a user with assigned bugs - verify the dashboard stays on "Assigned to You"
- Select a query manually, reload - verify the saved preference is respected and no fallback occurs

### DIFF
--- a/extensions/MyDashboard/lib/Queries.pm
+++ b/extensions/MyDashboard/lib/Queries.pm
@@ -91,6 +91,16 @@ sub QUERY_DEFS {
       }
     },
     {
+      name        => 'reportedbugs',
+      heading     => 'All Reported by You',
+      description => 'All bugs you reported, including resolved and closed ones.',
+      params      => {
+        'emailreporter1' => 1,
+        'emailtype1'     => 'exact',
+        'email1'         => $user->login
+      }
+    },
+    {
       name        => 'openccbugs',
       heading     => "You Are CC'd On",
       description => 'You are in the CC list of the bug, so you are watching it.',
--- a/extensions/MyDashboard/lib/Queries.pm
+++ b/extensions/MyDashboard/lib/Queries.pm
@@ -93,7 +93,7 @@ sub QUERY_DEFS {
     {
       name        => 'reportedbugs',
       heading     => 'All Reported by You',
-      description => 'All bugs you reported, including resolved and closed ones.',
+      description => 'All bugs you reported, both open and closed ones.',
       params      => {
         'emailreporter1' => 1,
         'emailtype1'     => 'exact',
--- a/extensions/MyDashboard/web/js/query.js
+++ b/extensions/MyDashboard/web/js/query.js
@@ -202,8 +202,17 @@ window.addEventListener('DOMContentLoaded', () => {
     }
   };
 
-  // Initial load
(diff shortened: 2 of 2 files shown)


extensions/MyDashboard/lib/Queries.pm
```diff
@@ -90,6 +90,16 @@ sub QUERY_DEFS {
         'email1'         => $user->login
       }
     },
+    {
+      name        => 'reportedbugs',
+      heading     => 'All Reported by You',
+      description => 'All bugs you reported, including resolved and closed ones.',
```
dklawren: The description using both resolved and closed is somewhat redundant IMO. Maybe instead say "All bugs you reported, both open and closed ones." or "All bugs you reported, including ones that are closed.".

extensions/MyDashboard/web/js/query.js
```diff
@@ -202,8 +202,17 @@ window.addEventListener('DOMContentLoaded', () => {
...
+  // Initial load — fall back to reportedbugs if assignedbugs is empty
+  // and no saved preference exists (helps new users with no assigned bugs).
+  updateQueryTable(default_query).then(() => {
+    if (default_query === 'assignedbugs' && !query && bugQueryTable.data.length === 0) {
+      const $option = document.querySelector('#query option[value="reportedbugs"]');
+      if ($option) {
+        $option.selected = true;
+        updateQueryTable('reportedbugs');
+      }
+    }
```
dklawren, line 215: Fallback doesn't update default_query, so auto-refresh reverts to the empty assigned list. The fallback sets $option.selected = true and calls updateQueryTable('reportedbugs'), but leaves default_query as 'assignedbugs'. auto_updateQueryTable() (line 207) schedules setInterval(() => updateQueryTable(default_query), …). So for a new user with auto-refresh enabled and no assigned bugs: the table initially shows their reported bugs, but ~10 minutes later the interval reloads assignedbugs (empty), wiping the reported-bugs view while the dropdown still reads "All Reported by You" — a confusing UI desync. Fix: set default_query = 'reportedbugs' inside the fallback block.

dklawren (changes requested):

extensions/MyDashboard/lib/Queries.pm
Xzzz: Yes I agree, I think I was a bit tired when I wrote this...
Updated to "All bugs you reported, both open and closed ones"

extensions/MyDashboard/web/js/query.js
Xzzz: Fixed

dklawren (approved):

PR #2655, opened 2026-06-29
bug 2051396: form.web.bounty should not add the sec-bounty? flag

dveditz:
The form has told people our bounty program moved for a long while now, but we need make it explicit that this form is NOT a substitute.

### DIFF
--- a/extensions/BMO/template/en/default/bug/create/create-web-bounty.html.tmpl
+++ b/extensions/BMO/template/en/default/bug/create/create-web-bounty.html.tmpl
@@ -22,6 +22,7 @@
 #web_bounty_form .field_desc {
   padding-bottom: 3px;
 }
+#message,
 #web_bounty_form .field_desc,
 #web_bounty_form .head_desc {
   width: 600px;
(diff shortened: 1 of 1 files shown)


dklawren (approved):

dveditz:
Thanks for the quick review, @dklawren !  I was going to get @fkiriakos07 to approve the text changes first if you don't mind holding off on the merge. She's in CEST so tomorrow at the realistic earliest

extensions/BMO/template/en/default/bug/create/create-web-bounty.html.tmpl
```diff
@@ -64,12 +65,13 @@ function validateAndSubmit() {
...
+<h1>Web Bounty Form (obsolete)</h1>
 
 <div id="message">
-  We have migrated our web [% terms.bug %] bounty program to HackerOne. Please submit your report to our 
-  program on <a href="https://hackerone.com/mozilla">HackerOne</a>. If you do not prefer to use HackerOne, 
-  you can still submit the [% terms.bug %] using this form.
+  <p>We have migrated our web [% terms.bug %] bounty program to HackerOne. Please submit
+  your report to <a href="https://hackerone.com/mozilla">our program on HackerOne</a>.
+  <p>You can still submit website security [% terms.bugs %] using this form, but they
+  will not be part of our [% terms.bug %] bounty program.
```
fkiriakos07:
```suggestion
  will not be eligible for bounty.
```

fkiriakos07: maybe that's clearer? what do you think?

dveditz: I thought about using that wording, but I wondered if "eligible" applied to the bugzilla entry or if it meant the security problem reported in the bugzilla entry even if the reporter then did go and report it to HackerOne. I wasn't sure how readers would interpret that, or even how WE meant it.

fkiriakos07: what we mean is that if the bug was submitted on bugzilla, it wouldn't be eligible for bounty, but it would be eligible on hackerone

how about? "if you are not able to submit the report through hackerone, you can still use the form but the bug wouldn't be eligible for bounty"

dveditz: after trying many different things I ended up with essentially your first suggestion

fkiriakos07: sounds good, thank you

fkiriakos07:
thanks Dan for making the changes, it looks good

PR #2658, opened 2026-07-06
Bug 2052188 - Move Close as Invalid button to the bottom action bar

Xzzz:
## Summary
- Added a `bottom_actions` hook point in the bug-modal comment-actions bar, between "Save Changes" and "Resolve As"
- Moved InvalidBugHelper's "Close as Invalid" button template from the `top_actions` hook to the new `bottom_actions` hook

## Test plan
- Load a bug as an editbugs user who can enter the "Invalid Bugs" product
- Confirm "Close as Invalid" appears at the bottom, between "Save Changes" and "Resolve As"


dklawren (changes requested):
From Slack:
dholbert  [1:09 PM]
It makes more sense there than in its current spot, but it will probably be a bit confusing to have ~adjacent buttons that look like they do the same thing:
[Close as Invalid] Resolve as [Fixed] [Invalid]
dholbert  [1:12 PM]
I wonder if the label should also change to be a bit more specific? e.g. maybe we should align with the language in the warning message, e.g. Close as invalid test/spam bug ?
dholbert  [1:13 PM]
(I know that's a bit longer, but it's likely helpful to make it a little more specific to disambiguate from the existing "Resolve as Invalid" UI)

Here is what I propose we do:
Move the "Closed as Invalid" button to the end of the the actions buttons with a extra gap after the last one ("DUPLICATE"). 

Image

Also please add a title tag to the button containing " Close as invalid test/spam bug" or something similar in description.

Thanks

Xzzz:
__What has been modified__

1. Moved the `Hook.process('bottom_actions', ...)` call (which renders the "Close as Invalid" button) to after the `#resolve-as` block in `edit.html.tmpl:1496`
2. Added `margin-left: 16px` on `#close-as-invalid-btn` in `bug_modal.css` for the extra gap after DUPLICATE
3. Added `title="Close as Invalid Test/Spam bug"` to the button in `edit-bottom_actions.html.tmpl:27`

extensions/InvalidBugHelper/template/en/default/hook/bug_modal/edit-bottom_actions.html.tmpl
```diff
@@ -24,5 +24,5 @@
    END;
 %]
 [% IF can_close_invalid %]
-<button type="button" id="close-as-invalid-btn" class="minor">Close as Invalid</button>
+<button type="button" id="close-as-invalid-btn" class="minor" title="Close as Invalid Test/Spam bug">Close as Invalid</button>
```
dklawren: https://bugzilla.mozilla.org/show_bug.cgi?id=2052993#c1

Not wanting to drag this out too long but we could change the button label to "Close and move to Invalid Bugs" as suggested in the above bug comment. Then we can consider this done IMO.

dklawren (changes requested):

dklawren (approved):
LGTM r=dkl

extensions/InvalidBugHelper/template/en/default/hook/bug_modal/edit-bottom_actions.html.tmpl
```diff
@@ -24,5 +24,5 @@
    END;
 %]
 [% IF can_close_invalid %]
-<button type="button" id="close-as-invalid-btn" class="minor">Close as Invalid</button>
+<button type="button" id="close-as-invalid-btn" class="minor" title="Close as Invalid Test/Spam bug">Close and move to Invalid Bugs</button>
```
dklawren: My fault for not mentioning in my previous comment. This will need to fixed to pass the sanity test for bugwords:

Failed test '/app/extensions/InvalidBugHelper/template/en/default/hook/bug_modal/edit-bottom_actions.html.tmpl contains invalid bare words (e.g. 'bug') --WARNING'

```
<button type="button" id="close-as-invalid-btn" class="minor" title="Close as Invalid Test/Spam [% terms.bug %]">Close and move to Invalid [% terms.Bugs%]</button>
```

dklawren (changes requested):

dklawren (approved):

PR #2662, opened 2026-07-08
Bug 2053269 - BMO API bugzilla.login helper needs to support Bugzilla_api_token + cookie for authentication support

dklawren:
Bug 2043733's native Mojo REST endpoint authenticates via the bugzilla.login helper (Bugzilla::App::Plugin::Login), which didn't understand the web UI's Bugzilla_api_token mechanism — so browser-originated requests (login cookie + Bugzilla_api_token query param, as sent by Bugzilla.API in JS) got a 401.

Fix — Bugzilla/App/Plugin/Login.pm

- Added use Bugzilla::Error; and use Bugzilla::Token;.
- Extended the cookie branch to also handle USAGE_MODE_MOJO_REST, mirroring the legacy Bugzilla::Auth::Login::Cookie (56-78):
  - If a Bugzilla_api_token param is present, validate it via GetTokenData — it must be an api_token whose userid matches the login cookie's user, else rate-limit + ThrowUserError('auth_invalid_token').
  - Login cookie present but no api-token → clear the cookie (REST requires the token as a CSRF guard), falling back to unauthenticated.
  - Valid cookie + token then flows through the existing DB cookie verification.

ThrowUserError works here because MOJO_REST sets ERROR_MODE_MOJO, which stashes the error and lets the app render it as JSON.

New test — qa/t/rest_native_login.t

A generic auth test (not tied to any feature) that drives a login-required native-Mojo REST endpoint (rest/search/needinfo_last_seen) and verifies all four paths: no-auth → 401, X-Bugzilla-API-Key header → 200, browser login cookie + Bugzilla_api_token → 200, cookie-without-token → 401, and invalid token → 400. It performs a real browser login over HTTP and reads the token from the page's #bugzilla-global config.

### DIFF
--- a/qa/t/rest_native_login.t
+++ b/qa/t/rest_native_login.t
@@ -84,7 +84,8 @@ $t->post_ok(
 # value here also confirms the browser login above succeeded.
 $t->get_ok($url . 'home')->status_is(200);
 my $global = $t->tx->res->dom->at('#bugzilla-global');
-my $bugzilla_config = $global ? decode_json($global->{'data-bugzilla'}) : {};
+my $bugzilla_json = $global ? $global->attr('data-bugzilla') : undef;
+my $bugzilla_config = $bugzilla_json ? decode_json($bugzilla_json) : {};
 my $api_token = $bugzilla_config->{api_token};
 ok($api_token, 'Logged in and obtained a Bugzilla_api_token');
 
--- a/qa/t/rest_native_login.t
+++ b/qa/t/rest_native_login.t
@@ -1,0 +1,104 @@
+#!/usr/bin/env perl
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+#
+# This Source Code Form is "Incompatible With Secondary Licenses", as
+# defined by the Mozilla Public License, v. 2.0.
+
+# Exercises authentication for native (Mojolicious) REST endpoints, which
+# authenticate through the bugzilla.login helper (Bugzilla::App::Plugin::Login)
+# rather than the legacy WebService stack. In particular this verifies that,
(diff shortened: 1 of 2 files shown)


qa/t/rest_native_login.t
```diff
@@ -0,0 +1,105 @@
...
+  ->status_is(200)->json_has('/result');
+
+# 3b. The login cookie alone (no api-token) is not sufficient for a REST
+#     request; the api-token acts as a required CSRF guard.
+$t->get_ok($url . $endpoint)->status_is(401);
+
+# 3c. Supplying an invalid Bugzilla_api_token is rejected as a bad request
+#     (auth_invalid_token) rather than silently falling back to anonymous.
+$t->get_ok($url . $endpoint . '&Bugzilla_api_token=invalid-token-value')
+  ->status_is(400);
```
zzzeid, line 103: Wonder if we should test when a token is provided without a cookie?

qa/t/rest_native_login.t
```diff
@@ -0,0 +1,105 @@
...
+#
+# 3. Authentication via the login cookie + Bugzilla_api_token parameter works.
+#    This is the mechanism the web UI (Bugzilla.API) uses, and is the path that
+#    native Mojo REST endpoints previously rejected with a 401.
+#
+
+# Perform a browser login so the user agent's cookie jar holds valid
+# Bugzilla_login and Bugzilla_logincookie cookies.
+
+# The first request seeds the Bugzilla_login_request_cookie (anti-CSRF) cookie.
```
zzzeid, line 61: Not clear what "anti-CSRF" means here, given that there is mention of CSRF prevention being tied to the token, not the cookie.

zzzeid (approved):

qa/t/rest_native_login.t
dklawren: The anti-CSRF token is for ensuring the login POST actually originated from Bugzilla's own login page rather than a forged form on another origin.

The test script is scraping the resulting Bugzilla_login_token out of the rendered form — purely so it can perform a legitimate browser-style login.

PR #2664, opened 2026-07-08
Bug 2052048 - Hide "Close and move to Invalid Bugs" button for bugs reported by editbugs members

Xzzz:
## Summary
- Refuse `close_as_invalid` in `WebService.pm` when the bug's reporter is an editbugs member, throwing `bug_status_unresolvable` before any mutation.
- Hide the "Close and move to Invalid Bugs" button in the bug modal for the same case, alongside the existing editbugs/product/status/security-group checks.
- Add test coverage exercising both the REST route wiring and the reporter-editbugs guard (bug from an editbugs reporter is refused and bug from a plain reporter is still closed and moved to Invalid Bugs). Please note the non-editbugs-reporter assertions are skipped gracefully in environments where the "Invalid Bugs" product doesn't exist.

## Test plan
- `t/bmo/invalid-bug-helper.t` passes (19/19, with 2 environment-dependent assertions skipped when the Invalid Bugs product is absent)
- Manually verified the button is hidden on a bug reported by an editbugs member and shown otherwise

## References

- Bugzilla link: [2052048](https://bugzilla.mozilla.org/show_bug.cgi?id=2052048)
- Original intent: [1684509](https://bugzilla.mozilla.org/show_bug.cgi?id=1684509)

### DIFF
--- a/extensions/InvalidBugHelper/lib/WebService.pm
+++ b/extensions/InvalidBugHelper/lib/WebService.pm
@@ -36,7 +36,7 @@ sub close_as_invalid {
   my $bug = Bugzilla::Bug->check({id => $bug_id});
 
   if ($bug->reporter->in_group('editbugs')) {
-    ThrowUserError('bug_status_unresolvable', {bug => $bug});
+    ThrowUserError('invalid_bug_reporter_is_trusted', {bug => $bug});
   }
 
   # Block non-members from closing bugs with mandatory security groups.
--- a/extensions/InvalidBugHelper/lib/WebService.pm
+++ b/extensions/InvalidBugHelper/lib/WebService.pm
@@ -35,6 +35,10 @@ sub close_as_invalid {
 
   my $bug = Bugzilla::Bug->check({id => $bug_id});
 
+  if ($bug->reporter->in_group('editbugs')) {
+    ThrowUserError('bug_status_unresolvable', {bug => $bug});
+  }
+
   # Block non-members from closing bugs with mandatory security groups.
   foreach my $group (@{$bug->groups_in}) {
     my $gc = $bug->product_obj->group_controls->{$group->id};
--- a/extensions/InvalidBugHelper/t/bmo/invalid-bug-helper.t
+++ b/extensions/InvalidBugHelper/t/bmo/invalid-bug-helper.t
@@ -70,7 +70,7 @@ ok($param_names{invalidbughelper_warning_text}, 'warning_text param defined');
 # member, regardless of who is performing the close (Bug 1684509).
 
 BEGIN { Bugzilla->extensions }
-Bugzilla->usage_mode(USAGE_MODE_CMDLINE);
+Bugzilla->usage_mode(USAGE_MODE_TEST);
 Bugzilla->error_mode(ERROR_MODE_DIE);
 
 my $product = Bugzilla::Product->new({name => 'Firefox'})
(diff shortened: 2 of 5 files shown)


extensions/InvalidBugHelper/lib/WebService.pm
```diff
@@ -35,6 +35,10 @@ sub close_as_invalid {
 
   my $bug = Bugzilla::Bug->check({id => $bug_id});
 
+  if ($bug->reporter->in_group('editbugs')) {
+    ThrowUserError('bug_status_unresolvable', {bug => $bug});
```
dklawren: bug_status_unresolvable is not good fit for this tyoe of error. Looking at others available and also did not find a good match so we will need to create a new one.

The nice part: the extension already owns its error template (extensions/InvalidBugHelper/template/en/default/hook/global/user-error-errors.html.tmpl), so this is a two-line addition, no core changes:

```
  [% ELSIF error == "invalid_bug_reporter_is_trusted" %]
    [% title = BLOCK %][% terms.Bug %] Cannot Be Closed as Invalid[% END %]
    This [% terms.bug %] was reported by a member of the 'editbugs' group and
    cannot be closed as invalid.
  [% END %]
```

(Note the existing file uses IF error == for its single block; a second block should be ELSIF, or a separate IF — either works, but ELSIF is cleaner.)

Then in WebService.pm:
  
```perl
 if ($bug->reporter->in_group('editbugs')) {
    ThrowUserError('invalid_bug_reporter_is_trusted', {bug => $bug});
  }
```

And update the test regex from qr/Cannot Be Closed as Invalid|already resolved/ to match the new message (the title is still "Cannot Be Closed as Invalid", so that half of the regex keeps working — but tighten it to the new body text to make the assertion meaningful).

Bottom line: create a new error. It's cheap (extension-local, no core touch), and it's the only way to give an accurate message on a path that's only reachable via direct API calls where accuracy matters most.

dklawren (changes requested):

extensions/InvalidBugHelper/lib/WebService.pm
Xzzz: Fixed following your advices (commit "Bug 2052048 - Use a dedicated error for the editbugs-reporter close_as_invalid guard")
Thanks!

dklawren (changes requested):
Some minor changes:
- Make the second test assertion unconditional so a regression can't be silently skipped.
- Switch the test to USAGE_MODE_TEST for consistency with sibling t/bmo tests
- Move the t/bmo/invalid-bug-helper.t to extensions/InvalidBugHelper/t/bmo/invalid-bug-helper.t. .github/workflows/ci.yml will look in extension dirs as well for t/bmo/*.t and t/*.t.

Xzzz:
Modified following your advices:
- File `t/bmo/invalid-bug-helper.t` moved to `extensions/InvalidBugHelper/t/bmo/invalid-bug-helper.t` to match .github/workflows/ci.yml's extension glob (`extensions/*/t/bmo/*.t`)
- Modified USAGE_MODE_CMDLINE to USAGE_MODE_TEST at line 73 for consistency with t/bmo tests
- Made product assertion unconditionnal and modified `$result->{product}` to `ref $result eq 'HASH' ? $result->{product} : undef` at line 144 to avoid an undef hashref dereference

dklawren (approved):
LGTM r=dkl

PR #2665, opened 2026-07-08
Bug 2053504 - Need to add index to tracking_flags_values table for better performance from search

dklawren:


extensions/TrackingFlags/Extension.pm
```diff
@@ -277,6 +277,7 @@ sub db_schema_abstract_schema {
     INDEXES => [
       tracking_flags_values_idx =>
         {FIELDS => ['tracking_flag_id', 'value'], TYPE => 'UNIQUE',},
+      tracking_flags_values_tracking_flag_id_idx => ['tracking_flag_id'],
```
cgsheeh, line 280: Looking at the plan from the bug, the filter passes all 7,014 rows it scans, and Query Insights has it averaging 3,499 rows returned out of a ~7,000 row table. At that selectivity MySQL will keep picking the table scan over a range scan, so I don't think this index will change the plan. Have you been able to try it on a copy of prod?

extensions/TrackingFlags/Extension.pm
```diff
@@ -341,6 +342,10 @@ sub install_update_db {
   $dbh->bz_add_column('tracking_flags_values', 'comment',
     {TYPE => 'TEXT', NOTNULL => 0,},
   );
+
+  $dbh->bz_add_index('tracking_flags_values',
+    'tracking_flags_values_tracking_flag_id_idx',
+    [qw(tracking_flag_id)]);
```
cgsheeh, line 348: The other thing that stands out in the screenshot from the bug is 2.4M calls at 9ms - `preload_all_the_things` re-runs this on every `Flag->match`, and `get_all` only caches in `request_cache`. Since the values change rarely, would caching them with memcached get us more than an index will?

cgsheeh (approved):

Xzzz (approved):
lgtm

Non-blocking follow-up: Connor's question above is still open. Would be good to get an `EXPLAIN` before/after on a prod copy to confirm this is actually changing the plan. If it turns out not to help, Connor's memcached suggestion for `preload_all_the_things`/`get_all` is probably the bigger lever.

Approving since the change itself is low-risk either way

Xzzz (approved):
LGTM => This is a solid response to our follow-up. I approve as-is

PR #2668, opened 2026-07-14
Bug 2054574 - Do not auto-assigning Hackbot when submitting a Phabricator revision

dklawren:


extensions/PhabBugz/template/en/default/admin/params/phabbugz.html.tmpl
```diff
@@ -16,5 +16,12 @@
      phabricator_enabled     => 'Enable Phabricator Integration',
      phabricator_base_uri    => 'Phabricator Base URI',
      phabricator_api_key     => 'Phabricator User API Key',
+     phabricator_auto_assign_exclude_list =>
+       'A space or comma delimited list of Bugzilla accounts (login names).'
```
shtrom, line 20: nit: is `space or comma` a perl built-in? Otherwise, I'd suggest to just pick one

shtrom (approved):
lgtm

extensions/PhabBugz/template/en/default/admin/params/phabbugz.html.tmpl
dklawren: Nothing Perl specific. We have this for other params in the BMO code. It uses a regex of `[\s,]+` in most places. I will keep the suggestion in mind for the next change.

PR #2676, opened 2026-07-21
Bug 2056682 - Added Net::CIDR as well as updated Perl and other dependencies to the base image

dklawren:
- Many updates to various dependencies
- Workarounds were added to fix some build issues with newer Perl version


cgsheeh (approved):
Seems fine to me, but worth calling out that the PR title doesn't quite align with the actual contents. The title makes it seem like we're adding a new dependency, but this PR also updates the snapshot file, bumps the base image tag, makes Dockerfile changes, etc.

PR #2683, opened 2026-07-24
Bug 2057679 - Update docker build process to pull in latest builds of vendered javascript libraries needed by the client

dklawren:


package.json
```diff
@@ -0,0 +1,24 @@
...
+  "version": "0.0.0",
+  "private": true,
+  "description": "Third-party front-end libraries for BMO, managed via npm so security updates flow through Dependabot instead of hand-vendored blobs.",
+  "license": "MPL-2.0",
+  "type": "module",
+  "scripts": {
+    "build": "node scripts/build-frontend.mjs",
+    "verify": "node scripts/build-frontend.mjs --check"
+  },
+  "//": "Versions are pinned to what was previously vendored so this change is behaviour-neutral. Dependabot proposes upgrades; treat majors (jquery 4.x, mermaid 11.x, devbridge 2.x) as manual, tested reviews. bpopup is not published to npm and remains vendored under js/jquery/plugins/bPopup/.",
```
cgsheeh: @dklawren Could we stop committing the generated libraries altogether and let the `assets` stage produce them at image build time? I went looking for something that would break and didn't find much. It would make the process of adding new JS deps much simpler, allow dependabot to fully update and run tests in CI, reduce the amount of irrelevant code in the repo, etc.

dklawren: I think the legacy reason for this is because bugzilla trying to not be docker only and also allow for packaging such rpm, deb, etc. So all files needed for proper operation would need to be in-repo. But since we are more than likely always going to be docker based and new images built at test and deploy time I am fine with making this change.

Xzzz (approved):
LGTM, nice cleanup and the design logic is already settled in-thread with Connor.

Few things I specifically checked:
- Dockerfile ordering: `COPY --from=assets /build/out /app/js` lands before the `chown -R app:app /app`, and after `COPY . /app`, so permissions are correct
- .gitignore entries line up exactly with TARGETS/TARGET_DIRS in build-frontend.mjs: no risk to see a generated file to be integrated into a commit
- js/duo-min.js is deleted with nothing replacing it: grepped for duo-min/DuoWeb across the repo => no template or JS references it. The Duo MFA flow is server-side OAuth redirect (Bugzilla/MFA/Duo.pm), not the old client-side SDK, so this is well dead-code removal
- The earlier Copilot comment about TARGET_DIRS not recursing into subdirectories has been correctly addressed: copyDir() in the current version does recurse now

PR #2687, opened 2026-07-27
Bug 2058190: [GitHub] fix race condition on adding PRs to bugs

justdave:
* [Bug 2058190](https://bugzilla.mozilla.org/show_bug.cgi?id=2058190)

Turns out there was already an attempt to avoid duplicates in the code, but it did not lock the bug around the duplicate check, which meant if you received two notifications in a row for the same PR, they could both pass the duplicate check before either one wrote to the database.

This PR fixes it by grabbing a row-level lock on the bugs table around the duplicate check, so that the second one coming in has to wait in line for the first one to finish before it checks if there's a duplicate.

### DIFF
--- a/Bugzilla/API/V1/Github.pm
+++ b/Bugzilla/API/V1/Github.pm
@@ -114,12 +114,34 @@ sub pull_request {
   eval {
     $dbh->bz_start_transaction;
 
-    # Serialize all PR-link writes for this bug to make opened-event handling
-    # idempotent under concurrent deliveries.
-    $dbh->selectrow_array(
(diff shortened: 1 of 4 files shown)


justdave:
I see ! vs eq precedence issues being reported in the test, you can fix that by cherry-picking https://github.com/bugzilla/harmony/commit/2be0c4ceae1238709af4d229ef40ec8f0b8878bc out of Bugzilla Harmony (we just fixed that last week)

justdave:
The remaining nit from CoPilot that I didn't fix yet is a timing issue with the test...  and due to the nature of the test, there will always be timing issues with it, and I don't see a way to force that to be concurrent. You get lucky sometimes, just like in production.

PR #2696, opened 2026-08-06
Bug 2061441 - Migrate Classification REST resource to native Mojo API

Xzzz:
## Summary

Ports `Bugzilla::WebService::Classification`'s `get` method into a native `Bugzilla::API::V1::Classification` Mojo controller, mirroring the pattern already used for Component/Teams/Reminders/Configuration/User/etc.

This is the first of a set of child bugs migrating REST resources to native Mojo API, see [2057358](https://bugzilla.mozilla.org/show_bug.cgi?id=2057358) for details.

## Changes

- Add `Bugzilla/API/V1/Classification.pm`: `GET /rest/classification/:id_or_name`, same permission logic (`useclassification` param / `editclassifications` group) and JSON response shape as the legacy endpoint
- Delete `Bugzilla/WebService/Classification.pm` and `Bugzilla/WebService/Server/REST/Resources/Classification.pm`
- Remove the `Classification` entry from `WS_DISPATCH` in `Bugzilla/WebService/Constants.pm` and drop the corresponding `use` line in `Bugzilla/WebService/Server/REST.pm`, since both `REST.pm::preload` and `_find_resource` iterate that dispatch table and would otherwise try to load a now-deleted module

## Test plan

- Verified against a local docker instance:
  - `GET /rest/classification/1` and `GET /rest/classification/Unclassified` (by id and by name) both return the expected `classifications` JSON
  - `GET /rest/classification/99999` and `.../NoSuchClass` both throw the correct `object_does_not_exist` error
  - Confirmed unmatched paths still fall through correctly to the legacy `rest.cgi` compat route
- Ran the sanity/bmo/webservices/selenium1-4 test suites clean

## References

- Bugzilla link: [2061441](https://bugzilla.mozilla.org/show_bug.cgi?id=2061441)


Bugzilla/API/V1/Classification.pm
```diff
@@ -0,0 +1,76 @@
...
+use Mojo::Base qw( Mojolicious::Controller );
+
+use Bugzilla::Classification;
+use Bugzilla::Constants;
+
+sub setup_routes {
+  my ($class, $r) = @_;
+  my $class_routes = $r->under(
+    '/classification' => sub { Bugzilla->usage_mode(USAGE_MODE_MOJO_REST); });
+  $class_routes->get('/:id_or_name')->to('V1::Classification#get');
```
dklawren: `:id_or_name` is a standard mojo placeholder, so it matches `[^/.]+` and will not match a classification name containing a dot. the legacy resource regex was `[^/]+`, so `GET /rest/classification/Foo.Bar` used to work and now falls through to the legacy rest.cgi route, which no longer knows this resource

use the relaxed placeholder instead:

```perl
$class_routes->get('/#id_or_name')->to('V1::Classification#get');
```

Bugzilla/API/V1/Classification.pm
```diff
@@ -0,0 +1,76 @@
...
+sub _classification_to_hash {
+  my ($self, $classification, $user) = @_;
+
+  my $products
+    = $user->in_group('editclassifications')
+    ? $classification->products
+    : $user->get_selectable_products($classification->id);
+
+  return {
+    id          => $classification->id,
```
dklawren: `id` and `sort_key` lose their numeric json type here. the legacy code ran them through `$self->type('int', ...)` which does a literal `int($value)` (Bugzilla/WebService/Server/JSONRPC.pm:211), guaranteeing `"id": 1`

now the raw DBI value goes straight to Mojo::JSON, so a value read from mysql serializes as `"id": "1"`. worse, `Bugzilla::Classification` is `IS_CONFIG => 1` so it is memcached, and a cache hit round-trips through json and comes back as a real number, meaning the type flips depending on cache state

docs/en/rst/api/core/v1/classification.rst:62 documents these as `int`. same applies to the product `id` in `_product_to_hash`. suggest forcing numeric context, e.g. `id => 0 + $classification->id`

Bugzilla/API/V1/Classification.pm
```diff
@@ -0,0 +1,76 @@
...
+
+  my @classifications;
+  if ($user->in_group('editclassifications')
+    || grep { $_->id == $classification->id }
+    @{$user->get_selectable_classifications})
+  {
+    @classifications = ($self->_classification_to_hash($classification, $user));
+  }
+
+  return $self->render(json => {classifications => \@classifications});
```
dklawren, line 54: `include_fields` / `exclude_fields` are no longer honored. the legacy method wrapped both hashes in `filter $params` and its POD documented these as accepted, so a client doing `?include_fields=name` now gets the full object back instead of a trimmed one

worth either supporting it here or calling the drop out explicitly in the PR description and docs

dklawren (commented):
Overall the port matches the established `Bugzilla::API::V1::*` pattern, and the cleanup of `WS_DISPATCH` plus the two deleted modules checks out. I verified there is no `jsonrpc.cgi` in the tree, so `WS_DISPATCH` is only consumed by `REST.pm::preload` and `_find_resource`, and   
  there are no dangling references to either deleted module. Route ordering also puts `API.pm` ahead of the `rest.cgi` fallback, so the new route wins.

Four things I would like addressed before this lands. The first three are contract regressions against the endpoint this replaces.

### 1. Route placeholder narrows what names match

`Bugzilla/API/V1/Classification.pm:20`

`:id_or_name` is a standard Mojolicious placeholder, so it matches `[^/.]+`. The legacy resource regex was `[^/]+`. A classification name containing a dot used to resolve and now falls through to the legacy `rest.cgi` route, which no longer knows this resource, so the client gets a confusing 404 instead of data.

```perl
$class_routes->get('/#id_or_name')->to('V1::Classification#get');
```

### 2. `id` and `sort_key` lose their numeric JSON type

`Bugzilla/API/V1/Classification.pm:58`

The legacy code ran these through `$self->type('int', ...)`, which does a literal `int($value)` (`Bugzilla/WebService/Server/JSONRPC.pm:211`), guaranteeing `"id": 1`. Now the raw DBI value goes straight to `Mojo::JSON`, so a value read from MySQL serializes as `"id": "1"`.        

It gets worse than a flat change: `Bugzilla::Classification` is `IS_CONFIG => 1`, so it is memcached, and a cache hit round-trips through JSON and comes back as a real number. The type flips depending on cache state, which is harder to catch than a consistent break.

`docs/en/rst/api/core/v1/classification.rst:62` documents both as `int`. The same applies to the product `id` in `_product_to_hash`. Forcing numeric context (`0 + $classification->id`) is enough.

### 3. `include_fields` and `exclude_fields` are no longer honored

`Bugzilla/API/V1/Classification.pm:46`

The legacy method wrapped both hashes in `filter $params`, and its POD documented these as accepted. A client calling `?include_fields=name` now gets the full object back. Either support it here or call the drop out explicitly in the PR description and the docs.

dklawren (changes requested):

Bugzilla/API/V1/Classification.pm
Xzzz: Change applied. Thanks!

Bugzilla/API/V1/Classification.pm
Xzzz: Changes applied => both now forced to numeric context (`0 + ...`) for classifications and products, so they serialize as JSON numbers regardless of whether the value came from DB or round-tripped through memcached.
Thanks!

Bugzilla/API/V1/Classification.pm
Xzzz: Modifications done: reused `Bugzilla::WebService::Util::filter` instead of dropping the feature (same helper the legacy method used). Query params are split the same way `REST.pm` used to, and the `products` sub-objects are filtered with the `products.` prefix to match old behavior.

Bugzilla/API/V1/Classification.pm
```diff
@@ -0,0 +1,92 @@
...
+use Bugzilla::Classification;
+use Bugzilla::Constants;
+use Bugzilla::WebService::Util qw(filter);
+
+sub setup_routes {
+  my ($class, $r) = @_;
+  my $class_routes = $r->under(
+    '/classification' => sub { Bugzilla->usage_mode(USAGE_MODE_MOJO_REST); });
+  $class_routes->get('/#id_or_name')->to('V1::Classification#get');
+}
```
dklawren, line 32: only GET is registered here, so `OPTIONS /rest/classification/<id>` no longer matches and backtracks to the `rest.cgi` fallback, which now throws `rest_invalid_resource` (404) because the resource module was deleted

before this change the legacy server answered that preflight with 200 plus `Allow: GET` and `Access-Control-Allow-Methods: GET` (see `Bugzilla/WebService/Server/REST.pm:55`), so cross-origin browser clients that trigger a preflight, for example anything sending `X-Bugzilla-API-Key`, will now be blocked

`Access-Control-Allow-Origin` and `Access-Control-Allow-Headers` are already set by `_prepare_rest_request`, so only `Allow` and `Access-Control-Allow-Methods` need adding

```suggestion
  $class_routes->get('/#id_or_name')->to('V1::Classification#get');
  $class_routes->options('/#id_or_name')->to('V1::Classification#options');
}

sub options {
  my ($self) = @_;

  $self->res->headers->header('Allow'                        => 'GET');
  $self->res->headers->header('Access-Control-Allow-Methods' => 'GET');

  return $self->rendered(200);
}
```

dklawren (changes requested):

Bugzilla/API/V1/Classification.pm
Xzzz: Fixed as suggested. Thanks!

Xzzz:
Found "unable to connect to dbi:mysql:database=bugs;host=bmo.db as bugs at /app/scripts/entrypoint.pl line 248." in selenium4 logs.
I then run selenium4 test locally and it succeeded... Maybe DB container (bmo.db) has been slow to come up?
As I dont have permissions to rerun test jobs, Ive created an empty commit and pushed it to rerun full test suite ;)

Bugzilla/API/V1/Classification.pm
```diff
@@ -0,0 +1,102 @@
...
+use Bugzilla::Classification;
+use Bugzilla::Constants;
+use Bugzilla::WebService::Util qw(filter);
+
+sub setup_routes {
+  my ($class, $r) = @_;
+  my $class_routes = $r->under(
+    '/classification' => sub { Bugzilla->usage_mode(USAGE_MODE_MOJO_REST); });
+  $class_routes->get('/#id_or_name')->to('V1::Classification#get');
+  $class_routes->options('/#id_or_name')->to('V1::Classification#options');
```
dklawren: the new mojo route only covers `/rest/classification/...`, but `/rest.cgi/*PATH_INFO` is still routed to rest.cgi in `Bugzilla::App::Controller::CGI`. with the resource module deleted and the `WS_DISPATCH` entry gone, `GET /rest.cgi/classification/1` no longer resolves and returns an unknown resource error instead of the classification. worth either adding a compat route or confirming no clients use the `rest.cgi` form

dklawren (approved):
LGTM r=dkl

PR #2698, opened 2026-08-12
Bug 1883428 - Improve the needinfo email situation

Xzzz:
## Summary

Needinfo and other flag notifications (request/grant/deny/clear) were previously sent as a separate email via `Flag::notify()`. This decouples that: flag events are now rendered as a section inside the normal bugmail, sent through `Bugzilla/BugMail.pm`, instead of a standalone notify() email.

- `Bugzilla/BugMail.pm`: `_get_flag_mail_events()` gathers flag activity (requested/answered, including needinfo auto-clear on reply - status `X`, which the old notify()-based flow handled but was initially missed here) and resolves the related attachment so templates can check `can_review` / `external_redirect`
- Core `template/en/default/email/bugmail.{txt,html}.tmpl` and the BMO override `extensions/BMO/template/en/default/email/bugmail.{txt,html}.tmpl` render a `flag_events` section, with a `Hook.process('flag_event', ...)` extension point
- `extensions/BMO` fully overrides the core bugmail templates on this instance, so the flag-event content is ported into both
- Needinfo and Splinter hook fragments (reporter-aware wording + user-guide link; review-tool link) are re-added under the new `flag_event` hook, replacing the ones orphaned by the earlier removal of `request/email.txt.tmpl`. Splinter's review link now gates on `attachment.can_review` instead of `ispatch`, so it also covers GitHub PR/Phabricator attachments, matching other call sites in the codebase
- BMO's own external-redirect attachment link is inlined directly in the BMO templates rather than hooked, since TT's `TRIM` strips each hook fragment's edges independently and concatenating BMO's + Splinter's fragments with no separator produced garbled text when both fire on the same event

## Test plan
- Verified end-to-end on a local instance: needinfo request, needinfo auto-clear (reply), needinfo requested-from-reporter wording, plain review flag (Splinter link), review flag on a GitHub PR attachment (both BMO and Splinter links render, correctly separated), both text and HTML bugmail formats
- `perl -c` on modified Perl files
- Grepped the repo for orphaned references to the removed `Flag::notify`/hook fragments

## References

- Bugzilla link: [1883428](https://bugzilla.mozilla.org/show_bug.cgi?id=1883428)
- Supersedes (at least): [1410772](https://bugzilla.mozilla.org/show_bug.cgi?id=1410772), [1774541](https://bugzilla.mozilla.org/show_bug.cgi?id=1774541) and [2025631](https://bugzilla.mozilla.org/show_bug.cgi?id=2025631)


Bugzilla/BugMail.pm
```diff
@@ -701,6 +760,140 @@ sub _generate_bugmail {
...
+      # 'X' is a flag cleared without +/- (e.g. needinfo auto-cleared when
+      # the requestee replies) -- notify.() treated that the same as an
+      # explicit answer, so we do too.
+      #
+      # flags.setter_id gets overwritten to whoever granted/denied/cleared the
+      # flag, so the requester has to be found by looking back at this
+      # flag's most recent '?' activity row.
+      my ($requester_id) = $dbh->selectrow_array(
+        "SELECT setter_id FROM flag_activity
+          WHERE flag_id = ? AND status = '?' AND flag_when <= ?
```
dklawren: this looks back to the most recent `?` row for the flag, but never checks the flag was *still* in `?` state when this row was written, so an already-answered flag that later gets cleared re-notifies the original requester

old `notify()` gated on `$old_flag->status eq '?'`, which excluded this case

repro: A sets `review?` on an attachment, B sets `review+`, then the attachment is obsoleted or the bug is moved to another product. both paths write a `status => 'X'` flag_activity row (`Flag::update_flags` and `Flag::force_retarget`). processing that `X` row finds A's original `?` row and A gets "Your request for review has been cleared by <whoever obsoleted it>", which never used to be sent. attachments get obsoleted constantly so this is a lot of spurious mail

suggest fetching the immediately preceding activity row for the flag instead and requiring its status to be `?`, rather than the most recent `?` row at any point in history

Bugzilla/BugMail.pm
```diff
@@ -482,6 +540,7 @@ sub sendMail {
     referenced_bugs    => $referenced_bugs,
     bugmailtype        => $bugmailtype,
     blocker_short_desc => $blocker_short_desc,
+    flag_events        => $flag_events,
```
dklawren, line 601: `flag_events` is added to `$vars` but `enqueue()` below never flattens it, so with `use_mailer_queue` on the TheSchwartz payload carries live blessed `Bugzilla::FlagType`, `Bugzilla::Attachment` and `Bugzilla::User` objects

every other object in `$vars` goes through `_flatten_object` for that reason, and `dequeue()` re-inflates them with `new_from_hash`. the `setter` here is also a shared `%user_cache` object that the recipient loop has already populated with lazily-built fields (`groups`, `mail_settings`, ...), so what gets frozen is a much larger and less predictable graph than a flattened hash

needs flattening in `enqueue()` plus matching inflation in `dequeue()`. that also gives you a place to re-check flag-event attachment visibility at send time, which the other TOCTOU re-checks in `dequeue()` already do

Bugzilla/BugMail.pm
```diff
@@ -199,6 +201,30 @@ sub Send {
...
+  # that's now been granted/denied, becomes a recipient even without any
+  # other role on the bug. Flag-type cc_list addresses (admin-configured,
+  # bypasses per-user opt-in) are added the same way notify() used to.
+  my @flag_events
+    = $params->{dep_only}
+    ? ()
+    : _get_flag_mail_events($bug, $start, $end, \%user_cache);
+
+  foreach my $event (@flag_events) {
+    if ($event->{action} eq 'requested') {
```
dklawren, line 219: the private-attachment check is missing on this path

old `notify()` only added the addressee when `(!$attachment_is_private || $addressee->is_insider)`. `_get_flag_type_cc` below keeps that check for cc_list addresses, but requestee/requester recipients no longer get it, so a non-insider can receive "<setter> has requested review for attachment 12345 from you"

the BMO template also renders `event.attachment.external_redirect` -> `event.attachment.data` (the URL), and the Splinter hook renders a review link for the attachment id, so it is more than just the id leaking. `can_see_bug` is still enforced, but insider gating on the attachment is not

simplest fix is to drop events whose attachment is private when the recipient is not an insider, in the per-recipient `@user_flag_events` grep and when deciding to add the recipient at all

template/en/default/email/bugmail-flagtype-cc.txt.tmpl
```diff
@@ -0,0 +1,32 @@
...
+  #%]
+
+[%# ponytail: minimal fallback for flag type cc_list addresses with no
+  # Bugzilla account -- the normal bugmail pipeline is keyed on user id
+  # throughout and has no path to a raw address. Revisit if this needs to
+  # carry more than a one-line-per-event summary (diffs, comments, HTML). %]
+
+From: [% Param('mailfrom') %]
+To: [% to %]
+Subject: [[% terms.Bug %] [%+ bug.id %]] [% bug.short_desc %]
```
dklawren, line 21: `terms` is not defined here, so this renders as `Subject: [ 12345] summary`

`global/variables.none.tmpl` has to be PROCESSed. the core bugmail templates get it transitively via `global/field-descs.none.tmpl`, and the deleted `request/email.txt.tmpl` did `[% PROCESS global/variables.none.tmpl %]` explicitly. this template only INCLUDEs `email/header-common.txt.tmpl`, which does not process it

same file, two more things while you are in here:
- line 30 renders status `X` as "denied"; the other templates say "cleared" for it
- no `@@body-headers@@` placeholder, so BMO's `_replace_placeholder_in_part` has nothing to substitute and these mails lose the body headers the old request mail carried

Bugzilla/BugMail.pm
```diff
@@ -701,6 +760,140 @@ sub _generate_bugmail {
...
+
+  my (%account_recipients, %raw_addresses);
+  foreach my $event (@$flag_events) {
+    my $cc_list = $event->{type}->cc_list;
+    next unless $cc_list;
+
+    my $attachment_is_private
+      = $event->{attachment_id}
+      ? Bugzilla::Attachment->new({id => $event->{attachment_id}, cache => 1})
+      ->isprivate
```
dklawren: this re-fetches an attachment that `$event->{attachment}` already holds, and derefs the result unconditionally, so it dies with "Can't call method isprivate on an undefined value" if the attachment row is gone (flag_activity is a historical log, the attachment it points at can have been deleted)

`$event->{attachment} ? $event->{attachment}->isprivate : 0` covers both

template/en/default/global/reason-descs.none.tmpl
```diff
@@ -24,6 +24,9 @@
   ${constants.REL_QA}       => "You are the QA Contact for the ${terms.bug}.",
   ${constants.REL_CC}       => "You are on the CC list for the ${terms.bug}.",
   ${constants.REL_GLOBAL_WATCHER} => "You are watching all $terms.bug changes.",
+  ${constants.REL_FLAG_REQUESTEE} => "A flag was requested of you on this ${terms.bug}.",
+  ${constants.REL_FLAG_REQUESTER} => "You requested a flag that was set on this ${terms.bug}.",
+  ${constants.REL_FLAG_TYPE_CC}   => "You are on the CC list for this flag type.",
```
dklawren, line 29: `watch_reason_descs` below did not get matching entries

flag recipients are added to `%recipients` before the `watch` query in `Send`, so user-watchers inherit `REL_FLAG_*` with `BIT_WATCHING`. the templates guard with `IF watch_reason_descs.$reason` so nothing errors, but such a recipient gets a bugmail with an empty "You are receiving this mail because" entry and no flag section (the per-recipient grep filters on their own user id)

worth deciding whether watchers should inherit these roles at all, and adding the three watch descriptions if they should

dklawren (changes requested):

Bugzilla/BugMail.pm
Xzzz: Fixed following your suggestion. I just used `flag_when < ?` instead of `<=`. If two `flag_activity` rows for the same flag can ever share an identical timestamp, `<` is a safer choice (it excludes current row).

Bugzilla/BugMail.pm
Xzzz: Fixed in "Bug 1883428 - Gate flag-event recipients and mail content on private-attachment insider access" commit. Thanks!

Bugzilla/BugMail.pm
Xzzz: Fixed following your suggestion. Thanks!

template/en/default/email/bugmail-flagtype-cc.txt.tmpl
Xzzz: Modifications done, three bugs fixed:
- added `[% PROCESS global/variables.none.tmpl %]` to fix `terms` issue
- status 'X' now render as "cleared" like the other templates
- added `@@body-headers@@` placeholder

Bugzilla/BugMail.pm
Xzzz: Modifications done:
- enqueue(): flatten flag_events before it hits the job queue
- dequeue(): inflate flag_events and re-check attachment visibility at send time

template/en/default/global/reason-descs.none.tmpl
Xzzz: I prefer excluding flag roles from watcher inheritance rather than adding the three descriptions.
The reason is flag events are addressed to a specific person (someone asked them, or they asked and got answered). Watching that person's general bug activity doesn't carry that same meaning, and the flag-event section itself is filtered per-recipient on the actual requestee/requester's user id, so a watcher would never see the flag content regardless. Adding watch_reason_descs entries would just give a non-empty reason line for a section that's always empty for them, which seemed worse than not sending the role at all.
=> Fixed by skipping REL_FLAG_REQUESTEE/REQUESTER/TYPE_CC in the watcher-inheritance loop in Send().

Bugzilla/BugMail.pm
```diff
@@ -199,6 +201,44 @@ sub Send {
...
+    my $attachment_private = $event->{attachment} && $event->{attachment}->isprivate;
+
+    if ($event->{action} eq 'requested') {
+      my $requestee_id = $event->{requestee_id};
+      my $requestee
+        = $user_cache{$requestee_id} ||= Bugzilla::User->new({id => $requestee_id, cache => 1});
+      $recipients{$requestee_id}->{+REL_FLAG_REQUESTEE} = BIT_DIRECT
+        if !$attachment_private || ($requestee && $requestee->is_insider);
+    }
+    elsif ($event->{action} eq 'answered') {
```
dklawren, line 227: `answered` adds `REL_FLAG_REQUESTER` even when the setter is the requester, so clearing your own needinfo mails you back. `wants_mail([EVT_REQUESTED_FLAG], REL_ANY)` also skips the `EVT_CHANGED_BY_ME` suppression that `wants_bug_mail` would have applied, and the filter at line 407 then drops the event, so the recipient gets a bugmail about their own change with an empty flag section. suggest mirroring the `setter->id != $requester_id` check here

Bugzilla/BugMail.pm
```diff
@@ -199,6 +201,44 @@ sub Send {
...
+        = $user_cache{$requester_id} ||= Bugzilla::User->new({id => $requester_id, cache => 1});
+      $recipients{$requester_id}->{+REL_FLAG_REQUESTER} = BIT_DIRECT
+        if !$attachment_private || ($requester && $requester->is_insider);
+    }
+  }
+
+  my ($flag_type_cc, $flag_type_cc_raw)
+    = _get_flag_type_cc($bug, \@flag_events, \%user_cache);
+  foreach my $user_id (keys %$flag_type_cc) {
+    $recipients{$user_id}->{+REL_FLAG_TYPE_CC} = BIT_DIRECT;
```
dklawren, line 245: `REL_FLAG_TYPE_CC` recipients never match the `@user_flag_events` filter at line 407, which only tests `requestee_id`/`requester_id`. so account holders on a flag type cc_list get a bugmail with no flag section, while account-less addresses on the same list do get the per-event summary via `bugmail-flagtype-cc.txt.tmpl`. same admin-configured list, two different outcomes

Bugzilla/BugMail.pm
```diff
@@ -373,6 +460,7 @@ sub sendMail {
   my $referenced_bugs    = $params->{referenced_bugs};
   my $dep_only           = $params->{dep_only};
   my $blocker_short_desc = $params->{blocker_short_desc};
+  my $flag_events        = $params->{flag_events} || [];
```
dklawren, line 491: `sendMail` returns 0 further down when `!@display_diffs && !@send_comments`, and `flag_events` is not consulted, so a flag-only notification can be dropped. repro: new bug with an insider-only description plus a `needinfo?` on a non-insider. `$start` is undef so `@diffs` is empty, the single comment is filtered out, and the requestee gets nothing. the old `notify()` sent a mail here

template/en/default/request/email.txt.tmpl
```diff
@@ -1,107 +0,0 @@
...
-[% END %]
-From: [% Param('mailfrom') %]
-To: [% to %]
-Subject: [% flagtype_name %] [%+ subject_status %]: [[% terms.Bug %] [%+ bug.bug_id %]] [% bug.short_desc %]
-[%- IF attachment %] :
-  [Attachment [% attachment.id %]] [% attachment.description FILTER clean_text %][% END %]
-Date: [% date %]
-X-Bugzilla-Type: request
-[%- IF flag.requestee %]
-X-Bugzilla-Flag-Requestee: [% flag.requestee.email %]
```
dklawren, line 61: this was the only place `X-Bugzilla-Flag-Requestee` was emitted. with the `needinfo requested:` subject prefix also going away, users who filter needinfo mail have nothing left to match on. worth emitting it from `flag_events` in `email/bugmail-header.txt.tmpl`

Bugzilla/BugMail.pm
```diff
@@ -701,6 +831,145 @@ sub _generate_bugmail {
...
+      # flags.setter_id gets overwritten to whoever granted/denied/cleared the
+      # flag, so the requester has to be found by looking back at this
+      # flag's activity. Only treat it as an answer if the immediately
+      # preceding row was still '?' -- old notify() gated on that too, so
+      # e.g. an already-answered flag later cleared by an attachment being
+      # obsoleted (also logged as status 'X') doesn't re-notify the
+      # original requester.
+      my ($prev_status, $requester_id) = $dbh->selectrow_array(
+        "SELECT status, setter_id FROM flag_activity
+          WHERE flag_id = ? AND flag_when < ?
```
dklawren: `flag_when < ?` against a `DATETIME` populated from `LOCALTIMESTAMP(0)` means a `?` and its answer landing in the same second make this lookup miss, `$prev_status` comes back undef and the event is skipped by the `next` below. plausible with automation that sets and clears flags. ordering by `(flag_when, id)` and breaking the tie on `id` would be exact

Bugzilla/BugMail.pm
```diff
@@ -228,6 +268,17 @@ sub Send {
...
     foreach my $watch (@$userwatchers) {
       while (my ($role, $bits) = each %{$recipients{$watch->[1]}}) {
+
+        # Flag roles are addressed to a specific person (someone asked
+        # them, or they asked and got answered) -- watching that person's
+        # other bug activity isn't the same thing, and without this the
+        # watcher got a content-free mail (empty reason line, no flag
+        # section, since the per-recipient flag_events filter only
+        # matches the actual requestee/requester's own user id).
+        next
```
dklawren, line 285: skipping the role inheritance is right, but `push(@{$watching{$watch->[0]}}, $watch->[1])` a few lines down still runs. a watcher who inherited no role at all still ends up in `%watching`, which gives them `X-Bugzilla-Watch-Reason: None <login>` and an extra `user_id_to_login` lookup in `sendMail`

Bugzilla/BugMail.pm
```diff
@@ -701,6 +831,145 @@ sub _generate_bugmail {
   return $email;
 }
 
+# Find flags that were requested of, or answered by, someone during this
+# window, so sendMail() can add a note at the top of the mail for that
+# person even if they hold no other role on the bug (bug 1883428).
+sub _get_flag_mail_events {
```
dklawren, line 876: no test coverage for any of this. there is no existing `BugMail` test in `t/` to extend so it would be a new file, but the recipient selection here (private attachment gating, the `X` prev-status rule, flag type cc) is cheap to unit test and expensive to get wrong in production

dklawren (changes requested):

Bugzilla/BugMail.pm
Xzzz: Fixed: added `$event->{setter}->id != $requester_id` before adding `REL_FLAG_REQUESTER` (same guard the content filter already has).

Bugzilla/BugMail.pm
Xzzz: Fixed: `_get_flag_type_cc` now tracks the actual events per account recipient instead of a bare 1, and `@user_flag_events` checks membership in that list.

Bugzilla/BugMail.pm
Xzzz: Fixed: `sendMail` "nothing to send" now also checks flag_events.

template/en/default/request/email.txt.tmpl
Xzzz: Fixed: emitting it from `flag_events` in `bugmail-header.txt.tmpl` for `requested` events. Had to carry the requestee/requester User objects through the enqueue/dequeue flatten-inflate cycle since only setter was preserved before.

Bugzilla/BugMail.pm
Xzzz: Fixed by adding `id` to the select and break ties with `(flag_when < ? OR (flag_when = ? AND id < ?)) ORDER BY flag_when DESC, id DESC`

Bugzilla/BugMail.pm
Xzzz: Fixed: `%watching` is now only populated when a role was actually inherited from the watched user.

Bugzilla/BugMail.pm
Xzzz: Added `t/bugmail-flag-events.t` covering `_flag_event_visible_to` (extracted the repeated private-attachment/insider check into one function so it's testable without a DB). The prev-status tie-break and flag-type-cc lookups are SQL-driven and still only covered by the manual test plan.

_Please note I could add a DB-backed test if you prefer have that than the manual plan_

Bugzilla/BugMail.pm
```diff
@@ -560,6 +688,27 @@ sub dequeue {
...
+        %$_,
+        type       => Bugzilla::FlagType->new_from_hash($_->{type}),
+        setter     => Bugzilla::User->new_from_hash($_->{setter}),
+        requestee  => $_->{requestee} ? Bugzilla::User->new_from_hash($_->{requestee}) : undef,
+        requester  => $_->{requester} ? Bugzilla::User->new_from_hash($_->{requester}) : undef,
+        attachment => $_->{attachment_id}
+          ? Bugzilla::Attachment->new({id => $_->{attachment_id}, cache => 1})
+          : undef,
+      }
+    } @{$vars->{flag_events}}
```
dklawren: unguarded deref of `$vars->{flag_events}`. any `bug_mail` job already sitting in TheSchwartz when this deploys was enqueued without a `flag_events` key, so dequeue dies with "Can't use an undefined value as an ARRAY reference" and those mails are lost across retries. use `@{$vars->{flag_events} || []}`

Bugzilla/BugMail.pm
```diff
@@ -521,6 +628,27 @@ sub enqueue {
...
+  # it can also re-check visibility at send time).
+  $vars->{flag_events} = [
+    map {
+      {
+        action        => $_->{action},
+        attachment_id => $_->{attachment_id},
+        requestee_id  => $_->{requestee_id},
+        requester_id  => $_->{requester_id},
+        status        => $_->{status},
+        type          => _flatten_object($_->{type}),
```
dklawren, line 655: `flatten_to_hash` reads `Bugzilla::Object::_serialization_keys`, a per-request cache only filled by `new_from_hash` / `new_from_list` / `_do_list_select`. `_get_flag_mail_events` builds the flag type with `Bugzilla::FlagType->new({id => ...})`, which does not fill it, so in a process that never list-selects flag types this hits `@{undef}` and dies. reachable via `scripts/sendunsentbugmail.pl` -> `BugMail::Send` -> `enqueue`. `Bugzilla::FlagType->new_from_list([$row->{type_id}])->[0]` in `_get_flag_mail_events` fixes it, or carry `type_id` and `name` as plain scalars instead of flattening the object

Bugzilla/BugMail.pm
```diff
@@ -424,7 +530,7 @@ sub sendMail {
     @send_comments = grep { !$_->is_private } @send_comments;
   }
 
-  if (!scalar(@display_diffs) && !scalar(@send_comments)) {
+  if (!scalar(@display_diffs) && !scalar(@send_comments) && !scalar(@$flag_events)) {
```
dklawren, line 543: sendMail can now emit a flag-only mail (no diffs, no comments), but dequeue's "if nothing remains to send" bail near line 784 still tests only `diffs` and `new_comments`, so on the mailer-queue path that mail is silently dropped. that guard needs `|| @{$vars->{flag_events}}` too

Bugzilla/BugMail.pm
```diff
@@ -272,11 +337,29 @@ sub Send {
       # Go through each role the user has and see if they want mail in
       # that role.
       foreach my $relationship (keys %{$recipients{$user_id}}) {
-        if ($user->wants_bug_mail(
-          $bug, $relationship, $start ? \@diffs : [],
-          $comments, $params->{dep_only}, $changer
-        ))
-        {
+        my $wants_mail;
+        if ($relationship == REL_FLAG_REQUESTEE) {
```
dklawren, line 351: flag recipients now run through the generic recipient loop, which skips on `$user->is_bug_ignored($id)`. the old `notify()` only checked `email_enabled`, so a needinfo request aimed at someone who has the bug on their ignore list now disappears with no mail at all. if that is intentional it is worth a note, otherwise the flag rels should bypass the ignore check the way they bypass `wants_bug_mail`

Bugzilla/Constants.pm
```diff
@@ -355,14 +356,21 @@ use constant REL_CC       => 3;
 # REL 4 was REL_VOTER, before it was moved ino an extension.
 use constant REL_GLOBAL_WATCHER => 5;
 
+# Bug 1883428: recipients added because of a flag change, independently of
+# any other role they may hold on the bug.
+use constant REL_FLAG_REQUESTEE => 6;    # A flag was requested of them
```
dklawren, line 361: putting these in RELATIONSHIPS is needed for the X-Bugzilla-Reason header, but it also feeds `Bugzilla::User::create` and `userprefs.cgi`, which both iterate `relationships()`. every new account now gets email_setting rows for rel 6/7/8 across all POS/NEG events, and userprefs deletes them again on the next save because the prefs template renders no checkboxes for them. these rels bypass `wants_bug_mail` entirely so the rows never do anything, worth skipping them in those two loops

template/en/default/email/bugmail-flagtype-cc.txt.tmpl
```diff
@@ -0,0 +1,37 @@
...
+  # throughout and has no path to a raw address. Revisit if this needs to
+  # carry more than a one-line-per-event summary (diffs, comments, HTML). %]
+
+[% PROCESS global/variables.none.tmpl %]
+
+From: [% Param('mailfrom') %]
+To: [% to %]
+Subject: [[% terms.Bug %] [%+ bug.id %]] [% bug.short_desc %]
+Date: [% date %]
+X-Bugzilla-Type: request
```
dklawren, line 23: no `threadingmarker` here, so flag-type cc_list mail stops threading with the rest of the bug's mail. the removed `request/email.txt.tmpl` emitted one

Bugzilla/BugMail.pm
dklawren: We can file as a follow up bug/task.

dklawren (changes requested):

Bugzilla/BugMail.pm
Xzzz: Fixed: guarded with `@{$vars->{flag_events} || []}`

Bugzilla/BugMail.pm
Xzzz: Fixed as suggested: use of `new_from_list(...)` to build flag type

Bugzilla/BugMail.pm
Xzzz: Fixed: added `|| @{$vars->{flag_events}}` to that guard, as in `sendMail`

Bugzilla/BugMail.pm
Xzzz: Fixed - It was not intentional, I then went with your second suggestion: flag relationships (`REL_FLAG_REQUESTEE`/`REQUESTER`/`TYPE_CC`) now bypass the ignore check the same way they already bypass `wants_bug_mail()`, matching old `notify()` behavior.

Bugzilla/Constants.pm
Xzzz: Fixed: skipped `REL_FLAG_REQUESTEE`/`REQUESTER`/`TYPE_CC` in both the `User::create` and `userprefs.cgi` loops over `relationships()`, so no more dead email_setting rows

template/en/default/email/bugmail-flagtype-cc.txt.tmpl
Xzzz: Fixed: added `build_thread_marker($bug->id, 0, 0)` to the vars and `[%+ threadingmarker %]` to the template

Bugzilla/BugMail.pm
Xzzz: New task [2070239](https://bugzilla.mozilla.org/show_bug.cgi?id=2070239) created

Bugzilla/Flag.pm
```diff
@@ -714,7 +706,6 @@ sub force_retarget {
     else {
       # Track deleted attachment flags.
       push(@removed, $class->snapshot([$flag])) if $flag->attach_id;
```
dklawren, line 708: dropping `notify()` here loses the only notification this path ever had.

`FlagType::update` calls `force_retarget($flag_ids)` with no `$bug` (Bugzilla/FlagType.pm:173 and :189), reached from editflagtypes.cgi:372 when an admin edits a flag type's inclusion/exclusion lists. flags that no longer apply fall into this `else` and get deleted with a `status => 'X'` flag_activity row, but `FlagType::update` never touches the bug row, so there is no bugs_activity, no delta_ts/lastdiffed change, and `BugMail::Send` is never called for those bugs.

two consequences:
- the requester is never told their `needinfo?`/`review?` was cleared, which `notify()` used to do here
- the orphan `X` row stays newer than `lastdiffed`, so the next time anyone edits that bug `_get_flag_mail_events` picks it up (the preceding row is still `?`), and "your request for needinfo has been cleared by &lt;admin&gt;" gets bundled into an unrelated bugmail, attributed to the admin, potentially weeks later

the product-move path is fine, `update_flags` passes `$bug` and the bug update writes activity. this call site needs its own trigger

Bugzilla/BugMail.pm
```diff
@@ -199,6 +201,50 @@ sub Send {
     }
   }
 
+  # Bug 1883428: someone a flag was requested of, or who requested a flag
+  # that's now been granted/denied, becomes a recipient even without any
+  # other role on the bug. Flag-type cc_list addresses (admin-configured,
+  # bypasses per-user opt-in) are added the same way notify() used to.
+  my @flag_events
```
dklawren, line 208: the guard you fixed in `sendMail` and in `dequeue` is still missing from the third copy at line 147, which runs before this block:

```perl
return {'sent' => []} unless scalar(@diffs) || scalar(@$comments);
```

flag changes normally write a `flagtypes.name` bugs_activity row, so `@diffs` is non-empty and i can't hand you a repro through the ui for this one on its own. still worth `|| @flag_events` for consistency with the other two, and because this return also skips `_send_flag_type_cc_raw_mail` at line 464. would need `@flag_events` hoisted above the return, `$start`/`$end` are already in scope there

template/en/default/email/bugmail.txt.tmpl
```diff
@@ -25,6 +25,20 @@
 
 [% isnew = bug.lastdiffed ? 0 : 1 %]
 
+[% FOREACH event = flag_events %]
+[% IF event.action == 'requested' %]
+[% IF event.type.name == 'needinfo' %]
```
dklawren, line 30: branching on `event.type.name == 'needinfo'` puts a bmo-only flag type into an upstream core template.

the `flag_event` hook this pr adds is the right place for it, and `extensions/Needinfo/template/en/default/hook/email/bugmail-flag_event.txt.tmpl` already fires on exactly this condition, it just contributes the follow-up sentences rather than the headline.

suggest moving the headline into the needinfo hook and letting the generic `ELSE` wording stand in core. same applies to `bugmail.html.tmpl` and to both bmo overrides, where the special case is duplicated again

dklawren (changes requested):

Bugzilla/Flag.pm
Xzzz: `FlagType::update` now records which bugs are affected by each of its two `force_retarget` calls, and after the transaction commits it calls `BugMail::Send` for each of those bugs, with the admin as changer.
The `X` rows fall in that bug's `lastdiffed` -> now window, so the requester is told their request was cleared, as `notify()` used to do. `lastdiffed` then moves past the row, so it can no longer show up weeks later in an unrelated bugmail.
Bugs whose flag was only retargeted produce no flag event, and `Send` returns without mailing. This relies on the `Send` guard fix from the next thread, since these mails have no diffs or comments.

=> Fixed in "Bug 1883428 - Send bugmail for flags cleared by a flag type edit"

template/en/default/email/bugmail.txt.tmpl
Xzzz: Removed the `needinfo` special case from the core text/HTML templates and from both BMO overrides, so every flag request uses the generic "has requested <flag> from you" line. The "needs more information from you to work on this bug" sentence now opens the Needinfo `flag_event` hook, followed by the existing reporter-aware wording and the user-guide link.

=> Fixed in "Bug 1883428 - Move the needinfo headline into the Needinfo flag_event hook"

PR #2718, opened 2026-08-20
Bug 2065244 - Fix several small issues with My Attention dashboard

dklawren:


Bugzilla/Memcached.pm
```diff
@@ -300,6 +300,13 @@ sub should_rate_limit {
     $memcached->add($key, 0, $rate_seconds + 1);
     my $tokens = $memcached->get_multi(@keys);
     my $cas    = $memcached->gets($key);
+
+    # gets() returns undef when memcached is unreachable (or the key expired
```
cgsheeh: This hunk is unrelated to the My Attention dashboard, and bug 2065244 doesn't mention rate limiting anywhere. Can it move to its own bug so it gets reviewed on its own terms?

Bugzilla/Memcached.pm
```diff
@@ -300,6 +300,13 @@ sub should_rate_limit {
...
+    # between the add and the gets).  Don't let $cas->[1]++ autovivify an
+    # arrayref holding an undef cas id, which would then be passed to cas().
+    # Fail open: no working memcached means no rate limiting.
+    return 0 unless ref $cas eq 'ARRAY' && defined $cas->[0];
```
cgsheeh: `return 0` exits the retry loop entirely, so one transient `gets()` miss disables rate limiting for the whole request. If the key genuinely expired between the `add` and the `gets`, `next` would let the loop do the retry it was written for and still fail open once `$tries` is exhausted. Was `return` deliberate?

extensions/BMO/lib/Reports/Attention.pm
```diff
@@ -374,7 +374,7 @@ sub other_needinfo_bugs {
               AND bugs.bug_severity NOT IN ('S1','S2')
               AND (keywords.keywordid IS NULL OR keywords.keywordid NOT IN (?, ?))
               AND (bug_group_map.group_id IS NULL OR bug_group_map.group_id NOT IN ("
-    . join ',', @{$cache->{sec_group_ids}} . '))
+    . (join ',', @{$cache->{sec_group_ids}}) . '))
```
cgsheeh, line 377: If `sec_group_ids` is empty, this will become `NOT IN ()` which I believe is a syntax error. Maybe we need a `NOT IN (0)` fallback?

extensions/BMO/lib/Reports/Attention.pm
```diff
@@ -405,31 +405,31 @@ sub report {
...
   my $dbh = Bugzilla->dbh;
 
-  # Here we load some values into cache that will be used later
-  # by the various queries. Process cache lasts til server restart.
-  my $process_cache = Bugzilla->process_cache->{attention} = {};
+  # Here we load some values into cache that will be used later by the
+  # various queries. Request cache lasts for this request only, so a
+  # keyword, flag type or security group added mid-session is picked up
+  # on the next page load rather than at the next server restart.
+  my $lookup_cache = $request_cache->{attention};
```
cgsheeh: Nit: we just assigned and created the `$request_cache` on line 403 above, maybe this should be collapsed into that section with a single comment instead of two.

cgsheeh (approved):

PR #2719, opened 2026-08-21
Bug 2061445 - Migrate Bugzilla (system info) REST resource to native Mojo API

Xzzz:
## Summary

Ports `Bugzilla::WebService::Bugzilla`'s `version`/`extensions`/`timezone`/`time`/`jobqueue_status` methods into a native `Bugzilla::API::V1::Bugzilla` Mojo controller, mirroring the pattern already used for Classification/Component/Teams/Reminders/Configuration.

## Changes

- Add `Bugzilla/API/V1/Bugzilla.pm`: `GET /rest/version`, `/rest/extensions`, `/rest/timezone`, `/rest/time` (public), and `/rest/jobqueue_status` (login required), same JSON response shape as the legacy endpoints
- Delete `Bugzilla/WebService/Bugzilla.pm` and `Bugzilla/WebService/Server/REST/Resources/Bugzilla.pm`
- Remove the `Bugzilla` entry from `WS_DISPATCH` in `Bugzilla/WebService/Constants.pm`, drop the corresponding `use` line in `Bugzilla/WebService/Server/REST.pm`, and remove the POD reference in `Bugzilla/WebService.pm`
- Update `qa/t/rest_user_login_logout.t`: drop the assertions that `Bugzilla_login`/`Bugzilla_password` are honored/rejected on `/rest/version`. Native Mojo REST routes only support cookie, `X-Bugzilla-API-Key` and OAuth2 bearer-token auth. Please note that this gap affects every resource already migrated to native Mojo, not just this one. Per discussion on 2057358 we're intentionally dropping username/password auth on native REST endpoints rather than porting it forward (cookie auth stays for web-UI JS, OAuth2 stays for Phabricator)

## Test plan

- Verified against a local docker instance:
  - `GET /rest/version`, `/rest/extensions`, `/rest/timezone`, `/rest/time` return the expected JSON, matching the legacy endpoint output
  - `GET /rest/jobqueue_status` requires login (`login_required` when anonymous) and returns the expected counts when authenticated via API key
  - Confirmed cookie and API-key-based auth continue to work
  - Confirmed `Bugzilla_login`/`Bugzilla_password` are now silently ignored
- Ran the sanity/bmo/webservices/selenium1-4 test suites clean

## References

- Bugzilla link: [2061445](https://bugzilla.mozilla.org/show_bug.cgi?id=2061445)


Bugzilla/API/V1/Bugzilla.pm
```diff
@@ -0,0 +1,143 @@
...
+      tz_short_name => 'UTC',
+    }
+  );
+}
+
+sub jobqueue_status {
+  my ($self) = @_;
+
+  my $user = $self->bugzilla->login;
+  $user->id || return $self->user_error('login_required');
```
dklawren, line 102: usage_mode is still USAGE_MODE_REST here (set by `_prepare_rest_request`), so `user_error` hits neither branch in `Bugzilla::App::Plugin::Error::_render_error` and nothing gets rendered, an anonymous GET returns a 404 page instead of the JSON login_required error. every other native controller (Reminders, Teams, User) sets `Bugzilla->usage_mode(USAGE_MODE_MOJO_REST)` as the first statement, before `login`. same applies to any ThrowUserError raised inside login itself, eg a bad api key

suggest moving line 98 above line 95, and adding an anonymous-request case to `qa/t/rest_bugzilla.t` since the current test only covers the api-key path

Bugzilla/API/V1/Bugzilla.pm
```diff
@@ -0,0 +1,143 @@
...
+        GROUP BY errors
+    };
+
+  my $status;
+  try {
+    $status = $dbh->selectrow_hashref($query);
+  }
+  catch {
+    ERROR($_);
+    return $self->code_error('jobqueue_status_error');
```
dklawren: `return` inside a Try::Tiny catch block only returns from the block, not from `jobqueue_status`, so after `code_error` renders the 500 the sub falls through to the render below and overwrites the body with `{"errors":0,"total":0}` while the status stays 500. the old code died out of the catch via ThrowCodeError so it never reached that point

eg `catch { ERROR($_) }` then `return $self->code_error('jobqueue_status_error') unless $status;`

Bugzilla/API/V1/Bugzilla.pm
```diff
@@ -0,0 +1,143 @@
...
+
+sub time {
+  my ($self) = @_;
+  Bugzilla->usage_mode(USAGE_MODE_MOJO_REST);
+
+  # All Webservices return times in UTC; Use UTC here for backwards compat.
+  my $dbh     = Bugzilla->dbh;
+  my $db_time = $dbh->selectrow_array('SELECT LOCALTIMESTAMP(0)');
+  $db_time = datetime_from($db_time, 'UTC')->iso8601();
+  my $now_utc = DateTime->now()->iso8601();
```
dklawren: response shape changes here: the legacy endpoint ran these through `type('dateTime')`, which is `Bugzilla::WebService::Server::JSONRPC::datetime_format_outbound` and appends a `Z`, so `/rest/time` used to return `2026-08-21T14:00:00Z`. plain `iso8601()` drops the `Z` and clients that parse these as UTC will now read them as local time. suggest `iso8601() . 'Z'` for db_time, web_time and web_time_utc

Bugzilla/API/V1/Bugzilla.pm
```diff
@@ -0,0 +1,143 @@
...
+sub version {
+  my ($self) = @_;
+  Bugzilla->usage_mode(USAGE_MODE_MOJO_REST);
+
+  return $self->render(json => {version => BUGZILLA_VERSION});
+}
+
+sub extensions {
+  my ($self) = @_;
+  Bugzilla->usage_mode(USAGE_MODE_MOJO_REST);
```
dklawren, line 52: `extensions` and `time` were not in the old `LOGIN_EXEMPT` list, so the legacy server called `Bugzilla->login()` (LOGIN_NORMAL) for them, meaning they were gated when the `requirelogin` param is on. neither handler calls `$self->bugzilla->login` now, so both become anonymously readable under requirelogin. Teams.pm calls `$self->bugzilla->login()` for exactly this reason even though it does not use the user

dklawren (changes requested):

Bugzilla/API/V1/Bugzilla.pm
Xzzz: Fixed => moved `Bugzilla->usage_mode(USAGE_MODE_MOJO_REST)` above the `login` call and added the anonymous-request case to `qa/t/rest_bugzilla.t:59`

Bugzilla/API/V1/Bugzilla.pm
Xzzz: Fixed as suggested. Thanks!

Bugzilla/API/V1/Bugzilla.pm
Xzzz: Fixed by restoring 'Z' suffix to `$db_time` (for db_time) and `$now_utc` (for web_time and web_time_utc)

Bugzilla/API/V1/Bugzilla.pm
Xzzz: Fixed: added `$self->bugzilla->login()` call to `extensions` and `time`

Bugzilla/WebService/Constants.pm
```diff
@@ -313,7 +313,6 @@ sub WS_DISPATCH {
   Bugzilla::Hook::process('webservice', {dispatch => \%hook_dispatch});
 
   my $dispatch = {
```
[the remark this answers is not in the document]
Xzzz: Note that `jsonrpc.cgi` was deleted and the JSON-RPC endpoint disallowed in bug 2058158 (already merged), and `Bugzilla::WebService::Server::JSONRPC` is only instantiated from that script => No live consumer left.
This PR also removes the legacy REST resource for `Bugzilla` (`Bugzilla/WebService/Server/REST/Resources/Bugzilla.pm`), so nothing else references the `Bugzilla` key in `WS_DISPATCH` now.

Bugzilla/API/V1/Bugzilla.pm
```diff
@@ -0,0 +1,147 @@
...
+    }
+  );
+}
+
+sub jobqueue_status {
+  my ($self) = @_;
+
+  Bugzilla->usage_mode(USAGE_MODE_MOJO_REST);
+
+  my $user = $self->bugzilla->login;
```
dklawren: **Query-param API-key and login-token auth silently stops working on `/rest/jobqueue_status`.**

`$self->bugzilla->login` (`Bugzilla/App/Plugin/Login.pm`) only reads the `x-bugzilla-api-key` header, the `Authorization` bearer token, and cookie + `Bugzilla_api_token`. Legacy REST ran `fix_credentials` (`Bugzilla/WebService/Util.pm:326`), which promoted `api_key` / `Bugzilla_api_key` (`Bugzilla/Auth/Login/APIKey.pm:54`) and `Bugzilla_token` (`Bugzilla/Auth/Login/Cookie.pm:143`) out of query params.

Concretely: a monitoring script doing `GET /rest/jobqueue_status?api_key=XXX` gets 200 on master and 401 `login_required` after this PR. Same for `?Bugzilla_token=` obtained from `/rest/login`.

The PR description only claims `Bugzilla_login`/`Bugzilla_password` were dropped, and `docs/en/rst/api/core/v1/general.rst:128` still documents the query-param key as supported.

Bugzilla/API/V1/Bugzilla.pm
```diff
@@ -0,0 +1,147 @@
...
+  my ($self) = @_;
+  Bugzilla->usage_mode(USAGE_MODE_MOJO_REST);
+
+  return $self->render(json => {version => BUGZILLA_VERSION});
+}
+
+sub extensions {
+  my ($self) = @_;
+  Bugzilla->usage_mode(USAGE_MODE_MOJO_REST);
+  $self->bugzilla->login();
```
dklawren: **The "restore requirelogin gating" change is a no-op here, and a bad API key now yields an anonymous 200.** (also applies to `time` at line 74)

In `USAGE_MODE_MOJO_REST` the login helper never enforces `LOGIN_REQUIRED`: it sets `$type = LOGIN_REQUIRED` when `requirelogin` is on, but the redirect/error branch is gated to `USAGE_MODE_BROWSER || USAGE_MODE_MOJO`, and it otherwise falls through to `return Bugzilla->user` (anonymous). Legacy `Bugzilla->login()` promoted to `LOGIN_REQUIRED` (`Bugzilla.pm:230`) and `fail_nodata` threw `login_required` for non-browser modes (`Bugzilla/Auth/Login/CGI.pm:72`).

Two concrete regressions from this one call:

1. With `requirelogin=1`, anonymous `GET /rest/extensions` and `GET /rest/time` return 200 with data, where master returned 401.
2. A revoked or mistyped API key hits `check_rate_limit` and then proceeds as anonymous, so master's `api_key_not_valid` 401 becomes a silent 200 — a client with a dead key can't tell auth failed.

If the gating is intended, it needs `$self->bugzilla->login(LOGIN_REQUIRED)` plus an explicit `$user->id || return $self->user_error('login_required')`, as `jobqueue_status` does.

Bugzilla/API/V1/Bugzilla.pm
```diff
@@ -0,0 +1,147 @@
...
+use Try::Tiny;
+
+use Bugzilla::Constants;
+use Bugzilla::Logging;
+use Bugzilla::Util qw(datetime_from);
+
+sub setup_routes {
+  my ($class, $r) = @_;
+
+  $r->get('/version')->to('V1::Bugzilla#version');
```
dklawren: **`/rest.cgi/version` (and `/rest.cgi/{extensions,timezone,time,jobqueue_status}`) now 404.**

These native routes are registered only under the `/rest` bridge (`Bugzilla/App/Controller/API.pm:45`), but `Bugzilla/App/Controller/CGI.pm:46` deliberately keeps `/rest.cgi/*PATH_INFO` -> `rest_cgi`, which resolves the resource through `WS_DISPATCH`. With the `Bugzilla` entry and the resource file deleted, `GET /rest.cgi/version` goes from 200 JSON on master (PATH_INFO becomes `/version`, `Bugzilla/App/Controller/CGI.pm:176`, matching `qr{^/version$}`) to `rest_invalid_resource` 404.

Either add the alias routes, or make dropping the `/rest.cgi/` form an explicit, documented decision.

Bugzilla/API/V1/Bugzilla.pm
```diff
@@ -0,0 +1,147 @@
...
+}
+
+sub extensions {
+  my ($self) = @_;
+  Bugzilla->usage_mode(USAGE_MODE_MOJO_REST);
+  $self->bugzilla->login();
+
+  my %extensions;
+  foreach my $extension (@{Bugzilla->extensions}) {
+    $extensions{$extension->NAME} = {version => $extension->VERSION || 0};
```
dklawren: **`/rest/extensions` changes the JSON type of `version` from string to number for versionless extensions.**

Legacy forced `$self->type('string', $version)`, so `0` serialized as `"0"`. `$extension->VERSION || 0` now yields an integer and Mojo::JSON emits a bare `0`. Four shipped extensions have no `$VERSION` (`extensions/Gravatar`, `JiraWebhookSync`, `RestrictComments`, `Webhooks`), so the response really does change from `"Gravatar":{"version":"0"}` to `"version":0` — contradicting the documented "version as a string".

```perl
$extensions{$extension->NAME} = {version => '' . ($extension->VERSION || 0)};
```

Bugzilla/API/V1/Bugzilla.pm
Xzzz: Fixed: extensions and time now call $self->bugzilla->login(LOGIN_REQUIRED) and return login_required if $user->id is false (same pattern as jobqueue_status).
In addition, I updated qa/t/rest_bugzilla.t to send the API key on both endpoints and assert anonymous access now gets 401.

=> Committed as "Bug 2061445 - Enforce login on extensions and time"

Bugzilla/API/V1/Bugzilla.pm
Xzzz: Fixed as suggested: `version => '' . ($extension->VERSION || 0)` now restores the string type for versionless extensions.

=> Committed as "Bug 2061445 - Keep extension version as a string in REST response"

dklawren (approved):
LGTM r=dkl

PR #2720, opened 2026-08-21
Bug 2065387 - Ensure more triage owners have access to security bugs

dklawren:


Bugzilla/Search.pm
```diff
@@ -1458,6 +1458,20 @@ sub _standard_joins {
       extra => ['security_cc.who = ' . $user->id],
     };
     push @joins, $security_cc_join;
+
+    # Triage owners can see all bugs in their component, but only if they are
+    # also a member of the mozilla-employee-confidential group.
+    if ($user->in_group('mozilla-employee-confidential')) {
```
cgsheeh: This grants the triage owner visibility into *every* group-restricted bug in their component, not just security ones (legal, HR, partner-restricted, etc.). Should we restrict this to just sec-related bugs?

Bugzilla/User.pm
```diff
@@ -1567,13 +1569,20 @@ sub visible_bugs {
 
     $sth->execute(@check_ids);
     my $use_qa_contact = Bugzilla->params->{'useqacontact'};
+
+    # Triage owners can see all bugs in their component, but only if they are
+    # also a member of the mozilla-employee-confidential group.
+    my $use_triage_owner = $self->in_group('mozilla-employee-confidential');
```
cgsheeh: Nit: maybe we should have a method on user like `$self->is_employee_confidential` instead of hard-coding the `mozilla-employee-confidential` group in a few places.

Bugzilla/User.pm
```diff
@@ -1548,12 +1548,14 @@ sub visible_bugs {
...
       "SELECT DISTINCT bugs.bug_id, reporter, assigned_to, qa_contact,
-                    reporter_accessible, cclist_accessible, cc.who,
-                    bug_group_map.bug_id
+                    components.triage_owner_id, reporter_accessible,
+                    cclist_accessible, cc.who, bug_group_map.bug_id
                FROM bugs
                     LEFT JOIN cc
                               ON cc.bug_id = bugs.bug_id
                                  AND cc.who = $user_id
+                    LEFT JOIN components
```
cgsheeh: `visible_bugs` is one of our hottest queries and this adds the `components` join for every caller, including logged-out users who can never reach the triage-owner branch. Since the statement handle is cached per user object anyway, could we hoist the `in_group` check above the `prepare` and only add the join when it is true?

t/bmo/triage-owner-security-visibility.t
```diff
@@ -0,0 +1,276 @@
...
+Bugzilla->set_user($admin);
+
+my ($product) = grep { @{$_->versions} } Bugzilla::Product->get_all;
+plan skip_all => 'Need a product with at least one version' unless $product;
+
+###############################################################################
+# Helpers
+###############################################################################
+
+sub add_to_group {
```
cgsheeh, line 79: Any reason not to use `$user->set_groups({add => [$group]})` plus `$user->update` here? That drops the hand-rolled SQL and the manual memcached invalidation.

t/bmo/triage-owner-security-visibility.t
```diff
@@ -0,0 +1,276 @@
...
+  can_see($owner_member, $bug->id),
+  'Re-granting mozilla-employee-confidential restores the triage owner bug access'
+);
+ok(
+  search_finds($owner_member, $bug->id),
+  'Re-granting mozilla-employee-confidential restores the bug in triage owner search results'
+);
+
+###############################################################################
+# Cleanup
```
cgsheeh: If any of the fixture setup or `Bugzilla::Bug->create` throws we never reach this, leaving the temp group, components and `group_control_map` rows behind. Moving the cleanup into an `END` block would make it unconditional.

cgsheeh (approved):
This PR seems to revert Bug 1790624, has the original reasoning around the restriction changed?

dklawren:
> This PR seems to revert Bug 1790624, has the original reasoning around the restriction changed?

Yes. There was discussion on Slack and dveditz explained why it was removed. So the agreement was that we bring it back but add the extra requirement of being mozilla-employee-confidential. This way when people leave and lose that group they also lose triage owner access.

Bugzilla/Search.pm
dklawren: Will address as a follow up bug.

PR #2722, opened 2026-08-22
Bug 1202436 - Document authentication method precedence

loganrosen:
Bug 1202436 tracks missing REST API documentation for requests that provide more than one authentication method. BMO evaluates legacy username/password credentials before API keys, so a valid API key can be ignored. When the account requires API-key authentication, the mixed request fails with an `API key authentication is required` error.

This documents the precedence, lack of fallback, and interaction with the `Require API key authentication for API requests` preference. It advises clients to send only the `X-BUGZILLA-API-KEY` header rather than disabling the preference. Authentication behavior is unchanged.

Bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1202436

Testing:

- Built the Sphinx HTML documentation with the pinned requirements. The build succeeded with six pre-existing warnings.

### DIFF
--- a/docs/en/rst/api/core/v1/general.rst
+++ b/docs/en/rst/api/core/v1/general.rst
@@ -110,6 +110,19 @@ any request, and you will be authenticated as that user if the key is correct an
 You can set up an API key by using the :ref:`API Keys tab <api-keys>` in the
 Preferences pages.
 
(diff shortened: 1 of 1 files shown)


dklawren (approved):
LGTM r=dkl

PR #2723, opened 2026-08-23
Bug 2059948 - Do not offer recovery codes to Duo users

loganrosen:
Commit 9e23f770d landed the production fix that prevents Duo users from generating BMO recovery codes. This follow-up fills the remaining gaps:

- Update the two-factor authentication guide, which still said Duo users were offered the recovery-code generator.
- Add Selenium coverage confirming the recovery-code control is absent for Duo users.
- Verify a forged recovery request is rejected even if it spoofs TOTP as the submitted provider.

Test:
- `qa/t/2_test_login_duo.t` (35 tests)

https://bugzilla.mozilla.org/show_bug.cgi?id=2059948

### DIFF
--- a/qa/t/2_test_login_duo.t
+++ b/qa/t/2_test_login_duo.t
@@ -32,7 +32,8 @@ $sel->title_is('User Preferences');
 $sel->click_ok('mfa-select-duo');
 $sel->type_ok('mfa-duo-user', $config->{admin_user_login});
 $sel->type_ok('mfa-password', $config->{admin_user_passwd});
-$sel->click_ok('update');
+$sel->driver->find_element('//form[@name="userprefsform"]')->submit;
+$sel->wait_for_page_to_load(WAIT_TIME);
 $sel->click_ok('//a[contains(text(),"Redirect Back")]',
   'Click Duo Security verification');
 $sel->title_is('User Preferences');
@@ -40,10 +41,54 @@ $sel->is_text_present_ok(
   'The changes to your two-factor authentication have been saved',
   'Duo successfully enabled');
 
+ok(
+  !$sel->is_element_present('mfa-recovery'),
+  'Recovery code generation is not offered for Duo'
+);
+
+# A forged recovery request must fail before opening the Duo prompt.
+$sel->driver->execute_script(
+  q{document.getElementById('mfa-auth-container').style.display = 'block';}
+);
+$sel->type_ok('mfa-password', $config->{admin_user_passwd});
+$sel->driver->execute_script(q{
+  const source = document.forms.userprefsform;
+  const form = document.createElement('form');
+  form.method = 'post';
+  form.action = source.action;
+
+  [
+    ['tab', 'mfa'],
+    ['token', source.elements.token.value],
+    ['dosave', '1'],
+    ['mfa_action', 'recovery'],
+    ['mfa', 'TOTP'],
+    ['password', source.elements.password.value],
+  ].forEach(([name, value]) => {
+    const input = document.createElement('input');
+    input.type = 'hidden';
+    input.name = name;
+    input.value = value;
+    form.appendChild(input);
+  });
+
+  document.body.appendChild(form);
+  form.submit();
+});
+sleep(2);
+$sel->wait_for_page_to_load(WAIT_TIME);
+$sel->title_is('Duo Security Error');
+$sel->is_text_present_ok(
+  'Recovery codes are not available when using Duo Security',
+  'Forged Duo recovery request rejected'
+);
+
 # Disable Duo for the admin user
+$sel->open_ok('/userprefs.cgi?tab=mfa');
 $sel->click_ok('mfa-disable');
 $sel->type_ok('mfa-password', $config->{admin_user_passwd});
-$sel->click_ok('update');
+$sel->driver->find_element('//form[@name="userprefsform"]')->submit;
+$sel->wait_for_page_to_load(WAIT_TIME);
 $sel->click_ok('//a[contains(text(),"Redirect Back")]',
   'Click Duo Security verification');
 $sel->title_is('User Preferences');
--- a/qa/t/2_test_login_duo.t
+++ b/qa/t/2_test_login_duo.t
@@ -32,8 +32,7 @@ $sel->title_is('User Preferences');
 $sel->click_ok('mfa-select-duo');
 $sel->type_ok('mfa-duo-user', $config->{admin_user_login});
 $sel->type_ok('mfa-password', $config->{admin_user_passwd});
-$sel->driver->find_element('//form[@name="userprefsform"]')->submit;
-$sel->wait_for_page_to_load(WAIT_TIME);
+$sel->click_ok('update');
 $sel->click_ok('//a[contains(text(),"Redirect Back")]',
   'Click Duo Security verification');
 $sel->title_is('User Preferences');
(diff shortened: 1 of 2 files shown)


qa/t/2_test_login_duo.t
```diff
@@ -32,18 +32,63 @@ $sel->title_is('User Preferences');
 $sel->click_ok('mfa-select-duo');
 $sel->type_ok('mfa-duo-user', $config->{admin_user_login});
 $sel->type_ok('mfa-password', $config->{admin_user_passwd});
-$sel->click_ok('update');
+$sel->driver->find_element('//form[@name="userprefsform"]')->submit;
```
dklawren: this swap from `click_ok('update')` to a raw `form.submit()` is unrelated to the recovery-code fix, and it happens again on line 90 for the disable flow

submitting the form directly skips the real Submit Changes button and any html5 validation on it, so the duo enable/disable paths lose the coverage they had. nothing in this PR touches the enroll branch of mfa.html.tmpl, so `click_ok('update')` should still work

could you revert both, or add a comment explaining what forced the change

dklawren:
https://github.com/mozilla/bmo/commit/9e23f770dd35b3023b483e7c57bfd1969b8b1112 as a Duo security fix that also had changes inside that remove recovery codes for Duo. Please take a look and if this PR is a duplicate then please close, if not see if anything was missed and update this PR to fill in the gaps. Thanks!

loganrosen:
Thanks! I compared this against 9e23f770d and rebased the PR onto current master. The production changes were duplicates, so I removed them and narrowed this PR to the two gaps left behind: correcting the stale user guide and adding Selenium coverage for the hidden Duo recovery control and a forged recovery request that spoofs TOTP. The focused Duo test passes all 35 tests.

qa/t/2_test_login_duo.t
dklawren: **Navigation race (medium).** `Bugzilla::Test::Selenium::wait_for_page_to_load` ([Selenium.pm:126](../blob/master/Bugzilla/Test/Selenium.pm#L126)) only calls `set_timeout('page load', ...)` — it never blocks. And under W3C, `Selenium::Remote::WebElement::submit` is `execute_script("return arguments[0].submit()")`, which returns as soon as navigation *starts*. The `click_ok('update')` this replaces routed to WebDriver Element Click, which waits for navigation to complete.

So line 37's `click_ok('//a[contains(text(),"Redirect Back")]')` now runs against the pre-submit DOM with only the 600ms implicit wait. On a slow Duo round trip that fails with element-not-found or a stale-element error. (Note you did add `sleep(2)` after the forged `form.submit()` on line 78, but not here.)

**Coverage loss (low).** `js/account.js` disables `#update` on load whenever `#mfa-action` exists, and only the `#mfa-select-duo` / `#mfa-disable` handlers re-enable it. Submitting the form directly skips that enable path and skips HTML5 constraint validation on `#mfa-password` / `#mfa-duo-user` — the test now passes even if either breaks.

**Nit.** `Bugzilla::Test::Selenium` already has `submit($locator)` ([Selenium.pm:436](../blob/master/Bugzilla/Test/Selenium.pm#L436)), which does the `_fix_locator` + `find_element(...)->submit()` dance. `$sel->submit('//form[@name="userprefsform"]')` would do here and on line 90.

qa/t/2_test_login_duo.t
```diff
@@ -32,18 +32,63 @@ $sel->title_is('User Preferences');
...
+$sel->wait_for_page_to_load(WAIT_TIME);
 $sel->click_ok('//a[contains(text(),"Redirect Back")]',
   'Click Duo Security verification');
 $sel->title_is('User Preferences');
 $sel->is_text_present_ok(
   'The changes to your two-factor authentication have been saved',
   'Duo successfully enabled');
 
+ok(
+  !$sel->is_element_present('mfa-recovery'),
```
dklawren, line 48: **No positive control (low).** `!is_element_present('mfa-recovery')` is satisfied by *any* page that lacks the button. If the navigation race above leaves the browser on the pre-submit page, an error page, or the Duo redirect page, this passes vacuously and the regression it guards goes unnoticed.

Asserting a control element that must be present on the Duo prefs tab — e.g. `ok($sel->is_element_present('mfa-disable'), ...)` — alongside it makes the absence meaningful.

(For what it's worth, the locator itself is fine: `_fix_locator` ([Selenium.pm:480](../blob/master/Bugzilla/Test/Selenium.pm#L480)) makes this an exact `@id`/`@name` match, so `mfa-recovery-container` / `mfa-recovery-frame` don't false-positive.)

qa/t/2_test_login_duo.t
```diff
@@ -32,18 +32,63 @@ $sel->title_is('User Preferences');
...
+    input.type = 'hidden';
+    input.name = name;
+    input.value = value;
+    form.appendChild(input);
+  });
+
+  document.body.appendChild(form);
+  form.submit();
+});
+sleep(2);
```
dklawren: **Fixed sleep is the only real wait (low).** Line 79's `wait_for_page_to_load` does not block (see the comment on line 35), and `title_is` on line 80 is a plain `get_title` with no retry. If the forged POST takes longer than 2s, the test reads the stale `User Preferences` title and fails spuriously in CI.

A short poll on the title or body text would be deterministic instead of timing-dependent.

qa/t/2_test_login_duo.t
```diff
@@ -32,18 +32,63 @@ $sel->title_is('User Preferences');
...
+  'Recovery codes are not available when using Duo Security',
+  'Forged Duo recovery request rejected'
+);
+
 # Disable Duo for the admin user
+$sel->open_ok('/userprefs.cgi?tab=mfa');
 $sel->click_ok('mfa-disable');
 $sel->type_ok('mfa-password', $config->{admin_user_passwd});
-$sel->click_ok('update');
+$sel->driver->find_element('//form[@name="userprefsform"]')->submit;
```
dklawren: Same navigation race as line 35: `->submit` returns as soon as navigation starts, line 91's `wait_for_page_to_load` is a no-op, and line 92 immediately clicks `//a[contains(text(),"Redirect Back")]`. Whatever fix you apply above applies here too.

dklawren (commented):
Automated code review. The docs change and the new Duo-recovery assertions look right — the guard they target is real and reachable (`userprefs.cgi:744-756` recomputes `$mfa` from `$user->mfa` first, so the forged `mfa=TOTP` is ignored and `duo_user_error` is thrown), and `mfa-recovery` is genuinely gone for Duo in `mfa.html.tmpl:90-97`. All findings below are about test reliability in `qa/t/2_test_login_duo.t`.

dklawren (changes requested):

qa/t/2_test_login_duo.t
loganrosen: Fixed in c5eec0760. The test now confirms `mfa-disable` is present on the Duo preferences page before asserting that `mfa-recovery` is absent.

qa/t/2_test_login_duo.t
loganrosen: Fixed in c5eec0760. I restored `click_ok('update')` for Duo enable so the test exercises the submit button and HTML5 validation, and removed the ineffective `wait_for_page_to_load` call. I did not switch to `$sel->submit(...)` because that would preserve the validation bypass.

qa/t/2_test_login_duo.t
loganrosen: Fixed in c5eec0760. The disable flow now uses `click_ok('update')` and no longer calls the ineffective page-load wait, matching the enable flow.

qa/t/2_test_login_duo.t
loganrosen: Fixed in c5eec0760. The fixed two-second sleep and ineffective page-load wait are replaced with bounded polling for the `Duo Security Error` title before the existing assertions.

dklawren (approved):
LGTM r=dkl

PR #2724, opened 2026-08-23
Bug 2065864 - Match webhook event selections exactly

loganrosen:
Bug 1979028 added the `attachment_change` webhook event, but `should_send` checked the comma-separated event setting with substring regular expressions. Because `attachment_change` contains both `change` and `attachment`, a webhook configured only for attachment modifications also received bug modification and new attachment events.

This parses the stored setting into exact event tokens before applying the existing routing-key checks. It also uses the exact `change` token for the moved-product/component fallback, where the same substring collision occurred.

The new connector test covers each event independently, verifies that `attachment_change` does not leak into the two shorter event names, and exercises combined selections and the moved-product path.

Bug: https://bugzilla.mozilla.org/show_bug.cgi?id=2065864

**Testing:** `test_sanity t/001compile.t t/002goodperl.t t/005whitespace.t extensions/Push/t/webhook.t`

### DIFF
--- a/extensions/Push/lib/Connector/Webhook.pm
+++ b/extensions/Push/lib/Connector/Webhook.pm
@@ -62,7 +62,7 @@ sub should_send {
   return 0 unless Bugzilla->params->{webhooks_enabled};
 
   my $webhook = Bugzilla::Extension::Webhooks::Webhook->new($self->{webhook_id});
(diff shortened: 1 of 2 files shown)


dklawren (approved):
Thanks r=dkl

PR #2726, opened 2026-08-25
Bug 2060932 - Support GitHub-style collapsible sections in comments

suhaibmujahid:
Comments render markdown with cmark's safe option and every `<` escaped beforehand, so raw HTML never reaches the parser. Instead of weakening that, convert the four disclosure tags back to real elements after rendering: mark the escaped tags in text nodes (skipping pre/code so the syntax can still be documented), then re-parse so the block level elements are lifted out of the paragraph markdown wrapped them in.

Only those four exact tags are recognized and they never carry attributes, so no other markup can be smuggled in. The marker characters are stripped from the input so they cannot be forged, and unbalanced tags cannot leak an unclosed element into the page.

### DIFF
--- a/Bugzilla/Markdown.pm
+++ b/Bugzilla/Markdown.pm
@@ -36,6 +36,30 @@ sub _build_markdown_parser {
 }
 
 my $MARKDOWN_OFF = quotemeta '#[markdown(off)]';
+
+# The only raw HTML allowed in comments: GitHub-style collapsible sections.
+# Markdown rendering escapes all tags, so the escaped text is swapped back to
+# real elements afterwards. Only these exact tags are recognized and they never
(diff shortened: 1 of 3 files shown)


dklawren:
@kyoshino Is this something that you could take over and finish for us? Given your past experience with markdown comments and rendering.

kyoshino:
Sure, will check it out!

kyoshino:
Tested quickly, locally. If the `open` attribute is attached, it doesn’t work. Nested `<details>` are fine.

https://gist.github.com/scmx/eca72d44afee0113ceb0349dd54a84a2

image

kyoshino:
Also a header in title needs `display: inline-block`:

```html
<details><summary>

# Summary (h1)
</summary>
content
</details>
```

image

PR #2733, opened 2026-09-01
Bug 2066252 - Require a target milestone when adding checkin-needed-tb

tobypilling:
- Prevent adding `checkin-needed-tb` without a target milestone for Calendar, MailNews Core, and Thunderbird bugs.
- Allow the keyword and milestone to be set in the same update.
- Do not block unrelated updates to bugs that already have the keyword.
- Add tests covering the validation behavior.

## Testing

`docker compose -f docker-compose.test.yml run --rm --build bmo.test test_bmo -q -f t/bmo/checkin-needed-tb-milestone.t`

All 8 tests passed.

### DIFF
--- a/extensions/MozChangeField/Extension.pm
+++ b/extensions/MozChangeField/Extension.pm
@@ -17,6 +17,7 @@ use Bugzilla::Constants;
 use Bugzilla::Logging;
 
 use Bugzilla::Extension::MozChangeField::Pre::CanConfirm;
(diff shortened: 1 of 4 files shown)


dklawren (commented):
I would rather not utilize a new hook and just use the already present evaluate_create() and evaluate_change() methods. I have submitted pr #2734 with the changes which I had already worked on. So you can see what I mean there. Much simpler.

PR #2734, opened 2026-09-01
Bug 2066252 - Prevent save if keyword checkin-needed-tb exists without target milestone

dklawren:


extensions/MozChangeField/lib/Post/CheckinNeededTbMilestone.pm
```diff
@@ -0,0 +1,52 @@
...
+
+sub _check_milestone {
+  my ($bug) = @_;
+  my $product = $bug->product_obj;
+
+  return if !PRODUCTS->{$product->name};
+  return if !$bug->has_keyword(KEYWORD);
+
+  # The product's default milestone (normally '---') means 'not set'.
+  return if $bug->target_milestone ne $product->default_milestone;
```
cgsheeh, line 28: Should we compare with `---` directly? Is there any risk in a silent pass if a product gets a real default milestone?

extensions/MozChangeField/lib/Post/CheckinNeededTbMilestone.pm
```diff
@@ -0,0 +1,52 @@
...
+  my ($self, $args) = @_;
+  _check_milestone($args->{bug});
+}
+
+sub evaluate_change {
+  my ($self, $args) = @_;
+  my $changes = $args->{changes};
+
+  # Only enforce on a save that touches one of the two fields, so bugs
+  # already in this state don't block unrelated edits.
```
cgsheeh: Should this be "one of the three fields", ie keywords/target_milestone/product?

cgsheeh (approved):

extensions/MozChangeField/lib/Post/CheckinNeededTbMilestone.pm
dklawren: I am not sure what silent pass means if it gets a real milestone. The products default milestone can be set to any valid milestone value and is what gets set if a user doesn't set anything for milestone.

PR #2736, opened 2026-09-04
Bug 2069143 - When retrieving comments for a bug(s) via REST API, remove comments that would be collapsed in the web UI unless explicitly asking for them

dklawren:


Xzzz (approved):
lgtm

PR #2740, opened 2026-09-08
Bug 2070247 - Support limiting who can needinfo the Hackbot account

dklawren:
What changed

extensions/Needinfo/Extension.pm
- _user_needinfo_blocked is now tri-state: on blocks everyone, editbugs blocks only requesters without editbugs, off blocks nobody. Self-needinfo is always allowed under editbugs.
- add_setting gains the editbugs option plus force_check => 1 - the setting already exists, so Bugzilla/User/Setting.pm:140 would otherwise skip the update. remove_setting only drops user choices outside the new option list, so existing on/off preferences survive checksetup.      
- _check_requestee throws a distinct needinfo_blocked_editbugs when that's the actual reason.
- user_preferences reads the posted value directly, mapping a stale 1 to on.

Prefs template - checkbox → three-option select, inside the existing shared Request blocking: row.

Error template - new needinfo_blocked_editbugs branch.

Both of your requirements fall out of the single existing needinfo_blocked choke point rather than needing separate work: the autocomplete greys the account as "Requests blocked" and alerts if you pick it (js/field.js:508-530, fed by extensions/Review/Extension.pm:951), and typing the address manually errors on save via object_end_of_create/object_end_of_update - which sit under every write path, web form and REST alike. The needinfo_role dropdown also drops blocked reporter/assignee/QA/mentor options for free.

Still needs doing

- checksetup.pl must run for the new option to appear in setting_value on prod, the pref page will render the select before then, but saving editbugs would fail validation.


extensions/Needinfo/Extension.pm
```diff
@@ -25,8 +25,18 @@ BEGIN {
...
+  my ($requestee) = @_;
+  my $setting = $requestee->settings->{block_needinfo} or return 0;
+
+  return 1 if $setting->{value} eq 'on';
+  return 0 unless $setting->{value} eq 'editbugs';
+
+  # Never block someone from needinfo'ing themselves.
+  return 0 if Bugzilla->user->id == $requestee->id;
+
+  # ponytail: global editbugs membership, not per-product -- needinfo_blocked
```
cgsheeh: `ponytail:` looks like a stray marker from your editor/agent session -- it's the only occurrence in the whole repo. Can you turn this into a plain `TODO:` (or just drop it)?

extensions/Needinfo/Extension.pm
```diff
@@ -63,10 +73,15 @@ sub install_update_db {
 sub install_before_final_checks {
   my ($self, $args) = @_;
   add_setting({
-    name     => 'block_needinfo',
-    options  => ['on', 'off'],
-    default  => 'off',
-    category => 'Reviews and Needinfo'
+    name        => 'block_needinfo',
+    options     => ['on', 'editbugs', 'off'],
```
cgsheeh: `block_needinfo` also renders on the generic Settings tab (`userprefs.cgi?tab=settings`), which builds its `<select>` from `legal_values` and labels each option with `setting_descs.$value`. `on`/`off` resolve to "On"/"Off", but `editbugs` has no entry so it'll show the raw string. Please add `setting_descs.editbugs` to `hook/global/setting-descs-settings.none.tmpl`, and while you're there `setting_descs.block_needinfo = "Block needinfo requests"` is no longer accurate for a tri-state.

extensions/Needinfo/Extension.pm
```diff
@@ -63,10 +73,15 @@ sub install_update_db {
...
+    default     => 'off',
+    category    => 'Reviews and Needinfo',
+
+    # The setting predates the 'editbugs' option, so force add_setting to
+    # rebuild it. remove_setting only drops user choices outside the new
+    # option list, so existing 'on'/'off' preferences survive.
+    force_check => 1,
```
cgsheeh: This is the only `force_check` in our codebase, and it's unconditional, so every `checksetup.pl` run from now on will drop and re-insert the `setting`/`setting_value` rows. It's idempotent so I'm not blocking on it, but I'd prefer the one-time fixup live in `install_update_db` so we can drop it later -- or at minimum a note here that it can be removed once prod has run checksetup.

extensions/Needinfo/Extension.pm
```diff
@@ -241,9 +256,13 @@ sub _check_requestee {
     = ref($requestee)
     ? $requestee
     : Bugzilla::User->new({name => $requestee, cache => 1});
-  if ($user->needinfo_blocked) {
-    ThrowUserError('needinfo_blocked', {requestee => $user});
-  }
+  return unless $user->needinfo_blocked;
+
+  my $blocked_editbugs
+    = $user->settings->{block_needinfo}->{value} eq 'editbugs';
```
cgsheeh: This re-reads the setting internals to recover a decision `needinfo_blocked` already made, so the two can drift (e.g. the self-needinfo exemption isn't reflected here). Could `needinfo_blocked` return the reason instead -- e.g. `''`/`'on'`/`'editbugs'` -- so this just maps reason to error tag?

extensions/Needinfo/Extension.pm
```diff
@@ -291,9 +310,15 @@ sub user_preferences {
...
-  $settings->{block_needinfo}->validate_value($value);
-  $settings->{block_needinfo}->set($value);
+  my $value = $input->{block_needinfo} // 'off';
+  $value = 'on' if $value eq '1';    # stale page still posting the old checkbox
+
+  my $setting = $settings->{block_needinfo};
+  my %legal_values = map { $_ => 1 } @{$setting->legal_values};
+  ThrowCodeError('setting_value_invalid',
+    {name => 'block_needinfo', value => $value})
+    unless $legal_values{$value};
```
cgsheeh: Looks like the existing `validate_value` call does this exact same thing? Any reason we couldn't continue with `validate_value` instead?

extensions/Needinfo/template/en/default/hook/account/prefs/account-field.html.tmpl
```diff
@@ -16,10 +16,19 @@
...
-      Block needinfo requests (please use only if you'll be unavailable for an extended period of time)
-    </label>
+    [% block_needinfo = user.settings.block_needinfo.value %]
+    <label for="block_needinfo">Needinfo requests:</label>
+    <select id="block_needinfo" name="block_needinfo">
+      <option value="off"[% " selected" IF block_needinfo == "off" %]>
+        Accept from anyone
+      </option>
+      <option value="editbugs"[% " selected" IF block_needinfo == "editbugs" %]>
+        Only accept from users with editbugs
```
cgsheeh: "editbugs" is internal jargon for most users. Something like "Only accept from users who can edit bugs (editbugs)" would read better on the prefs page.

cgsheeh (approved):

PR #2742, opened 2026-09-11
Bug 1877201 - Fix for a wrong comment_id returned after POST to bug/{bug_id}/comment

ksy36:
I came across an issue where after posting comments to `bug/{bug_id}/comment`, the endpoint returns wrong comment id.

For example, for https://bugzilla.mozilla.org/show_bug.cgi?id=2068734, it returned 4120899, 4120900, 4120901 for the last 3 comments, while the actual ids are 18336000, 18336001 and 18336002.

It's currently passing `comment_id` to `bz_last_key `to get the id of just inserted comment

```
my $new_comment_id = $dbh->bz_last_key('longdescs', 'comment_id');
```
https://github.com/mozilla/bmo/blob/master/Bugzilla/WebService/Bug.pm#L1332-L1342

but on MySQL these parameters appears to be unused, it just selects last inserted id, which in this case is not a comment id, but auto incremented id from another table

```
sub bz_last_key {
  my ($self) = @_;

  my ($last_insert_id) = $self->selectrow_array('SELECT LAST_INSERT_ID()');

  return $last_insert_id;
}
```
https://github.com/mozilla/bmo/blob/master/Bugzilla/DB/Mysql.pm#L95-L101

I couldn't reproduce this locally though, so it might be something that's happening only on production.


dklawren (approved):
LGTM r=dkl

hamzahamidi (commented):
No actionable issue found.

PR #2743, opened 2026-09-14
Bug 2065171 - Migrate BugUserLastVisit REST resource to native Mojo API

Xzzz:
## Summary

Ports `Bugzilla::WebService::BugUserLastVisit`'s `get`/`update` methods into a native `Bugzilla::API::V1::BugUserLastVisit` Mojo controller, mirroring the pattern already used for Classification/Component/Teams/Reminders/Configuration/Bugzilla (system info).

## Changes

- Add `Bugzilla/API/V1/BugUserLastVisit.pm`: `GET`/`POST /rest/bug_user_last_visit` and `/rest/bug_user_last_visit/<id>` (login required), same JSON response shape as the legacy endpoints
- Delete `Bugzilla/WebService/BugUserLastVisit.pm` and `Bugzilla/WebService/Server/REST/Resources/BugUserLastVisit.pm`
- Remove the `BugUserLastVisit` entry from `WS_DISPATCH` in `Bugzilla/WebService/Constants.pm` and drop the corresponding `use` line in `Bugzilla/WebService/Server/REST.pm`

**Breaking change:** removing the `WS_DISPATCH` entry also removes `BugUserLastVisit.get`/`update` from JSON-RPC and XML-RPC, not just the legacy REST dispatcher, since all three share that table. Native Mojo routes only serve REST. This matches the same tradeoff already made in the Classification and Bugzilla (system-info) migrations earlier in this series.

## Test plan

- `GET /rest/bug_user_last_visit` (anonymous => `login_required`, authenticated => list of last-visited bugs)
- `GET /rest/bug_user_last_visit/<id>`
- `GET /rest/bug_user_last_visit?ids=<id>&ids=<id>`
- `POST /rest/bug_user_last_visit/<id>`
- `POST /rest/bug_user_last_visit` with `{"ids":[...]}` body
- `OPTIONS` on both routes returns `Allow: GET, POST`
- Confirm response shape (bare JSON array, `last_visit_ts` with trailing `Z`) matches the legacy endpoint

## References

- Bugzilla link: [2065171](https://bugzilla.mozilla.org/show_bug.cgi?id=2065171)


Bugzilla/API/V1/BugUserLastVisit.pm
```diff
@@ -0,0 +1,174 @@
...
+}
+
+sub _ids_from_request {
+  my ($self) = @_;
+
+  if (my $id = $self->param('id')) {
+    return [$id];
+  }
+
+  if ($self->req->method eq 'POST') {
```
dklawren: **`ids` on POST is now only accepted from a JSON body.**

The legacy layer took it from the query string too — `Bugzilla::WebService::Server::REST::_retrieve_json_params` deliberately merged query-string params for non-GET requests ("Allow parameters in the query string if request was non-GET"), and CGI parsed urlencoded bodies. So both of these used to work and now fail:

- `POST /rest/bug_user_last_visit?ids=123` → `param_required` (empty body decodes to `{}`)
- `POST /rest/bug_user_last_visit` with `Content-Type: application/x-www-form-urlencoded` and body `ids=123&ids=456` → `decode_json` throws → `rest_malformed_json`

Suggest falling back to `$self->every_param('ids')` when the body is absent or isn't JSON.

Bugzilla/API/V1/BugUserLastVisit.pm
```diff
@@ -0,0 +1,174 @@
...
+  return $params;
+}
+
+sub _bug_user_last_visit_to_hash {
+  my ($self, $bug_id, $last_visit_ts, $params) = @_;
+
+  return filter(
+    $params,
+    {
+      id            => 0 + $bug_id,
```
dklawren, line 165: **`0 + $bug_id` numifies the raw request value rather than the resolved bug.**

The new `/:id` placeholder accepts non-numeric ids, whereas the legacy resource regex was `^/bug_user_last_visit/(\d+)$` (404 `rest_invalid_resource`). So `POST /rest/bug_user_last_visit/my-bug-alias` now resolves the alias via `Bugzilla::Bug->check`, records the visit, and responds `{"id":0,"last_visit_ts":...}` with an "Argument isn't numeric" warning in the log.

Use `$bug->id` here (and pass `$bug` rather than `$bug_id` from `update`).

Bugzilla/API/V1/BugUserLastVisit.pm
```diff
@@ -0,0 +1,174 @@
...
+    return [$id];
+  }
+
+  if ($self->req->method eq 'POST') {
+    my $params;
+    my $error;
+    try { $params = decode_json($self->req->body || '{}'); }
+    catch { $error = 'rest_malformed_json'; };
+    return (undef, $error) if $error;
+    my $ids = $params->{ids} // [];
```
dklawren: **Input type validation was dropped along with `validate(@_, 'ids')`.**

`ref $ids ? $ids : [$ids]` passes *any* reference straight through, so `{"ids":{"a":1}}` returns a hashref and the `@$ids` check in `update` dies with "Not an ARRAY reference" → HTTP 500 / unknown-fatal. Legacy returned a clean `invalid_params` user error.

Suggest `ref $ids eq 'ARRAY'` with a user error otherwise.

Bugzilla/API/V1/BugUserLastVisit.pm
```diff
@@ -0,0 +1,174 @@
...
+    # skip aliases.
+    $user->visible_bugs([grep {/^[0-9]+$/} @$ids]);
+  }
+
+  my @last_visits = @{$user->last_visited};
+
+  if ($ids) {
+
+    # remove bugs that we are not interested in if ids is passed in.
+    my %id_set = map { ($_ => 1) } @$ids;
```
dklawren, line 66: **Same root cause on the read path: `%id_set` is keyed on the raw path/query value**, while the `$user->last_visited` records carry numeric `bug_id`s.

`GET /rest/bug_user_last_visit/my-bug-alias` therefore returns a successful `200 []` rather than the visit record (legacy: 404 `rest_invalid_resource`). Resolve aliases to bug ids before building `%id_set`, or reject non-numeric ids up front.

Bugzilla/API/V1/BugUserLastVisit.pm
```diff
@@ -0,0 +1,174 @@
...
+  }
+
+  my $ids = $self->every_param('ids');
+  return @$ids ? $ids : undef;
+}
+
+sub _filter_params {
+  my ($self) = @_;
+
+  my $params = $self->req->params->to_hash;
```
dklawren, line 142: **`include_fields`/`exclude_fields` sent in the JSON POST body are now silently ignored**, since this only reads `$self->req->params` (query string + form body).

Legacy merged the decoded body into `$params` before `rest_include_exclude`/`filter`, so `POST /rest/bug_user_last_visit {"ids":[123],"include_fields":["last_visit_ts"]}` used to return only `last_visit_ts` and now also returns `id`. Worth merging the decoded body in here.

Bugzilla/WebService/Constants.pm
```diff
@@ -317,7 +317,6 @@ sub WS_DISPATCH {
     'User'             => 'Bugzilla::WebService::User',
     'Product'          => 'Bugzilla::WebService::Product',
     'Group'            => 'Bugzilla::WebService::Group',
-    'BugUserLastVisit' => 'Bugzilla::WebService::BugUserLastVisit',
```
dklawren, line 320: Dropping this entry removes the methods from JSON-RPC and XML-RPC as well, not just from the legacy REST layer — anything calling `jsonrpc.cgi` with `method=BugUserLastVisit.get`/`update` will start getting an unknown-method error.

This matches the pattern of the earlier migrations in the series, so probably intentional, but it's an undocumented breaking change that seems worth calling out in the PR description / API docs.

dklawren (changes requested):

Bugzilla/API/V1/BugUserLastVisit.pm
Xzzz: Fixed in "Bug 2065171 - Merge query-string and JSON body params for ids/include_fields"

Bugzilla/API/V1/BugUserLastVisit.pm
Xzzz: Fixed in "Bug 2065171 - Merge query-string and JSON body params for ids/include_fields"

Bugzilla/API/V1/BugUserLastVisit.pm
Xzzz: Fixed following your suggestion: `ref $ids eq 'ARRAY' ? $ids : [$ids];` and added user_error "ids must be an array" if `$ids ne 'ARRAY'`.
Committed in "Bug 2065171 - Reject non-array ids with invalid_params"

Bugzilla/API/V1/BugUserLastVisit.pm
Xzzz: I finally went with `$bug->id` at the call site rather than pass `$bug` and call `->id` inside the helper, as it is also shared with `get()` which has a plain numeric `bug_id` and not a `$bug` object.
=> `update()` now calls `_bug_user_last_visit_to_hash($bug->id, ...)` and `$bug_id` is always `$bug->id` (from `update`) or `$_->bug_id` (from `get`).

Fixed in "Bug 2065171 - Use resolved bug id, restrict :id route to digits"

Bugzilla/API/V1/BugUserLastVisit.pm
Xzzz: Finally went with reject, enforced at the route level instead of in `get()`: `/:id` now takes `[id => qr/\d+/]`, so a non-numeric alias in the URL never reaches this method => 404 before `%id_set` is built.

Left query-string `ids` alone: legacy never resolved aliases there either (it just compared raw values against `last_visited`'s numeric `bug_id`s), so a non-numeric value there was already a silent no-match in both versions, not a regression.

Fixed in "Bug 2065171 - Use resolved bug id, restrict :id route to digits"

Bugzilla/WebService/Constants.pm
Xzzz: Correct, and yes intentional: same tradeoff as the Classification/Bugzilla(system-info) migrations before this one => dropping the `WS_DISPATCH` entry removes `BugUserLastVisit.get`/`update` from JSON-RPC and XML-RPC as well as from the legacy REST dispatcher, since all three share that table. Native Mojo routes only serve REST.

I hadn't called this out explicitly in the description => added a note to it now
Happy to raise that on [2057358](https://bugzilla.mozilla.org/show_bug.cgi?id=2057358) too, as it'd apply to every resource in the series, not just this one.

And yes, I totally agree, API documentation needs an update too, and even a big one ;)

Bugzilla/API/V1/BugUserLastVisit.pm
```diff
@@ -0,0 +1,183 @@
...
+}
+
+sub _ids_from_request {
+  my ($self) = @_;
+
+  if (my $id = $self->param('id')) {
+    return [$id];
+  }
+
+  my $ids = $self->_request_params->{ids} // [];
```
dklawren: defaulting to `[]` here changes behavior for `GET /rest/bug_user_last_visit` with no ids. the old code left `$ids` undef, so both `if ($ids)` blocks in `get` were skipped and every visited bug was returned. an empty arrayref is truthy, so now the id filter runs with an empty `%id_set` and the endpoint returns `[]`

suggest returning undef when no ids were supplied, or switching the guards in `get` to `if (@$ids)`

Bugzilla/API/V1/BugUserLastVisit.pm
```diff
@@ -0,0 +1,183 @@
...
+sub get {
+  my ($self) = @_;
+
+  my $user = $self->bugzilla->login;
+  $user->id || return $self->user_error('login_required');
+
+  my ($ids, $error, $vars) = $self->_ids_from_request;
+  return $self->user_error($error, $vars) if $error;
+
+  if ($ids) {
```
dklawren: same root cause as the `// []` default below: `[]` is truthy, so this and the filter block at line 63 always run. `if (@$ids)` is the safe check

Bugzilla/API/V1/BugUserLastVisit.pm
```diff
@@ -0,0 +1,183 @@
...
+
+=head1 NAME
+
+Bugzilla::API::V1::BugUserLastVisit - Find and Store the last time a user
+visited a bug.
+
+=head1 DESCRIPTION
+
+This part of the Bugzilla REST API allows you to lookup and update the last
+time a user visited a bug.
```
dklawren: no qa test covers this endpoint, before or after the migration. worth adding `qa/t/rest_bug_user_last_visit.t` alongside this change, mirroring `qa/t/rest_classification.t`

minimum cases: POST with `ids` in a json body, GET by path id, GET with no ids (returns everything), and unauthenticated access returning `login_required`

Bugzilla/API/V1/BugUserLastVisit.pm
dklawren, line 125: defaulting to `[]` here changes behavior for `GET /rest/bug_user_last_visit` with no ids. the old code left `$ids` undef, so both `if ($ids)` blocks in `get` were skipped and every visited bug was returned. an empty arrayref is truthy, so now the id filter at line 63 runs with an empty `%id_set` and the endpoint returns `[]`

suggest returning undef when no ids were supplied, or switching both guards in `get` (lines 53 and 63) to `if (@$ids)`

dklawren (changes requested):

Bugzilla/API/V1/BugUserLastVisit.pm
Xzzz: Good catch :+1: Changed to return `undef` when `ids`is absent (matching legacy).

Fixed in "Bug 2065171 - Return undef when ids is absent"

Bugzilla/API/V1/BugUserLastVisit.pm
```diff
@@ -0,0 +1,184 @@
...
+
+  if (my $id = $self->param('id')) {
+    return [$id];
+  }
+
+  my $ids = $self->_request_params->{ids};
+  return undef unless defined $ids;
+  return (undef, 'invalid_params', {type_error => 'ids must be an array'})
+    if ref $ids && ref $ids ne 'ARRAY';
+  return ref $ids eq 'ARRAY' ? $ids : [$ids];
```
dklawren: two precedence differences from the legacy layer here

1. path id short-circuits `ids`, but legacy merged the request body *over* path params (`_retrieve_json_params` applies `%$extra_params` last), so `POST /rest/bug_user_last_visit/123` with body `{"ids":[456]}` used to update 456 and now updates 123. query-string precedence is unchanged, just the body

2. `if (my $id = $self->param('id'))` is a truthiness test, so `/bug_user_last_visit/0` falls through to the no-ids branch and `get` returns the whole last-visited list instead of an empty one. `$self->param` also falls back to request params, so `?id=5` is now treated as a filter where legacy only looked at `ids`

checking `defined $self->stash('id')` instead would fix both halves of 2

Bugzilla/API/V1/BugUserLastVisit.pm
```diff
@@ -0,0 +1,184 @@
...
+}
+
+sub _request_params {
```
dklawren, line 138: this param-merging layer is hand-rolled and has no automated coverage

the other native migrations in this series each have a qa test (`rest_classification.t`, `rest_components.t`, `rest_reminders.t`) and 65642f5ff updated `rest_bugzilla.t`, so `qa/t/rest_bug_user_last_visit.t` would fit the convention

worth covering the cases already in the PR test plan plus the query-string-vs-json-body precedence, since three frontend call sites depend on this endpoint (`bug_modal.js`, `MyDashboard/query.js`, `show-header.html.tmpl`) and they post a json body with no `Content-Type`, which is exactly the path `decode_json` here handles

dklawren (changes requested):

Xzzz:
Pushed a follow-up commit: `_request_params` now calls a shared `Bugzilla::WebService::Util::merge_request_params` helper instead of doing the query-string/JSON-body merge inline. Extracted a single reusable helper rather than keeping two copies.
The `include_fields`/`exclude_fields` split stays local here since it's specific to this resource.
=> No behavior change

Bugzilla/API/V1/BugUserLastVisit.pm
Xzzz: Fixed:
- ids precedence now matches legacy:
  -> for POST, a body/query-string `ids` overrides the path id (falls back to the path id only when the request has no `ids` at all)
  -> for GET, no change (the path id still wins)
- Switched from `$self->param('id')` to `defined $self->stash('id')`, which fixes both id=0 (was falsy, fell through to "no ids") and the stray `?id=5` query param being misread as a path id.

Bugzilla/API/V1/BugUserLastVisit.pm
Xzzz: Added `qa/t/rest_bug_user_last_visit.t`, matching the `rest_classification.t`, `rest_components.t`, and `rest_reminders.t` convention.
It covers the PR test plan cases plus the query-string-vs-json-body/path precedence (including the no-Content-Type case the three frontend callers rely on).

=> Fixed in "Bug 2065171 - Add qa/t/rest_bug_user_last_visit.t"

Bugzilla/WebService/Util.pm
```diff
@@ -297,6 +300,29 @@ sub params_to_objects {
...
+  # Bugzilla::WebService::Server::REST) and the documented behavior in
+  # docs/en/rst/api/core/v1/general.rst.
+  my $params = $c->req->params->to_hash;
+
+  if (length $c->req->body) {
+    my $body_params;
+    try { $body_params = decode_json($c->req->body); }
+    catch { $body_params = undef; };
+    $params = {%$body_params, %$params} if ref $body_params eq 'HASH';
+  }
```
dklawren, line 321: a malformed JSON body is silently discarded here, which can produce a wrong result instead of an error

legacy `_retrieve_json_params` threw `json_rpc_invalid_params` on a decode failure. now the `catch` sets `undef` and the request continues.

concrete case: a client sends `POST /rest/bug_user_last_visit/123` with an intended body of `{"ids":[456]}`, but the body is truncated. decode fails, `$body_params` is undef, and the path id wins, so the endpoint returns 200 and records a visit against bug **123** instead of 456. no error reaches the caller.

the bare `POST /rest/bug_user_last_visit` variant degrades into a confusing `param_required` for the same reason.

suggestion: when the body is non-empty and the content type is JSON but decoding fails, raise `rest_malformed_json` (already used in `Bugzilla/API/V1/Reminders.pm:69`). keep ignoring genuinely non-JSON bodies so urlencoded and multipart form posts still work.

worth also calling this file out in the PR description, since it is shared and the change affects more than this endpoint

qa/t/rest_bug_user_last_visit.t
```diff
@@ -0,0 +1,138 @@
...
+        severity    => 'blocker',
+        description => $summary,
+      })->status_is(200)->json_has('/id');
+  return $t->tx->res->json->{id};
+}
+
+my $bug_id_1 = create_bug('bug_user_last_visit test bug 1');
+my $bug_id_2 = create_bug('bug_user_last_visit test bug 2');
+
+### Section 1: Anonymous access requires login
```
dklawren, line 47: the negative cases that would actually prove behavioral parity are missing

- a **group-restricted bug the user cannot see** - this is the security relevant one, and the parity claim in the PR description rests on it. `qa/config` already has restricted-group fixtures used by other `qa/t/rest_*.t` files
- anonymous `POST` (only anonymous GET is covered)
- a nonexistent bug id, including that the transaction is rolled back mid-loop
- `POST /rest/bug_user_last_visit` with no `ids` anywhere, expecting `param_required`
- `include_fields` / `exclude_fields`, which `_request_params` explicitly handles

the first one plus the malformed-body case on `Util.pm` are the same underlying gap: failures that quietly turn into plausible-looking successes

Bugzilla/API/V1/BugUserLastVisit.pm
```diff
@@ -0,0 +1,175 @@
...
+  my @last_visits = @{$user->last_visited};
+
+  if ($ids) {
+
+    # remove bugs that we are not interested in if ids is passed in.
+    my %id_set = map { ($_ => 1) } @$ids;
+    @last_visits = grep { $id_set{$_->bug_id} } @last_visits;
+  }
+
+  my $params = $self->_request_params;
```
dklawren, line 67: the request body gets read four times per POST on a hot path

`_request_params` runs twice per request, once inside `_ids_from_request` (line 47 in `get`, line 84 in `update`) and again here, and `merge_request_params` itself calls `$c->req->body` twice (`Bugzilla/WebService/Util.pm:316,318`). for a file-backed request asset each `body` call re-slurps from disk, so one POST slurps 4x and JSON-decodes 2x.

`extensions/BugModal/web/bug_modal.js:1699` fires a GET plus a POST on every bug page view, so this runs constantly.

suggestion: compute `my $params = $self->_request_params` once in `get` and `update` and pass it into `_ids_from_request`, and hoist `my $body = $c->req->body` into a variable in `merge_request_params`

Bugzilla/API/V1/BugUserLastVisit.pm
```diff
@@ -0,0 +1,175 @@
...
+  my $routes = $r->under(
+    '/bug_user_last_visit' => sub { Bugzilla->usage_mode(USAGE_MODE_MOJO_REST); });
+  $routes->get('/')->to('V1::BugUserLastVisit#get');
+  $routes->get('/:id' => [id => qr/\d+/])->to('V1::BugUserLastVisit#get');
+  $routes->post('/')->to('V1::BugUserLastVisit#update');
+  $routes->post('/:id' => [id => qr/\d+/])->to('V1::BugUserLastVisit#update');
+
+  foreach my $path ('/', '/:id') {
+    $routes->options($path)->to('V1::BugUserLastVisit#options');
+  }
```
dklawren, line 29: the OPTIONS routes use a bare `/:id` while GET and POST on lines 23 and 25 constrain `[id => qr/\d+/]`

so `OPTIONS /rest/bug_user_last_visit/abc` answers `200 Allow: GET, POST`, advertising methods that 404 on that path

suggestion: apply the same `[id => qr/\d+/]` constraint in the OPTIONS loop

dklawren (changes requested):

Bugzilla/WebService/Util.pm
Xzzz: Fixed. The underlying issue is gone now that #2751 has merged: the shared `merge_request_params` in master already returns `($params, $error)` and raises `rest_malformed_json`, and it only attempts a decode when the body was not already parsed as form params, so urlencoded and multipart posts still work. This branch has been rebased onto that and `_request_params` now propagates the error, which `get/update` pass to `user_error`.

That rebase did surface a regression worth flagging. The collapse-to-scalar you asked for in #2751 uses `$c->req->param`, which keeps only the last value, so `?ids=1&ids=2` silently became a single id and `qa/t/rest_bug_user_last_visit.t`'s query-string case failed. I did not want to work around it locally, because the old `->to_hash` had the mirror-image flaw: it returned a scalar for one occurrence and an arrayref for two, which is why `Group.pm` has to run everything back through `validate()` to recoerce. So `merge_request_params` now takes an optional arrayref of parameter names that are lists; those always come back as arrayrefs and everything else always as a scalar, with no dependence on how many times a key was sent.
Called as `merge_request_params($self, ['ids'])` here. Default behaviour is unchanged when the argument is omitted. This also fixes the same latent bug in bug 2065173, where `?ids=1&ids=2` currently returns one group.

Pleasde note that Bugzilla/WebService/Util.pm is called out in the PR description as you asked.

Bugzilla/API/V1/BugUserLastVisit.pm
Xzzz: Fixed, both halves. get and update now call `_request_params` once and pass the result into `_ids_from_request`, and `merge_request_params` reads `$c->req->body` into a variable.
=> That takes a POST from four body reads and two JSON decodes down to one of each

Bugzilla/API/V1/BugUserLastVisit.pm
Xzzz: Fixed: The `OPTIONS` routes are now declared explicitly instead of in a loop, with the same `[id => qr/\d+/]` constraint as `GET/POST`, so `OPTIONS /rest/bug_user_last_visit/abc` 404s rather than advertising methods that do not exist on that path. Covered in `qa/t/rest_bug_user_last_visit.t`

qa/t/rest_bug_user_last_visit.t
Xzzz: Added, all five :
- a bug in the QA-Selenium-TEST group, created by the one user who belongs to it, then requested by the `editbugs` user: `POST` returns 401 / code 102 `bug_access_denied`, and a `GET` filtered on that id returns an empty list rather than leaking it. This is the parity case you singled out, and it confirms the new code behaves like the legacy one: `Bugzilla::Bug->check` runs `check_is_visible` and throws before the next unless `$user->can_see_bug` line is ever reached, in both versions.
- anonymous `POST`, alongside the anonymous `GET` that was already there,
- a nonexistent bug id, asserting 404 / code 101 and that the visit recorded earlier in the same loop is rolled back. `Bugzilla::Error::_throw_error` issues an `explicit bz_rollback_transaction()` before dying, so this is real behaviour rather than an accident of timing. The test sleeps one second first, because `last_visit_ts` has second granularity and a broken rollback could otherwise produce an identical timestamp and pass spuriously.
- `POST /rest/bug_user_last_visit` with no `ids` in the path, query string or body, expecting `param_required`,
- include_fields and exclude_fields, the two that _request_params splits by hand

Xzzz:
The branch has been rebased onto master now that #2751 landed, so it no longer carries its own copy of `merge_request_params`. That rebase needed one change to the shared helper, which is described in the thread on `Bugzilla/WebService/Util.pm` and in a dedicated section of the PR description: the collapse-to-scalar kept only the last value of a repeated key, so `?ids=1&ids=2` silently became a single id. Callers now declare which parameters are lists. It is backward compatible, and it also fixes the same latent bug in #2745.

Two things surfaced while writing the tests. Neither belongs in this PR, and I would rather not repeat the `i_am_webservice` situation, so flagging them here instead. Happy to file either or both if you agree.

**Unmatched `/rest/` paths fall through to the legacy CGI rather than 404.**

`Bugzilla/App/Controller/CGI.pm:47` declares `$r->any('/rest/*PATH_INFO')->to('CGI#rest_cgi')`. When a route constraint rejects a path, the request is not answered with a 404 by the Mojo router, it falls through to that catch-all and reaches the legacy `rest.cgi`, which no longer knows a resource once it has been migrated. In practice the connection is closed with no response at all.

I hit this with the `[id => qr/\d+/]` constraint you asked for on the `OPTIONS` routes: `OPTIONS /rest/bug_user_last_visit/abc` used to answer a misleading `200 Allow: GET, POST` and now drops the connection. The constraint itself is correct and is in place, but I dropped the test assertion rather than pin broken behaviour. Worth noting that `GET` and `POST` on the same path have carried this constraint since earlier in this branch and behave the same way, so this is not something the `OPTIONS` change introduced.

This affects every native REST controller in the 2057358 series that constrains a placeholder, so it seems worth its own bug.

**A nonexistent bug id produces a misleading error.**

`Bugzilla::Bug->new` only sets `error => 'NotFound'` when it is handed a bare scalar. Given a hashref, as both this controller and the legacy one pass it (`Bugzilla::Bug->check({id =>$bug_id, cache => 1})`), it sets `error => 'InvalidBugId'` and `bug_id => $param->{name}`, which is undef. `check()` therefore reports `improper_bug_id_field_value` with no bug id, and the caller gets 400 / code 100 "You must enter a valid bug number!" instead of 404 / code 101 "Bug XXXXXXX does not exist."

Behaviour is identical in the legacy endpoint, so this PR preserves it deliberately and the new test asserts the real behaviour with a comment explaining why. But it is a confusing response for API callers, and it applies to any endpoint that calls `check()` with a hashref.

Bugzilla/WebService/Util.pm
```diff
@@ -301,29 +301,38 @@ sub params_to_objects {
...
+  # re-slurps it from disk.
+  my $body = $c->req->body;
+  if (length $body && !@{$c->req->body_params->names}) {
     my $body_params;
     my $error;
-    try { $body_params = decode_json($c->req->body); }
+    try { $body_params = decode_json($body); }
     catch { $error = 'rest_malformed_json'; };
     return (undef, $error) if $error;
     $params = {%$body_params, %$params} if ref $body_params eq 'HASH';
```
dklawren: json body values skip the list coercion, so declared list params can still come back as scalars. wrap them after the merge or narrow the pod

dklawren (approved):
LGTM r=dkl

PR #2750, opened 2026-09-17
Bug 2072305 - Accept query-string params on Component create/update

Xzzz:
## Summary

`Component.pm`'s `create`/`update` only read parameters from the JSON POST/PUT body. Adds the query-string merge that the REST docs promise (query string overrides the body on a key collision), reusing the same `merge_request_params` helper as bugs 2065171/2065173, and fixes a stray-key crash in `update()`'s `set_all()` call that this change would otherwise expose.

## Changes

- Add `Bugzilla::WebService::Util::merge_request_params` (third copy of this helper across open PRs -- will collapse to one once 2065171 or 2065173 merges)
- `Component.pm` `create`/`update`: query string + JSON body merge instead of JSON-body-only; drop the now-unused `_get_params`
- `Component.pm` `update()`: whitelist the 8 documented update fields before `set_all()`, since form-urlencoded cookie-auth requests (which include `Bugzilla_api_token`) will now actually reach it instead of failing JSON parsing first
- `qa/t/rest_components.t`: cover query-string-only create and query-string-overrides-body update

## Test plan

- POST /rest/component/Firefox with fields entirely in the query string, no JSON body
- PUT /rest/component/Firefox/<name> with a query-string field overriding a JSON-body field
- Existing qa/t/rest_components.t cases unaffected

## References

- Bugzilla link: [2072305](https://bugzilla.mozilla.org/show_bug.cgi?id=2072305)


Bugzilla/WebService/Util.pm
```diff
@@ -297,6 +300,29 @@ sub params_to_objects {
...
+
+  # $c->req->params already covers the query string plus, for POST/PUT, an
+  # application/x-www-form-urlencoded or multipart body. Layer a JSON body
+  # underneath that (silently ignored if absent or not valid JSON), so
+  # params work from either the query string or a JSON request body.
+  # Query-string values win on a key collision, matching the legacy REST
+  # layer (see fix_credentials/_retrieve_json_params in
+  # Bugzilla::WebService::Server::REST) and the documented behavior in
+  # docs/en/rst/api/core/v1/general.rst.
+  my $params = $c->req->params->to_hash;
```
dklawren, line 314: `req->params` is `body_params->clone->append(query_params)`, and `Mojo::Parameters::to_hash` turns a repeated key into an arrayref instead of letting one side win. so a form-urlencoded body `description=B` plus `?description=A` gives `description => ['B','A']`, which `set_description` stringifies to `ARRAY(0x...)` in the db -- the "query string wins" claim in the comment above doesn't hold, and the same applies to any repeated query param. the legacy `_retrieve_json_params` avoids this by assigning `url_param` values one at a time. suggest layering each source explicitly, e.g. json body, then `$c->req->body_params->to_hash`, then `$c->req->query_params->to_hash`

Bugzilla/WebService/Util.pm
```diff
@@ -297,6 +300,29 @@ sub params_to_objects {
...
+
+  if (length $c->req->body) {
+    my $body_params;
+    try { $body_params = decode_json($c->req->body); }
+    catch { $body_params = undef; };
```
dklawren, line 319: swallowing the decode error drops the `rest_malformed_json` response `_get_params` used to return. a PUT with a broken json body now falls through to `set_all({})` and returns 200 with an unchanged component, which reads as success. `Reminders.pm` still throws `rest_malformed_json` for the same input, and the legacy layer cited above throws `json_rpc_invalid_params`, so this is a behaviour regression rather than a match. consider signalling the parse failure to the caller

Bugzilla/API/V1/Component.pm
```diff
@@ -95,21 +94,24 @@ sub update {
...
-  my ($params, $error) = $self->_get_params();
-  return $self->user_error($error) if $error;
+  my $params = merge_request_params($self);
+
+  # Whitelist the documented update fields; set_all() throws unknown_method
+  # for any stray key (e.g. Bugzilla_api_token, include_fields).
+  my %values = map { $_ => $params->{$_} }
+    grep { exists $params->{$_} }
+    qw(name description default_assignee default_qa_contact default_bug_type
+    is_active triage_owner team_name bug_description_template);
```
dklawren, line 104: `is_active` is only a real boolean when it comes from a json body. from the query string it is the string `"false"`, and `isactive` validates through `Bugzilla::Object::check_boolean`, which is `$_[1] ? 1 : 0` -- so `?is_active=false` activates the component, the opposite of the documented value in component.rst. the query-string path needs an explicit string-to-boolean coercion

dklawren (changes requested):

Bugzilla/WebService/Util.pm
Xzzz: Resolved by merging master: the shared `merge_request_params` from bug 2065171 now reads each param with `$c->req->param`, so a repeated key keeps only its last value, and since `req->params` appends the query string after the body, the query string wins. Params that really are lists have to be declared explicitly (`merge_request_params($c, ['ids'])`). Component has none, so every field comes back as a plain string.

Added a test sending a form-urlencoded `description` together with a `?description=` query parameter, asserting that the query-string value is stored.

=> Fixed in "Bug 2072305 - Test that the query string wins over a form body"

Bugzilla/WebService/Util.pm
Xzzz: Also resolved by merging master: `merge_request_params` now returns `($params, $error)`, with `rest_malformed_json` for a body that fails to decode. Both `create` and `update` in Component now return that error to the caller (done in the merge commit, since the helper's return value changed).

Added a test that a `PUT` with a broken JSON body returns 400 with the malformed-JSON message, instead of 200 with an unchanged component.

=> Fixed in "Bug 2072305 - Test that a malformed JSON body is rejected"

Bugzilla/API/V1/Component.pm
Xzzz: `update()` now converts a non-JSON `is_active` explicitly: `true`/`1` becomes 1 and `false`/`0` becomes 0 (case-insensitive), and any other value is rejected with `invalid_params`. JSON booleans are unchanged.

Added tests for `?is_active=false`, `?is_active=true` and an invalid value.

=> Fixed in "Bug 2072305 - Coerce is_active from query-string and form values"

dklawren (approved):
LGTM. r=dkl

PR #2751, opened 2026-09-17
Bug 2072313 - Accept query-string params on Reminders add

Xzzz:
## Summary

Reminders.pm's add() only read parameters from the JSON POST body. Adds the query-string merge that the REST docs promise, reusing the same merge_request_params helper as bugs 2065171/2065173/2072305.

Depends on [2072305](https://bugzilla.mozilla.org/show_bug.cgi?id=2072305) per the bug dependency (both introduce the same shared helper; will collapse to one copy once either merges).

## Changes

- Reuse Bugzilla::WebService::Util::merge_request_params (see 2072305)
- Reminders.pm add(): query string + JSON body merge instead of JSON-body-only
- qa/t/rest_reminders.t: cover query-string-only reminder creation

**Side effect on bug 2072318:** removes the try/catch this change replaces, which included the `return` inside `catch` that bug 2072318 was filed against (it never actually aborted `add()` on malformed JSON). That code no longer exists after this change, so 2072318 is resolved here rather than needing its own patch.

## Test plan

- POST /rest/reminder with bug_id/note/reminder_ts entirely in the query string, no JSON body
- Existing qa/t/rest_reminders.t cases unaffected

## References

- Bugzilla link: [2072313](https://bugzilla.mozilla.org/show_bug.cgi?id=2072313)
- Depends on [2072305](https://bugzilla.mozilla.org/show_bug.cgi?id=2072305)
- Resolves [2072318](https://bugzilla.mozilla.org/show_bug.cgi?id=2072318) as a side effect


Bugzilla/WebService/Util.pm
```diff
@@ -297,6 +300,29 @@ sub params_to_objects {
...
+  # layer (see fix_credentials/_retrieve_json_params in
+  # Bugzilla::WebService::Server::REST) and the documented behavior in
+  # docs/en/rst/api/core/v1/general.rst.
+  my $params = $c->req->params->to_hash;
+
+  if (length $c->req->body) {
+    my $body_params;
+    try { $body_params = decode_json($c->req->body); }
+    catch { $body_params = undef; };
+    $params = {%$body_params, %$params} if ref $body_params eq 'HASH';
```
dklawren, line 320: the empty `catch` swallows decode errors, so `rest_malformed_json` can never be raised anymore

with `Content-Type: application/json` and a truncated body like `{"bug_id": 123, "note": "x"`, decoding fails, `$params` ends up empty, and `_check_bug_id(undef)` reports "You must enter a valid bug number" — the client is told the bug id is wrong instead of the JSON

the PR body is right that the old `return` inside `catch` never aborted `add()` (`user_error` renders and returns 0, see `Bugzilla/App/Plugin/Error.pm:27`), but the fix also drops the error entirely rather than making it work

suggest signalling decode failure back to the caller when the body is non-empty and the content type is JSON, so the controller can still call `user_error('rest_malformed_json')`

Bugzilla/WebService/Util.pm
```diff
@@ -297,6 +300,29 @@ sub params_to_objects {
...
+
+  # $c->req->params already covers the query string plus, for POST/PUT, an
+  # application/x-www-form-urlencoded or multipart body. Layer a JSON body
+  # underneath that (silently ignored if absent or not valid JSON), so
+  # params work from either the query string or a JSON request body.
+  # Query-string values win on a key collision, matching the legacy REST
+  # layer (see fix_credentials/_retrieve_json_params in
+  # Bugzilla::WebService::Server::REST) and the documented behavior in
+  # docs/en/rst/api/core/v1/general.rst.
+  my $params = $c->req->params->to_hash;
```
dklawren, line 314: `to_hash` returns an arrayref for any key that appears more than once

`POST /rest/reminder?bug_id=123&note=a&note=b&reminder_ts=2024-06-08` makes `note` `['a','b']`, and `note` has no entry in `Bugzilla::Reminder::VALIDATORS`, so `run_create_validators` passes it straight through and the DB gets the literal string `ARRAY(0x...)`

`bug_id` and `reminder_ts` are shielded by their validators, `note` is not — and since this helper is meant for reuse by 2065171/2065173/2072305, every future caller inherits it

suggest collapsing to scalars, e.g. building the hash from `$c->req->params->names` plus `$c->req->param($_)`

Bugzilla/API/V1/Reminders.pm
```diff
@@ -63,13 +63,7 @@ sub add {
...
     if !$user->in_group(Bugzilla->params->{reminders_group});
 
-  my $params = {};
-  try {
-    $params = decode_json($self->req->body);
-  }
-  catch {
-    return $self->user_error('rest_malformed_json');
-  };
+  my $params = merge_request_params($self);
```
dklawren, line 66: this is where the `rest_malformed_json` response is lost — see the comment on `merge_request_params`

if the helper keeps swallowing decode errors, this call site needs its own check so a bad JSON body still returns the right error

qa/t/rest_reminders.t
```diff
@@ -75,6 +75,20 @@ $t->delete_ok($url . "rest/reminder/invalid" => {'X-Bugzilla-API-Key' => $api_ke
 $t->delete_ok($url . "rest/reminder/$id" => {'X-Bugzilla-API-Key' => $api_key})
   ->status_is(200)->json_is('/success' => 1);
 
+### Section 3b: Fields may also be passed entirely via the query string,
+### with no JSON body
+
+$t->post_ok($url
+    . "rest/reminder?bug_id=$bug_id&note=Query%20String%20Reminder"
+    . '&reminder_ts=2024-06-08' => {'X-Bugzilla-API-Key' => $api_key})
+  ->status_is(200)->json_is('/note' => 'Query String Reminder');
```
dklawren, line 84: only the happy path is covered here

the deliberate behavior change for malformed JSON (claimed to resolve 2072318) has no test — worth adding a POST with an invalid JSON body asserting the expected status and code, once the handling above is settled

dklawren (changes requested):

Bugzilla/WebService/Util.pm
Xzzz: Fixed: `merge_request_params` now returns `($params, $error)`, setting `$error = 'rest_malformed_json'` when the body is present but fails to decode, instead of silently discarding it. Callers check `$error` and call `user_error`

Bugzilla/WebService/Util.pm
Xzzz: Good catch. Fixed as sugested: replaced `$c->req->params->to_hash` with `$params->{$_} = $c->req->param($_) for @{$c->req->params->names}` so a repeated key collapses to a scalar instead of an arrayref

Bugzilla/API/V1/Reminders.pm
Xzzz: Addressed by the `merge_request_params` change above => `add()` now does `my ($params, $error) = merge_request_params($self); return $self->user_error($error) if $error;`, so a malformed body still gets `rest_malformed_json`.

qa/t/rest_reminders.t
Xzzz: Added a case posting a truncated JSON body and asserting `status 400`, `code 32000` and the malformed-JSON message.

Bugzilla/WebService/Util.pm
```diff
@@ -297,6 +300,34 @@ sub params_to_objects {
...
+  # legacy REST layer (see fix_credentials/_retrieve_json_params in
+  # Bugzilla::WebService::Server::REST) and the documented behavior in
+  # docs/en/rst/api/core/v1/general.rst.
+  #
+  # ->to_hash would turn a repeated key (e.g. ?note=a&note=b) into an
+  # arrayref, which validators don't expect, so collapse to scalars instead.
+  my $params = {};
+  $params->{$_} = $c->req->param($_) for @{$c->req->params->names};
+
+  if (length $c->req->body) {
```
dklawren, line 319: the decode is gated on `length $c->req->body`, not on the content type, so *any* non-JSON body is now rejected

`POST /rest/reminder?bug_id=123` with `Content-Type: application/x-www-form-urlencoded` and body `note=hello` returns `rest_malformed_json`, because `decode_json` is handed the raw form string

that contradicts the comment three lines above, which promises `$c->req->params` "already covers the query string plus, for POST/PUT, an application/x-www-form-urlencoded or multipart body" — the helper collects those params and then rejects the request they came from. same for multipart

the legacy layer doesn't have this problem: CGI.pm only populates `POSTDATA`/`PUTDATA` for non-form content types, so `_retrieve_json_params` never tries to decode a form body

and since 2065171/2065173/2072305 all reuse this helper, every future caller inherits it

suggest gating the decode on the content type — only decode when `$c->req->headers->content_type` matches `application/json` — or skipping it when `$c->req->body_params->names` is non-empty

Bugzilla/Util.pm
```diff
@@ -298,7 +298,9 @@ sub i_am_cgi {
 
 sub i_am_webservice {
   my $usage_mode = Bugzilla->usage_mode;
-  return $usage_mode == USAGE_MODE_JSON || $usage_mode == USAGE_MODE_REST;
+  return $usage_mode == USAGE_MODE_JSON
+    || $usage_mode == USAGE_MODE_REST
+    || $usage_mode == USAGE_MODE_MOJO_REST;
```
dklawren, line 303: this is a real bug and worth fixing — without it every native Mojo REST error message over 72 chars gets `\n`s injected by the `error_message` filter in `Bugzilla/Template.pm:866`, which is exactly what the new test would trip over — but it doesn't belong in bug 2072313

`i_am_webservice` has 14 call sites and this flips all of them for every `USAGE_MODE_MOJO_REST` endpoint at once, not just Reminders

most are unreachable under `MOJO_REST` — native REST auth goes through `Bugzilla/App/Plugin/Login.pm`, not `Bugzilla::Auth::Login::Cookie`/`APIKey` — so the practical risk is low. but `extensions/RestrictComments/Extension.pm:58` does change: `object_end_of_set_all` fires for `Bugzilla/API/V1/Github.pm:364` and `extensions/PhabBugz/lib/API/V1/PhabBugz.pm:235`, and will now stop force-clearing `restrict_comments` on those updates. that's arguably the correct behavior, but it's untested and unrelated to this bug

suggest either splitting it into its own bug, or dropping the commit here and relaxing the new assertion to `json_like` so the test doesn't depend on the wrapping

dklawren (changes requested):

Bugzilla/WebService/Util.pm
Xzzz: Fixed by skipping the decode when `$c->req->body_params->names` is not empty. Gating based solely on `application/json` would have rejected it, representing a silent regression compared to the legacy system. A test has been added in section 3b-bis (POST form => verifies that the reminder is indeed created).

Bugzilla/Util.pm
Xzzz: Good catch! Fixed by reverting the `i_am_webservice` change (returning to `USAGE_MODE_JSON || USAGE_MODE_REST`), and the assertion for test 3c has been switched to `json_like` using the pattern `qr/JSON data used for the request was malformed` (as this pattern is unaffected by 72-column wrapping).

Xzzz: Submitted new bug [2074854](https://bugzilla.mozilla.org/show_bug.cgi?id=2074854) "i_am_webservice() does not recognize USAGE_MODE_MOJO_REST"

dklawren (approved):
LGTM. r=dkl

PR #2753, opened 2026-09-18
Bug 2073282 - Add deprecated api_key=<key> query-parameter fallback to native Mojo REST auth

Xzzz:
## Summary

Adds a fallback to `Bugzilla::App::Plugin::Login`'s Mojo login check so `?api_key=<key>` query parameters are accepted the same way the legacy WebService dispatcher already accepts them, in case undiscovered callers rely on it (like k8s monitoring service actually do).

_Please note that this is a deprecation-pending stopgap, not a first-class supported method (docs already warn this method is likely to be deprecated due to security concerns)._

## Changes

- `Bugzilla::App::Plugin::Login`'s `bugzilla.login` helper: fall back to the `api_key` query/body parameter when the `X-Bugzilla-API-Key` header is absent, same precedence and validation path (revoked/sticky-IP checks, rate limiting) as the header case
- `qa/t/rest_native_login.t`: cover authentication via `?api_key=`
- `docs/en/rst/api/core/v1/general.rst`: note that native Mojo REST resources also accept the deprecated `api_key` query parameter as a fallback

## Test plan

- GET a native Mojo REST endpoint with `?api_key=<valid key>` and no header -> authenticates
- Existing `qa/t/rest_native_login.t` cases (header, cookie + token) unaffected

## References

- Bugzilla link: [2073282](https://bugzilla.mozilla.org/show_bug.cgi?id=2073282)


Bugzilla/App/Plugin/Login.pm
```diff
@@ -107,7 +107,15 @@ sub register {
...
       # For api requests, we check for the api key in the header
       if ($usage_mode == USAGE_MODE_REST || $usage_mode == USAGE_MODE_MOJO_REST) {
-        if (my $api_key_text = $headers->header('x-bugzilla-api-key')) {
+
+        # Deprecated fallback for the legacy ?api_key=<key> query parameter,
+        # same as the legacy WebService dispatcher (see
+        # Bugzilla::WebService::Util::fix_credentials). This is a
+        # deprecation-pending stopgap, not a first-class supported method.
+        my $api_key_text
+          = $headers->header('x-bugzilla-api-key') || $c->param('api_key');
```
dklawren, line 116: two gaps versus the legacy dispatcher this is meant to match

1. `fix_credentials` accepts both `api_key` **and** `Bugzilla_api_key` (Bugzilla/WebService/Util.pm:330 promotes the short form, and Bugzilla/Auth/Login/APIKey.pm:54 reads the long form directly). the docs warning this PR now points to also lists both. a k8s-style caller sending `?Bugzilla_api_key=<key>` still gets a 401 on native resources, so the "undiscovered callers" case is only half covered

Looking through the logs, there are unfortunately some use of the Bugzilla_api_key query parameter in addition to api_key so we need to support both for now.

2. `$c->param` reads urlencoded/multipart **body** params too, not just the query string. the code comment above says "query parameter", the docs change says "query parameter", and the test only covers the query string — so the deprecated surface being reopened is larger than what is described. either use `$c->req->query_params->param('api_key')` to keep the stopgap minimal, or say plainly that body params are accepted as well

qa/t/rest_native_login.t
```diff
@@ -49,6 +49,13 @@ $t->get_ok($url . $endpoint)->status_is(401);
 $t->get_ok($url . $endpoint => {'X-Bugzilla-API-Key' => $api_key})
   ->status_is(200)->json_has('/result');
 
+#
+# 2a. Deprecated fallback: an API key passed as the ?api_key= query parameter
+#     also works (bug 2073282), same as the legacy WebService dispatcher.
+#
+$t->get_ok($url . $endpoint . '&api_key=' . $api_key)
```
dklawren, line 56: worth adding a case where both the header and `?api_key=` are sent, asserting the header wins

that precedence is security relevant — `fix_credentials` documents it as stopping body-injected params from overriding gateway auth headers, and it already regressed once (t/webservice-fix-credentials.t:85, bug 2035598). the `||` here preserves it but nothing locks it in

docs/en/rst/api/core/v1/general.rst
```diff
@@ -114,8 +114,10 @@ Send only one authentication method with each request. BMO does not combine
 credentials or choose the strongest method when more than one is supplied.
 
 Most resources have been migrated off the legacy authentication path onto BMO's
-native REST framework, which accepts only a cookie, an ``X-Bugzilla-API-Key``
-header, or an OAuth2 bearer token. Legacy ``Bugzilla_login`` and
+native REST framework, which accepts a cookie, an ``X-Bugzilla-API-Key``
```
dklawren, line 117: "described in the warning below" overstates it — that warning covers `Bugzilla_api_key` **or** `api_key`, but only `api_key` works on native resources after this change. so we need to implement the long form too.

dklawren (changes requested):
Thanks for jumping on this quickly.

Bugzilla/App/Plugin/Login.pm
Xzzz: 1. `Bugzilla_api_key` is now accepted as well as `api_key`, with the same precedence as `fix_credentials`: header, then `Bugzilla_api_key`, then `api_key`. Added a `?Bugzilla_api_key=` case to `rest_native_login.t`.
=> Fixed in "Bug 2073282 - Also accept the Bugzilla_api_key query parameter"

2. The fallback now reads only the query string (`$c->req->query_params`). Urlencoded/multipart body params are no longer accepted, so the behaviour matches what the code comment, docs and test describe.
=> Fixed in "Bug 2073282 - Read the deprecated API key from the query string only"

qa/t/rest_native_login.t
Xzzz: Added a case sending a valid `X-Bugzilla-API-Key` header together with `?api_key=bogus-key-value`, expecting 200. If the query parameter won, the bogus key would be rejected with a 401, so this locks in the header precedence.
=> Fixed in "Bug 2073282 - Test that the API key header wins over the query parameter"

docs/en/rst/api/core/v1/general.rst
Xzzz: The long form is now implemented (see the Login.pm thread), and the paragraph names both `Bugzilla_api_key` and `api_key`, so it matches the warning below.
=> Fixed in "Bug 2073282 - Also accept the Bugzilla_api_key query parameter"

Bugzilla/App/Plugin/Login.pm
Xzzz: 1. `Bugzilla_api_key` is now accepted as well as `api_key`, with the same precedence as `fix_credentials`: header, then `Bugzilla_api_key`, then `api_key`. Added a `?Bugzilla_api_key=` case to `rest_native_login.t`. Fixed in "Bug 2073282 - Also accept the Bugzilla_api_key query parameter".
2. The fallback now reads only the query string (`$c->req->query_params`). Urlencoded/multipart body params are no longer accepted, so the behaviour matches what the code comment, docs and test describe. Fixed in "Bug 2073282 - Read the deprecated API key from the query string only".

qa/t/rest_native_login.t
Xzzz: Added a case sending a valid `X-Bugzilla-API-Key` header together with `?api_key=bogus-key-value`, expecting 200. If the query parameter won, the bogus key would be rejected with a 401, so this locks in the header precedence. Fixed in "Bug 2073282 - Test that the API key header wins over the query parameter".

docs/en/rst/api/core/v1/general.rst
Xzzz: The long form is now implemented (see the Login.pm thread), and the paragraph names both `Bugzilla_api_key` and `api_key`, so it matches the warning below. Fixed in "Bug 2073282 - Also accept the Bugzilla_api_key query parameter".

dklawren (approved):
LGTM. r=dkl

